PR #20, opened 2019-11-23
Experimental: Streams

mikearnaldi:


mikearnaldi:
just got the first version to compile, this is far from being a small one given I had to touch internal components like Deferred. We should write tests for it and check also that "R" gets correctly passed.

Some typings don't really match (even in the original there are forced casts)

Except for this api looks very nice

mikearnaldi:
Merge to avoid conflicts on main index.ts, note:
- missing refactor to convention (pipe/non-pipe for Stream & Managed monads)
- tests for streams (getting to decent coverage)

PR #34, opened 2019-11-27
bootstrap rxjs integration

mikearnaldi:


mikearnaldi:
@sledorze you did ask for a rxjs integration, does this cover the scope?

PR #36, opened 2019-11-27
add provideM & provideSomeM

sledorze:


packages/effect/src/index.ts
```diff
@@ -459,6 +459,26 @@ export const provideAll = <R>(r: R) => <E, A>(
...
 
+/**
+ * Provides all environment necessary to the child effect via an effect
+ *
+ * Note that this ***should*** be typically used at ***startup time***, not dynamically
+ */
+
+export const provideM = <R2, R, E2>(f: Effect<R2, E2, R>) => <E, A>(
+  ma: Effect<R, E, A>
+): Effect<R2, E | E2, A> => chain(f, r => provide(r)(ma));
```
mikearnaldi: if this provides all it can use `provideAll` and be safe for runtime

packages/effect/src/index.ts
```diff
@@ -459,6 +459,26 @@ export const provideAll = <R>(r: R) => <E, A>(
...
+
+/**
+ * Provides some of the environment necessary to the child effect via an effect
+ *
+ * Note that this should be typically used at startup time, not dynamically
+ */
+
+export const provideSomeM = <R2, R, E2>(f: Effect<R2, E2, R>) => <E, A, R3>(
+  ma: Effect<R & R3, E, A>
```
mikearnaldi: same here

PR #37, opened 2019-11-27
first list implem for rxjs

sledorze:


sledorze:
@mikearnaldi maybe you do not want it in effect; I've not exposed it from index as it does not has an Effect interface

sledorze:
@mikearnaldi actually I will do some rename and comments

mikearnaldi:
we don't expose all from index, fine accessing from /lib

PR #49, opened 2019-12-01
StreamEither (alpha)

mikearnaldi:


packages/effect/src/stream/estream.ts
```diff
@@ -0,0 +1,263 @@
...
+}
+
+export function fromArray<A>(as: readonly A[]): EStream<T.NoEnv, T.NoErr, A> {
+  return S.stream.map(S.fromArray(as), a => Ei.right(a));
+}
+
+export function fromIterator<A>(
+  iter: Lazy<Iterator<A>>
+): EStream<T.NoEnv, T.NoErr, A> {
+  return S.stream.map(S.fromIterator(iter), a => Ei.right(a));
```
sledorze: General remark, we should get ride of reducible calls like this (eta)

packages/effect/src/stream/estream.ts
```diff
@@ -0,0 +1,263 @@
...
+
+export const URI = "matechs/EStream";
+export type URI = typeof URI;
+declare module "fp-ts/lib/HKT" {
+  interface URItoKind3<R, E, A> {
+    [URI]: EStream<R, E, A>;
+  }
+}
+
+export const stream: Monad3E<URI> = {
```
sledorze: Shouldn't we directly implement Bifunctor and MonadThrow ?

mikearnaldi: probably yes, I planned to finish porting all the api of stream to here and after finish the instance

PR #50, opened 2019-12-01
add effectToObservable

sledorze:


packages/rxjs/src/index.ts
```diff
@@ -115,3 +115,40 @@ export function toObservable<R, E, A>(
       })
   );
 }
+
+export function effectToObservable<A>(
+  eff: T.Effect<T.NoEnv, never, A>
+): Rx.Observable<A> {
+  return new Rx.Observable(sub => {
+    let unsubs = false;
+    T.run(
```
mikearnaldi: this returns a `Lazy<void>` that you should call to interrupt

packages/rxjs/tests/RxJS.test.ts
```diff
@@ -210,3 +210,51 @@ describe("RxJS", () => {
...
+      .bind("obs", T.pure(O.effectToObservable(eff)))
+      .bindL("res", ({ obs, counters }) =>
+        T.asyncTotal(cb => {
+          obs.subscribe(
+            n => {
+              counters.values.push(n);
+            },
+            e => {
+              counters.errors.push(e);
+              // cb("ok");
```
mikearnaldi: ?

packages/rxjs/tests/RxJS.test.ts
```diff
@@ -210,3 +210,51 @@ describe("RxJS", () => {
...
+
+  it("effectToObservable returns value", async () => {
+    const counters = await T.runToPromise(test(T.pure("a")));
+    assert.deepEqual(counters.values, ["a"]);
+    assert.deepEqual(counters.errors, []);
+  });
+
+  it("effectToObservable returns error", async () => {
+    const counters = await T.runToPromise(
+      test(T.raiseError("err") as any) // forces accepting errors with any
```
mikearnaldi: ?

packages/rxjs/tests/RxJS.test.ts
sledorze: sorry there was another push
(was fixed meanwhile)

packages/rxjs/tests/RxJS.test.ts
sledorze: Signature prevent to make Observable from failing Effect so that cast is to force Errors that are not allowed by the type checker (like execution errors)

sledorze: @mikearnaldi  May I write a throw?

mikearnaldi: everytime you throw you end up in hell...

mikearnaldi: an execution error will never be catched properly to be propagated here, it will resolve to an abort in the fiber

sledorze: ok so you're saying no T.raiseError would ever be generated so no reason for this test as it test an impossible case, right?

sledorze: make sense, will remove it

mikearnaldi:
Isn't all of this actually: `Rx.from(T.runToPromise(eff))` ?

sledorze:
@mikearnaldi 
The first I did would run the effect for each new subscriber of the observable.
This last one will run the effect Once.

Issue #65, opened 2019-12-06
Design of HTTP client


mikearnaldi:
Basic points:
* support for all HTTP methods
* works in browser and node
* wraps response/request data in both success and error case
* supports typed request/response
* integrates with a generic tracing instrumentation
* websocket? 

Any ideas both on the API and on the features are open for discussion!

mikearnaldi:
Further points made by slack conversations:
- avoid wrapping libraries that are promise-based (will slow down everything due to promise performance)
- avoid wrapping any library with big footprint

Ideally build upon the platform api (xmlhttprequest in browser & http/https in node)

PR #67, opened 2019-12-07
HTTP Client & Libcurl implementation (to follow browser/fetch)

mikearnaldi:
tracks #65 

* interfaces exposed in http-client
* libcurl implementation in http-client-libcurl (supports http 1.0/1.1/2 and all that libcurl supports)
* fetch based for browser
* whatwg-fetch based for isomorphic


packages/http-client/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  return "headers" in r;
+}
+
+export function request<R, I, E, O>(
+  method: Method,
+  url: string,
+  body?: I
+): T.Effect<Http & R, HttpError<E>, Response<O>> {
+  return T.accessM(({ http }: Http) =>
+    T.accessM((r: R) => {
```
sledorze: Why do we need two accessM ? can't we only do one?

packages/http-client-libcurl/src/index.ts
```diff
@@ -0,0 +1,124 @@
...
+        const reqHead = [
+          "Content-Type: application/json",
+          ...pipe(
+            headers,
+            R.collect((k, v) => `${k}: ${v}`)
+          )
+        ];
+
+        req.setOpt("URL", url);
+        req.setOpt("CAINFO", certfile);
```
sledorze: Pondering if should not be a dep.. :)

packages/http-client-libcurl/src/index.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+function getResponse<A>(
+  statusCode: number,
+  bodyStr: string,
+  headers: Buffer | C.HeaderInfo[]
+): H.Response<A> {
+  return {
+    status: statusCode,
+    body: parseOrUndefined(bodyStr),
```
sledorze: JSON only ?

packages/http-client/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+function hasHeaders(r: any): r is HttpHeaders {
+  return "headers" in r;
+}
+
+export function request<R, I, E, O>(
+  method: Method,
+  url: string,
+  body?: I
+): T.Effect<Http & R, HttpError<E>, Response<O>> {
+  return T.accessM(({ http }: Http) =>
```
sledorze: Can't we only have one `accessM` ?

packages/http-client-libcurl/src/index.ts
sledorze: Having it fed by env would not be nice from a DX point of view, okish

packages/http-client-libcurl/src/index.ts
sledorze: only support for Json (no TEXT then), should me made explicit

mikearnaldi: this json bias is temporary I plan to extract serde out and support more, it was a quick one to develop the api first

packages/http-client/src/index.ts
mikearnaldi: probably

packages/http-client-libcurl/src/index.ts
mikearnaldi: as below temporary

packages/http-client-libcurl/src/index.ts
mikearnaldi: this is to be though there are pro and cons of shipping the bundle I think we should allow consumer to overwrite it but we should also ship it given libcurl doesn't integrate well with systems

mikearnaldi: made at least overridable

mikearnaldi: wouldn't be against making it env but I think param is more appropriate because it is really relative to the libcurl implementation and it is not a generic http thing, for example the browser implementation will not have it

packages/http-client-libcurl/src/index.ts
mikearnaldi: extracted generic deserializer to env

packages/http-client-libcurl/src/index.ts
```diff
@@ -0,0 +1,158 @@
+import { effect as T } from "@matechs/effect";
+import * as H from "@matechs/http-client";
+import * as E from "fp-ts/lib/Either";
+import { pipe } from "fp-ts/lib/pipeable";
+import * as R from "fp-ts/lib/Record";
+import * as C from "node-libcurl";
+import path from "path";
+import querystring from "querystring";
+
+function isJson(requestType?: H.RequestType) {
```
DenisFrezzato: This function is missing the return type.

packages/http-client-libcurl/src/index.ts
```diff
@@ -0,0 +1,158 @@
...
+      )
+  }
+});
+
+function customReq<I>(
+  method: string,
+  req: C.Curl,
+  body?: I,
+  requestType?: H.RequestType
+) {
```
DenisFrezzato: This function is missing the return type.

packages/http-client/src/index.ts
```diff
@@ -0,0 +1,202 @@
...
+      method: Method,
+      url: string,
+      headers: Record<string, string>,
+      body?: I,
+      requestType?: RequestType
+    ) => T.Effect<HttpDeserializer, HttpError<E>, Response<O>>;
+  };
+}
+
+function hasHeaders(r: any): r is HttpHeaders {
```
DenisFrezzato: This doesn't look type safe, if this function could actually take a parameter of any type, consider this type safe implementation:
```typescript
function hasHeaders(r: unknown): r is HttpHeaders {
  return typeof r === 'object' && !!r && "headers" in r;
}
```

mikearnaldi: this takes an environment that is always defined, worst case is {}

mikearnaldi: still I prefer your signature.

PR #71, opened 2019-12-10
fromEffect does not rely on Promise

sledorze:


mikearnaldi:
closing in favor of #73

PR #72, opened 2019-12-10
add _E and _A inhabitants to Effect

sledorze:


mikearnaldi:
already in https://github.com/mikearnaldi/matechs-effect/commit/aaf5a86574176f156a48983ec7d67573f6320d64

Issue #95, opened 2019-12-28
Multitenant Logger


mikearnaldi:
Implement a base logger using freeEnv and provide an implementation backed by winston

PR #96, opened 2019-12-28
Logger

mikearnaldi:


mikearnaldi:
provides module for #95 winston impl to follow

PR #132, opened 2020-02-17
Enhance interrupt expressivity

mikearnaldi:


mikearnaldi:
better approach in #134

PR #142, opened 2020-03-01
upgrade cqrs-es peerDependencies of @matechs/logger

sledorze:


sledorze:
The problem was during compilation (even on fresh repo):

lerna ERR! yarn run build exited 1 in '@matechs/cqrs-es'
lerna ERR! yarn run build stdout:
yarn run v1.21.0
$ rimraf lib && tsc --build tsconfig.build.json
src/index.ts(35,14): error TS2742: The inferred type of 'aggregate' cannot be named without a reference to '@matechs/cqrs/node_modules/@matechs/logger/lib/logger'. This is likely not portable. A type annotation is necessary.
src/index.ts(35,14): error TS2742: The inferred type of 'aggregate' cannot be named without a reference to '@matechs/cqrs/node_modules/@matechs/orm'. This is likely not portable. A type annotation is necessary.
src/offset.ts(26,14): error TS2742: The inferred type of 'ormOffsetStore' cannot be named without a reference to '@matechs/cqrs/node_modules/@matechs/orm'. This is likely not portable. A type annotation is necessary.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

mikearnaldi:
Arguably this doesn't change anything as the version updated is on the same range as the latest. Also it is building fine in ci, do you have a repro?

sledorze:
Yes the repro was local compilation (even on a fresh project).
With that change, no more compilation error.
I can understand why there was a failure, I don't understand why it was only happening locally.

sledorze:
Hyp: The signature was exporting a version of a lib (from deps) that was not matching it's peerdeps and needed to name it differently.
By converging those versions it solves the issue.

sledorze:
@mikearnaldi ok to merge?

mikearnaldi:
Without a repro no. The change will be useless as version update will break it again

mikearnaldi:
No Repro on Fresh Install.
```
lerna success run Ran npm script 'build' in 22 packages in 78.3s:
lerna success - @matechs/apollo
lerna success - @matechs/browser
lerna success - @matechs/console
lerna success - @matechs/cqrs-es
lerna success - @matechs/cqrs
lerna success - @matechs/effect
lerna success - @matechs/epics
lerna success - @matechs/express
lerna success - @matechs/fancy
lerna success - @matechs/graceful
lerna success - @matechs/http-client-fetch
lerna success - @matechs/http-client-libcurl
lerna success - @matechs/http-client
lerna success - @matechs/logger-winston
lerna success - @matechs/logger
lerna success - @matechs/orm
lerna success - @matechs/rpc-client
lerna success - @matechs/rpc
lerna success - @matechs/rxjs
lerna success - @matechs/tracing
lerna success - @matechs/uuid
lerna success - @matechs/zoo
Done in 78.91s.
```

sledorze:
for the record, was a problem with my way to use lerna;
was calling `lerna link` which should not be called in a monorepo setting

PR #190, opened 2020-04-09
Add Koa support

patroza:
- Koa support similar to Express, but with `withRouter` and `withSubRouter` support
- Enhanced demo compared to Express, by using a `randomMessage` module
- Express demo: Use `../src` import so that live code is used.

## Thoughts

For follow-up?

- What would be needed to make the DSL more protective.
   ie:
    - `withRouter` should not be called within another `withRouter` call. Instead one should call `withSubRouter`
Or is it fine that someone may compose an invalid program? - well invalid from a certain perspective.


mikearnaldi:
closing in favour of #192

PR #195, opened 2020-04-11
Matechs Prelude

mikearnaldi:
Prelude is the next generation API of effect, it aims to simplify the user-land surface as much as possible cutting the underlying jargon.

It is supposed to support smoothly multiple effect type signatures tracking sync/async failability/non-failability.

It should include default exports of the basic components from fp-ts to smooth adoption


mikearnaldi:
Hit a wall with dynamic retyping, will probably move to 2 variants of effect , one fully sync & one classic having the classic extend the sync

PR #196, opened 2020-04-12
Streams: takeUntil 

rjdestigter:
Implementation of _takeUntil_ for streams:

ToDo:
- [x] Update documentation
- [x] Tests

### DIFF
--- a/packages/effect/src/stream/index.ts
+++ b/packages/effect/src/stream/index.ts
@@ -444,6 +444,10 @@ function filter_<R, E, A>(stream: Stream<R, E, A>, f: F.Predicate<A>): Stream<R,
   );
 }
 
+export function filter<A, B extends A>(
+  f: F.Refinement<A, B>
+): <R, E>(s: Stream<R, E, A>) => Stream<R, E, B>;
+export function filter<A>(f: F.Predicate<A>): <R, E>(s: Stream<R, E, A>) => Stream<R, E, A>;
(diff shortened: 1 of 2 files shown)


rjdestigter:
I'm working on tests but I'm having trouble getting the current tests to run:

image

mikearnaldi:
@rjdestigter you need to build first

rjdestigter:
What are the build and test commands @mikearnaldi ?

mikearnaldi:
- lerna run build (build)
- yarn jest (test)

mikearnaldi:
they are quite long to run if you want to scope it only to effect package you can:
- yarn jest packages/effect

packages/effect/src/stream/index.ts
```diff
@@ -444,6 +444,10 @@ function filter_<R, E, A>(stream: Stream<R, E, A>, f: F.Predicate<A>): Stream<R,
   );
 }
 
+export function filter<A, B extends A>(
```
rjdestigter, line 447: I also added a definition to _filter_ allowing you to refine the type of output a stream produces. To see an example, it is used on line `919` of this file where a stream of `Wrapped ` values is refined to a stream of `WrappedStream ` values.

mikearnaldi: This is nice, there was a variant in here:

https://github.com/Matechs-Garage/matechs-effect/blob/master/packages/effect/src/stream/index.ts#L458

But this overloading is better

PR #198, opened 2020-04-17
Bindings for pino logger

qlonik:
I tried implementing bindings for pino logger.  In this package there are implementations for two types of usages:
* With error/warn/... functions defined in `@matechs/logger`
* With functions more idiomatic to pino, such as fatal/error/debug/info

In pino api, each of the log functions has overloaded signature:
```ts
function log(o: object, msg?: string, ...rest: unknown[]): void;
function log(msg: string, ...rest: unknown[]): void;
```

I'm not sure if effect allows for overloaded functions, but I implemented those overloads as two separate functions `fatal` and `fatalS` and others. If is possible to merge those, it would be nice. 

I ran out of time and didn't have time to finish the test. Also, I dont know how to run tests, so I was not able to properly finish it. When I try `yarn jest package/pino/tests/Pino.test.ts`, it says `Cannot find module '@matechs/effect' from 'Pino.test.ts'`.

### DIFF
--- a/packages/apollo/package.json
+++ b/packages/apollo/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@matechs/apollo",
-  "version": "2.0.0-beta.0",
+  "version": "2.0.0-beta.1",
   "license": "MIT",
   "private": false,
   "repository": "https://github.com/mikearnaldi/matechs-effect.git",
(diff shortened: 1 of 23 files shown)


mikearnaldi:
@qlonik before running tests you need to build the project

qlonik:
I forgot I had pull bot setup for forks. I think it messed up this pull request. I disabled that bot for this repository.

@mikearnaldi  Could you please reopen it and I will keep working on pino bindings?

PR #215, opened 2020-04-25
Par zip managed

qlonik:
Zips two managed in parallel. Not sure if this function is needed in the library, but at least there are extra overloads to sequenceT and sequenceS allowing it to work with Effect Validation.


mikearnaldi:
This implementation is not parallel, also none of the parallel implementations of effect uses the validation applicative (that is sequential)

mikearnaldi:
Also, this is buggy in case of an error happening during acquisition of either the first or second resource there is no release on the other

PR #228, opened 2020-05-03
Fork all

qlonik:
Added comments and tests for `forkAll`. Does this look good?

### DIFF
--- a/packages/effect/src/effect.ts
+++ b/packages/effect/src/effect.ts
@@ -5,6 +5,7 @@ import {
   task as TA,
   taskEither as TE,
   array as Ar,
+  nonEmptyArray as NEA,
   semigroup as Sem,
   monoid as Mon,
   tree as TR,
@@ -12,6 +13,7 @@ import {
 } from "fp-ts";
 import { pipeable, pipe } from "fp-ts/lib/pipeable";
 import * as ex from "./original/exit";
+import { fold as exFold } from "./exit";
 import { Runtime } from "./original/runtime";
 import {
   Monad4E,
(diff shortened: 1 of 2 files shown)


mikearnaldi:
Few comments:
1: doesn't look correct in always returning the first error, please add a test where the first effect in the array passed into forkAll is T.never and the second is an error

2: I would prefer having the signatures varying E and R given the effects passed to forkAll may have different environments & errors

mikearnaldi:
Additionally it doesn't seems to cancel when something fails

mikearnaldi:
@qlonik is this actually a sub-case of https://github.com/Matechs-Garage/matechs-effect/pull/229 ?

qlonik:
It does look very similar to the `parFastZipWith` in #229. The only difference is that this zips fibers and returns the fiber and 229 zips effects and return an effect. But it is possible to fork effect returned from parFastZipWith.

Yep, it doesn't cancel other effects if one of them fails. But does parZipWith cancel the other one when first fails?

In either case, it seems that this functions could be implemented simpler if it is done with the use of parZipWith or parFastZipWith. So, I'm not exactly sure this function is needed anymore?

_very_ rough code:
```typescript
export function forkAll<S, R, E, A>(ios: Effect<S, R, E, A>[]): SyncR<R, Fiber<E, A[]>> {
  const arrMon = Ar.getMonoid<A>();
  const x = pipe(ios, Ar.map(map((x) => Ar.of(x))));
  const y = Mon.fold<AsyncRE<R, E, A[]>>({
    empty: pure(arrMon.empty),
    concat: (fa, fb) => parZipWith(fa, fb, arrMon.concat)
  })(x);

  return fork(y);
}

export function forkAllFast<S,R,E,A>(ios: Effect<S,R,E,A>[]):SyncR<R,Fiber<E,A[]>> {
  const arrMon = Ar.getMonoid<A>();
  const x = pipe(ios, Ar.map(map((x) => Ar.of(x))));
  const y = Mon.fold<AsyncRE<R, E, A[]>>({
    empty: pure(arrMon.empty),
    concat: (fa, fb) => parFastZipWith(fa, fb, arrMon.concat)
  })(x);

  return fork(y);
}
```

mikearnaldi:
Fast cancel the rest at first failure, it is implemented for that reason

mikearnaldi:
`const forkAll = flow(T.parFastSequenceArray, T.fork)` and it also has proper cancel, I don't think it is needed in the library as it's a simple combinator, if you want to add more fiber specific utilities it might go in the fiber module but until that is done we can safely close this.

qlonik:
Yea, this is a good idea. No need to pollute the library with each 2 function compositions. :)

PR #240, opened 2020-05-06
Add lint to pre-commit

patroza:


mikearnaldi:
this is too slow, it lints all not only the changed

Issue #273, opened 2020-06-16
Align effect.filterOrElse with either.filterOrElse


jsmacka:
**Is your feature request related to a problem? Please describe.**
`filterOrElse` seems to have a different logic for `Effect` and `Managed` than it does for `Either`. It would be less confusing if that was not the case. 

**Describe the solution you'd like**
When the predicate returns true I would expect to end up in the successful rather than the unsuccessful case.

In core/src/Effect/index.ts I would like to see:
```
export const filterOrElse: {
  <E, A, B extends A>(refinement: Refinement<A, B>, onFalse: (a: A) => E): <S, R>(
    ma: Effect<S, R, E, A>
  ) => Effect<S, R, E, B>
  <E, A>(predicate: Predicate<A>, onFalse: (a: A) => E): <S, R>(
    ma: Effect<S, R, E, A>
  ) => Effect<S, R, E, A>
} = <E, A>(predicate: Predicate<A>, onFalse: (a: A) => E) => <S, R>(
  ma: Effect<S, R, E, A>
): Effect<S, R, E, A> =>
  chain_(ma, (a) => (predicate(a) ? completed(done(a)) : completed(raise(onFalse(a)))))
// instead of chain_(ma, (a) => (predicate(a) ? completed(raise(onFalse(a))) : completed(done(a))))
```

and in core/src/Managed/index.ts:
```
export const filterOrElse: {
  <E, A, B extends A>(refinement: Refinement<A, B>, onFalse: (a: A) => E): <S, R>(
    ma: Managed<S, R, E, A>
  ) => Managed<S, R, E, B>
  <E, A>(predicate: Predicate<A>, onFalse: (a: A) => E): <S, R>(
    ma: Managed<S, R, E, A>
  ) => Managed<S, R, E, A>
} = <E, A>(predicate: Predicate<A>, onFalse: (a: A) => E) => <S, R>(
  ma: Managed<S, R, E, A>
): Managed<S, R, E, A> =>
  chain_(ma, (a) =>
    predicate(a)
      ? encaseEffect(T.completed(done(a)))
      : encaseEffect(T.completed(raise(onFalse(a))))
  )
```

Compare to core/src/Either/index.ts:
```
export const filterOrElse: {
  <E, A, B extends A>(refinement: Refinement<A, B>, onFalse: (a: A) => E): <E2>(
    ma: Either<E2, A>
  ) => Either<E | E2, B>
[...]

PR #274, opened 2020-06-17
#273  Add some tests and fix issue (breaking)

jsmacka:
This commit changes `filterOrElse` for `Managed` and `Effect` such that the logic agrees with `filterOrElse` for `Either`: When the predicate returns true, we continue in the successful branch in all three cases, otherwise we switch to the error branch.


jsmacka:
I hope it's ok to simply open a pull request - not so familiar with the local customs on git. Feedback is very welcome :)

mikearnaldi:
Yes, it's always welcome! thanks

PR #284, opened 2020-07-17
Publish source maps

qlonik:
I'm working on bringing source maps in the published module. I realized that in order to have correct jump to definition and debugging to work, the sources of the library need to be included. The set of scripts added in this pr includes the original source files without modifying the structure of the published package (so all imports like `from '@matechs/core/Either';` should still work).

Since the `.js` files are unchanged, compared to all previous versions, I'm guessing that bundlers would not care about original source files and will completely skip bundling them into the final application. So it shouldn't affect the bundle size of compiled apps. 

Additionally, with these changes, it is not possible to see how various methods are implemented internally, and so the effect will act as a sort of documentation of itself. As well as it shows the original source during the debug session, allowing to learn inner workings of the library.

@mikearnaldi What do you think about including source code in the published package?

EDIT: also, for now only core has proper source maps enabled. Other packages might have source maps generated, but those are broken, since they don't link correctly


mikearnaldi:
I think we can include the maps I don't think they will end up in bundles

mikearnaldi:
keep in mind `@matechs/effect/either` is needed for bundlers to properly index es6/commonjs

qlonik:
> 
> 
> keep in mind `@matechs/effect/either` is needed for bundlers to properly index es6/commonjs

I'm not sure what you mean here. Is it `Either` from `@matechs/core` or is it something else? Could you clarify, please?

mikearnaldi:
the import `@matechs/core/Either` is mapped to `/Either` where there is a `package.json` indexing both the commonjs and the es6 modules

qlonik:
Oh yes, I understand. The structure of `build/` dir is still the same as before these changes: there is `build/Either` folder with package.json pointing to cjs and to es6 versions:
```json
{
  "sideEffects": false,
  "main": "./index.js",
  "module": "../esm/Either/index.js",
  "typings": "./index.d.ts"
}
```

This pr places .d.ts.map and .js.map files beside each of those and adds source files in the separate folder. I tried putting source files in the same folder, but then tsc compiler gets confused with the import.

qlonik:
Is it possible to publish preview of these changes? This would allow us to ensure it works before merging.

mikearnaldi:
not really, I have publish integrated in the ci and have found no way to get preview working properly. Usually I use a local registry to do this type of test

mikearnaldi:
I've added a babel step to auto-add __PURE__ comments , does it break this in any way? The reason being is writing manually PURE comments is noisy and we definitely need to pay attention to have those comments everywhere in top-level call sites (the "next" is already quite fat by default, around 80k, because of fiber tracing & additional functionality)

qlonik:
When babel is executed without `-s` flag (which enables source maps) then babel strips out source maps comments. When I add that flag, everything seems alright. I'm going to rebase and bring those changes into this merge.

Out of curiosity: why is there 2 different modules for pure annotations? It seems only one is being used. Can we get rid of the other?

mikearnaldi:
yes we can it's a leftover

PR #289, opened 2020-07-24
Tracing error tags

qlonik:
* Adjusted tags
* added extra handling of Error in JSON.stringify
* simplified some code (removed unneeded Dos)
* moved functions around


qlonik:
I also wanted to add another function to public api which allows to add extra tags into span. However, I'm not clear on one moment in that code, so I will submit it as a next pr.

PR #290, opened 2020-07-25
Allow adding extra tags into span

qlonik:
This allows addition of extra tags to span.

I'm uncertain about current implementation. Currently, the code first adds tags into span and then executes the target effect:

```typescript
pipe(
  T.sync(() => {
    const span = r[SpanContext].spanInstance
    for (const [k, v] of Object.entries(extraTags)) {
      span.setTag(k, v)
    }
  }),
  T.apSecond(ma)
)
```

Should this be other way around?

Also, it is unclear to me to which span the tags will apply between these two cases:
```typescript
pipe(
  T.unit,
  withExtraTags({ name: "hello" }),
  withChildSpan("operation")
)
```
and 
```typescript
pipe(
  T.unit,
  withChildSpan("operation"),
  withExtraTags({ name: "hello" })
)
```

### DIFF
--- a/packages_be/tracing/src/Tracing/index.ts
+++ b/packages_be/tracing/src/Tracing/index.ts
@@ -43,6 +43,9 @@ export interface TracerOps {
   withChildSpan(
     operation: string
   ): <S, R, E, A>(ma: T.Effect<S, R, E, A>) => T.Effect<S, R, E, A>
+  addSpanTags(
+    extraTags: Record<string, unknown>
+  ): <S, R, E, A>(ma: T.Effect<S, R, E, A>) => T.Effect<S, R, E, A>
 }
 
 export interface Tracer {
@@ -206,6 +209,17 @@ export const Tracer = (factory: T.Sync<OT> = T.sync(() => new OT())) =>
                     )
                   : ma
               )
+          },
+          addSpanTags(extraTags: Record<string, unknown>) {
+            return <S, R, E, A>(ma: T.Effect<S, R, E, A>): T.Effect<S, R, E, A> =>
+              T.accessM((r: R) =>
+                hasSpanContext(r)
+                  ? T.applySecond(
+                      T.sync(() => r[SpanContext].spanInstance.addTags(extraTags)),
+                      ma
+                    )
+                  : ma
+              )
           }
         }
       })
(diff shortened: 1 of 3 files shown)


mikearnaldi:
it looks like this will add to the parent, I agree it's not a clear implementation. The only thing you can do is access the current span so the api would be more like AddExtraTag and not wrap a child

qlonik:
I tried implementing both ways: `pipe(ma, T.apFirst(T.sync(() => /* addTag */ )))` and `pipe(T.sync(() => /* addTag */), T.apSecond(ma))`. It doesn't seem that there is a difference in which span tags are added, but it is slightly easier to do with apSecond approach, since apFirst might not execute addition of tags on error in ma, since ap is implemented in terms of chain.

However, it seems that tags need to be added into child before wrapping into `withChildSpan` call, or else tags will get added into parent. I'm not sure how improve api in order to make it more obvious other than by adding the comment about it.

mikearnaldi:
I don't think you are getting the point both apFirst & apSecond have access to the same env the only env you have access to is the "current" env (parent), you cannot add it to the child span because the environment of the child is only existing for the child, it is created inside "withChildSpan" and provided to "ma". The only way to add it to the child span would be to have "withChildSpan" be aware of tags, for example by adding a second optional argument to "withChildSpan" or create a separated function that also accepts tags or create a ChildTags environment that is fed from parent and consumed by "withChildSpan"

qlonik:
Yes, I understand that apFirst and apSecond have access to same current environment, which might or might not have child span.

Yes, `withChildSpan` creates the new span and provides it to ma, so that ma can set tags and when it is doing so, they will get added to that child span. I didn't mean to have an ability of passing tags created in parent span to child span.

With current implementation, if I have something like this (not real code, just mockup):

```typescript
const tracedEff = pipe(
  T.unit,
  T.liftDelay(100),
  addExtraTags({ tag: 'custom tag' })
)

const main1 = withControllerSpan(tracedEff)
const main2 = withControllerSpan(withChildSpan(tracedEff))
```

The execution of main1 will add the extra tag into parent span, but in main2 that extra tag will be in the child span.

mikearnaldi:
Have you thought about failure scenarios? we might still want the tag to apply even if "ma" fails

qlonik:
Yea, i quickly checked: applySecond/applyFirst are implemented in terms of chain, so it executes first and if it succeeds, it executes second. So if I use applyFirst anywhere like so:

```typescript
const x = applyFirst(ma, T.sync(() => { /* set tags */ }))
```

Then if ma fails, then tags wont get set, since ma will be executed and then tags will be chained after ma.

But if I use applySecond, like so:

```typescript
const x = applySecond(T.sync(() => { /* set tags */ }), ma)
```

Then tags will get set first and as long as setting tags doesn't fail, ma will get chained after tags set and executed. 


To be fair though, I think it is a good idea to add a test case for this scenario, when ma fails. I will add it later tonight.

mikearnaldi:
targeting legacy

PR #293, opened 2020-08-02
Fix record

qlonik:
Addressing the same issue as #292 

I added test case to verify this issue. It seems that there is some convention about how test files are named and where they go. Could you adjust the test case file that I've added in the way that you prefer?


qlonik:
Oops, I forgot to rename interfaces in Overloads. I'm going to push that change too.

PR #296, opened 2020-08-05
guard against nullable value

qlonik:
I noticed that trace sometimes causes an unexpected crash because of an unchecked property access on `undefined`/`null` values


mikearnaldi:
Nice catch!

PR #299, opened 2020-08-14
Allow effectful jest mock acquisition

qlonik:
I'm not sure why mock acquisition was limited to only be `IO`. This extends it to be any effect.


qlonik:
I'm guessing this would be a breaking change?

mikearnaldi:
mmm, yes it's breaking but also correct

mikearnaldi:
targeting legacy

PR #310, opened 2020-08-29
Stream work

mikearnaldi:


packages/system/src/Managed/releaseMap.ts
```diff
@@ -8,6 +8,8 @@ export type Finalizer = (
   exit: T.Exit<any, any>
 ) => T.Effect<unknown, unknown, never, any>
 
+export const noopFinalizer: Finalizer = () => T.unit
```
mikearnaldi, line 11: dupe of https://github.com/Matechs-Garage/matechs-effect/blob/master/packages/system/src/Managed/managed.ts#L4

mikearnaldi:
@qlonik Looks very good, only a dupe on `noop` finalizer as mentioned, I would probably also add the comments from zio as the rest of the combinators and then we can merge this (to avoid getting master detached too much)

PR #311, opened 2020-08-31
Add handoff

mikearnaldi:


mikearnaldi:
Absurd should not return `never` otherwise the generic `A` is useless:
https://github.com/gcanti/fp-ts/blob/master/src/function.ts#L277

Await is duplicated (and also the implementation with do-while is weird):
https://github.com/Matechs-Garage/matechs-effect/blob/master/packages/system/src/Promise/wait.ts

mikearnaldi:
I now get why it's like that, in ZIO the compareAndSet is because of multi-threading

PR #312, opened 2020-08-31
BufferedPull pullArray

mikearnaldi:
I believe https://github.com/zio/zio/blob/master/streams/shared/src/main/scala/zio/stream/ZStream.scala#L4002 contains a bug, what is basically going on is when we call `update.as(chunk.drop(idx))` for the last time the `update` will fail (signaling end) de-facto losing the last element. In `pullElement` calling `update` is delegated to the subsequent call


mikearnaldi:
Confirmed: https://github.com/zio/zio/pull/4164

PR #321, opened 2020-10-07
Add Filterable and FilterableWithIndex typeclasses

0x706b:


packages/core/src/Prelude/Filterable/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    Q,
+    W,
+    X,
+    I,
+    S,
+    R,
+    E
+  >(
+    fa: HKT.Kind<F, C, N, K, Q, W, X, I, S, R, E, A>
+  ) => HKT.Kind<F, C, string, K, Q, W, X, I, S, R, E, B>
```
qlonik, line 31: Shouldn't this `string` be an `N` type?

mikearnaldi: Arguable, I do think N would be a better type here too but the concept of filter is to remove stuff, meaning this N would actually be oversized

mikearnaldi (changes requested):

packages/core/src/Prelude/Filterable/index.ts
```diff
@@ -0,0 +1,75 @@
+import type { Either } from "../../Classic/Either"
+import type { Option } from "../../Classic/Option"
+import type { Predicate, Refinement } from "../../Function"
+import type { Separated } from "../../Utils"
+import type { Compactable } from "../Compactable"
+import type { Covariant } from "../Covariant"
+import type * as HKT from "../HKT"
+
+export interface Filterable<F extends HKT.URIS, C = HKT.Auto>
```
mikearnaldi, line 9: Filterable should be a `type` that intersect its childs like `Filterable = Covariant & Filter & FilterMap & Partition`

mikearnaldi (changes requested):

0x706b:
I believe this should resolve the requested changes. I separated the functions out into their own folders, similar to `Foldable`, and changed the type of the `N` parameter in the return type from `string` back to `N` for the functions that needed it

mikearnaldi (approved):

mikearnaldi:
Perfect @0x706b thanks a lot!

mikearnaldi:
Reduced dependencies on Covariant & Compactable (not necessary to pack on filterable instance)

PR #322, opened 2020-10-18
implement generator for Stream effect

mattiamanzati:
As stated in https://github.com/gcanti/fp-ts/issues/261#issuecomment-711168155 , I came up with a way of replaying generators locally in order to support do notation for stream

### DIFF
--- a/packages/system/src/Stream/Stream/gen.ts
+++ b/packages/system/src/Stream/Stream/gen.ts
@@ -3,6 +3,8 @@ import type { _E, _R } from "../../Utils"
 import { chain_ } from "./chain"
 import { Stream } from "./definitions"
 import { succeed } from "./succeed"
+import { fromEffect } from "./fromEffect"
+import { die } from "../../Effect/die"
 
 export class GenStream<R, E, A> {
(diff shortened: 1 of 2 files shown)


packages/system/src/Stream/Stream/gen.ts
```diff
@@ -25,21 +27,24 @@ export function gen<Eff extends GenStream<any, any, any>, AEff>(
...
-    const state = iterator.next()
     function run(
-      state: IteratorYieldResult<Eff> | IteratorReturnResult<AEff>
+      replayStack: any[]
     ): Stream<any, any, AEff> {
+      const iterator = f(adapter as any)
+      let state = iterator.next()
+      for(let i = 0; i < replayStack.length; i++){
+        if(state.done){
+          return fromEffect(die("Something very wrong has happened. Replaying values resulted in a premature end of the generator execution. Provided generator should be pure and perform effects only by yielding them, so that the generator can safely be re-run without side effects."))
```
mikearnaldi, line 37: Better to make a dedicated Tagged Exception maybe in the GlobalExceptions module

PR #329, opened 2020-10-21
simplify some ReaderT transformation functions

qlonik:
Noticed that if I push return type of some transformers into `HKT.instance` generic, then I don't need to specify types explicitly. and in some places I was able to skip pipe and simplify even more (e.g. provide())

This doesn't fail in my editor, but not sure what CI will report.

Not sure, if there was some actual reason to have those types specified explicitly.


mikearnaldi:
Interesting catch I will have to understand why is that... BTW no no real meaning in adding the types except making it compile,

PR #332, opened 2020-10-22
filter added to system/stream

curiosityfoundation:
am not sure where to add a test. based on system/stream/mapConcat


packages/system/src/Stream/Stream/filter.ts
```diff
@@ -0,0 +1,17 @@
...
+import { mapChunks } from "./mapChunks"
+
+/**
+ * Applies the predicate to each element and allows passing elements 
+ * to reach the output of this stream.
+ */
+export const filter = <O>(f: Predicate<O>) => <R, E>(
+  self: Stream<R, E, O>
+): Stream<R, E, O> =>
+  pipe(
```
qlonik, line 14: Looks like this pipe is not necessary.

packages/system/src/Stream/Stream/filter.ts
```diff
@@ -0,0 +1,17 @@
...
+    self,
+    mapChunks((o) => Array_.chain_(o, (o) => (f(o) ? [o] : [])))
```
qlonik, line 16: It seems that it is possible to use curried version `Array_.chain` instead of uncurried one.

qlonik (commented):
Looks pretty good. You also need to reexport this method in one of the index files. Let's see what Michael says about it too.

packages/system/src/Stream/Stream/filter.ts
curiosityfoundation: happy to make this change - original thought was to model it after mapConcat which uses pipe and Array_.chain_

packages/system/src/Stream/Stream/filter.ts
mikearnaldi: we should use here the Array filter function as they do in zio:
https://github.com/zio/zio/blob/master/streams/shared/src/main/scala/zio/stream/ZStream.scala#L1245

mikearnaldi:
@curiosityfoundation thanks, I will merge it for now and patch in master directly. We are trying to keep implementations in line with ZIO as much as possible so I will change A.chain to use the A.filter as it is in https://github.com/zio/zio/blob/master/streams/shared/src/main/scala/zio/stream/ZStream.scala#L1245

Issue #337, opened 2020-10-22
Provide capability of fixing "A" in X.gen


mattiamanzati:
When making exensive use of X.gen (with X being anything supported), it works great thanks to its ability of automatically inferring the corret types of E and R. Unfortunately, as far as I know there is no syntax available to fix the return type and let the infer continue to work on the E and R type.

It would be nice to also have alongside `T.gen(function*(_){ ... })`, something like `T.genA<MyType>()(function*(_){ ... })`.

Without this syntax, alternatives are to manually put a "return bla as MyType", but when you have multiple return types that means multiple manual casts to perform.

What do you think about this feature?

mikearnaldi:
Would be useful yes

mikearnaldi:
we can do better:

```ts

// T.Effect<A & B, string, number>
const programNumber = T.gen<number>()(function* (_) {
  const a = yield* _(T.access((_: A) => _.a))
  const b = yield* _(T.access((_: B) => _.b))

  const c = a + b

  if (c > 10) {
    yield* _(T.fail(`${c} should be lower then x`))
  }

  return c
})

// T.Effect<A & B, string, number>
const programNumberFailString = T.gen<string, number>()(function* (_) {
  const a = yield* _(T.access((_: A) => _.a))
  const b = yield* _(T.access((_: B) => _.b))

  const c = a + b

  if (c > 10) {
    yield* _(T.fail(`${c} should be lower then x`))
  }

  return c
})

// T.Effect<A & B, string, number>
const programNumberFailStringR = T.gen<A & B & { foo: string }, string, number>()(
  function* (_) {
    const a = yield* _(T.access((_: A) => _.a))
    const b = yield* _(T.access((_: B) => _.b))

    const c = a + b

    if (c > 10) {
      yield* _(T.fail(`${c} should be lower then x`))
    }

    return c
  }
)
```

mikearnaldi:
@mattiamanzati is it left only the genF to do? there I was thinking only A would be useful to put in an overload

mattiamanzati:
Yep! That one was the only missing one I found

PR #338, opened 2020-10-22
Begin work on overloaded gen

mikearnaldi:
References issue #337


mikearnaldi:
@mattiamanzati I'll merge this batch to avoid getting it too out of sync, let's split the rest in a second

PR #366, opened 2020-11-01
preliminary work

mikearnaldi:


mikearnaldi:
impossible for combinators.

PR #368, opened 2020-11-02
Fix Async Fold

0x706b:
While reading through `Async`, I noticed the third parameter of `IFold` (`g`) was unused in `runInternal`, which was returning the success value from running `self`. Also changed the _asyncTag from `"ICatch"` to `"IFold"` for consistency


mikearnaldi:
@0x706b Thanks!

PR #396, opened 2020-11-18
Some stream methods

0x706b:


packages/system/src/Managed/methods/api.ts
```diff
@@ -2554,3 +2554,32 @@ export function mergeAllParN_(n: number) {
...
+    readonly apply: <R, E, A>(
+      ma: Managed<R, E, A>
+    ) => T.Effect<R, E, readonly [RM.Finalizer, A]>
+  ) {}
+}
+
+/**
+ * Creates a scope in which resources can be safely allocated into together with a release action.
+ */
+export function scope(): Managed<unknown, never, Scope> {
```
mikearnaldi, line 2574: this should probably be a constant

mikearnaldi:
Great work @0x706b !

PR #409, opened 2020-11-22
Some stream combinators

r-cyr:


packages/system/src/Stream/Stream/_internal/bufferSignal.ts
```diff
@@ -0,0 +1,80 @@
...
+              T.bind("p", () => P.make<never, void>()),
+              T.bind("added", ({ p }) => queue.offer([take, p])),
+              T.tap(({ added, p }) => T.when_(ref.set(p), () => added)),
+              T.asUnit
+            )
+        )
+
+      return pipe(
+        Take.fromPull(as),
+        T.tap((take) => offer(take as Ex.Exit<O.Option<E1>, A.Array<O1>>)),
```
r-cyr, line 52: I would need help with this one.

packages/system/src/Stream/Stream/collectWhile.ts
```diff
@@ -0,0 +1,75 @@
...
+import type { MutableArray } from "../../Mutable"
+import * as O from "../../Option"
+import * as Ref from "../../Ref"
+import * as Pull from "../Pull"
+import { Stream } from "./definitions"
+
+/**
+ * Maps an array until `none` is returned
+ */
+function collectWhileArray<X, Y>(
```
r-cyr, line 14: Is there something in the lib to do that? The original collectWhile implementation uses a PartialFunction.

mikearnaldi: I think this is actually missing but it is a good addition to the array module

PR #451, opened 2020-12-15
More steams combinators, up to TakeWhile

r-cyr:


packages/system/src/Stream/Stream/partition.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Partition a stream using a predicate. The first stream will contain all element evaluated to true
+ * and the second one will contain all element evaluated to false.
+ * The faster stream may advance by up to buffer elements further than the slower one.
+ */
+export function partition_<R, E, O>(
+  self: Stream<R, E, O>,
+  p: (o: O) => boolean,
+  buffer = 16
+): M.Managed<R, E, readonly [Stream<unknown, E, O>, Stream<unknown, E, O>]> {
```
mikearnaldi, line 16: given `partitionEither` has `Managed<R, never, ...>` shouldn't this be like `M.Managed<R, never, readonly [Stream<unknown, E, O>, Stream<unknown, E, O>]>`

Also tuples should be flagged readonly

mikearnaldi (changes requested):

packages/system/src/Stream/Stream/partition.ts
r-cyr: Good catch, and it's actually the inferred type... :/

PR #461, opened 2020-12-20
Call onError on promise rejection

### LINKED ISSUE #460: As.promise onError not working as intended?
First of all, thanks for the great package!

I'm trying to build something from the [react example](https://github.com/Matechs-Garage/effect-ts-react), but I'm facing some issues with As.promise. I'm trying to map rejections to some Error type, and the type inference tells me that the Failure case can only contain my type afterwards. However, the actual rejections bubble up and when I match the Exit on Failure<my error type>, it contains a TypeError (that was thrown from Browser fetch originally) and not my error type.

I've added a [test case](https://github.com/pschyska/effect-ts/commit/b3f6ae1e015866b3f0df385776b54beaf739f1aa) that shows the problem. The inferred type here is `As.Exit<string,never>` as expected, but the contained value is actually `Error("err")` or `rejected` and not `mapped u`.

Am I holding it wrong?

Cheers

pschyska:
This would fix #460. Not 100% sure if that would be the right place to do it.

### DIFF
--- a/packages/core/test/async.test.ts
+++ b/packages/core/test/async.test.ts
@@ -41,4 +41,28 @@ describe("Async", () => {
       )
     ).toEqual(As.successExit([1, 2, 3]))
   })
+
+  it("onError", async () => {
+
+    const throwP = async () => {
+      throw new Error("err");
+    }
+    const rejectP = async () => Promise.reject("reject");
+
+    expect(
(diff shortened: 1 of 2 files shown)


mikearnaldi:
Perfect place to do it, thanks!

PR #489, opened 2021-01-09
Selective tracing

mikearnaldi:


qlonik (approved):
I guess I'm a bit confused about the change set. I thought that all user functions need to be wrapped into tracer code and all exposed public functions should trace user functions, since we don't know which functions they will use. Why do we no longer need to wrap each function into traceAs?

packages/tracing-plugin/src/tracer.ts
```diff
@@ -39,6 +39,8 @@ export default function tracer(
   const tracingOn = !(_opts?.tracing === false)
   const checker = _program.getTypeChecker()
 
+  checker
```
qlonik, line 42: accidental change?

qlonik:
Also, looks like 2 out of 3 of changes in #488 do not apply here anymore

packages/tracing-plugin/src/tracer.ts
mikearnaldi: accidental yes

packages/tracing-utils/src/index.ts
```diff
@@ -1,25 +1,47 @@
 import { isTracingEnabled } from "./Global"
 
-export function traceAs<F extends Function>(g: any, f: F): F {
-  if (g["$trace"] && isTracingEnabled()) {
-    f["$trace"] = g["$trace"]
+let tmp: any
```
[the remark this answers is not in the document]
mikearnaldi: it's a global variable to hold up the trace for the `traceCall` method, fundamentally we have 2 types of traces, one attached to function arguments and one to trace the call-site of a function and given there is really no way to pass implicit arguments I used a global variable so it's like `set global variable -> run function (that will collect variable) -> cleanup`

mikearnaldi:
> I guess I'm a bit confused about the change set. I thought that all user functions need to be wrapped into tracer code and all exposed public functions should trace user functions, since we don't know which functions they will use. Why do we no longer need to wrap each function into traceAs?

So `traceAs` is a re-tracing mechanism it copies over the trace from one function to the other, it is used in places like `map` to trace the continuation of the chain that is internally used, ideally you want to use it to trace effects as the caller rather then exposing the internal (similar to zio's ZIOFn)

The functions that need tracing are definitely not all, basically, we only need traces on those functions that will appear as a continuation on the fiber, the old plugin was adding traces to everything. 

This change-set uses hints defined as tags in the methods that says what should contain a trace, the downside is for example `T.chain(myMethod)` will be traced pointing to `T.chain(***here***myMethod)` rather than at the definition site of `myMethod` it's basically tracing usage instead of definition.

The above is not a problem in my view as traces are really used only in case of errors and it is enough to detect the usage-site of things especially because you use the constructors directly, e.g: `T.chain(myFn)` is a thing I see a use for but `T.effectAsync(myFn)` most of the time doesn't make sense as you want to inline `myFn` like `T.effectAsync((cb) => {...})` meaning the definition and usage site are the same.

PR #490, opened 2021-01-10
Persistent Hashmap

mikearnaldi:


qlonik:
I was thinking: it would be cool to implement basic functionality only using basic JS types and provide it as a standalone library (but part of effect-ts ecosystem) and then add extra methods/functions integrating Options/Eithers and such from the rest of `system` package. What do you think?

mikearnaldi:
unsure the basic option data-type is kind of central and still don't know if I am going to use some other components mainly for optimizations, let's see where it leads and how many data types we can build persistent (I am not opposed though just unsure)

PR #536, opened 2021-02-03
Broadcast bug

mikearnaldi:


mikearnaldi:
To check in zio (probably it over-emits due to an off by one in range)

mikearnaldi:
Thanks to @r-cyr for the catch!

PR #576, opened 2021-02-26
[WIP] Adding a cahedInvalidate combinator to cache an effect with the…

mikearnaldi:
… possibility to invalidate the cache before timeout expiry


remiguittaut:
You mean that my commit message was too long? ^^ Haven't had time to continue. I'll see that later

PR #624, opened 2021-03-20
Move type classes

qlonik:
Tentatively move things around. 

1. Associative, Closure, Commutative, Identity and Inverse are moved into `Structure/` since they represent algebraic structures. (Another suggestion was to name this folder as `Structural/`)
2. Bounded, Equal, Ord and Show are moved into `Prelude/` since they are type-classes similar to Covariant, Filterable, Applicative and others. Another idea was to place them into `Structure/` or `Structural/` since they define structure for various data types like string, Array and others.
3. EitherT, IndexedT, InvariantT, OptionT, ReaderT and StateT are moved into `Transformer/`. There is also a possibility to remove `T` from their name, although it seems like a pretty well established convention to distinguish them from actual data types. Additionally, we could move XReaderT together into `Transformer/` or to create another folder `XPure/Transformer/`, specifically for XPure transformers.


mikearnaldi:
Closing, we may re-evaluate the structure post HKT update to the new encoding

PR #643, opened 2021-04-03
Merge tracing

qlonik:
The resolution is a bit hacky and adds special behavior when it detects package named '@effect-ts/system', 'effect-ts' and '@effect-ts/core', but the entire tracing is a bit hacky, so hopefully okay :) 

EDIT: I also tested it locally by enabling this version of tracing on a local project. seems to work


qlonik:
I realized that packages can import from themselves via `@effect-ts/system`, so I simplified resolution a bit and removed relative imports.

qlonik:
Note, I also modified tests such that they run against real sources now instead of built files. It additionally allows us to run tests in parallel with build and type check.

packages/tracing-plugin/src/tracer.ts
```diff
@@ -53,6 +56,25 @@ export default function tracer(
...
+  const packageJsonDir = findUp(programDir, "package.json")
+
+  if (packageJsonDir) {
+    try {
+      const packageJsonContent = fs.readFileSync(
+        path.join(packageJsonDir, "package.json"),
+        "utf-8"
+      )
+      const obj = JSON.parse(packageJsonContent)
+      if (typeof obj === "object" && INTERNAL_PACKAGES.includes(obj.name)) {
```
qlonik, line 70: Potentially, name could be a false positive match for some private packages. We could potentially, instead of checking for the name of the package, see if there is a config field set to specify that it is an effect internal package. It would be kinda of "hidden feature" - that is if someone set's that option in package.json, they will shoot themselves in a foot and break their own traces.

qlonik:
weird, jobs are failing now. I'll fix it tomorrow.

mikearnaldi:
> @effect-ts/system

Packages can self-reference but those self-references will end up in the built where relative paths should be instead

packages/tracing-plugin/src/tracer.ts
mikearnaldi: I don't get this, there are only 3 cases:
1) system => needs to be local paths
2) core => needs to reference system
3) any other need to reference core

qlonik:
What is the negative of packages self-referencing in the built files?

packages/tracing-plugin/src/tracer.ts
qlonik: That's what I thought too, but I had a bug with tests.

In short, tests run from top level folder, where the package name is `'effect-ts'`, so the import that's being used in tests is `@effect-ts/core/Tracing`. The paths in tsconfig.json would reference built files rather than real sources, so then there would actually be 2 different instances of `Tracing/index.ts` files being used, and so one trace was being lost (the one in tuple test cases). I'm guessing that since tests are now linked against real source, we can actually return to this behaviour and resolve to `@effect-ts/core/Tracing` even for tests and it should work. I could try that.

mikearnaldi:
@qlonik can break if you have aliases locally

packages/tracing-plugin/src/tracer.ts
qlonik: I changed it, so now in system it should be all relative paths. I also removed exception in tests, so it now links against `@effect-ts/core`. This seems to work locally. But it means that in jest configuration we have to link against `src` rather than `build`, which is good in my opinion anyway.

mikearnaldi:
We should really weigh benefits against price here, I don't see what we gain doing this and on the other side we end up with a more complex implementation that may have edge cases, i.e the findUp already looks suspicious and the test config changes too.

qlonik:
Yea, findUp is a bit weird, but usually tsconfig.json is in the same folder as package.json, so it should be fine.

As for the tests, I think it is a better idea to run tests against the source code. It runs faster and allows us to run tests earlier - before the build is done. The change to run tests against the source could be done in the separate pr anyway.

mikearnaldi:
I backported tests configs, unsure about the rest it seems to be too tricky to rely on

qlonik:
Okay. I changed it so it is specified via "hidden" config. No more findUp and reading fs.

PR #669, opened 2021-04-24
Stream combinators (Stage 1)

r-cyr:


packages/system/src/Experimental/Stream/_internal/api.ts
```diff
@@ -0,0 +1,883 @@
...
+  C
+>(
+  self: Stream<R, E, A>,
+  sink: SK.Sink<R1, E1, A1, E2, A1, B>,
+  schedule: SC.Schedule<R2, O.Option<B>, C>
+): Stream<R & R1 & R2 & Has<CL.Clock>, E2, E.Either<C, B>> {
+  type HandoffSignal = HO.HandoffSignal<C, E1, A>
+  type SinkEndReason = SER.SinkEndReason<C>
+
+  const deps = T.mapN_(
```
mikearnaldi, line 144: isn't this T.tuple?

packages/system/src/Experimental/Stream/_internal/api.ts
```diff
@@ -0,0 +1,883 @@
...
+ */
+export function broadcastedQueuesDynamic(maximumLag: number) {
+  return <R, E, A>(self: Stream<R, E, A>) => broadcastedQueuesDynamic_(self, maximumLag)
+}
+
+/**
+ * Allows a faster producer to progress independently of a slower consumer by buffering
+ * up to `capacity` chunks in a queue.
+ *
+ * @note Prefer capacities that are powers of 2 for better performance.
```
mikearnaldi, line 425: The pow2 comment doesn't apply to the current impl of queue, we should either take a look if it makes sense to port over the pow2 code in queues too

PR #677, opened 2021-05-02
refactor(scripts): fix deprecated lib calls

ahrjarrett:
This is my first commit, so any feedback is absolutely welcome. Nothing in this PR should change or break any functionality -- these updates / tiny function additions won't have an immediate impact on code quality except to (_maybe_) marginally improve tree-shaking due to a previous over-dependence on `TE.taskEither`, which was unnecessary.

Please let me know if you have any feedback, working through stuff like this is how I like to get familiar with a new library. Lots of great stuff in this library! 

🖤

### DIFF
--- a/scripts/package.ts
+++ b/scripts/package.ts
@@ -1,15 +1,19 @@
-import * as A from "fp-ts/Array"
-import { parseJSON } from "fp-ts/Either"
-import { flow, pipe } from "fp-ts/function"
+import * as ROA from "fp-ts/Array"
+import * as J from "fp-ts/Json"
 import * as TE from "fp-ts/TaskEither"
+import * as E from "fp-ts/Either"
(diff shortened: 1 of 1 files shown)


ahrjarrett:
Also, I was expecting prettier to run via CI -- it looks like that happens prepublish? Let me know if you'd like me to run prettier on the changed file and re-submit.

mikearnaldi:
First of all thanks for the contribution, addressing your points:

1) it is good practice to run "yarn lint" before commit, the CI will catch formatting errors but won't autofix

2) the scripts are only used in this module to avoid self dependencies the rest of the ecosystem uses packages/build-utils that are the same scripts built using effect so indeed the quality of those in scripts is kind of rough

I'll merge as soon as CI passes, looks good overall!

PR #788, opened 2021-07-24
Feat(experimental-stream): More Stream and Sink combinators

r-cyr:
Work In Progress


packages/system/src/Effect/failCause.ts
```diff
@@ -0,0 +1,9 @@
+import type * as CS from "../Cause"
+import * as T from "../Effect"
+
+/**
+ * Returns an effect that models failure with the specified `Cause`.
+ */
+export function failCause<E>(cause: CS.Cause<E>): T.IO<E, never> {
```
mikearnaldi, line 7: This should be a rename really (we don't need to keep halt)

PR #871, opened 2021-09-24
Interruption investigation

r-cyr:


mikearnaldi:
Reverting the changes to workaround interruption it seems the skipped test now throws an exception somewhere.

Screenshot 2021-09-28 065925

packages/system/src/Stream/Stream/effectAsyncInterrupt.ts
```diff
@@ -17,12 +21,79 @@ import { effectAsyncInterruptEither } from "./effectAsyncInterruptEither"
...
+            toEmit(async (k) => {
+              try {
+                return runtime.runPromise(
+                  pipe(
+                    Take.fromPull(k),
+                    T.chain((_) => Q.offer_(output, _))
+                  )
+                )
+              } catch (e) {
+                if (e instanceof CS.FiberFailure) {
```
mikearnaldi, line 50: better to avoid instanceof comparism

mikearnaldi: also is there a specific reason to use a promise? runPromise hides the full cause

mikearnaldi (changes requested):

packages/system/src/Effect/effectAsyncInterrupt.ts
```diff
@@ -16,7 +22,7 @@ import { effectMaybeAsyncInterruptBlockingOn } from "./effectMaybeAsyncInterrupt
  * provide better diagnostics.
  */
 export function effectAsyncInterrupt<R, E, A>(
-  register: (cb: Cb<Effect<R, E, A>>) => Canceler<R>,
+  register: (cb: Cb<Effect<R, E, A>>) => E.Either<Canceler<R>, Effect<R, E, A>>,
```
mikearnaldi, line 25: didn't we have the *Maybe variants that returned an either?

r-cyr:
Removed Stream effectAsync implementations

PR #949, opened 2021-11-11
Async collectAll environment fix

ax-hkightlinger:
# Description
Currently, the environment is lost when running variants of `IAll` in async. This PR creates a test (which currently fails) for the preservation of environments when running these computations and provides a bug fix. All tests pass in this PR.

# Results from test with current code
Screen Shot 2021-11-11 at 8 32 05 AM

### DIFF
--- a/packages/core/test/async.test.ts
+++ b/packages/core/test/async.test.ts
@@ -76,7 +76,7 @@ describe("Async", () => {
     ).toEqual(As.failExit("reject"))
   })
 
-  it("preserves envioronment in collectAll", async () => {
+  it("preserves environment in collectAll", async () => {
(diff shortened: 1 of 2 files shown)


packages/core/test/async.test.ts
```diff
@@ -75,4 +75,16 @@ describe("Async", () => {
       )
     ).toEqual(As.failExit("reject"))
   })
+
+  it("preserves envioronment in collectAll", async () => {
```
jbmusso, line 79:
```suggestion
  it("preserves environment in collectAll", async () => {
```

jbmusso (changes requested):
Just a small typo being caught :-)

PR #950, opened 2021-11-11
Added Sync derive utilities

ax-hkightlinger:
This PR adds `derive` utilities for the `Sync` effect type. This is largely a copy past of the derive utilities in `effect`. I can make an attempt at trying to unify derive utilities across effect types if you that think that is a better approach.


mikearnaldi:
Better to keep them separated as it is already quite complex :)

Thanks for the PR!

PR #955, opened 2021-11-16
Initial pass at forEach and forEachPar in Async

ax-hkightlinger:
This is an initial pass at adding `forEach` and `forEachPar` variants into the `Async` module in order to more closely resemble the api of `effect`. This would be a breaking change.

Not expecting this to be a final draft but I wanted to get your thoughts before I spent any more time on it.

### DIFF
--- a/packages/core/test/async.test.ts
+++ b/packages/core/test/async.test.ts
@@ -1,4 +1,7 @@
 import * as As from "../src/Async"
+import { range } from "../src/Collections/Immutable/Array"
+import * as Chunk from "../src/Collections/Immutable/Chunk"
+import * as Tp from "../src/Collections/Immutable/Tuple"
 import { identity, pipe } from "../src/Function"
 
 describe("Async", () => {
@@ -35,11 +38,13 @@ describe("Async", () => {
   })
   it("collectAll", async () => {
     expect(
-      await pipe(
-        As.collectAll([As.succeed(1), As.succeed(2), As.succeed(3)] as const),
-        As.runPromiseExit
+      Chunk.toArray(
+        await pipe(
+          As.collectAll([As.succeed(1), As.succeed(2), As.succeed(3)]),
+          As.runPromise
+        )
       )
-    ).toEqual(As.successExit([1, 2, 3]))
+    ).toEqual([1, 2, 3])
   })
 
   it("onError", async () => {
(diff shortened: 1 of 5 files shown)


mikearnaldi:
Seems reasonable! I wanted to add a note, the reason Async is encoded initially (as a tagged union of primitives) is to have stack-safety but after encoding it I realised it is possible to get safety by using microtasks to suspend potentially recursive calls in a final encoding in the same way as I did in fp-ts https://github.com/gcanti/fp-ts/pull/1591 

As a background Async exists mostly for bundle-size constrained environments where `Effect` may be overkill and it may be good to "re-encode" the type to be final so we aren't forcing all the primitives in the bundle

ax-hkightlinger:
Thanks for the note! I convinced my team (they would probably say forced) to use `@effect-ts` in our project and are trying to keep it to just `Sync` and `Async` on the frontend. I am a huge fan of this project and `Zio` as well and would love to help it move forward in whatever way I can.

I actually took a look at your `fp-ts` PR as well. I personally think there is value in the initial encoding beyond stack safety. I think there is a clearer implementation of effect interruption and environment. Primitives that can be run in sync are done so without putting each within a microtask. And the entire implementation is only a few hundred lines of typescript as well. We are talking about a pretty small difference in final bundle size.

mikearnaldi:
I was unclear, what I meant is using final encoding preserving environment & interruption, I would agree with you on the style, initial encodings are nicer but they do take space, Async is a few hundreds lines of code this is true but adds up to roughly 15/20k where Effect itself if not used with managed & layers is just shy of 35k if we encode it finally I believe we could get it down to a few kb.

Lovely to hear you are adopting :) and looking forward to build it up together!

ax-hkightlinger:
The initial encoding of `Async` by itself is only ~4.5kb minified (non-gzipped) as it stands today. `import-cost` in VSCode actually tells me 3.7kb when I import only `runPromiseExitEnv` which would include all the primitives. This PR would reduce it slightly by removing the `IAll` primitive and corresponding runtime code in `runPromiseExitEnv`.

I have the entire, non-treeshaken `Async` module sized at 27.4KB before compression. 

Smaller would obviously still be better though :)

mikearnaldi:
Then my brain has fused at some point... It was way bigger I must have optimized and forgot... :) well ignore.

ax-hkightlinger:
All good, just didn't want you to spend time trying to improve things at the margin. My guess is you have bigger fish to fry.

mikearnaldi:
Shall we stage this? I think it's a good basis to start

PR #957, opened 2021-11-18
Feature/sync tuple

ax-hkightlinger:
This PR adds an implementation of `tuple` for the `Sync` effect type that does not rely on `Prelude.tupleF`.


mikearnaldi:
Any specific reason why not to use tupleF?

ax-hkightlinger:
Very good point. My initial reasoning was to make it more closely resemble `Effect`, but I didn't stop to think about why `Effect` didn't use `tupleF` and how it is inherently different. I will withdraw the PR.

Issue #961, opened 2021-11-19
Provide way to get index in T.forEach (...) functions


schickling:
Currently there's no way to get the current index in `T.forEach`, `T.forEachPar` etc.

Note sure if possible but ideally this could be done with a function overload and therefor avoiding a duplication of the related exports.

adrian-gierakowski:
How about a solution similar to https://ramdajs.com/docs/#addIndex?

mikearnaldi:
Not sure there is a valid typing for an addIndex, doing it via function overload isn't advisable because you could pass to forEach a function with an optional second argument (like succeed, fail, or anything that has tracing) and there is no way of distinguishing at runtime if the function you pass in wants an index or not.

The easiest would be to add `T.forEachWithIndex` and similar variants, or maybe think of returning extended functions and like from forEach return a function with a `.indexed` method like `T.forEach.indexed(() => ` even though that could have perf implication in constructing the forEach

wesselvdv:
I'd say this can be closed no? With #968 being merged?

PR #968, opened 2021-11-24
add forEachWithIndex variants

tatchi:
Opening this one as a draft to get some early feedback. Mainly to know if I'm on the right track.

For instance, I copy/pasted the `forEachPar*` functions but the versions with `index` don't require much changes. I'm wondering if that would be better to try extracting the part that change and create a more general function that can create both versions?

Also, should we create a `withIndex` version for all of them?

Related issue: #961


mikearnaldi:
We could try to generalise a bit further but generally speaking there isn't much to be done, in Effect those functions are usually heavy optimized and generalising requires additional closure calls (like you could derive forEach from forEachWith index but you'd need a function mapping `forEach(f) = forEach((x) => f(x))`)

packages/system/src/Effect/excl-forEach.ts
```diff
@@ -83,6 +83,32 @@ export function forEach_<A, R, E, B>(
...
+  f: (a: A, i: number) => Effect<R, E, B>,
+  __trace?: string
+): Effect<R, E, Chunk.Chunk<B>> {
+  return core.suspend(() => {
+    let index = 0
+    const acc: B[] = []
+
+    return map.map_(
+      forEachUnit_(as, (a) =>
+        map.map_(f(a, index), (b) => {
```
mikearnaldi, line 102: we could use the length of `acc` as index instead of keeping another variable

mikearnaldi:
Lets stage this in multiple PRs, it's a good start for next release

PR #969, opened 2021-11-25
feat: Unobtrusive ServiceId

patroza:


packages/system/src/Has/index.ts
```diff
@@ -33,16 +36,16 @@ export type Flat<A> = { readonly [k in keyof A]: A[k] } extends infer X ? X : ne
 export function service<T extends PropertyKey, X extends Record<PropertyKey, unknown>>(
   t: T,
   x: X
-): Flat<X & { readonly serviceId: T }> {
+): Flat<X & { readonly [ServiceId]: T }> {
   // @ts-expect-error
-  return { ...x, serviceId: t }
+  return x
```
qlonik, line 41: is keeping the `serviceId` at runtime inside the service unnecessary?

mikearnaldi: it is currently but I am still unsure it will remain unnecessary, I wanted to remove the requirements of passing a tag to providers like layers

PR #1004, opened 2021-12-14
Update install instructions

### LINKED ISSUE #998: Intellij - Autosuggestion not working without explicit @effect-ts/system install
In Intellij Idea  typescript only detects the direct exports from `@effect-ts/core/effect`, not the re-exported modules from `@effect-ts/system` (See below)

Without explicit install only direct exports are found:
341F72A9-08DA-4457-A97F-A5FB1E0BDB35

With explicit install, re-exports from system are found:
037CDA49-2222-4D4C-B0AF-F19ACB2066D0

## Debug:
- The dependency `@effect-ts/system` is  indeed installed
- Autosuggestion works fine in vs code
- I haven't had this problem with any other package but `effect-ts` relies heavily on re-exports
- Deps: `@effect-ts/core (latest)  typescript (4.5.2)`

## Possible Solutions
Simply add the package as an explicit installation in the `Readme` "How to install" section. 
Drawbacks: 
- Freezes the dependency on `system` which could hurt later if you plan on changing how things work
- Won't hurt anyone but this is solving a few people's problem by making everybody "pay"
Advantages:
- Won't hurt anybody 🤷‍♂️

I haven't been able to solve the issue any other way (editor config) for now. 

If this works for you the PR #1004 is already opened :)

isthatcentered:
Fixes #998: Autosuggestion not working without explicit /system dependency for Intellij Idea

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Within the ecosystem of packages you will find:
 We recommend the usage of `yarn` and if you have multiple packages the usage of `yarn workspaces` that handles by default hoisting of dependencies:
 
 ```sh
-yarn add @effect-ts/core
+yarn add @effect-ts/core @effect-ts/system
 ```
 
 ## tsconfig


mikearnaldi:
We may switch to peer dependencies flagged as optional that should work

PR #1026, opened 2022-01-13
Fix: option getApplyIdentity

tatchi:
First commit adds a test. Second commit fixes the issue.

I was not sure where to create the tests nor how to structure them so please let me know if I should do it differently


mikearnaldi:
Lovely

PR #1027, opened 2022-01-13
Add isSuccess and isFailure for Sync

bfulop:
We noticed that the `isSuccess` and `isFailure` were missing from the Sync effect type.
This PR copies over the implementation from the Async equivalent. It does not include tracing. I'm not sure if it's needed in the sync context?


mikearnaldi:
That's perfect, in regards to tracing both sync and async don't currently support tracing, they will in a future release but for now no need!

Thanks for the pr

PR #1030, opened 2022-01-14
fix(exp-stream,exp-channel): missing T.transplants for zipWithLatest, mergeWith

wesselvdv:
Initial idea to use `T.transplant` came from @r-cyr, I moved the transplant up so it would apply for the entire zip. 

I am not sure what kind of test I should add for this @mikearnaldi? I think the case we're solving by grafting is one where a `Stream` isn't closed, which I don't have a clue how to craft a test case that's supposed to end after a while.


wesselvdv:
I added a test case which is failing, so that invalidates my fix unfortunately. I am not sure if the test case is valid, but I think so.

wesselvdv:
It seems the grafting fix only changes the actual fiber that's causing the interrupt.

wesselvdv:
I traced the problem to `Channel.mergeWith`, I had to apply the exact same `Effect.transplant` to extend their lifespan into the parent scope.

mikearnaldi:
Adding context from discord (for posterity):

Screenshot 2022-01-14 124340

Issue #1035, opened 2022-01-17
Migrate the Effect-TS codebase to mirror ZIO 2.x


IMax153:
The [`feat/system-next`](https://github.com/Effect-TS/core/tree/feat/system-next) branch is a work-in-progress port of the changes introduced in ZIO `v2.x` to Effect-TS. This is very early work and is by no means feature-complete - many features are missing, do not work properly, or are implemented inefficiently. However, the basic ability to run effects has been built.

I would like to ask the rest of the Effect-TS community to help continue the work started on this branch. The plan will be that individuals who would like to contribute to this effort should select something either from the list below (or an unlisted feature that they would like to port from ZIO 2.x) and state that they will be working on that feature in the comments. This way we don't have efforts being duplicated. Pull requests should be opened against the `feat/system-next` branch (not `master`).

|  Status  | Feature  |  Assignee | Additional Details  |
|:--------:|:--------:|:---------:|:-------------------:|
| In-Progress  | Layer  | @maxwellbrown1990  |  |
|   | Tracing  |   | Most of the tracing implementation has been stubbed out. It should be re-implemented and tests added.   |
|   | Schedule  |   |  The signature of the `Schedule` datatype has been reworked in 2.x  |
|   | Scope  |   | The basic `Scope` has been implemented but it is feature incomplete  |
|   | Metrics  |   | The metric framework needs a second set of eyes - there's a lot of duplication. And it needs tests.   |
|   | Metrics  |   | The metric framework needs a second set of eyes - there's a lot of duplication. And it needs tests.   |
|   | XPure  |   | The `XPure` datatype needs to be integrated into the `FiberContext` run loop.   |
|   | Async  |   | The `Async` datatype should be reviewed to see if it is needed.  |
|   | STM Datatypes |   | There are many STM datatypes that are missing currently. |
|   | STM API |   | The API of STM should be audited against the ZIO 2.x branch. |
|   | Collections API |   | The naming of the functions in the collections directory should be unified with ZIO 2.x (i.e. `...M` functions should be `...Effect`, `...Unit` should be `...Discard`) |
|   | Stack Traces |   | In ZIO 2.x, error stacktraces closely mirror Java stacktraces - we should see if we can acoomplish the same or similar.  |

I will try to update the list above as more issues that need to be addressed are identified. If it is determined that the scope of this issue is more suited to a GitHub project, it can be enabled at a later date.

patroza:
> `...M` functions should be `...Effect`

[...]

PR #1043, opened 2022-01-17
feat(scope): implement Fiber roots

0x706b:
#1035 

## Changelog

- feat(scope): implement Fiber roots
- feat(effect): implement `repeatUntilEffect` and `repeatUntil`
- feat(hashset): implement `from`
- test(fiber): add `roots` test


mikearnaldi (approved):

packages/system-next/src/Fiber/context.ts
```diff
@@ -124,6 +124,8 @@ export type FiberRefLocals = Map<FiberRef.Runtime<any>, any>
 
 export const currentFiber = new AtomicReference<FiberContext<any, any> | null>(null)
 
+export const _roots = new Set<FiberContext<any, any>>()
```
IMax153, line 127: Should we make this a `LazyValue`? @mikearnaldi @0x706b

IMax153 (changes requested):

packages/system-next/src/Fiber/context.ts
0x706b: Couldn't hurt 👍🏻

mikearnaldi: doesn't hurt but also it is a Set so it shouldn't be a big deal (doesn't depend on anything)

IMax153 (commented):
Thank you for your work on this @0x706b !

PR #1046, opened 2022-01-19
Add Zio 2.x implementation of TPriorityQueue

wesselvdv:
This PR adds the Zio 2.x implementation of TPriorityQueue, SortedMap and some missing combinators for STM.

Related to #1035

### DIFF
--- a/packages/system-next/src/Collections/Immutable/SortedMap/index.ts
+++ b/packages/system-next/src/Collections/Immutable/SortedMap/index.ts
@@ -1,0 +1,173 @@
+import { pipe } from "@effect-ts/system/Function"
+
+import * as It from "../../../Iterable"
+import type * as O from "../../../Option"
+import type * as Ord from "../../../Ord"
+import * as ST from "../../../Structural"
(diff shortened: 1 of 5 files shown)


mikearnaldi (approved):

packages/system-next/src/Collections/Immutable/SortedMap/index.ts
```diff
@@ -0,0 +1,173 @@
+import { pipe } from "@effect-ts/system/Function"
```
IMax153, line 1: This should be a relative import `"../../../Function"`.

IMax153 (changes requested):
@wesselvdv - this PR looks great! There's only one tiny change I would make. Once you push it we can go ahead and merge 👍 

Thanks again for your work on this

packages/system-next/src/Collections/Immutable/SortedMap/index.ts
wesselvdv: nice catch!

IMax153 (approved):

PR #1048, opened 2022-01-20
perf: decrease atomic references

0x706b:
## Summary

Since we don't have to worry about parallelism in a single-threaded runtime, `AtomicReference` and related are used in many places that they don't strictly need to be. This PR removes uses of `AtomicReference` in some hot code paths to avoid the extra function call.


0x706b:
After running some (micro-) benchmarks, it appears there isn't really a difference. Closing for now

PR #1051, opened 2022-01-21
Port `List` and `ListBuffer` from Scala

0x706b:
`List` has very good performance in terms of `prepend`, `reverse`, and iteration. The mutable `ListBuffer` makes building lists by `append`-ing fast as well. I think they could be very useful for the `Channel` implementation, and to replace the current `List` in some performance-sensitive areas.

I also added the scala license to a `licenses` folder in the root of the repo, and added a command to the `pack` script to copy licenses into the `dist` directory, to comply with my understanding of the Apache license terms.

## Benchmarks

These were run on transpiled ESM files, with `node --experimental-specifier-resolution=node`. I noticed some drastic performance decreases when I ran with `ts-node`, possibly due to the one-function-per-file model, and how `commonjs` exports are transpiled.

```
LinkedList.prepend (1000) x 227,161 ops/sec ±2.77% (81 runs sampled)
List.prepend (1000) x 53,329 ops/sec ±0.96% (86 runs sampled)
LinkedList.reverse (1000) x 147,642 ops/sec ±1.21% (88 runs sampled)
List.reverse (1000) x 55,135 ops/sec ±0.77% (96 runs sampled)
LinkedList.forEach (1000) x 229,070 ops/sec ±0.47% (90 runs sampled)
List.forEach (1000) x 151,727 ops/sec ±0.33% (92 runs sampled)
LinkedList.filter (1000) x 141,112 ops/sec ±0.86% (89 runs sampled)
List.filter (1000) x 97,757 ops/sec ±0.60% (96 runs sampled)
LinkedList.map (1000) x 158,324 ops/sec ±1.24% (89 runs sampled)
List.map (1000) x 495,225 ops/sec ±1.03% (92 runs sampled)
```


mikearnaldi:
Is there a reason why not to remove the current List module in favour of this one?

mikearnaldi (approved):

0x706b:
> Is there a reason why not to remove the current List module in favour of this one?

 @mikearnaldi The existing `List` could have some use in terms of immutable `append`. `map` seems to be superior to `LinkedList` as well. I have ported both of these in `principia`, and I renamed the original list to `Vector`. We could probably do something similar here, if you think the existing List is worth keeping

mikearnaldi:
Mmm for fast immutable append we also have Chunk which is also similar to a vector probably we can get rid of the list

0x706b:
I just ran some benchmarks. `Chunk` performance seems to be not as good as `Vector` (List) in most cases, other than repeated `concat`. I believe this is because for most ops, `Chunk` uses its iterator. Maybe this could be mitigated with some optimization; I have written an [unrolled recursive version of `Chunk.map`](https://github.com/0x706b/principia.ts/blob/733d38725256c68dbaea70c3393b5575b64d91d3/packages/base/src/collection/immutable/Conc/core.ts#L1143) which seems to work well.

All that to say, I don't know if `Chunk` replaces the current `List` in its current state.

```
Vector.map (1000) x 461,544 ops/sec ±1.07% (94 runs sampled)
Chunk.map (1000) x 21,298 ops/sec ±0.73% (94 runs sampled)
Vector.append (1000) x 52,034 ops/sec ±0.92% (84 runs sampled)
Chunk.append (1000) x 17,014 ops/sec ±0.60% (94 runs sampled)
Vector.prepend (1000) x 53,886 ops/sec ±0.75% (91 runs sampled)
Chunk.prepend (1000) x 17,819 ops/sec ±1.26% (90 runs sampled)
Vector.concat (1000 x 100) x 3,587 ops/sec ±1.01% (92 runs sampled)
Chunk.concat (1000 x 100) x 5,646 ops/sec ±1.00% (94 runs sampled)
```

mikearnaldi:
Indeed seems to be better with List, in fact the algorithm itself is radix
balanced might be worth to keep as it has really good perf for the average
case, maybe worth investigating if an equivalent (better maintained) exist
in scala. I like the naming with Vector though.

0x706b:
Sounds good! I'll work on renaming `List` to `Vector`, and `LinkedList` to `List`

mikearnaldi:
@0x706b if you don't mind I'll close this and directly port the changes in `system-next` we've been moving things around a bit and rebasing may be complcated

0x706b:
> if you don't mind I'll close this and directly port the changes in `system-next` we've been moving things around a bit and rebasing may be complicated

@mikearnaldi yup! that works for me

mikearnaldi:
reopening to keep it in mind :)

mikearnaldi:
Ported in tsplus/stdlib

Issue #1005, opened 2021-12-15
Proposal for HKT restructuring


mikearnaldi:
Since the time we implemented HKTs we found a new encoding that doesn't require having a central registry of URIs and that should have far better performances while allowing for more flexibility.

The encoding is implemented in:

https://github.com/mikearnaldi/hkt-new

I propose we change the current encoding to reflect the new implementation, as part of the change we will also need to remove the flexibility of choosing the variance of a type parameter namely in the current encoding we are using a config param to specify how a type parameter should be mixed, the problem with that approach is that it may generate inconsistencies, namely we do implementations on a reduced base using `HKT/HKT2` and we have a generic overload that considers all the params with their respective rules but there is no guarantee that the base impl would actually work with that set of rules, namely I could implement a base using R as contravariant and having R as covariant in the target type. Also we actually never used any other setting other than `S` invariant, `R` contravariant, etc.

isthatcentered:
If I may add my two cents,

I re-implemented it, to understand how you managed to make it work and it is a pleasure to work with ! (especially coming from `fp-ts`'s way of doing HKTs)

The only "confusing" thing was the separate declaration of `{Option|Either|...}F` for the `R, E, A`mapping. Maybe the naming convention could be  `{Option|Either|...}Mapping` or something to get a better feel of what this does... But that's a bit of a nitpick😅 (and nothing a few lines of documentation wouldn't make easy).

I'm not sure choosing the variance is required (besides the complexity it might add), I tend to expect R to be contravariant, E & A to be covariant and would probably be confused otherwise.

mikearnaldi:
it could be `OptionHKT` or `OptionRepr` or similar, at the end it is a generic representation of a type

mikearnaldi:
(overall I agree, it is refreshing how simpler it is)

mikearnaldi:
> I'm not sure choosing the variance is required (besides the complexity it might add), I tend to expect R to be contravariant, E & A to be covariant and would probably be confused otherwise.

My guess is it may be restored as a config param over the typeclass that doesn't get propagated into Kind at all (or it may)

Arguably it isn't too useful

IMax153:
I also agree - I really enjoy the new encoding! Excellent work on this @mikearnaldi 👏 

With respect to variance, maybe it would be helpful to see a use case in user-land where variance may help avoid pitfalls? 

[...]

PR #1055, opened 2022-01-22
#1005 - HKT restructuring

isthatcentered:
Please note this is just a draft, things are voluntarily kept together to ease refactoring while I am discovering things. 

### 01/22/2022 
Unproven decisions to note:
- Variance is now fixed per parameter and doesn't take into account a fixed type (at least for now)

Example HKT instances
```typescript
export interface AsyncF extends PV2.HKT {
  readonly type: A.Async<this["R"], this["E"], this["A"]>
}

export const Covariant = P.instance<PV2.Covariant<AsyncF>>({
  map: A.map
})

export const Any = P.instance<PV2.Any<AsyncF>>({
  any: () => A.succeed({})
})

export const AssociativeBoth = P.instance<PV2.AssociativeBoth<AsyncF>>({
  both: A.zip
})

...
```

### DIFF
--- a/packages/core/src/PreludeV2/FX/index.ts
+++ b/packages/core/src/PreludeV2/FX/index.ts
@@ -1,0 +1,7 @@
+// ets_tracing: off
+
+import type * as HKT from "../HKT"
+
+export interface Fail<F extends HKT.HKT> {
+  readonly fail: <E = never>(e: E) => HKT.Kind<F, any, any, any, unknown, E, never>
+}
--- a/packages/core/src/PreludeV2/HKT/index.ts
+++ b/packages/core/src/PreludeV2/HKT/index.ts
@@ -1,0 +1,77 @@
+import type { InitialVariance, MixVariance } from "./variance"
+
+export declare const URI: unique symbol
+
+export interface Typeclass<F extends HKT> {
+  readonly [URI]?: F
+}
+
+type Param = "X" | "I" | "S" | "R" | "E" | "A"
+type ParamVariance<P extends Param> = {
+  X: "_"
+  I: "_"
+  S: "_"
+  R: "-"
+  E: "+"
+  A: "+"
+}[P]
+
+export interface HKT {
+  readonly X?: unknown
+  readonly I?: unknown
+  readonly S?: unknown
+  readonly R?: unknown
+  readonly E?: unknown
+  readonly A?: unknown
+  readonly type?: unknown
+}
+
+export type Kind<F extends HKT, X, I, S, R, E, A> = F extends { readonly type: unknown }
+  ? (F & {
+      readonly X: X
+      readonly I: I
+      readonly S: S
+      readonly R: R
+      readonly E: E
+      readonly A: A
+    })["type"]
+  : {
+      readonly _F: F
+      readonly X: X
+      readonly I: I
+      readonly S: S
+      readonly _R: (_: R) => void
+      readonly _E: () => E
+      readonly _A: () => A
+    }
+
+export interface ComposeF<F extends HKT, G extends HKT> extends HKT {
+  readonly type: Kind<
+    F,
+    this["X"],
+    this["I"],
+    this["S"],
+    this["R"],
+    this["E"],
+    Kind<G, this["X"], this["I"], this["S"], this["R"], this["E"], this["A"]>
+  >
+}
+
+// Initial type pere parameterized type depending on variance
+export type Initial<P extends Param> = {
+  X: InitialVariance<ParamVariance<"X">>
+  I: InitialVariance<ParamVariance<"I">>
+  S: InitialVariance<ParamVariance<"S">>
+  R: InitialVariance<ParamVariance<"R">>
+  E: InitialVariance<ParamVariance<"E">>
+  A: InitialVariance<ParamVariance<"A">>
+}[P]
+
+export type Mix<P extends Param, X extends [any, ...any[]]> = {
+  X: MixVariance<ParamVariance<"X">, X>
+  I: MixVariance<ParamVariance<"I">, X>
+  S: MixVariance<ParamVariance<"S">, X>
+  R: MixVariance<ParamVariance<"R">, X>
+  E: MixVariance<ParamVariance<"E">, X>
+  A: MixVariance<ParamVariance<"A">, X>
+}[P]
--- a/packages/core/src/PreludeV2/HKT/variance.ts
+++ b/packages/core/src/PreludeV2/HKT/variance.ts
@@ -1,0 +1,33 @@
+import type { UnionToIntersection } from "@effect-ts/system/Utils"
+
+type OrNever<K> = unknown extends K ? never : K
+
+type Variance = "+" | "-" | "_"
+
+export type InitialVariance<V extends Variance> = {
+  "+": never
+  "-": unknown
+  _: any
+}[V]
+
+// composes types according to variance specified in V
+export type MixVariance<V extends Variance, X extends [any, ...any[]]> = V extends "_"
+  ? X[0]
+  : V extends "+"
+  ? X[number]
+  : V extends "-"
+  ? X extends [any]
+    ? X[0]
+    : X extends [any, any]
+    ? X[0] & X[1]
+    : X extends [any, any, any]
+    ? X[0] & X[1] & X[2]
+    : X extends [any, any, any, any]
+    ? X[0] & X[1] & X[2] & X[3]
+    : X extends [any, any, any, any, any]
+    ? X[0] & X[1] & X[2] & X[3] & X[4]
+    : X extends [any, any, any, any, any, any]
+    ? X[0] & X[1] & X[2] & X[3] & X[4] & X[5]
+    : UnionToIntersection<{ [k in keyof X]: OrNever<X[k]> }[keyof X]>
+  : X[0]
+
--- a/packages/core/src/PreludeV2/index.ts
+++ b/packages/core/src/PreludeV2/index.ts
@@ -1,0 +1,97 @@
+import type * as Tp from "@effect-ts/system/Collections/Immutable/Tuple"
+
+import type * as HKT from "./HKT"
+
+export * from "./HKT"
+export * as FX from "./FX"
+
+/**
+ * An associative binary operator that combines two values of types `F<A>`
+ * and `F<B>` to produce an `F<[A, B]>`.
+ */
+export interface AssociativeBoth<F extends HKT.HKT> {
+  both: <X2, I2, S2, R2, E2, B>(
+    fb: HKT.Kind<F, X2, I2, S2, R2, E2, B>
+  ) => <X, I, S, R, E, A>(
+    fa: HKT.Kind<F, X, I, S, R, E, A>
+  ) => HKT.Kind<
+    F,
+    HKT.Mix<"X", [X2, X]>,
+    HKT.Mix<"I", [I2, I]>,
+    HKT.Mix<"S", [S2, S]>,
+    HKT.Mix<"R", [R2, R]>,
+    HKT.Mix<"E", [E2, E]>,
+    Tp.Tuple<[A, B]>
+  >
+}
+
+export interface Any<F extends HKT.HKT> {
+  readonly any: <
+    X = HKT.Initial<"X">,
+    I = HKT.Initial<"I">,
+    S = HKT.Initial<"S">,
+    R = HKT.Initial<"R">,
+    E = HKT.Initial<"E">
+  >() => HKT.Kind<F, X, I, S, R, E, unknown>
+}
+
+/**
+ * Given a function A => B, go from F<A> to F<B>
+ *
+ * ```typescript
+ * declare const user: User
+ * const userName: Option<string> = pipe(O.some(user), O.map(user => user.username))
+ * ```
+ */
+export interface Covariant<F extends HKT.HKT> {
+  readonly map: <A, B>(
+    f: (a: A) => B
+  ) => <
+    X = HKT.Initial<"X">,
+    I = HKT.Initial<"I">,
+    S = HKT.Initial<"S">,
+    R = HKT.Initial<"R">,
+    E = HKT.Initial<"E">
+  >(
+    fa: HKT.Kind<F, X, I, S, R, E, A>
+  ) => HKT.Kind<F, X, I, S, R, E, B>
+}
+
+/**
+ * Describes a type that can be "flattened" in an
+ * associative way.
+ *
+ * ```typescript
+ * declare const user: User
+ * const userName: Option<string> = pipe(
+ *  O.some(user),
+ *  O.map(user => user.email), // Option<Option<string>>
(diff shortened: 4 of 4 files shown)


mikearnaldi:
The whole MixVariance/InitialVariance thing should be removed

mikearnaldi:
`Variance is now fixed per parameter and doesn't take into account a fixed type (at least for now)`

Parameters like `S` that are considered invariant should not mix at all, meaning in a signature you should not have `S1` and `S2` at all

mikearnaldi:
Also ideally every parameter has a meaning, it almost feel like we could get away with:

```
S = State (Invariant)
I = Category Input (Contravariant)
R = Requirements (Contravariant)
E = Error (Covariant)
A = Output (Covariant)
```

Any fixed type can be supported by transformers by having parametric URIs like:

```ts
interface KeyValueT<Key, Value, F extends HKT> {
  type: KeyValue<Key, Value, Kind<F, this[...]>
}
```

isthatcentered:
> I am not even too sure we want to keep the invariant `S` around, you could always have a:
> 
> ```ts
> interface StateT<S, F> {
>    type: (s: S) => Kind<F, this["I"], this["R"], this["E"], [S, this["A"]]>
> }
> ```
> 
> and would even allow mixing multiple different states

Oh, that would be cool, the simpler the `Kind` signature the better! 

And If we go down this road, should we keep the `I` parameter then?  The parameters would kind of lose their "usage" meaning (requirements/error/output) but we still expose a  Contravariant input, and a covariant output (+ error Covariant output for convenience even though it kind of goes against the idea).

isthatcentered:
> I am not even too sure we want to keep the invariant `S` around, you could always have a:
> 
> ```ts
> interface StateT<S, F> {
>    type: (s: S) => Kind<F, this["I"], this["R"], this["E"], [S, this["A"]]>
> }
> ```
> 
> and would even allow mixing multiple different states

I gave it a try but it seems that would force us to fix the S type no? 

```typescript
import { pipe, tuple } from "../../Function/index.js"
import * as PV2 from "../../PreludeV2"

interface StateT<S, F> extends PV2.HKT {
  readonly type: (
    s: S
  ) => PV2.Kind<F, this["X"], this["I"], this["R"], this["E"], readonly [this["A"], S]>
}

export function monad<F extends PV2.HKT>(M: PV2.Monad<F>) {
  return <S>() =>
    PV2.instance<PV2.Monad<StateT<S, F>>>({
      any: () => (s) =>
        pipe(
          M.any(),
          M.map((m) => tuple(m, s))
        ),
      flatten: (ffa) => (x) =>
        pipe(
          x,
          ffa,
          PV2.DSL.chainF(M)(([f, us]) => f(us))
        ),
      map: (f) => (fa) => (x) =>
        pipe(
          x,
          fa,
          M.map(([a, s]) => tuple(f(a), s))
        )
    })
}
```

(Also, If `S`goes away, it should probably take `X` with it, now that I'm re-reading this)

mikearnaldi:
```
I gave it a try but it seems that would force us to fix the S type no?
```

That is the objective, `S` needs to be invariant anyway and given the lack of generics in constants it end up being forced to appear in every signature even things like `unit` that are then forced to be functions like `unit<S>()`

Even with `S` invariant you can't have a computation that has both state `A` and state `B` because you have a single `S`, "fixing" the param allows you to have many `StateT<A, StateT<B, Bla>>` they won't conflict and usually when you use state you really want to use a fixed state

mikearnaldi:
> Oh, that would be cool, the simpler the `Kind` signature the better!
> 
> And If we go down this road, should we keep the `I` parameter then? The parameters would kind of lose their "usage" meaning (requirements/error/output) but we still expose a Contravariant input, and a covariant output (+ error Covariant output for convenience even though it kind of goes against the idea).
> 
> Edit: I might have said something dumb here, I have never used the `I` before. I'll give it a go on `xpure` this should make me sweat enough to fix that :D

Thought of `I` to model the input type of the category typeclass, that basically encodes `I => A` and `A => B` enables `I => B`

isthatcentered:
Notable changes:
- Removed IndexFor and used params for all {...}WithIndex
- Removed all implementXWithIndex functions

shroomist:
is anything blocking this? I'm really looking forward to #1055

mikearnaldi:
Not really I am planning it for when system next is ready given it is a breaking change

mikearnaldi:
@isthatcentered merge commits are problematic can you please use rebase?

isthatcentered:
> @isthatcentered merge commits are problematic can you please use rebase?

Sorry about that. I'll rework the history to have something cleaner.

isthatcentered:
I was thinking about what kind of docs could be added to help the discovery of the various type classes, I know that when I started (and still sometimes), the main pain point I have is understanding "What's that thing for"/not knowing that there's already a data type for my use case.

Maybe something like creating a custom HKT and seeing how you can leverage the various type classes to implement the various features could be very nice to take a tour of everything.

Ex: decoder/tiny cli dsl/schedule/retry/tiny jenkins/...

We would probably need three different examples since 
a. It's hard to cover everything in a single example, 
b. You need 3 examples to spot a pattern and, 
c. You can increase complexity/go further with each example

Other random docs ideas: 
- It would also be nice to demonstrate the different ways to use the data types (pipe & flow/do/gen)
- Having a full scaffold example of an app with Effect (including Layer, Managed, Ref, tests).
- FP-TS has a ["find out what you need by use case" recipes doc](https://grossbart.github.io/fp-ts-recipes/#/) that's very nice for newcomers

Most of those are already in the tests examples but they are a little hidden and some are probably a bit hardcore for newcomers. 

(I know there's an ongoing docs effort, I haven't had the time to check It yet so I very well might be repeating something already said/discarder)

shroomist:
@isthatcentered I did find the jsdoc `@example` helpful
https://gcanti.github.io/fp-ts/modules/Apply.ts.html#apply-overview
https://github.com/gcanti/fp-ts/blob/master/src/Apply.ts#L13
The real problem was that the most interesting modules (or use cases) didn't provide those examples.

I'd vote (If I was asked) for the next-to-code jsdoc `@examples`.

isthatcentered:
Currently updating all the examples in `core/tests` (I used a different test file while moving the instances) and, while I'm at It I'm also adding explicit types and expectations (It's always nice when you're reading the code on github and can't execute the code). Renaming/categorizing the examples would be cool too.


Few issues I'll have to investigate tomorrow:
1. P.anyDSLMethod() fails with `TypeError` in tests (no clue why)
Ex:
```typescript
// effect/instances.ts
import * as P from "../../src/PreludeV2/index.js

export const { match, matchIn, matchMorph, matchTag, matchTagIn } =
  DSL.matchers<EffectF>()

// test.spect.ts
someMethodImportingEffectFiles() // 💥 TypeError
````

Can be fixed by accessing it via DSL root rather than prelude root
```typescript
// effect/instances.ts
import * as DSL from "../../src/PreludeV2/DSL/index.js

export const { match, matchIn, matchMorph, matchTag, matchTagIn } =
  DSL.matchers<EffectF>()

// test.spect.ts
someMethodImportingEffectFiles() // 👌 All fine
````

2. Transformers instance factory names (Ex: `fail(F): Fail<EitherT<F>>` can clash with the already exported function with that same name. I renamed Uppercased them (`applicative(F)` -> `Applicative(F)`). If anyone has a better idea please don't hesitat^^

3. Transformers provide types don't match implementation. After provide, the nested `HKT` `R`param has a type of  
 `unknown` (aka, provided too), but we only provide the transformer. I'm assuming that's fine since this is the current implementation but I wanted to point It out just in case.


4. I'll have to go back to XPure tomorrow, usage feels weird now that every dsl method has to take a `State` parameter be `struct<S>()(record)`.

File: packages/core/src/StateT/index.ts [hunk 1]
```diff
@@ -2,4 +2,67 @@
 
 import "../Operator/index.js"
 
-export * from "./Classic/index.js"
+import { pipe } from "../Function"
+import * as DSL from "../PreludeV2/DSL/index.js"
+import * as P from "../PreludeV2/index.js"
+
+interface StateT<A, B> {
```
isthatcentered, line 9: Had to do this otherwise type inference for `A` param was not working when using chain. Please let me know If there Is there a better way to fix this?

packages/core/test/examples/12.test.ts
```diff
@@ -1,22 +0,0 @@
-import { pipe } from "@effect-ts/system/Function"
-
```
isthatcentered, line 2: Couldn't see any difference with example 10 besides InvariantT which, from what I understand, can be deleted with the new HKT system. (I'll test this assumption further though)

packages/core/test/examples/17.test.ts
```diff
@@ -1,75 +0,0 @@
-import * as M from "@effect-ts/system/Collections/Immutable/Map"
-import * as Tp from "@effect-ts/system/Collections/Immutable/Tuple"
```
isthatcentered, line 2: Couldn't see any difference with example 11

packages/core/src/StateT/index.ts
mikearnaldi: very strange indeed, also to avoid errors down the line it is good to export all the used symbols

mikearnaldi:
> Currently updating all the examples in `core/tests` (I used a different test file while moving the instances) and, while I'm at It I'm also adding explicit types and expectations (It's always nice when you're reading the code on github and can't execute the code). Renaming/categorizing the examples would be cool too.
> 
> Few issues I'll have to investigate tomorrow:
> 
> 1. P.anyDSLMethod() fails with `TypeError` in tests (no clue why)
>    Ex:
> 
> ```ts
> // effect/instances.ts
> import * as P from "../../src/PreludeV2/index.js
> 
> export const { match, matchIn, matchMorph, matchTag, matchTagIn } =
>   DSL.matchers<EffectF>()
> 
> // test.spect.ts
> someMethodImportingEffectFiles() // 💥 TypeError
> ```
> 
> Can be fixed by accessing it via DSL root rather than prelude root
> 
> ```ts
> // effect/instances.ts
> import * as DSL from "../../src/PreludeV2/DSL/index.js
> 
> export const { match, matchIn, matchMorph, matchTag, matchTagIn } =
>   DSL.matchers<EffectF>()
> 
> // test.spect.ts
> someMethodImportingEffectFiles() // 👌 All fine
> ```
> 
> 2. Transformers instance factory names (Ex: `fail(F): Fail<EitherT<F>>` can clash with the already exported function with that same name. I renamed Uppercased them (`applicative(F)` -> `Applicative(F)`). If anyone has a better idea please don't hesitat^^
> 3. Transformers provide types don't match implementation. After provide, the nested `HKT` `R`param has a type of
>    `unknown` (aka, provided too), but we only provide the transformer. I'm assuming that's fine since this is the current implementation but I wanted to point It out just in case.
> 4. I'll have to go back to XPure tomorrow, usage feels weird now that every dsl method has to take a `State` parameter be `struct<S>()(record)`.

1. Sounds really strange but it would be a blocker if not solved.

2. I don't understand this, what are the functions already exported with the same name, can you point to the code of 2 things that would conflict? function names only make sense in a single module not in general so "fail" in the context of Effect or on the context of OptionT may mean different things, I don't like the uppercase naming for functions and it would conflict with type names. The convention could be something like `getApplicative` but I do prefer simply `applicative`.

3. please provide a code example I don't understand what you are referring to.

4. Just remove the param `S` from the `HKT` and drop support for `XPure`. If someone wants to use State you can only use the parametric one with a fixed state

mikearnaldi:
> True, as an added bonus I think the full example appears when you hover over the method!

I have the opposite opinion here, the reason there are no inline examples in jsdoc is because I truly don't like to see them in the on-hover. I think we will find a middle ground with paka where @schickling is planning embeddable code in docs

isthatcentered:
> 1. Sounds really strange but it would be a blocker if not solved.
> 2. I don't understand this, what are the functions already exported with the same name, can you point to the code of 2 things that would conflict? function names only make sense in a single module not in general so "fail" in the context of Effect or on the context of OptionT may mean different things, I don't like the uppercase naming for functions and it would conflict with type names. The convention could be something like `getApplicative` but I do prefer simply `applicative`.
> 3. please provide a code example I don't understand what you are referring to.
> 4. Just remove the param `S` from the `HKT` and drop support for `XPure`. If someone wants to use State you can only use the parametric one with a fixed state

1. I'll investigate and report, this is suspicious, If this is really an issue how come no one has surfaced It before.


2. Actually after thinking a bit more about this the only time this can happen is with `XPure`, now that It accepts an `S` parameter, we have the original `fail` method + the `fail` method for making an instance of `Fail`. Can't happen with other transformers since they only expose instances. This is solved by your 4th point so **back to the original naming** which felt much better! 


3. One example would be `XReader` 
```typescript
export function Provide<F extends P.HKT>(_: P.Monad<F>) {
  return HKT.instance<FX.Provide<XReaderTF<F>>>({
    provide:
      <R>(r: R) =>
      <X, I, E, A>(fa: XR.XReader<R, HKT.Kind<F, X, I, R, E, A>>) => {
        return pipe(
          fa,
          XR.provideSome(() => r)
        ) as XR.XReader<unknown, HKT.Kind<F, X, I, unknown, E, A>> // <- Cast required, the result is typed XR.XReader<unknown, HKT.Kind<F, X, I, unknown, E, A>>
      }
  })
}
```


4. I'll rephrase just to be sure, do you mean:
a. Remove the instances in `XPure/index.ts`, If someone wants to use `S` they can default to `XState`. 
b. Modify XPure to drop state handling, people can use `StateT` if needed.

isthatcentered:
So, the DSL import issue definitely comes from something I've done. Not sure what yet but I'll find it

isthatcentered:
Annnd It's fixed, my IDE was drunk, I had imports of `@effect-ts/core` inside core.

packages/core/src/IndexedT/index.ts
```diff
@@ -207,57 +26,70 @@ export class Ix<I, O, A> {
   constructor(readonly value: A) {}
 }
 
+export interface IndexedT<F extends P.HKT, _I, _O> {
```
isthatcentered, line 29: @mikearnaldi I removed the initial Indexed representation and exposed directly an IndexedT, let me know if that's an issue and I'll rework it :).

(also, this type is pretty cool and simple! I got lost in the types with the old version but the new system makes what happens very explicit)

packages/core/src/index.ts
```diff
@@ -48,7 +48,6 @@ export * as IO from "./IO/index.js"
 export * as Id from "./Id/index.js"
 export * as Identity from "./Identity/index.js"
 export * as IndexedT from "./IndexedT/index.js"
-export * as InvariantT from "./InvariantT/index.js"
```
isthatcentered, line 51: @mikearnaldi Removed InvariantT, unless I'm mistaken this isn't needed with the new system

isthatcentered:
@mikearnaldi I think we're there, I'll have to take at everything again for the peace of mind but everything compiles, and all tests are passing (Ill update the PR with all the notable changes).

PR #1056, opened 2022-01-24
Update Typescript Extensions: Support unions, overloads, type unification, and global pipe function

mikearnaldi:


mikearnaldi:
> This looks great to me - one question though, do we still need to manually pass through `__etsTrace`?

For posterity, we only need to fill the `__etsTrace` in the primitives, where they should be required fields of type `string | undefined`, everything else will be autofilled

PR #1062, opened 2022-01-25
improve type signature of NonEmptyArray.make

tatchi:
I personally find this signature better.


packages/system/src/Testing/FailureDetails/index.ts
```diff
@@ -16,7 +16,7 @@ export class FailureDetails {
 export function label_(self: FailureDetails, str: string): FailureDetails {
   const [h, ...tail] = self.assertion
 
-  return new FailureDetails(NEA.make([l.label_(h, str), ...tail]))
+  return new FailureDetails([l.label_(h, str), ...tail])
```
tatchi, line 19: Noticed that it works without the `make` functions so I just removed it.

mikearnaldi:
cherry-picked and rebased in main

PR #1079, opened 2022-02-07
feat(Ord): add tuple

tatchi:
See https://discord.com/channels/795981131316985866/795983589644304396/940166457323561002

I couldn't find such a function so decided to open a PR.

### DIFF
--- a/packages/system/src/Ord/operations.ts
+++ b/packages/system/src/Ord/operations.ts
@@ -3,6 +3,7 @@
 import type { Equal } from "../Equal/index.js"
 import { makeEqual } from "../Equal/index.js"
 import type { Ordering } from "../Ordering/index.js"
+import type { ForcedTuple } from "../Utils/index.js"
 import type { Ord } from "./definition.js"
 
 /**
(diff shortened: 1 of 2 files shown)


packages/system/src/Ord/operations.ts
```diff
@@ -94,3 +95,25 @@ export function inverted<A>(O: Ord<A>) {
 export function getEqual<A>(O: Ord<A>): Equal<A> {
   return makeEqual((x, y) => O.compare(x, y) === 0)
 }
+
+/**
+ * Given a tuple of `Ord`s returns an `Ord` for the tuple
+ */
+export function tuple<T extends ReadonlyArray<Ord<any>>>(
```
tatchi, line 102:
```ts
export function tuple<T extends ReadonlyArray<Ord<any>>>(
  ...ords: T
): Ord<
  ForcedTuple<{
    [K in keyof T]: T[K] extends Ord<infer A> ? A : never
  }>
> {
  return makeOrd((x, y) => {
    return pipe(
      ords,
      A.findFirstMapWithIndex((i, ord) => {
        const r = ord.compare(x.get(i), y.get(i))
        return r !== 0 ? O.some(r) : O.none
      }),
      O.getOrElse<Ordering>(() => 0)
    )
  })
}
```

Don't think it's really better though?

mikearnaldi: Kind of ok both ways, the current one seems slightly better perf wise

mikearnaldi (approved):

PR #1088, opened 2022-02-12
chore: introduce konfik

mikearnaldi:


IMax153 (approved):
This PR looks good, but the CI-CD pipe is failing.

schickling:
What was the problem with the `postinstall`?

mikearnaldi:
The target ES version is too high for the node version used, for
compatibility the bundled output should be plain es5/6

mikearnaldi:
Given the dependency between konfik and effect we won't be able to use konfik in this repo

schickling:
I think that's a misunderstanding. Effect shouldn't be a "transparent" dependency of Konfik as it is bundled into the Konfik CLI.

mikearnaldi:
Not if the model changes to allow plugins to use effect and given the model is still in discussion I'd delay until stable

PR #1100, opened 2022-02-18
chore: begin restructuring of builds

mikearnaldi:


mikearnaldi:
babel plugin to add extensions isn't working fine

PR #1154, opened 2022-03-31
adding hash caching for hashmap, hashset and generally on objects

sledorze:


mikearnaldi:
As per discord conclusion we can't assume transitive immutability

PR #1162, opened 2022-04-07
HKT restructuring

### LINKED ISSUE #1005: Proposal for HKT restructuring

isthatcentered:
Fixes #1005 

## What It looks like
Combined with fixed types and the ability to have multiple mappings/URIs for an HKT, we found `R`, `E`, `A` to be enough for pretty much every use case.

1. Declaring an HKT 
```typescript
export type Reader<R, A> = (r: R) => A

export interface ReaderF extends P.HKT {
  readonly type: Reader<this["R"], this["A"]>
}
```

3.  Declaring a transformer/Fixed types
```typescript
export interface ReaderT<F extends P.HKT> extends P.HKT {
  readonly type: Reader<this["R"], P.Kind<F, unknown, this["E"], this["A"]>>
}

export interface StringReader extends P.HKT {
  readonly type: Reader<string, this["A"]>
}

// or
export interface StringReader extends P.HKT {
  readonly type: P.Kind<ReaderF, string, this["E"], this["A"]>
}
```

4. Abstracting over HKTs
```typescript 
export interface AssociativeBoth<F extends P.HKT> extends P.Typeclass<F> {
  both: <R2, E2, B>(
    fb: P.Kind<F, R2, E2, B>
  ) => <R, E, A>(
    fa: P.Kind<F, R, E, A>
  ) => P.Kind<F, R2 & R, E2 | E, Tp.Tuple<[A, B]>>
}

export const AssociativeBoth = P.instance<P.AssociativeBoth<ReaderF>>({
  both: zip
})
```

5. But how do I implement my `Thing<S, R, E, A>` if there is only `R`, `E`, `A`?
A type can have multiple URIS/parameter mappings if needed. 
```typescript 
interface MyThingF<S> extends P.HKT {
    readonly type: Thing<S, this["R"], this["E"], this["A']>
}

interface MyThingStateCategoryF<R> extends P.HKT {
    readonly type: Thing<this["R"], R, this["E"], this["A']>
}

export const AssociativeComposeReader = <S>() => P.instance<P.AssociativeCompose<MyThingF<S>>>(...)
export const AssociativeComposeState =  <R>() => P.instance<P.AssociativeCompose<MyThingStateCategoryF<R>>>(...)
```

## (already known) Gotchas
1. No way to infer fixed types
```typescript 
export const FailValidation = P.instance<P.Fail<P.Validation<ReaderF, string[]>>({
    fail: <E>(e: E) => // implementation cannot use `e`since it can be anything
})

FailValidation.fail("totally_ok_to_pass_something_else_than_string_array")
```

Technically we could allow to fix a type at the instance declaration, but not sure we want to go there since (as far as I can think of) there aren't that many use cases where It would be useful. Aka significant toll, "ok at best" reward.

(History for this PR can be found [there](https://github.com/Effect-TS/core/pull/1055))


mikearnaldi:
superseded by work on `fp-ts`

PR #1175, opened 2022-04-14
chore: fix OneShot methods are used as properties, so should be properties.

patroza:


mikearnaldi:
superseded by https://github.com/Effect-TS/core/pull/1176

PR #1198, opened 2022-04-23
chore: remove error prone __call

mikearnaldi:


mikearnaldi:
cc @IMax153 the `__call` is kind of cryptic and has the awkward counter of being error prone when applied with plain pipe, given latest work on ts+ we can have a pipe operator as `/` that looks nicer for aspects `effect / summary`

PR #1205, opened 2022-05-10
Add custom lookup on map keys

isthatcentered:
Adds the ability to pass a custom matcher (eq) for map keys.

Added functions
- lookupEq (system)
- removeEq (system)
- getIdentityEq (core)
- getIdentity (core)

I'm afraid this might make the API for map more confusing than anything (also, @mikearnaldi made a very good point about using `HashMap` for this use case) so don't hesitate to reject the PR ;).

### DIFF
--- a/packages/core/src/Collections/Immutable/Map/index.ts
+++ b/packages/core/src/Collections/Immutable/Map/index.ts
@@ -3,3 +3,4 @@
 import "../../../Operator/index.js"
 
 export * from "@effect-ts/system/Collections/Immutable/Map"
+export * from "./instances.js"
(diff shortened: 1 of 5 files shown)


mikearnaldi:
I am a bit mixed, yes `HashMap` uses the custom equality but we don't have structures that deal with explicit Eq and those may become even more useful with the work that @0x706b has done for auto-parameters and the work we have done on derivation.

I think we should re-evaluate the api of HashMap and co and maybe of how custom equals and hash is implemented

isthatcentered:
Thanks for taking the time to review it @mikearnaldi :)

PR #1228, opened 2022-06-07
feat(TMap): add TMap for STM

wesselvdv:
Everything should be here


packages/core/_src/stm/TMap/definition.ts
```diff
@@ -0,0 +1,32 @@
...
+export const _K = Symbol.for("@effect/core/stm/TMap/K")
+export type _K = typeof _K
+
+export const _V = Symbol.for("@effect/core/stm/TMap/V")
+export type _V = typeof _V
+
+/**
+ * @tsplus type ets/TMap
+ */
+export interface TMap<K, V> {
```
IMax153, line 13: Depending on what you prefer, you could make `TMap` extend `Collection`, which gives you access to `Collection` methods in TS+.

You just then have to define a `Symbol.iterator` for the internal map representation.

See example: https://github.com/Effect-TS/core/blob/2128087016bd77d5ca8ce0709cd60b709d707f80/packages/core/_src/stm/TArray/definition.ts#L10-L13

packages/core/_src/stm/TMap/operations/get.ts
```diff
@@ -0,0 +1,29 @@
...
+  concreteTMap(self)
+  return STM.Effect((journal) => {
+    const buckets = self.tBuckets.unsafeGet(journal)
+
+    concreteTArray(buckets)
+
+    const idx = TMap.indexOf(k, buckets.chunk.length)
+    const bucket = buckets.chunk.unsafeGet(idx)!.unsafeGet(journal)
+
+    return bucket.find((_) => _.get(0) === k).map((_) => _.get(1))
```
IMax153, line 20: I think we may want to use `Equals.equals` for the equality check here since Scala is doing `==` which doesn't technically use reference equality.

packages/core/_src/stm/TMap/operations/delete.ts
```diff
@@ -0,0 +1,41 @@
...
+  concreteTMap(self)
+  return STM.Effect((journal) => {
+    const buckets = self.tBuckets.unsafeGet(journal)
+
+    concreteTArray(buckets)
+
+    const idx = TMap.indexOf(k, buckets.chunk.length)
+    const bucket = buckets.chunk.unsafeGet(idx)!.unsafeGet(journal)
+
+    const partitions = bucket.partition((_) => _.get(0) === k)
```
IMax153, line 20: Would use `Equals.equals` for the equality check here since Scala's `==` doesn't technically perform reference equality.

packages/core/_src/stm/TMap/operations/delete.ts
```diff
@@ -0,0 +1,41 @@
...
+    const toRemove = partitions.get(0)
```
IMax153, line 21: You could also destructure the tuple above:

```ts
const { tuple: [toRemove, toRetain] } = bucket.partition((_) => Equals.equals(_.get(0), k))
```

packages/core/_src/stm/TMap/operations/deleteAll.ts
```diff
@@ -0,0 +1,41 @@
...
+  return STM.Effect((journal) => {
+    Chunk.from(ks).forEach((k) => {
+      const buckets = self.tBuckets.unsafeGet(journal)
+
+      concreteTArray(buckets)
+
+      const idx = TMap.indexOf(k, buckets.chunk.length)
+      const bucket = buckets.chunk.unsafeGet(idx)!.unsafeGet(journal)
+
+      const partitions = bucket.partition((_) => _.get(0) === k)
```
IMax153, line 21: Same comment as others re: equality checks.

packages/core/_src/stm/TMap/operations/put.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const LOAD_FACTOR = .75 as const
+
+/**
+ * Stores new binding into the map.
+ *
+ * @tsplus fluent ets/TMap put
+ */
+export function put_<K, V>(self: TMap<K, V>, k: K, v: V): USTM<void> {
+  function resize(journal: Journal, buckets: TArray<List<Tuple<[K, V]>>>): void {
```
IMax153, line 14: I would extract this to a non-exported function in the global scope so that we are not re-creating this function every time `put` is called. You can pass any additional arguments required to the function that it may be accessing now because it's in scope.

packages/core/_src/stm/TMap/operations/put.ts
```diff
@@ -0,0 +1,82 @@
...
+    const newBuckets: Array<List<Tuple<[K, V]>>> = []
+    let i = 0
+
+    while (i < capacity) {
+      const pairs = buckets.chunk.unsafeGet(i)!.unsafeGet(journal)
+
+      for (const pair of pairs) {
+        const idx = TMap.indexOf(pair.get(0), newCapacity)
+
+        newBuckets[idx] = List.cons(pair, newBuckets[idx] as List<Tuple<[K, V]>>).asList()
```
IMax153, line 27: I don't think the call to `asList` is necessary here. And you can avoid the type-cast by just performing a non-null assertion: `newBuckets[idx]!`.

packages/core/_src/stm/TMap/operations/put.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Stores new binding into the map.
+ *
+ * @tsplus fluent ets/TMap put
+ */
+export function put_<K, V>(self: TMap<K, V>, k: K, v: V): USTM<void> {
+  function resize(journal: Journal, buckets: TArray<List<Tuple<[K, V]>>>): void {
+    concreteTArray(buckets)
+    const capacity = buckets.chunk.length
+    const newCapacity = capacity << 1
+    const newBuckets: Array<List<Tuple<[K, V]>>> = []
```
IMax153, line 18: In the ZIO implementation, they do:

```scala
val newBuckets  = Array.fill[List[(K, V)]](newCapacity)(Nil)
```

which is equivalent to:

```ts
const newBuckets: Array<List<Tuple<[K, V]>>> = Array.from({ length: newCapacity }, () => List.nil())
```

packages/core/_src/stm/TMap/operations/put.ts
```diff
@@ -0,0 +1,82 @@
...
+        const idx = TMap.indexOf(pair.get(0), newCapacity)
+
+        newBuckets[idx] = List.cons(pair, newBuckets[idx] as List<Tuple<[K, V]>>).asList()
+      }
+
+      i += 1
+    }
+
+    const newIdx = TMap.indexOf(k, newCapacity)
+    newBuckets[newIdx] = List.cons(Tuple.make(k, v), newBuckets[newIdx] as List<Tuple<[K, V]>>)
```
IMax153, line 34: Same comment as above.

packages/core/_src/stm/TMap/operations/put.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    const newArray: Array<TRef<List<Tuple<[K, V]>>>> = []
```
IMax153, line 36: In the ZIO implementation, they do:

```scala
val newArray = Array.ofDim[TRef[List[(K, V)]]](newCapacity)
```

which is equivalent to:

```ts
const newArray: Array<TRef<List<Tuple<[K, V>>>> = new Array(newCapacity)
```

packages/core/_src/stm/TMap/operations/put.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    i = 0
+    while (i < newCapacity) {
+      newArray[i] = TRef.unsafeMake(newBuckets[i]) as TRef<List<Tuple<[K, V]>>>
```
IMax153, line 40: Same comment as above.

packages/core/_src/stm/TMap/operations/put.ts
```diff
@@ -0,0 +1,82 @@
...
+  concreteTMap(self)
+  return STM.Effect((journal) => {
+    const buckets = self.tBuckets.unsafeGet(journal)
+
+    concreteTArray(buckets)
+
+    const capacity = buckets.chunk.length
+    const idx = TMap.indexOf(k, capacity)
+    const bucket = buckets.chunk.unsafeGet(idx)!.unsafeGet(journal)
+    const shouldUpdate = bucket.exists((_) => _.get(0) === k)
```
IMax153, line 58: Same comment re: equality checks.

packages/core/_src/stm/TMap/operations/put.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    if (shouldUpdate) {
+      const newBucket = bucket.map((kv) => kv.get(0) === k ? Tuple.make(k, v) : kv)
+      buckets.chunk.unsafeGet(idx)!.unsafeSet(newBucket, journal)
+    } else {
+      const newSize = self.tSize.unsafeGet(journal) + 1
+
+      self.tSize.unsafeSet(newSize, journal)
+
+      if (capacity * LOAD_FACTOR < newSize) resize(journal, buckets)
```
IMax153, line 68: Please put this statement in brackets with a newline

```ts
if (capacity * LOAD_FACTOR < newSize) {
  resize(journal, buckets)
} else {
  // ...
}
```

IMax153 (changes requested):
I didn't go through every method in detail, but I did go through `put` line by line with the ZIO implementation.

Additionally, migrating the `TMap` tests from ZIO will help sort out any bugs.

packages/core/_src/stm/TMap/definition.ts
wesselvdv: hmm... I just realised that I am not sure if this will work. The internal representation of the key value tuples are inside a `TRef`, which means I need a STM to access them, even when using `unsafeGet`. How would I go about doing this?

packages/core/_src/stm/TMap/operations/indexOf.ts
```diff
@@ -0,0 +1,14 @@
+import { hashUnknown } from "@tsplus/stdlib/structure/Hash"
+
+function hash<K>(k: K): number {
+  const h = hashUnknown(k)
```
IMax153, line 4: `Hash` is available globally, so you should do `Hash.unknown` - this way it will automatically resolve to the most specific file when compiling the import statement via TS+.

IMax153 (changes requested):

packages/core/_src/stm/TMap/operations/fromIterable.ts
```diff
@@ -0,0 +1,47 @@
...
+
+function nextPowerOfTwo(size: number): number {
+  const n = -1 >>> Math.clz32(size - 1)
+
+  return n < 0 ? 1 : n + 1
+}
+
+function allocate<K, V>(capacity: number, data: List<Tuple<[K, V]>>): USTM<TMap<K, V>> {
+  const buckets: Array<List<Tuple<[K, V]>>> = Array.from({ length: capacity }, () => List.nil())
+  const distinct = ImmutableMap.from<Array<Tuple<[K, V]>>>(...data)
```
IMax153, line 13: Please change to a native `Map` for performance.

packages/core/_src/stm/TMap/operations/takeFirst.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @tsplus fluent ets/TMap takeFirst
+ */
+export function takeFirst_<K, V, A>(
+  self: TMap<K, V>,
+  pf: (kv: Tuple<[K, V]>) => Option<A>
+): USTM<A> {
+  concreteTMap(self)
+  return STM.Effect<never, Option<A>>((journal) => {
+    let result = Option.emptyOf<A>()
```
IMax153, line 15: Can we use `Option.none` here and assert the type? Just to avoid the extra allocation.

IMax153 (changes requested):

IMax153 (approved):

PR #1232, opened 2022-06-11
Add multiple STM datatypes & fixes

wesselvdv:


packages/core/_test/stm/STM/taps.test.ts
```diff
@@ -46,42 +46,51 @@ describe.concurrent("STM", () => {
...
-      //     result      <- failedStm.tapError(e => errorRef.succeed(e)).either
-      //     tappedError <- errorRef.await
-      //   } yield (result, tappedError)
-      // assertM(tx.commit)(equalTo((Left("error"), "error")))
+      const result = await tx.unsafeRunPromise()
+
+      assert.isTrue(result)
     })
+
+    // it("tapError should apply the transactional function to the error result while keeping the effect itself", async () => {
```
wesselvdv, line 81: @mikearnaldi This test is failing expectedly (https://github.com/zio/zio/blob/a822917ad730910fd69f38c341ab134995cfae95/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala#L1197), but I can't seem to find a way to indicate this in a way that it actually doesn't crash and burn.

packages/core/_test/stm/TReentrantLock/locks.test.ts
```diff
@@ -0,0 +1,234 @@
...
+        const count = $(reader2.join())
+
+        return count === 1
+      })
+
+      const result = await program.unsafeRunPromise()
+
+      assert.isTrue(result)
+    })
+    it("1 write lock then 1 read lock, different fibers", async () => {
```
wesselvdv, line 64: @mikearnaldi or @IMax153 This test only has a difference with ZIO on the return value of option. Where ZIO expects it to be `Option.None`, whereas I am just getting a `Option.Some(Exit.Success(1))`. 

(https://github.com/zio/zio/blob/a822917ad730910fd69f38c341ab134995cfae95/core-tests/shared/src/test/scala/zio/stm/TReentrantLockSpec.scala#L57)

I couldn't really find a tangible reason why'd it would make sense for it to be a `Option.None`.

packages/core/_test/stm/TReentrantLock/locks.test.ts
```diff
@@ -0,0 +1,234 @@
...
+          }, ${locks})`
+        )
+        return rcount === 1 && option.isNone() && locks === 1
+      })
+
+      const result = await program.unsafeRunPromise()
+
+      assert.isTrue(result)
+    })
+    it("1 write lock then 1 write lock, different fibers", async () => {
```
wesselvdv, line 105: @mikearnaldi or @IMax153 This test only has a difference with ZIO on the return value of option. Where ZIO expects it to be `Option.None`, whereas I am just getting a `Option.Some(Exit.Success(1))`. 

(https://github.com/zio/zio/blob/a822917ad730910fd69f38c341ab134995cfae95/core-tests/shared/src/test/scala/zio/stm/TReentrantLockSpec.scala#L57)

I couldn't really find a tangible reason why'd it would make sense for it to be a `Option.None`.

packages/core/_test/stm/TReentrantLock/locks.test.ts
```diff
@@ -0,0 +1,234 @@
...
+        const wcount = $(ref.get())
+
+        return rcount === 1 && wcount === 1
+      })
+
+      const result = await program.unsafeRunPromise()
+
+      assert.isTrue(result)
+    })
+    it("upgrade read lock to write lock from same fiber", async () => {
```
wesselvdv, line 166: @mikearnaldi or @IMax153 This test only has a difference with ZIO on the value of the variable `option`. Where ZIO expects it to be `Option.None`, whereas I am just getting a `Option.Some(Exit.Success(1))`. 

(https://github.com/zio/zio/blob/a822917ad730910fd69f38c341ab134995cfae95/core-tests/shared/src/test/scala/zio/stm/TReentrantLockSpec.scala#L88)

I couldn't really find a tangible reason why'd it would make sense for it to be a `Option.None`.

PR #1237, opened 2022-06-23
fix(SubscriptionRef): fix implementation & add missing tests

wesselvdv:


wesselvdv:
I'll see if I can rework this to what was discussed on Discord. (follow ZIO more closely with (abstract) class inheritance and subtyping)

PR #1249, opened 2022-07-09
Add Gen and Sample data types

IMax153:


mikearnaldi (approved):

mikearnaldi (changes requested):
- Divergent from ZIO, Sample is tight to a concrete number while ZIO uses a generic Integral/Fractional.

- Tests are too slow, 4m+ added to test time isn't ok

PR #1264, opened 2022-08-03
Add `Differ` datatype

IMax153:
This PR is still WIP - the tests for `Differ.hashMap` are flaky, but they should not be. Further investigation is needed.


mikearnaldi:
superseded by https://github.com/ts-plus/stdlib/pull/198

PR #1289, opened 2022-08-30
Add `Reloadable`

IMax153:
This PR adds the data types required for `Cached` and `Reloadable` (namely `ScopedRef`.

Of note, we may want to review the `Reloadable` constructors for a more concise/streamlined API. Perhaps we can extract one of the required tags from the `Layer` that is also passed in?


mikearnaldi:
Not sure how we would be able to extract such tag, I think if the tag for Reloadable<A> is unique meaning there is no possibility of 2 being in th  Env (whihh should be the case) then we could derive the tag of Reloadable<A> from the Tag<A>.

Additionally i believe all this modules should be natively fluent like Ref is,  and we should also remove setAsync as we did in Ref.

mikearnaldi (approved):

PR #1441, opened 2023-09-29
auto install pnpm version from package.json

fubhy:
`pnpm/setup` can automatically install the right pnpm version using corepack with the version defined in package.json. Same with the corepack install in `nix`. The benefit of using `corepack` & defining the package manage rin package.json is that it then alerts incorrect use of e.g. `yarn` within the repo.

Caching is also solved by `actions/setup-node` already.

... We could of course also use `nix` for the complete setup ...


tim-smart:
If you resolve the conflicts we can merge :)

PR #1450, opened 2023-09-29
hotfix exports types condition for consumers with `NodeNext` module resolution

fubhy:
Discussion: https://discord.com/channels/795981131316985866/1157260511474360381

### DIFF
--- a/scripts/gen-packagejson.mjs
+++ b/scripts/gen-packagejson.mjs
@@ -2,7 +2,7 @@ import * as Fs from "node:fs";
 import Package from "../package.json" assert { type: "json" };
 
 const modules = Fs.readdirSync("src").filter(
-  (_) => _.endsWith(".ts") && _ !== "index.ts",
+  (_) => _.endsWith(".ts") && _ !== "index.ts"
 );
 const files = ["dist", "internal", ...modules.map((_) => _.slice(0, -3))];
 const exports = {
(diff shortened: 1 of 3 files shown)


scripts/gen-packagejson.mjs
```diff
@@ -21,6 +22,7 @@ modules
       module: `./${module}/dist/effect-${module}.esm.js`,
       import: `./${module}/dist/effect-${module}.cjs.mjs`,
       default: `./${module}/dist/effect-${module}.cjs.js`,
+      types: `./dist/declarations/src/${module}.d.ts`,
```
Andarist, line 25: all `exports` keys are order-sensitive so this should be the first one here

fubhy: Hm, in which situations would this matter? Because it already works with this in my repro.

Andarist: I'd expect that this relies on the "fallback condition" thing or however it's called but this doesn't list it as a problem: https://arethetypeswrong.github.io/?p=effect%402.0.0-next.39 . Note that this "fallback" thing is actually a bug in TS - for the time being it continues to work with it but it shouldn't. 

The issue here might be a different one. it's hard to tell on the spot without analyzing further - it's not worth spending time on it though ;p Rules-wise - it should be on the top (or at least it's usually advised for it to be at the top). If it works right now... maybe it is not worth changing it. 🤷‍♂️  Personally I'd probably change it though.

PR #1455, opened 2023-10-01
feat: add Streamable constructor

TylorS:
I figured I'd just open a PR to start some discussion on an equivalent to Effectable for Stream. 

The modification to `toChannel` is to favor choosing the Stream instance for any custom type which might be both a Stream and an Effect. This felt kinda odd, but I wasn't sure what else to do without larger changes to Stream's representation.

### DIFF
--- a/.changeset/mighty-tomatoes-visit.md
+++ b/.changeset/mighty-tomatoes-visit.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+add Streamable for creating custom Streams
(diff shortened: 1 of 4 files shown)


tim-smart:
I think this solution is fine :)

If you fix docgen then we can merge this.

TylorS:
@tim-smart Thanks, I've updated the PR with the updated documentation

Hub/package.json
tim-smart, line 1: Could you remove the build artifacts? Thanks.

tim-smart (changes requested):

Hub/package.json
TylorS: Updated the PR without the build artifacts

tim-smart (approved):

PR #1460, opened 2023-10-02
add bigint metrics

fubhy:
Opening this for comments ... I don't like this approach. I think I'd prefer a more generic internal API by adding value subtypes for counters, gauges, etc. so we do not inflate the public api further by introducing whole bigint-specific metric types.

### DIFF
--- a/src/Metric.ts
+++ b/src/Metric.ts
@@ -89,7 +89,23 @@ export declare namespace Metric {
    * @since 2.0.0
    * @category models
    */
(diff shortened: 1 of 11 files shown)


tim-smart:
I had a play around to see if we could avoid create new apis: https://github.com/Effect-TS/effect/pull/1465

tim-smart:
Did some more refinement in the other PR. Will get this supported in the otel package shortly.

PR #1463, opened 2023-10-02
Rename Hub to PubSub, closes #1462

### LINKED ISSUE #1462: Rename Hub to PubSub

gcanti:


tim-smart:
The name does feel a little weird to me - pub/sub is a software pattern, rather than a concrete "thing". But if @mikearnaldi is happy then we can merge.

mikearnaldi:
I had a bit of thinking, I kind of agree with Tim but also think that PubSub will help people easily grasp what the type is about, maybe worth the rename. Can you pls rebase it @gcanti , feel free to merge

PR #1500, opened 2023-10-08
feat: allow log annotations to be any object

sukovanej:
For example, in a project which we're migrating from fp-ts to effect, we are using a winston json logger for the production with Datadog that is able to deal with structured logging where annotation values are arbitrary JSON objects.


sukovanej:
For example, @tim-smart do you think such a change is reasonable? If so, could you please review? Thanks in advance.

tim-smart:
I think this is OK as we are already accepting unknown for the message. I'm just going to hijack this and take a look at the Tracer at the same time.

PR #1503, opened 2023-10-09
Fix Stream.groupAdjacentBy when group spans multiple chunks

VenomAV:
I noticed that Stream.groupAdjacentBy failed when a group spans multiple chunks. I added some tests to highlight the behavior and fixed the code.
Let me know if I can improve my PR.


tim-smart:
LGTM! Thanks for the contribution.

PR #1521, opened 2023-10-14
Enforce that members passed to TaggedEnum do not have a `_tag` property

ahrjarrett:
Closes https://github.com/Effect-TS/effect/issues/1497

Note: I was not intending to work on this issue -- originally I was actually contributing a minor type-level performance improvement to TaggedEnum, and when I went to open the PR, I checked to see if there were any related issues or pull requests, and saw #1497 (and happened to know how to make that work).

Note that:
1. users will see a type error if they try to create a `TaggedEnum` when one their members has a `_tag` property
2. even though they get the type error, the `TaggedEnum` still gets instantiated, and the `_tag` property is overwritten

Not 100% sure if this is the behavior that makes the most sense, but it seemed the most reasonable to me.

[TS v5.2 Playground](https://tsplay.dev/Wk3GjW)
[TS v4.8 Playground](https://tsplay.dev/WvQEYm)

Happy to make any requested adjustments, and I apologize if any of the names aren't clear -- I wrote this rather quickly, and figured better to get feedback before spending too long on it :)

### DIFF
--- a/src/Data.ts
+++ b/src/Data.ts
@@ -175,7 +175,6 @@ type TaggedEnum<
   : never
   ;
 
-/** @internal */
 interface _phantom<A> {
   _A: A
   keys: (this[`_A`][keyof this[`_A`]]) extends infer T ? T extends T ? keyof T : never : never;
--- a/src/Data.ts
+++ b/src/Data.ts
@@ -164,31 +164,23 @@ export const Structural: new<A>(
  * @since 2.0.0
  * @category models
  */
-type TaggedEnum<
-  A extends Record<string, Record<string, any>> & Untagged<_>,
-  _ extends _phantom<A> = _phantom<A>
->
-  = keyof A extends | infer Tag
-  ? Tag extends keyof A
-  ? Data<{ readonly [K in `_tag` | keyof A[Tag]]: K extends `_tag` ? Tag : A[Tag][K] }>
+export type TaggedEnum<
+  A extends Record<string, Record<string, any>> & _,
+  _ extends ChildrenUntagged<A> = ChildrenUntagged<A>
+> = keyof A extends infer Tag
+  ? Tag extends keyof A ? Data<{ readonly [K in `_tag` | keyof A[Tag]]: K extends `_tag` ? Tag : A[Tag][K] }>
   : never
   : never
-  ;
 
-interface _phantom<A> {
-  _A: A
-  keys: (this[`_A`][keyof this[`_A`]]) extends infer T ? T extends T ? keyof T : never : never;
-  isTagged: `_tag` extends this[`keys`] ? true : false
-}
+type ChildHasDiscriminant<A> = `_tag` extends ChildKeys<A> ? true : false
 
-/** @internal */
-type Untagged<_ extends { isTagged: any }>
-  = [_[`isTagged`]] extends [true]
+type ChildKeys<A> = A[keyof A] extends infer M ? M extends M ? keyof M : never : never
+
+type ChildrenUntagged<_> = [ChildHasDiscriminant<_>] extends [true]
   ? `It looks like you're trying to create a tagged enum, but one or more of its members already has a \`_tag\` property.`
   : unknown
-  ;
 
-/**
+/** ,
  * @since 2.0.0
  */
 export declare namespace TaggedEnum {
--- a/src/Data.ts
+++ b/src/Data.ts
@@ -164,12 +164,30 @@ export const Structural: new<A>(
  * @since 2.0.0
  * @category models
  */
-export type TaggedEnum<A extends Record<string, Record<string, any>>>
-  = keyof A extends | infer Tag 
-  ? Tag extends keyof A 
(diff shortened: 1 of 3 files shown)


ahrjarrett:
Re-reading the changes, I'm not particularly happy with the name `CheckUntagged`. Maybe it makes more sense to just it `Untagged`?

src/Data.ts
```diff
@@ -164,11 +164,30 @@ export const Structural: new<A>(
...
+interface _phantom<A> {
+  _A: A
+  keys: (this[`_A`][keyof this[`_A`]]) extends infer T ? T extends T ? keyof T : never : never;
+  isTagged: `_tag` extends this[`keys`] ? true : false
+}
+
+/** @internal */
+type CheckUntagged<_ extends { isTagged: any }>
+  = [_[`isTagged`]] extends [true]
+  ? `It looks like you're trying to create a tagged enum, but one or more of its members already have a \`_tag\` property.`
```
ahrjarrett, line 188:
```suggestion
  ? `It looks like you're trying to create a tagged enum, but one or more of its members already has a \`_tag\` property.`
```

src/Data.ts
```diff
@@ -164,11 +164,30 @@ export const Structural: new<A>(
  * @since 2.0.0
  * @category models
  */
-export type TaggedEnum<A extends Record<string, Record<string, any>>> = {
-  readonly [Tag in keyof A]: Data<
-    Readonly<Types.Simplify<A[Tag] & { _tag: Tag }>>
-  >
-}[keyof A]
+type TaggedEnum<
+  A extends Record<string, Record<string, any>> & CheckUntagged<_>,
```
ahrjarrett, line 168:
```suggestion
  A extends Record<string, Record<string, any>> & Untagged<_>,
```

src/Data.ts
```diff
@@ -164,11 +164,30 @@ export const Structural: new<A>(
...
+
+/** @internal */
+interface _phantom<A> {
+  _A: A
+  keys: (this[`_A`][keyof this[`_A`]]) extends infer T ? T extends T ? keyof T : never : never;
+  isTagged: `_tag` extends this[`keys`] ? true : false
+}
+
+/** @internal */
+type CheckUntagged<_ extends { isTagged: any }>
```
ahrjarrett, line 186:
```suggestion
type Untagged<_ extends { isTagged: any }>
```

src/Data.ts
```diff
@@ -164,11 +164,30 @@ export const Structural: new<A>(
...
+  _ extends _phantom<A> = _phantom<A>
+>
+  = keyof A extends | infer Tag
+  ? Tag extends keyof A
+  ? Data<{ readonly [K in `_tag` | keyof A[Tag]]: K extends `_tag` ? Tag : A[Tag][K] }>
+  : never
+  : never
+  ;
+
+/** @internal */
```
tim-smart, line 178: Will need to remove the internal comments otherwise the types will be stripped away.

tim-smart (changes requested):
Almost there thanks! Definitely an interesting concoction of types :)

I wonder if it can be simplified at all? My only concern is from a maintainability point of view - in a few months time I can imagine myself staring at it for a few minutes before figuring out what is going on.

ahrjarrett:
> Almost there thanks! Definitely an interesting concoction of types :)
> 
> I wonder if it can be simplified at all? My only concern is from a maintainability point of view - in a few months time I can imagine myself staring at it for a few minutes before figuring out what is going on.

Fair enough, I can take another stab at it.

src/Data.ts
ahrjarrett, line 178:
```suggestion
```

gcanti:
@ahrjarrett what about

```ts
export type TaggedEnum<A extends Record<string, Record<string, any>> & UntaggedChildren<A>> = {
  readonly [Tag in keyof A]: Data<
    { readonly [K in keyof A[Tag] | "_tag"]: K extends "_tag" ? Tag : A[Tag][K] }
  >
}[keyof A]

type ChildrenAreTagged<A> = { [K in keyof A]: "_tag" extends keyof A[K] ? true : false }[keyof A]

type UntaggedChildren<A> = ChildrenAreTagged<A> extends true
  ? "It looks like you're trying to create a tagged enum, but one or more of its members already has a `_tag` property."
  : unknown
```

type-level tests to add in `/dtslint/Data.ts`

```ts
import * as Data from 'effect/Data'

type HttpError = Data.TaggedEnum<{
  BadRequest: { status: 400, a: string }
  NotFound: { status: 404, b: number }
}>
// $ExpectType Data<{ readonly a: string; readonly status: 400; readonly _tag: "BadRequest"; }>
type BadRequest = Extract<HttpError, { _tag: 'BadRequest' }>
// $ExpectType Data<{ readonly b: number; readonly status: 404; readonly _tag: "NotFound"; }>
type NotFound = Extract<HttpError, { _tag: 'NotFound' }>

// @ts-expect-error
type Err = Data.TaggedEnum<{
  BadRequest: { _tag: 'BadRequest' }
}>
```

tim-smart (approved):

PR #1525, opened 2023-10-16
Removes unnecessary type parameter from TaggedEnum

ahrjarrett:
Per conversation in https://github.com/Effect-TS/effect/pull/1521 (thank you @gcanti, I didn't realize type parameters could be self-referential like that 😮), this PR removes the phantom type introduced in that PR.

I think we should remove it if possible since it introduces the possibility of confusing users. I previously thought that removing it wasn't possible, and that we were making a tradeoff for which behavior we felt was less surprising, but if the type and its constraint can be embedded into one, I'm all for it.

### DIFF
--- a/.changeset/short-crabs-beg.md
+++ b/.changeset/short-crabs-beg.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+removes unnecessary type parameter from TaggedEnum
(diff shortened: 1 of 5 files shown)


ahrjarrett:
Also I wanted to socialize the original change that I came here to make, which was to remove the unnecessary object instantiations (only to throw them away).

Here's a really long-winded explanation with examples talking about it, but if you're like me (skeptical), I'd recommend installing a tool like [attest](https://github.com/arktypeio/arktype/tree/main/dev/attest) locally and using it to see how many instantiations are created, and how those instantiations grow in proportion to the size of their inputs.

https://tsplay.dev/wRPRvm

PR #1528, opened 2023-10-16
Add BigDecimal module

fubhy:
Next steps (preferably follow-up):

- more math functions (sqrt, exp, etc.)
- rounding modes & precision rounding

### DIFF
--- a/src/BigDecimal.ts
+++ b/src/BigDecimal.ts
@@ -2,6 +2,15 @@
  * This module provides utility functions and type class instances for working with the `BigDecimal` type in TypeScript.
  * It includes functions for basic arithmetic operations, as well as type class instances for `Equivalence` and `Order`.
  *
+ * A `BigDecimal` allows storing any real number to arbitrary precision; which avoids common floating point errors
+ * (such as 0.1 + 0.2 ≠ 0.3) at the cost of complexity.
+ *
+ * Internally, `BigDecimal` uses a `BigInt` object, paired with a 64-bit integer which determines the position of the
+ * decimal point. Therefore, the precision *is not* actually arbitrary, but limited to 2<sup>63</sup> decimal places.
+ *
+ * It is not recommended to convert a floating point number to a decimal directly, as the floating point representation
+ * may be unexpected.
+ *
  * @since 2.0.0
  */
 
@@ -15,6 +24,8 @@ import * as Option from "./Option"
 import * as order from "./Order"
 import type { Ordering } from "./Ordering"
 import { type Pipeable, pipeArguments } from "./Pipeable"
+
+const DEFAULT_PRECISION = 100
 
 /**
  * @since 2.0.0
@@ -36,8 +47,6 @@ export interface BigDecimal extends Equal.Equal, Pipeable, Inspectable {
   readonly [TypeId]: TypeId
   readonly value: bigint
   readonly scale: number
-  /** @internal */
-  normalized?: BigDecimal | undefined
 }
 
 const BigDecimalProto: Omit<BigDecimal, "value" | "scale" | "normalized"> = {
@@ -110,6 +119,7 @@ export const make = (value: bigint | number): BigDecimal => {
 }
 
 const bigint0 = BigInt(0)
+const bigint1 = BigInt(1)
 const bigint10 = BigInt(10)
 const zero = make(bigint0)
 
@@ -128,35 +138,28 @@ const zero = make(bigint0)
  * @category constructors
  */
 export const normalize = (self: BigDecimal): BigDecimal => {
-  if (self.normalized !== undefined) {
-    return self.normalized
-  }
-
   if (self.value === bigint0) {
-    self.normalized = zero
-  } else {
-    const digits = `${self.value}`.split("")
+    return zero
+  }
 
-    let trail = 0
-    for (let i = digits.length - 1; i >= 0; i--) {
-      if (digits[i] === "0") {
-        trail++
-      } else {
-        break
-      }
-    }
+  const digits = `${self.value}`
 
-    if (trail === 0) {
-      self.normalized = self
+  let trail = 0
+  for (let i = digits.length - 1; i >= 0; i--) {
+    if (digits[i] === "0") {
+      trail++
     } else {
-      digits.splice(digits.length - trail)
-      const value = BigInt(digits.join(""))
-      const scale = self.scale - trail
-      self.normalized = scaled(value, scale)
+      break
     }
   }
 
-  return self.normalized
+  if (trail === 0) {
+    return self
+  }
+
+  const value = BigInt(digits.substring(0, digits.length - trail))
+  const scale = self.scale - trail
+  return scaled(value, scale)
 }
 
 /**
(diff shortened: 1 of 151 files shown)


src/BigDecimal.ts
```diff
@@ -0,0 +1,717 @@
...
+export interface BigDecimal extends Equal.Equal, Pipeable, Inspectable {
+  readonly [TypeId]: TypeId
+  readonly value: bigint
+  readonly scale: number
+}
+
+const BigDecimalProto: Omit<BigDecimal, "value" | "scale"> = {
+  [TypeId]: TypeId,
+  [Hash.symbol](this: BigDecimal): number {
+    return Hash.structure({ value: this.value, scale: this.scale })
```
tim-smart, line 40: You will need to first normalize the value, otherwise you will get false negatives using Equal.equals.

src/BigDecimal.ts
```diff
@@ -0,0 +1,717 @@
...
+export const make = (value: bigint | number): BigDecimal => {
+  if (typeof value === "number") {
+    const [lead, trail = ""] = `${value}`.split(".")
+    return scaled(BigInt(`${lead}${trail}`), trail.length)
+  }
+
+  return scaled(value, 0)
+}
+
+const zero = make(0n)
```
tim-smart, line 102: If you use the `0n` syntax, the module won't work with Hermes for React Native (unless they have added support recently)

src/BigDecimal.ts
```diff
@@ -0,0 +1,717 @@
...
+} = dual(2, (self: BigDecimal, that: BigDecimal): BigDecimal => {
+  if (that.value === 0n) {
+    return self
+  }
+
+  if (self.value === 0n) {
+    return that
+  }
+
+  if (self.scale > that.scale) {
```
tim-smart, line 239: Getting equivalent scales is quite common, so maybe a helper could be added?

```ts
const [a, b] = whateverify(self, that)
```

src/BigDecimal.ts
```diff
@@ -0,0 +1,717 @@
...
+  return scaled(scale(self, max).value % scale(divisor, max).value, max)
+})
+
+/**
+ * @category instances
+ * @since 2.0.0
+ */
+export const Equivalence: equivalence.Equivalence<BigDecimal> = equivalence.make((self, that) => {
+  if (self.scale > that.scale) {
+    return scale(that, self.scale).value === that.value
```
tim-smart, line 657:
```suggestion
    return scale(that, self.scale).value === self.value
```

test/BigDecimal.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  it("isBigDecimal", () => {
+    expect(BigDecimal.isBigDecimal(BigDecimal.make(1n))).toEqual(true)
+    expect(BigDecimal.isBigDecimal(1)).toEqual(false)
+    expect(BigDecimal.isBigDecimal("a")).toEqual(false)
+    expect(BigDecimal.isBigDecimal(true)).toEqual(false)
+  })
+
+  it("sum", () => {
+    deepStrictEqual(pipe(BigDecimal.make(1n), BigDecimal.sum(BigDecimal.make(2n))), BigDecimal.make(3n))
```
tim-smart, line 21: Will need to add some more test coverage for each operator with different scales.

tim-smart (changes requested):
Here is some early feedback. I think the data type itself is the right direction :)

src/BigDecimal.ts
tim-smart, line 40:
```suggestion
    const normalized = normalize(this)
    return pipe(
      Hash.hash(normalized.value),
      Hash.combine(Hash.number(normalized.scale))
    )
```

test/BigDecimal.ts
fubhy: Oh yeah these are just the ones I copied over so far. I am going to copy the test cases from the bigdecimal.rs rust crate.

src/BigDecimal.ts
fubhy: I thought about that but since these are operations that might get called countless times I also wanted to keep things as optimized as possible. Allocating a new array instance (sometimes even needlessly if the scale is already identical) seemed wasteful. Microoptimization, sure, but it is at the data type level after all.

src/BigDecimal.ts
fubhy: this seems to be fixed in Hermes

src/BigDecimal.ts
tim-smart: 👍🏼

src/BigDecimal.ts
```diff
@@ -0,0 +1,759 @@
...
+import * as equivalence from "./Equivalence"
+import { dual, pipe } from "./Function"
+import * as Hash from "./Hash"
+import { type Inspectable, NodeInspectSymbol, toString } from "./Inspectable"
+import * as Option from "./Option"
+import * as order from "./Order"
+import type { Ordering } from "./Ordering"
+import { type Pipeable, pipeArguments } from "./Pipeable"
+
+const TypeId: unique symbol = Symbol.for("effect/BigDecimal")
```
tim-smart, line 19: For consistency with the other modules, lets export this :)

src/BigDecimal.ts
```diff
@@ -0,0 +1,759 @@
...
+ *
+ * assert.deepStrictEqual(normalize(make(123.456)), make(123.456))
+ * assert.deepStrictEqual(normalize(make(123.456000)), make(123.456))
+ * assert.deepStrictEqual(normalize(make(123.000456)), make(123.000456))
+ * assert.deepStrictEqual(normalize(make(123.000000)), make(123))
+ *
+ * @since 2.0.0
+ * @category constructors
+ */
+export const normalize = (self: BigDecimal): BigDecimal => {
```
tim-smart, line 169: We could probably cache the normalized result inside the BigDecimal structure.

That should improve performance of the Hash implementation.

src/BigDecimal.ts
```diff
@@ -0,0 +1,759 @@
...
+    return pipe(
+      Hash.hash(normalized.value),
+      Hash.combine(Hash.number(normalized.scale))
+    )
+  },
+  [Equal.symbol](this: BigDecimal, that: unknown): boolean {
+    return isBigDecimal(that) && equals(this, that)
+  },
+  toString(this: BigDecimal) {
+    return toString(this.toJSON())
```
tim-smart, line 50: I think we should use `format` here?

src/BigDecimal.ts
```diff
@@ -0,0 +1,759 @@
...
+ * import { format, make } from 'effect/BigDecimal'
+ *
+ * assert.deepStrictEqual(format(make(-5n)), '-5')
+ * assert.deepStrictEqual(format(make(123.456)), '123.456')
+ * assert.deepStrictEqual(format(make(-0.00000123)), '-0.00000123')
+ *
+ * @since 2.0.0
+ * @category constructor
+ */
+export const format = (n: BigDecimal): string => {
```
tim-smart, line 735: For interop with the js ecosystem, an `unsafeToNumber` could be useful.

src/BigDecimal.ts
```diff
@@ -0,0 +1,759 @@
...
+ *
+ * @example
+ * import { format, make } from 'effect/BigDecimal'
+ *
+ * assert.deepStrictEqual(format(make(-5n)), '-5')
+ * assert.deepStrictEqual(format(make(123.456)), '123.456')
+ * assert.deepStrictEqual(format(make(-0.00000123)), '-0.00000123')
+ *
+ * @since 2.0.0
+ * @category constructor
```
tim-smart, line 733: I think "conversions" is a better docs category for this.

tim-smart (changes requested):

src/BigDecimal.ts
fubhy: You mean in the original `self` in case `normalize` is called on it again? Yeah that's clever!

src/BigDecimal.ts
fubhy: Simply as `(_) => Number(format(_))`?

src/BigDecimal.ts
tim-smart: Yeah, that seems reasonable.

src/BigDecimal.ts
tim-smart: Yeah I guess that is the best way, otherwise unsafely casting the bigint then scaling it down.

fubhy: Can't scale down if it has a negative scale (cut off trailing zeros)

tim-smart (approved):

tim-smart:
If you run pnpm gen one more time, then we can merge :)

PR #1534, opened 2023-10-17
Remove `Number` module `Option` variants

### LINKED ISSUE #1531: From Discord: Moving `Option` Math Utils to Data Primitive Modules for Better Organization and Potential Additi...
# Summary
The user is referring to a specific section of code in the `Option` module of the Effect-TS library on GitHub. They are suggesting that the math utility functions within this section should be moved to the data primitive modules for better organization and potential additional functionality.

# Discord thread

https://discord.com/channels/795981131316985866/1163767122745376828

fubhy:
Closes #1531 

https://discord.com/channels/795981131316985866/1163767122745376828/1163811579557580870


src/Option.ts
```diff
@@ -1117,98 +1103,6 @@ export const exists: {
...
-      const a: number = oa.value
-      if (a === 0) {
-        return 0
-      }
-      out *= a
-    }
-  }
-  return out
-}
-
```
fubhy, line 1211: We might want to keep these ... thoughts?

jessekelly881: Why do you want to remove lift2?

jessekelly881: multiplyCompact is the only fn here that can't be easily implemented as an alias of another fn. `sumCompact` is easily implemented in terms of `reduceCompact` but multiplyCompact has a performance improvement vs using `reduceCompact` because it ends computation when it finds a 0. I think for that reason it should be kept. Otherwise, the rest are somewhat useless.

tim-smart: We can look at adding it back to the Number module if someone feels it is missing.

docs/modules/Option.ts.md
```diff
@@ -63,15 +63,7 @@ Added in v2.0.0
   - [isOption](#isoption)
   - [isSome](#issome)
 - [lifting](#lifting)
-  - [lift2](#lift2)
```
tim-smart, line 66: Was removing lift2 intentional?

tim-smart: Going to add it back, to keep an alternative api for users to migrate to if required.

PR #1545, opened 2023-10-21
fix vitest type resolution

fubhy:
This fixes type resolution in the test suite (currently vitest types are not being resolved)


IMax153 (approved):

IMax153:
cc @tim-smart @mikearnaldi for further input before merging in case I missed something

PR #1546, opened 2023-10-21
remove unused dependencies

fubhy:


mikearnaldi:
Doesn't this break typing for test expect?

fubhy:
> Doesn't this break typing for test expect?

No. The expect types are included & re-exported from vitest

PR #1555, opened 2023-10-24
ReadonlyArray / List / Chunk: merge mapNonEmpty with map

gcanti:
@tim-smart there are many other "NonEmpty" functions that can be merged, I wonder if it's better to test the merge technique (`Infer` + `With`) in userland on this limited portion of APIs, or proceed with the merge of all the other APIs, what do you think?

### DIFF
--- a/.changeset/honest-peaches-drop.md
+++ b/.changeset/honest-peaches-drop.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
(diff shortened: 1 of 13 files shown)


tim-smart:
Yeah we can release this and do the others in another PR.

tim-smart (approved):

PR #1560, opened 2023-10-24
use Proxy for TaggedEnum constructors

tim-smart:


mikearnaldi:
We may want to change:

```ts
const Result = Data.taggedEnum<ResultDefinition>()
const Success = Result.Success
const Failure = Result.Failure
```

to:

```ts
const { Success, Failure } = Data.taggedEnum<ResultDefinition>()
```

as a convention

tim-smart:
Sounds good. I added back support for the previous syntax - but I feel like the api is too complex with both.

tim-smart:
> Sounds good. I added back support for the previous syntax - but I feel like the api is too complex with both.

Yeah if we push for destructuring as the recommended usage pattern, I think dropping the function constructors is reasonable.

mikearnaldi:
> > Sounds good. I added back support for the previous syntax - but I feel like the api is too complex with both.
> 
> Yeah if we push for destructuring as the recommended usage pattern, I think dropping the function constructors is reasonable.

I kind of agree yeah

PR #1562, opened 2023-10-24
add RequestResolver.fromEffectTagged

tim-smart:


mikearnaldi:
Not sure it works type wise, the request type could be a union of different requests and a generic error can't be used to complete all the potential requests

tim-smart:
Yes in that case in wouldn't make sense, the thought was to have an api for easily lifting an `Effect<never, E, Iterable<A>>` into a batched resolver.

mikearnaldi:
> Yes in that case it wouldn't make sense, the thought was to have an api for easily lifting an `Effect<never, E, Iterable<A>>` into a batched resolver.

but how do you prevent usage with multiple request types? maybe some kind of check in the param

tim-smart:
This could be cool:

```ts
const Resolver = Resolver.fromEffectTags<UserById | AllUserIds>()({
  UserById: reqs => Effect.forEach(reqs, getUser, { concurrency: "unbounded" }),
  AllUserIds: reqs => Effect.forEach(reqs, () => allUserIds),
})
```

mikearnaldi:
> This could be cool:
> 
> ```ts
> const UserResolver = Resolver.fromEffectTags<UserById | AllUserIds>()({
>   UserById: reqs => usersById(reqs.map(_ => _.id)),
>   AllUserIds: reqs => Effect.forEach(reqs, () => allUserIds),
> })
> ```

It should be possible, requests have a tag

Issue #1498, opened 2023-10-07
ConfigProvider.fromEnv not compatible with vite. 


jessekelly881:
https://github.com/Effect-TS/effect/blob/56ddb9c7d6a2d655038fcf6696e74466c9dd8440/src/internal/configProvider.ts#L107
https://vitejs.dev/guide/env-and-mode.html

Vite stores env variables under import.meta.env, not under process.env. Can we update this line to test for import.meta.env as well?

mikearnaldi:
isn't import.meta available only in esm?

jessekelly881:
Maybe.. The [mdn docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta) don't mention esm specifically but I'm honestly not sure.

mikearnaldi:
> Maybe.. The [mdn docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta) don't mention esm specifically but I'm honestly not sure.

Yup it's esm only:

```
michaelarnaldi@Michaels-MacBook-Pro effect % node 
Welcome to Node.js v18.16.1.
Type ".help" for more information.
> import.meta;
import.meta;
       ^^^^

Uncaught SyntaxError: Cannot use 'import.meta' outside a module
> 
```

mikearnaldi:
Only way I can think of working around this is to have a sub-module, something like `effect/env` that internally contains a conditional package.json and two files:
- index.cjs
- index.mjs
where each file exports the `env` object, either found it via process or import.meta or better a merge of the two

cc @tim-smart for ideas

mikearnaldi:
@jessekelly881 would you mind giving a try to #1567 and see if it fixes the issue

mikearnaldi:
Ah it would actually not work because vite needs the statically know the keys https://vitejs.dev/guide/env-and-mode.html#production-replacement

mikearnaldi:
Actually given https://vitejs.dev/guide/env-and-mode.html#production-replacement there is no way to make it compatible, the only way would be using the Map provider and manually listing import.meta.env.KEY entries

fubhy:
> Actually given https://vitejs.dev/guide/env-and-mode.html#production-replacement there is no way to make it compatible, the only way would be using the Map provider and manually listing import.meta.env.KEY entries

[...]

PR #1567, opened 2023-10-26
Prototype vite env integration

mikearnaldi:
Addresses https://github.com/Effect-TS/effect/issues/1498


mikearnaldi:
Wont't work unfortunately due to https://vitejs.dev/guide/env-and-mode.html#production-replacement

PR #1579, opened 2023-10-30
add support for deno in test suite

tim-smart:
- [x] docgen fix for markdown-toc dependency
- [x] add `deno task test` to CI
- [ ] generate Deno import map

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,3 +67,16 @@ jobs:
       - name: Install bun
         uses: oven-sh/setup-bun@v1
       - run: bun vitest
+
+  deno:
+    name: Deno
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+
+    steps:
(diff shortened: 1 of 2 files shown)


pixeleet:
@tim-smart With latest deno 1.38 tests seems to work with the `unstable-byonm` flag enabled

image

pixeleet:
Alternatively just doing `deno --unstable-byonm task test` does the same trick

tim-smart:
> Alternatively just doing `deno --unstable-byonm task test` does the same trick

Well that certainly makes it easier :)

ryanleecode:
So for deno we can just use `platform-node` and it will work?

pixeleet:
that is what you're doing effectively when doing `import { ... } from
"npm:effect"`

Ryan Lee ***@***.***> ezt írta (időpont: 2024. márc. 23.,
Szo, 6:47):

PR #1582, opened 2023-10-30
simplify build

fubhy:
- [x] Reverted to pre-`preconstruct` build setup
- [x] Bumped to `NodeNext` and `ES2022`
- [x] Set `type: module` (to enforce `.js` extension for `NodeNext` module resolution)
- [x] Added `.js` extension to all relative imports
- [x] Use the actual repository structure, `./dist` directory and package.json in place (don't "pack" it)
- [x] Use `tsc` for building.
- [x] Uses `type: commonjs` for `./dist/cjs`. So far no issues detected with that
- [x] Standardize structure & naming / casing for internal module (by @IMax153)
- [x] Switch to explicit exports in package.json for all non-internal modules
- [x] Adjust build-utils to support the build output structure proposed in this PR (./cc @IMax153)

In follow-ups:
- [ ] Document our decision for dual package publishing
- [ ] Add consumer tests (webpack, rollup, parcel, swc, vite, tsc, etc.)

EDIT: As discussed with @tim-smart today, we are going to push this PR over the finish line instead of doing those extra steps in a follow-up.


tim-smart:
Look at using `allowImportingTsExtensions` in tsconfig.

fubhy:
This should be ready for review now (I'll replace `@fubhy/docgen` with `@effect/docgen` again once https://github.com/Effect-TS/docgen/pull/41 is merged).

@mikearnaldi you mentioned that you ran into issues with using ESM-first and dropping a `type: commonjs` package.json in the `cjs` path. I've tried various consumers including (I believe) all the popular bundlers and latest Node versions without running into any issues with this setup though.

IMax153 (approved):
This looks good from my perspective - kudos to @fubhy for all the hard work and testing that went into this 👍 

@mikearnaldi @tim-smart - you folks are up!

mikearnaldi:
Amazing work, thanks for bringing this over the finish line!

PR #1583, opened 2023-10-30
bump node & pnpm version

fubhy:
./cc @IMax153

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,7 +19,7 @@
       pkgs = nixpkgs.legacyPackages.${system};
       corepackEnable = pkgs.runCommand "corepack-enable" {} ''
         mkdir -p $out/bin
-        ${pkgs.nodejs-18_x}/bin/corepack enable --install-directory $out/bin
+        ${pkgs.nodejs}/bin/corepack enable --install-directory $out/bin
       '';
     in {
       formatter = pkgs.alejandra;
@@ -28,7 +28,7 @@
         default = pkgs.mkShell {
           buildInputs = with pkgs; [
             bun
-            nodejs-18_x
+            nodejs
             corepackEnable
           ];
         };
--- a/flake.nix
+++ b/flake.nix
@@ -19,7 +19,7 @@
       pkgs = nixpkgs.legacyPackages.${system};
       corepackEnable = pkgs.runCommand "corepack-enable" {} ''
         mkdir -p $out/bin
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -19,7 +19,7 @@
       pkgs = nixpkgs.legacyPackages.${system};
       corepackEnable = pkgs.runCommand "corepack-enable" {} ''
         mkdir -p $out/bin
-        ${pkgs.nodejs-18_x}/bin/corepack enable --install-directory $out/bin
+        ${pkgs.nodejs}/bin/corepack enable --install-directory $out/bin
```
IMax153, line 22: We should always pin to a specific version of NodeJS with Nixpkgs.

```suggestion
        ${pkgs.nodejs_20}/bin/corepack enable --install-directory $out/bin
```

flake.nix
```diff
@@ -28,7 +28,7 @@
         default = pkgs.mkShell {
           buildInputs = with pkgs; [
             bun
-            nodejs-18_x
+            nodejs
```
IMax153, line 31: Same comment as above.

IMax153 (changes requested):

flake.nix
fubhy: Hm. Did that, but can you clarify why that's necessary given that we have that locked in flake.lock? Protecting against accidentally upgrading past the major version?

IMax153: Precisely - because Nixpkgs `nodejs` package is not pinned to a specific version of NodeJS, if we update our flake inputs (which we really should be doing regularly and there's CI to help us with this) we'd risk accidentally upgrading to an unintended version of NodeJS.

Also I had no idea there was even a `nodejs` (sans version) package in Nixpkgs.

IMax153 (approved):

jessekelly881:
Any reason you guys don't use dependabot out of curiosity?

fubhy:
> Any reason you guys don't use dependabot out of curiosity?

That doesn't support Nix. Does it?

IMax153:
We used it a long time ago - we found it more annoying than anything else given that we usually know when we want to version bump, etc. And we're not managing a vast amount of deps to be honest.

jessekelly881:
Yeah, I have found it a bit annoying. Although you can apparently have dependabot create just one PR with all dep updates which would be less annoying. I don't think it supports nix though. At least I didn't see anything after a bit of research. Although, maybe support could be added.

PR #1594, opened 2023-11-03
fix regression in process.hrtime detection

tim-smart:
introduced by #1592

### DIFF
--- a/src/internal/clock.ts
+++ b/src/internal/clock.ts
@@ -4,7 +4,6 @@ import * as Duration from "../Duration"
 import type * as Effect from "../Effect"
 import * as Either from "../Either"
 import { constFalse } from "../Function"
-import { hasProperty, isFunction } from "../Predicate"
 import * as core from "./core"
 import * as timeout from "./timeout"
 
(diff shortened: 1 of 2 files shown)


src/internal/clock.ts
```diff
@@ -55,9 +54,10 @@ const performanceNowNanos = (function() {
   return () => origin + BigInt(Math.round(performance.now() * 1_000_000))
 })()
 const processOrPerformanceNow = (function() {
-  const processHrtime = hasProperty(process, "hrtime") && isFunction(process.hrtime.bigint) ?
-    process.hrtime :
-    undefined
+  const processHrtime =
```
gcanti, line 57: @tim-smart weird, they should be equivalent, what's the regression?

tim-smart:
```
> const predicate = (_) => true
undefined
> predicate(asdf)
Uncaught ReferenceError: asdf is not defined
> typeof asdf
'undefined'
```

You can use typeof on an undefined variable, but you can't use it for a function call.

PR #1603, opened 2023-11-05
Introduce Effect.andThen

mikearnaldi:


IMax153 (approved):
I think this is a good addition 👍

gcanti:
I like it. 

Do we also want to add `andThen` to all the other data types implementing `flatMap`?

mikearnaldi:
Seems like a good choice, I am thinking of we can do something to
consolidate tap, like a discard option to andThen

mikearnaldi:
Would it be nuts to add auto-flattening of Effect so that andThen "replaces" map too?

IMax153:
> Would it be nuts to add auto-flattening of Effect so that andThen "replaces" map too?

I kind of really like this idea honestly - then we only have to teach users one thing.

mikearnaldi:
Given we are committing crimes, I've also added auto-flattening to `tap` based on @schickling suggestion and given that some folks are expecting sync functions to be valid. Let me know your view, I am not 100% sure but think it's worth a try

tim-smart:
Auto-flattening Promise with Effect.promise? Too far?

mikearnaldi:
Maybe kind of unsure though it would either introduce an unknown error or
be a defect

tim-smart:
I think if we did promises we would have to introduce a built in error type to wrap rejections.

PR #1606, opened 2023-11-05
add Logger.mapInputOptions

tim-smart:


mikearnaldi:
use case for this?

tim-smart:
It came from https://discord.com/channels/795981131316985866/1170728631060922499/1170836884935684227

We could then use it to add APIs like Logger.withSpanAnnotations etc.

mikearnaldi:
LGTM

PR #1616, opened 2023-11-07
properly encapsulate `src` and `test` tsconfig

fubhy:


src/internal/effectable.ts
```diff
@@ -69,7 +69,19 @@ export const EffectPrototype: Effect.Effect<never, never, never> = {
...
+  [ChannelTypeId]: channelVariance,
+  [Equal.symbol](that: any) {
+    return this === that
+  },
+  [Hash.symbol]() {
+    return Hash.random(this)
+  },
+  pipe() {
+    return pipeArguments(this, arguments)
+  },
```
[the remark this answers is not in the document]
fubhy: @mikearnaldi Is this okay or do you have an idea how to fix this?

src/internal/effectable.ts
```diff
@@ -69,7 +69,19 @@ export const EffectPrototype: Effect.Effect<never, never, never> = {
 
 /** @internal */
 export const CommitPrototype = {
```
tim-smart, line 71:
```ts
/** @internal */
export const CommitPrototype: Effect.Effect<never, never, never> = {
  ...EffectPrototype,
  _op: OpCodes.OP_COMMIT
} as any
```

Maybe something like this.

fubhy: Yea... That works. No idea why I didn't think of that, lol. Thanks @tim-smart!

PR #1620, opened 2023-11-08
check size of tree shaked module imports

fubhy:
The current ESM build has a problem with tree-shaking due to subpath package.json (with `type: module`).

We have to add `sideEfects: false` there too: https://github.com/Effect-TS/build-utils/pull/31

We should likely add a script to `build-utils` to generate skeletons for these size limit checks (whilst allowing manual tweaking / prevenint overrides on re-runs).


IMax153 (approved):
@fubhy - perhaps we should also add some check into CI-CD if the package size grows by more than `X%` - it doesn't have to fail CI-CD, just a warning to log to folks.

In any case, that can always be a separate PR - this looks good to me for now 👍

fubhy:
> @fubhy - perhaps we should also add some check into CI-CD if the package size grows by more than `X%` - it doesn't have to fail CI-CD, just a warning to log to folks.
> 
> Maybe something that logs a comment to the PR with size changes would be even better.
> 
> In any case, that can always be a separate PR - this looks good to me for now 👍

You mean for the tree shaking size changes? Or for the overall size of the full package tarball? Because the latter can be achieved (pretty much for free) via this: https://github.com/preactjs/compressed-size-action

PR #1622, opened 2023-11-08
create snapshot releases on command

fubhy:
@mikearnaldi this could be a good alternative.

In order to create a snapshot release from a user's pull request (or the `changeset-release/main` PR), we can just write

```sh
/snapshot <user defined tag>
```

It'll then create releases versioned like so: 

```sh
effect@0.0.0-{{user defined tag}}-{{commit sha}}
```

We could also adds a prefixed tag (e.g. `snapshot-{{ user defined tag }}`) to the npm release if we wanted to ...

We might also want to add another workflow that *always* publishes the `main` branch to npm as a snapshot using a `main` tag. That way, every commit on the `main` branch would always have a snapshot tagged version available on npm.

NOTE: This is currently untested. Will verify that it works as expected in a different repo with a test package first.

EDIT: If we don't want to allow user-provided tag names we can also just use the github commit hash ...

```sh
effect@0.0.0-snapshot-{{commit sha}}
```

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -6,5 +6,9 @@
   "access": "restricted",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": [],
+  "snapshot": {
+    "useCalculatedVersion": false,
+    "prereleaseTemplate": "{tag}-{commit}"
+  }
 }
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -1,0 +1,122 @@
+name: Snapshot
+on:
+  issue_comment:
+    types:
+      - created
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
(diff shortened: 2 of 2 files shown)


fubhy:
If we actually want sth. like this, I think I'll actually just turn that into a proper github action (https://github.com/actions/typescript-action) instead of hacking together a bunch of workflow steps.

tim-smart:
I just remembered this: https://github.com/tim-smart/effect-ts-github-action

I went ahead and updated everything :)

mikearnaldi:
Github based release automation can be very useful, especially if we start
having separates branches that may end up being long running (after 2.0
breaking changes are packed for the next cycle, etc)

fubhy:
This should work now ... I pimped it a bit to also comment back on the issue with updates from the workflow.

It's a bummer that `changeset` doesn't have any JSON output option. Parsing all that from `stdout` is quite painful.

fubhy:
Maybe you want to test this in the `cli` repo, @IMax153?

PR #1629, opened 2023-11-09
DX: Merged module names with their types

patroza:
supersedes https://github.com/Effect-TS/effect/pull/1623

this version only touches the external modules but leaves internal and implementation remain using `import * as`.
will make a separate PR for consideration once this succeeds.

TODOS

- [ ] Docgen @fubhy ❤️


datner:
approved


also inb4 reverted within a week

fubhy:
I am going to wait until after Mike's given his seal of approval before going ahead with docgen :-)

.changeset/quiet-lies-share.md
```diff
@@ -0,0 +1,19 @@
+---
+"effect": major
```
gcanti, line 2: -> "patch"?

patroza: It’s certainly a major change, but I agree just a patch version 😂

mikearnaldi: that's a good point, if we are pre-release we should probably flag majors too

mikearnaldi:
How would a user go about extending a module such as Effect? for example it is fairly common to have in user-land preludes that re-export `export * from "effect/Effect"` together with a set of custom functions?

patroza:
> How would a user go about extending a module such as Effect? for example it is fairly common to have in user-land preludes that re-export `export * from "effect/Effect"` together with a set of custom functions?

They would end up with an Effect.Effect i suppose until pressure is exerted on typescript to finally address the merging of type and module symbol

mikearnaldi:
> > How would a user go about extending a module such as Effect? for example it is fairly common to have in user-land preludes that re-export `export * from "effect/Effect"` together with a set of custom functions?
> 
> They would end up with an Effect.Effect i suppose until pressure is exerted on typescript to finally address the merging of type and module symbol

As much as I'd like to remove `Effect.Effect` this isn't unfortunately acceptable, I guess only one road remains ahead, make a proposal and a POC impl for TS upstream

patroza:
> As much as I'd like to remove `Effect.Effect` this isn't unfortunately acceptable, I guess only one road remains ahead, make a proposal and a POC impl for TS upstream

fair enough

PR #1647, opened 2023-11-14
turn on exactOptionalPropertyTypes

gcanti:


gcanti:
@fubhy honestly I'm not sure it actually worth it, adding `undefined` everywhere seems a lot of work with little benefit

fubhy:
It adds a little bit of maintenance burden on our end but ensures that consumers that have `exactOptionalProperties` enabled will not have to constantly write

```ts
{
   ...(optionalProp === undefined ? undefined : { optionalProp })
}
```

... or similar.

PR #1668, opened 2023-11-19
Config: propagate the path in validation, closes #1667

### LINKED ISSUE #1667: From Discord: Trouble with Propagating Path in Config Validation
# Summary
The user is experiencing trouble with propagating the path in Config validation while defining a Config. They noticed that the Config.validate function erases the path, making it difficult to report the source of a validation error. They provided an example code snippet to illustrate the issue. The user is looking for a solution or workaround to handle this use case. It is suggested that this may be a bug.

# Discord thread

https://discord.com/channels/795981131316985866/1175599242463674440

gcanti:
@mikearnaldi It seems that `OpCodes.OP_MAP_OR_FAIL` is losing the path. I spent a couple of hours studying the code, but I'm uncertain if there's a better solution for this bug. Additionally, I've included some tests.

### DIFF
--- a/.changeset/metal-humans-agree.md
+++ b/.changeset/metal-humans-agree.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Config: propagate the path in validation, closes #1667
--- a/src/internal/configProvider.ts
+++ b/src/internal/configProvider.ts
@@ -218,6 +218,19 @@ const extend = <A, B>(
   return [leftExtension, rightExtension]
 }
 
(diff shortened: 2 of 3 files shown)


mikearnaldi:
cc @tim-smart @IMax153 @patroza you probably have more context on config then me

tim-smart:
LGTM, but I'm also not that familiar with Config.

IMax153 (approved):
@gcanti - this LGTM! 👍 Thank you very much - feel free to merge when you feel it's ready.

PR #1681, opened 2023-11-21
forbid excess properties when matching tags

matheuspuel:


tim-smart:
Thanks for the contribution :)

PR #1687, opened 2023-11-22
fix YieldableError.toString crashing on react-native

matheuspuel:
Apparently, when accessing `Error.stack` on react-native it makes a call to the `toString` method.
So using `stack` on `toString` causes an infinite loop and crashes the application.

Additionally, AFAIK the standard behavior of `Error.toString` is `${name}: ${message}` without stack.

### DIFF
--- a/.changeset/hot-crabs-boil.md
+++ b/.changeset/hot-crabs-boil.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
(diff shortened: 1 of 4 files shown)


tim-smart:
Thanks, I have just pushed a tweak to ensure a custom toString is reflected in node's inspect output. Will this still work for react native?

matheuspuel:
Yes, it's working

PR #1694, opened 2023-11-23
Add Either.filterOrLeft

extremegf:
Hello, 

Here's a proposal for `Either.filterOrLeft` based on `Effect.filterOrFail`. This is particularly useful in non-effectful pipelines, to validate the produced result. I considered names `filterOrFail`, `filterOrElse`, `validate`, `validateWith` - I'm open to suggestions.

 Also I don't know how to handle `@since `

### DIFF
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -11,6 +11,7 @@ import type { Inspectable } from "./Inspectable.js"
 import * as either from "./internal/either.js"
 import type { Option } from "./Option.js"
 import type { Pipeable } from "./Pipeable.js"
+import type { Predicate, Refinement } from "./Predicate.js"
 import { isFunction } from "./Predicate.js"
 import type * as Types from "./Types.js"
 import type * as Unify from "./Unify.js"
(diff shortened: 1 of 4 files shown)


src/Either.ts
```diff
@@ -375,6 +376,59 @@ export const match: {
...
+ * )
+ * assert.deepStrictEqual(
+ *   pipe(
+ *     E.right(0),
+ *     E.filterOrLeft(isPositive, n => `${n} is not positive`)
+ *   ),
+ *   E.left("0 is not positive")
+ * )
+ *
+ * @since TODO(horban): add since
```
tim-smart, line 404: You can use 2.0.0 here. This represents the major version of when the api was introduced.

tim-smart (changes requested):

gcanti:
If we are going to add this, I vote for `filterOrFail`

gcanti:
yeah, if we are going to add back some biased APIs I would slightly prefer consistency with Effect, though I guess filterOrLeft is also acceptable since there's already mapLeft

src/Either.ts
extremegf: Done.

tim-smart:
My vote is with Left, to keep consistency with the other APIs in the module.

tim-smart (approved):

extremegf:
Let me know if you have some small stuff to take care of.

PR #1703, opened 2023-11-24
improved Duration.toString

### LINKED ISSUE #1702: Better Duration.toString()

It would be nice if Duration.toString() rendered something more human friendly than a json object. It would improve how Schema errors are rendered for Duration and would probably be useful in other places as well. 

Something like this: https://www.npmjs.com/package/pretty-ms or really even something as simple as `${x}ms` or `${x}ns` would likely be better. 

_No response_

jessekelly881:
Closes #1702


src/Duration.ts
```diff
@@ -123,7 +123,7 @@ const DurationProto: Omit<Duration, "value"> = {
     return isDuration(that) && equals(this, that)
   },
   toString(this: Duration) {
-    return toString(this.toJSON())
```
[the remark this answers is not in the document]
gcanti: IMO showing a stringified JSON is more a default when we can't come up with sth better than a rule.
A recent example is BigDecimal (I guess this PR is along the lines of that data type), but I agree that `format` would be a better name than `toString` (also in the BigDecimal module)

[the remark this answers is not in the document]
tim-smart: Maybe the .toString should be `Duration(150ms)` etc.

src/Duration.ts
```diff
@@ -123,7 +123,7 @@ const DurationProto: Omit<Duration, "value"> = {
     return isDuration(that) && equals(this, that)
   },
   toString(this: Duration) {
-    return format(this.toJSON())
+    return format(this)
```
tim-smart, line 126: I thing only toString should add the `Duration()` notation.

src/Duration.ts
```diff
@@ -588,3 +588,55 @@ export const equals: {
...
+ *
+ * Duration.format(Duration.millis(1000)) // "Duration(1s)"
+ * Duration.format(Duration.millis(1001)) // "Duration(1s 1ms)"
+ */
+export const format = (self: DurationInput): string => {
+  const duration = decode(self)
+  const parts = []
+
+  if (duration.value._tag === "Infinity") {
+    return "Duration(Infinity)"
```
tim-smart, line 607: Remove `Duration()` here, as then it makes `format` more general purpose.

tim-smart (changes requested):

src/Duration.ts
```diff
@@ -588,3 +588,55 @@ export const equals: {
...
 } = dual(2, (self: DurationInput, that: DurationInput): boolean => Equivalence(decode(self), decode(that)))
+
+/**
+ * Converts a `Duration` to a human readable string.
+ * @since 2.0.0
+ *
+ * @example
+ * import * as Duration from "effect/Duration"
+ *
+ * Duration.format(Duration.millis(1000)) // "Duration(1s)"
```
tim-smart, line 599: Just need to update the example :)

tim-smart (changes requested):

src/Duration.ts
jessekelly881: oops

tim-smart (approved):

PR #1704, opened 2023-11-25
add `.getLefts` and `.getRights` and rename `.compact` to `.getSomes`

fubhy:
./cc @tobiaslins

`.compact` is not the worst name but also not the best.

### DIFF
--- a/src/ReadonlyArray.ts
+++ b/src/ReadonlyArray.ts
@@ -1649,13 +1649,13 @@ export const partitionMap: {
  * @category filtering
  * @since 2.0.0
  */
-export const filterSome: <A>(self: Iterable<Option<A>>) => Array<A> = filterMap(identity)
+export const getSome: <A>(self: Iterable<Option<A>>) => Array<A> = filterMap(identity)
 
 /**
  * @category filtering
  * @since 2.0.0
  */
-export const filterLeft = <E>(self: Iterable<Either<E, any>>): Array<E> => {
+export const getLeft = <E, A>(self: Iterable<Either<E, A>>): Array<E> => {
   const out: Array<E> = []
   for (const a of self) {
     if (E.isLeft(a)) {
@@ -1670,7 +1670,7 @@ export const filterLeft = <E>(self: Iterable<Either<E, any>>): Array<E> => {
  * @category filtering
  * @since 2.0.0
  */
-export const filterRight = <A>(self: Iterable<Either<any, A>>): Array<A> => {
+export const getRight = <E, A>(self: Iterable<Either<E, A>>): Array<A> => {
   const out: Array<A> = []
   for (const a of self) {
     if (E.isRight(a)) {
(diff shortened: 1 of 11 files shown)


src/ReadonlyArray.ts
```diff
@@ -1649,7 +1649,37 @@ export const partitionMap: {
...
  * @since 2.0.0
  */
-export const compact: <A>(self: Iterable<Option<A>>) => Array<A> = filterMap(identity)
+export const filterSome: <A>(self: Iterable<Option<A>>) => Array<A> = filterMap(identity)
+
+/**
+ * @category filtering
+ * @since 2.0.0
+ */
+export const filterLeft = <E>(self: Iterable<Either<E, any>>): Array<E> => {
```
tim-smart, line 1658:
```suggestion
export const filterLeft = <E, A>(self: Iterable<Either<E, A>>): Array<E> => {
```

src/ReadonlyArray.ts
```diff
@@ -1649,7 +1649,37 @@ export const partitionMap: {
...
+  }
+
+  return out
+}
+
+/**
+ * @category filtering
+ * @since 2.0.0
+ */
+export const filterRight = <A>(self: Iterable<Either<any, A>>): Array<A> => {
```
tim-smart, line 1673: Same as above

tim-smart (changes requested):

src/ReadonlyArray.ts
fubhy: @tim-smart just for my understanding: whats the benefit of capturing `A` if we are eliminating it?

gcanti:
I'm not sure I like using "filter" with this king of APIs, they look more getters than filters: getRights, getLefts, getSomes.

src/ReadonlyArray.ts
tim-smart: We have run into subtle issues with inference and assignability before.

So where possible we just leave unused generics unused. They might be called `X` in some places to indicate that they are not used.

fubhy: Ok. I'd be curious to know more about that. Happy to make the adjustment now but I'm always curious to understand these edge cases with typescript ;-)

fubhy:
> I'm not sure I like using "filter" with this king of APIs, they look more getters than filters: getRights, getLefts, getSomes.

I'd be happy to call it something else. But after the third time of Tobias asking me what the name of the function was that filters and maps out `Option.Some` values it clicked for me that we need a better name here.

src/ReadonlyArray.ts
mikearnaldi: It's due to the TS inference algorithm that prefers weaker types as a shortcut, basically having `any` here will make TS back-infer to `any` the value of `A` in for example `pipe(a, filterLeft(f))`, we had hard to spot issues because of this kind of stuff so now we always put a generic even if ignored

fubhy: Ah ok. That actually makes sense. It's a bit quirky but kinda makes sense ...

tim-smart (approved):

PR #1705, opened 2023-11-25
chore: make `Effect.request` dual

FedericoBiccheddu:
Context: https://discord.com/channels/795981131316985866/1157603010038468679


tim-smart:
As discussed on Discord, the second argument accepts an Effect so we don't need this :)

PR #1714, opened 2023-11-29
Refactor try, promise, async

ricardo-valero:
I feel like these changes make the affected functions easier to read, maintain, and understand. I've added some utility functions that I think are helpful. I believe these should be separated because they aren't strictly tied to the original function, making the code more straightforward.


src/internal/core-effect.ts
```diff
@@ -1620,59 +1636,24 @@ export const tryPromise: {
...
-          .then((a) => resolve(core.exitSucceed(a)))
-          .catch((e) =>
-            resolve(core.fail(
-              catcher ? catcher(e) : new core.UnknownException(e)
-            ))
-          )
-      })
-  )
-}
+  arg:
```
tim-smart, line 1641: If you rebase with the latest main, we can take a look at simplifying that.

src/internal/core.ts
```diff
@@ -440,19 +447,10 @@ export const async = <R, E, A>(
 ): Effect.Effect<R, E, A> =>
   suspend(() => {
     let cancelerRef: Effect.Effect<R, never, void> | void = undefined
-    let controllerRef: AbortController | void = undefined
+    let controllerRef: AbortController | undefined = undefined
     const effect = new EffectPrimitive(OpCodes.OP_ASYNC) as any
-    if (register.length !== 1) {
```
tim-smart, line 445: I prefer the previous version, as it it was self-contained.

If `withSignalAndCallback` was used in 3 or more places, then I think the change could be worth it.

tim-smart (commented):
I'm not convinced about the `async` and `promise` changes, but I think we could take another look at `tryPromise` once you have rebased :)

PR #1715, opened 2023-11-29
simplify async constructors

tim-smart:


src/internal/core-effect.ts
```diff
@@ -1638,40 +1637,38 @@ export const tryPromise: {
...
           .then((a) => resolve(core.exitSucceed(a)))
           .catch((e) =>
             resolve(core.fail(
               catcher ? catcher(e) : new core.UnknownException(e)
             ))
           )
-      })
-  )
+      } catch (error) {
+        throw core.makeEffectError(internalCause.fail(
```
tim-smart, line 1650: Ah, should just use resolve(fail(...)) here.

PR #1726, opened 2023-12-01
fix the tacit use of some NonEmpty unfied APIs

gcanti:


steida:
Yeah, that's why flatten stopped working...

PR #1729, opened 2023-12-04
updated BigDecimal.toString

jessekelly881:
Inline with Duration


IMax153:
cc @fubhy since you did quite a bit of work on this IIRC.

tim-smart:
This was from the Duration PR, to make toString consistent. LGTM :)

PR #1730, opened 2023-12-04
replace throw with symbol in run loop

tim-smart:


package.json
```diff
@@ -2,6 +2,10 @@
   "name": "effect",
   "version": "2.0.0-next.58",
   "type": "module",
+  "exports": {
```
IMax153, line 5: Was leaving this in intentional? I think @fubhy had removed it purposefully.

tim-smart: No it wasn't :)

IMax153 (approved):

tim-smart:
superseded by #1734

PR #1731, opened 2023-12-04
remove value accumulation in histogram implementation

fubhy:
The current implementation appears to be wrong. AFAIK, the accumulation shouldn't happen in the producing app: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram

### DIFF
--- a/.changeset/tender-yaks-leave.md
+++ b/.changeset/tender-yaks-leave.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Removed accumulation of values in histogram implementation
--- a/src/internal/metric/hook.ts
+++ b/src/internal/metric/hook.ts
@@ -165,12 +165,8 @@ export const histogram = (key: MetricKey.MetricKey.Histogram): MetricHook.Metric
 
   const getBuckets = (): Chunk.Chunk<readonly [number, number]> => {
     const builder: Array<readonly [number, number]> = Array(size)
-    let cumulated = 0
     for (let i = 0; i < size; i++) {
-      const boundary = boundaries[i]
-      const value = values[i]
-      cumulated = cumulated + value
-      builder[i] = [boundary, cumulated]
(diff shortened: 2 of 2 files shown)


tim-smart:
Using the accumulated values seem correct to me (It also matches the zio impl). prometheus expects accumulated values too.

fubhy:
So as I learned the @effect/opentelemetry package is supposed to unwind the accumulated values again.

The reason is that prometheus expects accumulated values whilst otel wants them un-accumulated.

So this actually makes sense. Still, it's broken for me in grafana atm (via otel exporter through otel collector). I'll debug a bit and report back.

mikearnaldi:
> So as I learned the @effect/opentelemetry package is supposed to unwind the accumulated values again.
> 
> The reason is that prometheus expects accumulated values whilst otel wants them un-accumulated.
> 
> So this (accumulating here) actually makes sense somewhat. I'll say that I still find it a bit odd to do this back & fort instead of adding the accumulation step in a hypothetical future prometheus exporter... But that's a debate for another day.
> 
> Still, it's broken for me in grafana atm (via otel exporter through otel collector). I'll debug a bit and report back.

Metrics were designed in ZIO with prometheus in mind, in fact the available metric types exactly match prometheus. So that's probably the illogical logic behind this. There is no reason why we should exactly follow zio though, if we believe something to be better we should diverge

mikearnaldi:
though we should think about reading current value of metrics, if we expect accumulated values in reading storing un-accumulated will hurt read perf

fubhy:
> though we should think about reading current value of metrics, if we expect accumulated values in reading storing un-accumulated will hurt read perf

Sure. But the opposite is true as well. Currently, the otel integration unwinds the accumulated values at read time. That also comes at a (albeit lesser) cost.

fubhy:
I'm going to close this for now.

The original issue I ran into was caused by a misconfiguration in Grafana & OTEL Collector

PR #1732, opened 2023-12-04
use native js data types in metrics

tim-smart:


src/internal/metric/hook.ts
```diff
@@ -163,7 +160,7 @@ export const histogram = (key: MetricKey.MetricKey.Histogram): MetricHook.Metric
     }
   }
 
-  const getBuckets = (): Chunk.Chunk<readonly [number, number]> => {
+  const getBuckets = (): ReadonlyArray<readonly [number, number]> => {
```
IMax153, line 163: I believe @fubhy had implemented a fix here https://github.com/Effect-TS/effect/pull/1731. We could potentially incorporate into this PR since the return type is changing?

IMax153:
@tim-smart - looks like the build step is failing due to docs?

IMax153:
Sorry for the accidental closure - clicked the incorrect button 😅

mikearnaldi:
```ts
(self, labels) => core.fiberRefLocallyWith(self, core.currentMetricLabels, (old) => ReadonlyArray.union(old, labels))
```

Aren't we compromising uniqueness of labels here?

tim-smart:
> ```ts
> (self, labels) => core.fiberRefLocallyWith(self, core.currentMetricLabels, (old) => ReadonlyArray.union(old, labels))
> ```
> 
> Aren't we compromising uniqueness of labels here?

No, ReadonlyArray.union does a dedupe with Equal.equals.

But maybe we switch back to hashset, as it makes the code less maintainable.

mikearnaldi:
HashSet is probably slower as it is tree based, if labels don't need to be
checked for presence maybe its faster to use RA

tim-smart:
There is no presence checking, just insertion with uniqueness.

IMax153 (changes requested):

IMax153:
I personally don't mind the use of `ReadonlyArray` for this purpose. I also don't feel as though the code is less maintainable when using the data type. I think the perf gains from using `ReadonlyArray` are worth any tiny extra bit of overhead we have to deal with.

src/internal/fiberRuntime.ts
```diff
@@ -744,6 +744,9 @@ export class FiberRuntime<in out E, in out A> implements Fiber.RuntimeFiber<E, A
   reportExitValue(exit: Exit.Exit<E, A>) {
     if (_runtimeFlags.runtimeMetrics(this._runtimeFlags)) {
       const tags = this.getFiberRef(core.currentMetricLabels)
+      const startTimeMillis = this.id().startTimeMillis
+      const endTimeMillis = new Date().getTime()
```
IMax153, line 748: **Question** 
In a previous conversation somewhere, I think you had suggested using `Date.now`?

tim-smart: That is in the other PR :)

PR #1735, opened 2023-12-05
expose Layer MemoMap apis

tim-smart:


tim-smart:
Will need to add ref counting to prevent cases where Layers could be released too early.

mikearnaldi:
Was about to say, that part is already needed

mikearnaldi:
Something is off here. The following:

```ts
const fromLayer = <RIn, E, ROut>(
  evaluate: Layer.Layer<RIn, E, ROut>
): Layer.Layer<RIn, E, ROut> => {
  const suspend = Object.create(proto)
  suspend._tag = OpCodes.OP_SUSPEND
  suspend.evaluate = () => evaluate
  return suspend
}
```

doesn't quite make sense, evaluate is not lazy so this should be a no-op. I see though that removing it makes the newly implemented tests fail which is strange

mikearnaldi:
Even replacing `fromLayer` with suspend makes the test fail. 100% something is broken

tim-smart:
We can remove it and use suspend instead.

We can't use fromEffectContect directly in fromEffect, as that doesn't hit the memomap.

mikearnaldi:
Using suspend doesn't work

mikearnaldi:
also fromEffectContext should hit the memo map, every layer should

tim-smart:
Ah no we can't, as then we lose referential equality...

tim-smart:
> also fromEffectContext should hit the memo map, every layer should

https://github.com/Effect-TS/effect/blob/af9fd6d79a234760e7a5d84778142cf216a8b5b4/src/internal/layer.ts#L380

AFAIK, that is the final point of construction, which can't use the memo map otherwise it would create a busy loop.

mikearnaldi:
```ts
/**
 * Constructs a layer from the specified effect, which must return one or more
 * services.
 *
 * @since 2.0.0
 * @category constructors
 */
export const effectContext: <R, E, A>(effect: Effect.Effect<R, E, Context.Context<A>>) => Layer<R, E, A> =
  internal.fromEffectContext
```

mikearnaldi:
would be "ok" if internal but this just allows users to create layers that are never memoized

tim-smart:
I don't think effectContext needs it, because there is no Scope and Context is immutable?

tim-smart:
Ah yeah, there could be side effects though.

mikearnaldi:
Precisely, well we could wrap the public facing version with fromLayer

tim-smart:
Fixed :)

PR #1768, opened 2023-12-12
Effect.mergeAll should work when Z is an iterable, closes #1765

### LINKED ISSUE #1765: From Discord: `Effect.mergeAll` failing in case of `Z` being an array

_No response_

The user described scenarios with `Effect.mergeAll` that yielded failures during runtime:

```ts
import { Effect, pipe,ReadonlyArray } from "effect";

const program = Effect.gen(function*($) {
    const result = yield* $(
        pipe([["a"],["aa"],["aaa"],["aaaa"]].map((a) => Effect.succeed(a)), Effect.mergeAll("", (b, a) => [...b,...a])))
      })

    Effect.runFork(program);
```

```
import { Effect, pipe,ReadonlyArray } from "effect";

const program = Effect.gen(function*($) {
    const result = yield* $(
        pipe([["a"],["aa"],["aaa"],["aaaa"]].map((a) => Effect.succeed(a)), Effect.mergeAll("", (b, a) => ReadonlyArray.appendAll(b,a))))
      })

    Effect.runFork(program);
```

Returns: [ 'a', 'aa', 'aaa', 'aaaa' ]
Yet Typewise, b should be a string.

```ts
import { Effect, pipe,ReadonlyArray } from "effect";

const program = Effect.gen(function*($) {
    const result = yield* $(
        pipe([["a"],["aa"],["aaa"],["aaaa"]].map((a) => Effect.succeed(a)), Effect.mergeAll([], (b, a) => [...b,...a])))
      })

    Effect.runFork(program);
```

```ts
import { Effect, pipe,ReadonlyArray } from "effect";

const program = Effect.gen(function*($) {
    const result = yield* $(
        pipe([["a"],["aa"],["aaa"],["aaaa"]].map((a) => Effect.succeed(a)), Effect.mergeAll(ReadonlyArray.empty<string>(), (b, a) => ReadonlyArray.appendAll(b,a))))
      })

    Effect.runFork(program);
```

Was described as being fine typewise, yet failing with `TypeError: ab is not a function` on runtime.

The user would expect the possibility of the return type being an array.

### What do you see instead?

_No response_

### Additional information

_No response_

gcanti:


src/internal/fiberRuntime.ts
```diff
@@ -2225,7 +2225,7 @@ export const mergeAll = dual<
     readonly batching?: boolean | "inherit" | undefined
   }) => Effect.Effect<R, E, Z>
 >(
-  (args) => Predicate.isIterable(args[0]),
+  (args) => Predicate.isFunction(args[2]),
```
tim-smart, line 2228: This will fail if Z is a function :) (even though that is a rare case)

Predicate.isIterable(args[0]) && Predicate.isFunction(args[2]) should cover all the cases.

gcanti: I think that `args[2]` can only be:

- a function `f: (z: Z, a: A, i: number) => Z` (data-first) 
- an object or `undefined` (`options?: { ... }`) (data-last)

tim-smart: Ah yes, nice 👍🏼

tim-smart (approved):

PR #1775, opened 2023-12-16
Secret/map

### LINKED ISSUE #1774: Add Secret.map

There is no Secret.map fn

It would be useful to add a Secret.map fn. For example, 

```ts
Secret.map(token, t => `Bearer ${t}`)
```

_No response_

jessekelly881:
Closes #1774


mikearnaldi:
Please re-open this against the new repo structure, apologies we didn't realise changing head would close the PR.

PR #1784, opened 2023-12-18
Ensure that both top-level fields and nested fields can be read with `ConfigProvider.fromJson`

IMax153:
This PR makes sure we add the top-level field (with nested fields `JSON.stringify`'ed) to the underlying `ConfigProvider` map. This allows external consumers to create more elaborate configs with `Config.primitive`, for example using `@effect/schema` and `Schema.parseEither`.
```ts
  it.effect("fromJson - should load configs from nested JSON", () =>
    Effect.gen(function*($) {
      const config = Config.primitive<Record<string, unknown>>(
        "a JSON record",
        (json) => Either.right(JSON.parse(json))
      ).pipe(Config.nested("hostPorts"))
      const result = yield* $(
        ConfigProvider.fromJson({
          hostPorts: {
            host: "localhost",
            port: 8080
          }
        }).load(config)
      )
      assert.deepStrictEqual(result, {
        host: "localhost",
        port: 8080
      })
    }))
```

### DIFF
--- a/.changeset/plenty-buttons-thank.md
+++ b/.changeset/plenty-buttons-thank.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+ensure that both top-level fields and nested fields can be read with ConfigProvider.fromJson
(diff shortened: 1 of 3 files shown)


IMax153:
Closing this as it doesn't align with other `ConfigProvider` implementations.

PR #1793, opened 2023-12-28
Improve Effect.retry/repeat options

mikearnaldi:
Missing:
- [x] tests
- [x] type tests
- [x] removal of variants


packages/effect/src/internal/channel.ts
```diff
@@ -1217,7 +1217,7 @@ export const mergeAllWith = (
                 )
             })
           ),
-          Effect.repeatUntil(Option.isSome),
+          Effect.repeat({ until: (_): _ is Option.Some<OutDone> => Option.isSome(_) }),
```
tim-smart, line 1220: This is a case where tacit usage of refinements can no longer be applied, but maybe this ok.

tim-smart (commented):
@mikearnaldi ready for a review

PR #1794, opened 2023-12-29
Schema: add eitherFromUnion

matheuspuel:
Helps decoding expected errors.
Example use case:
```typescript
import * as PR from "@effect/schema/ParseResult"
import { Effect, pipe } from "effect"

const fetchData: Effect.Effect<never, "FetchError", string> = Effect.succeed(
  JSON.stringify({ error: "Custom error occurred!!!" })
)
const ResponseSchema = S.eitherFromUnion(
  S.transformOrFail(
    S.struct({ error: S.string.pipe(S.startsWith("Custom error occurred")) }),
    S.literal("CustomError"),
    () => PR.succeed("CustomError" as const),
    (_) => PR.fail(PR.forbidden(_))
  ),
  S.struct({ data: S.number })
)
// Effect<never, "FetchError" | "CustomError" | ParseError, { readonly data: number }>
const getData = pipe(
  fetchData,
  Effect.flatMap(S.decode(S.parseJson(ResponseSchema))),
  Effect.flatten
)
```

Note that it allows creating a schema that doesn't follow the `a = decode(encode(a))` rule, e.g.:
```typescript
const schema = S.eitherFromUnion(S.string, S.string )
expect(S.decodeSync(schema)(S.encodeSync(schema)(E.left("a")))).toEqual(E.right("a"))
```


packages/schema/src/Schema.ts
```diff
@@ -3641,6 +3641,39 @@ export const either = <IE, E, IA, A>(
...
+ * // Schema<"A" | "E", Either<"E", "A">>
+ * Schema.eitherFromUnion({
+ *   right: Schema.literal("A"),
+ *   left: Schema.literal("E")
+ * })
+ *
+ * @category Either transformations
+ * @since 1.0.0
+ */
+export const eitherFromUnion = <EI, EA, AI, AA>(schemas: {
```
matheuspuel, line 3657: I chose to use a struct parameter  `eitherFromUnion({ left, right })` because in my use cases the schemas tend to be long and it's easier to read this way.
But it doesn't conform to the existing functions, that take two parameters `either(left, right)`.
So you may choose what you prefer and I can adjust.

gcanti: @matheuspuel let's use the standard `eitherFromUnion(left, right)` for now

gcanti (changes requested):
I think the error messages in the tests need to be fixed, there have been some changes in `main` in the meantime

gcanti (approved):

gcanti:
Thank you @matheuspuel

PR #1847, opened 2024-01-01
Micro optimization to avoid inline creation & spreading of objects / arrays

fubhy:


tim-smart:
Just need a changeset :)

PR #1850, opened 2024-01-02
add index argument to many functions in ReadonlyArray

matheuspuel:


gcanti (changes requested):
@matheuspuel I would add some dtslint tests, based on my experiments these tests fail:

```ts
// takeWhile

ReadonlyArray.takeWhile(numbersOrStrings, (
  _item // $ExpectType string | number
) => true)

// fails: Parameter '_item' implicitly has an 'any' type.ts(7006)
ReadonlyArray.takeWhile(numbersOrStrings, (
  _item, // $ExpectType string | number
  _i // $ExpectType number
) => true)

pipe(
  numbersOrStrings,
  ReadonlyArray.takeWhile((
    _item // $ExpectType string | number
  ) => true)
)

// fails: Parameter '_item' implicitly has an 'any' type.ts(7006)
pipe(
  numbersOrStrings,
  ReadonlyArray.takeWhile((
    _item, // $ExpectType string | number
    _i // $ExpectType number
  ) => true)
)

```

unless you add a `i: number` parameter to refinements too:

```ts
export const takeWhile: {
  <A, B extends A>(refinement: (a: A, i: number) => a is B): (self: Iterable<A>) => Array<B>
  <B extends A, A = B>(predicate: (a: A, i: number) => boolean): (self: Iterable<B>) => Array<B>
  <A, B extends A>(self: Iterable<A>, refinement: (a: A, i: number) => a is B): Array<B>
  <A>(self: Iterable<A>, predicate: (a: A, i: number) => boolean): Array<A>
} = dual(2, <A>(self: Iterable<A>, predicate: (a: A, i: number) => boolean): Array<A> => {
  let i = 0
  const out: Array<A> = []
  for (const a of self) {
    if (!predicate(a, i)) {
      break
    }
    out.push(a)
    i++
  }
  return out
})
```

gcanti (approved):

PR #1855, opened 2024-01-03
Add `Cron` module and `Schedule.cron` constructor

### LINKED ISSUE #1524: Consider adding support for cron expressions for Schedule

There is currently no good way to translate cron expressions into an instance of Schedule. Although a "cron-like" api exists (i.e. `dayOfWeek`, `dayOfMonth`, etc.) it isn't very convenient. 

A cron string would be more convenient and familiar vs writing using the current set of `dayOfMonth`, `dayOfWeek`, etc. api. 
```ts
Schedule.cron("28 2 7 * *")
```

_No response_

fubhy:
- [x] Add cron expression parser
- [x] Add cron `match` calculator
- [x] Add cron `next` calculator
- [x] Add `Schedule.cron` constructor
- [x] Write tests for `Schedule.cron`

Fixes #1524

### DIFF
--- a/packages/effect/src/internal/schedule.ts
+++ b/packages/effect/src/internal/schedule.ts
@@ -430,7 +430,7 @@ export const cron = (expression: string): Schedule.Schedule<never, unknown, numb
         const [next, start, end] = previous
         const interval = Interval.make(start, end)
         return core.succeed([
-          [false, next],
+          [false, [next, start, end]],
           next,
           ScheduleDecision.continueWith(interval)
         ])
--- a/packages/effect/src/internal/schedule.ts
+++ b/packages/effect/src/internal/schedule.ts
@@ -427,12 +427,10 @@ export const cron = (expression: string): Schedule.Schedule<never, unknown, numb
     [true, [Number.MIN_SAFE_INTEGER, 0, 0]],
     (now, _, [initial, previous]) => {
       if (now < previous[0]) {
-        const [next, start, end] = previous
-        const interval = Interval.make(start, end)
         return core.succeed([
-          [false, [next, start, end]],
-          next,
-          ScheduleDecision.continueWith(interval)
+          [false, previous],
+          previous[0],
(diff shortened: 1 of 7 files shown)


packages/effect/src/internal/schedule.ts
```diff
@@ -419,6 +420,40 @@ export const mapInputEffect = dual<
...
+        next = result.getTime()
+      }
+
+      const start = beginningOfMinute(next)
+      const end = endOfMinute(next)
+      const interval = Interval.make(start, end)
+      return core.succeed([[false, next], new Date(next), ScheduleDecision.continueWith(interval)])
+    }
+  )
+}
```
fubhy, line 455: @mikearnaldi Can you review this bit specifically? I've no idea what I'm doing ;-)

mikearnaldi:
Is there any usage expectation of Cron apart from Schedule? if not it could be an internal module

fubhy:
> Is there any usage expectation of Cron apart from Schedule? if not it could be an internal module

Would that be your preference? I don't have a strong preference either way. I can /see/ this being useful outside of `Schedule` in user land but I don't have an immediate use-case myself so would be happy either way.

mikearnaldi:
> > Is there any usage expectation of Cron apart from Schedule? if not it could be an internal module
> 
> Would that be your preference? I don't have a strong preference either way. I can /see/ this being useful outside of `Schedule` in user land but I don't have an immediate use-case myself so would be happy either way.
> 
> There are several cron npm packages with a few million weekly downloads:
> 
> * https://www.npmjs.com/package/cron
> * https://www.npmjs.com/package/croner
> * https://www.npmjs.com/package/node-cron
> 
> Hard to say for what these are mostly used.
> 
> I don't see much harm in maintaining this as a public module other than the maintenance burden. Your call ;-)

My question was more out of curiosity, I agree regarding the maintenance burden, if this is stable it can be publicly exposed, only downside is if this is not stable and leads to breaking changes. I don't have a strong opinion, we can start having it public for now and reserve the right to decide otherwise if we feel like.

PR #1856, opened 2024-01-03
docs: add Url example to schema readme

nikgraf:
Needed an Url combinator to validate that a string is a valid URL and thought this might be useful to others as well. Tested it here in isolation to make sure the code works as expected: https://stackblitz.com/edit/effect-ts-examples-q6pwrk?file=package.json,src%2Findex.ts

Note: there could be another example to actually parse it to a URL instance. Happy to add this as well if there is interest.

### DIFF
--- a/packages/schema/README.md
+++ b/packages/schema/README.md
@@ -2787,33 +2787,61 @@ const Email = S.pattern(/^(?!\.)(?!.*\.\.)([A-Z0-9_+-\.]*)[A-Z0-9_+-]@([A-Z0-9][
 
 ## Url
 
-Multiple environments like the Browser or Node provide a built-in `URL` class that can be used to validate URLs. Here we demonstrate how to leverage it to validate if a string is a valid URL while the returned value is still a string.
+Multiple environments like the Browser or Node provide a built-in `URL` class that can be used to validate URLs. Here we demonstrate how to leverage it to validate if a string is a valid URL.
 
 ```ts
 import * as S from "@effect/schema/Schema";
 import * as ParseResult from "@effect/schema/ParseResult";
 
-const Url: S.Schema<string, string> = S.transformOrFail(
+const UrlString: S.Schema<string, string> = S.string.pipe(
+  S.filter((value) => {
+    try {
+      new URL(value);
+      return true;
+    } catch (_) {
+      return false;
+    }
+  }),
+);
+
+const parser = S.parseSync(UrlString);
+const validUrlString = parser("https://www.effect.website");
+```
+
+In case you prefer to normalize URLs you can combine `transformOrFail` with `URL`:
+
+```ts
+const NormalizedUrlString: S.Schema<string, string> = S.string.pipe(
+  S.filter((value) => {
+    try {
+      return new URL(value).toString() === value;
+    } catch (_) {
+      return false;
+    }
+  }),
+);
+
+const NormalizeUrlString: S.Schema<string, string> = S.transformOrFail(
   S.string,
-  S.string,
+  NormalizedUrlString,
   (value, _, ast) =>
     ParseResult.try({
       try: () => new URL(value).toString(),
-      catch: (err) => {
-        return ParseResult.parseError([
+      catch: (err) =>
+        ParseResult.parseError([
           ParseResult.type(
             ast,
             value,
             err instanceof Error ? err.message : undefined,
           ),
-        ]);
-      },
+        ]),
     }),
-  (value) => ParseResult.succeed(value),
+  ParseResult.succeed,
 );
 
-const parser = S.parseSync(Url);
-const validUrlString = parser("https://www.effect.website/");
+const parser = S.parseSync(NormalizeUrlString);
+const validUrlString = parser("https://www.effect.website");
+// validUrlString will be "https://www.effect.website/"
 ```
 
 # Technical overview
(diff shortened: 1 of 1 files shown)


fubhy (approved):

packages/schema/README.md
```diff
@@ -2785,6 +2785,37 @@ import * as S from "@effect/schema/Schema";
 const Email = S.pattern(/^(?!\.)(?!.*\.\.)([A-Z0-9_+-\.]*)[A-Z0-9_+-]@([A-Z0-9][A-Z0-9\-]*\.)+[A-Z]{2,}$/i);
 ```
 
+## Url
+
+Multiple environments like the Browser or Node provide a built-in `URL` class that can be used to validate URLs. Here we demonstrate how to leverage it to validate if a string is a valid URL while the returned value is still a string.
```
gcanti, line 2790: this is an actual transformation rather than a validation, for example the string `"https://www.effect.website"` is transformed into `"https://www.effect.website/"` by `Schema.parse(Url)`

packages/schema/README.md
```diff
@@ -2785,6 +2785,37 @@ import * as S from "@effect/schema/Schema";
...
+
+```ts
+import * as S from "@effect/schema/Schema";
+import * as ParseResult from "@effect/schema/ParseResult";
+
+const Url: S.Schema<string, string> = S.transformOrFail(
+  S.string,
+  S.string,
```
gcanti, line 2798: this should be a schema representing a valid URL, `S.string` represents *any* string

gcanti (changes requested):

packages/schema/README.md
nikgraf: good point, I will return the original

packages/schema/README.md
nikgraf: This was intentional. To clarify my intend: In the past 1,5 years working with Zod for example I never needed a URL instance. I only used Zod to validate it's a valid URL to then store the data as a string in SQL db as a string.

I see several options:
1. keep the string
2. keep the string and add a second example returning a URL instance
3. only provide the URL example
4. something else you haven in mind?

Let me know what you think makes the most sense

packages/schema/README.md
gcanti: but then you are defining a filter rather than a transformation so you can use the `filter` API

packages/schema/README.md
gcanti: Sorry, I wasn't clear. Above, I meant to say a schema representing a valid URL **string** (not in the sense of using the `URL` constructor), which is nothing more than what you are about to define using `filter`

gcanti: To recap: the example could include only the definition of the filter (`UrlString` below), but if you want to include a transformation, it must use the filter as the target schema, something like:

```ts
import * as ParseResult from "@effect/schema/ParseResult"

// a string filter, should be enough
const UrlString: S.Schema<string, string> = S.string.pipe(
  S.filter((value) => {
    try {
      new URL(value)
      return true
    } catch (_) {
      return false
    }
  })
)

// a string transformation
const NormalizeUrl: S.Schema<string, string> = S.transformOrFail(
  S.string,
  UrlString,
  (value, _, ast) =>
    ParseResult.try({
      try: () => new URL(value).toString(),
      catch: (err) =>
        ParseResult.parseError([ParseResult.type(ast, value, err instanceof Error ? err.message : undefined)])
    }),
  ParseResult.succeed
)
```

fubhy: @gcanti I suggested the `transformOrFail` approach to also normalize the URLs (e.g. trailing forward slash). I tend to prefer normalization in schemas to reduce possible permutations of the same value e.g. for caching purposes (where applicable). 

I can see however, how `filter` might be a simpler showcase here ...

gcanti: mmh I think I got the filter wrong, I guess should be

```ts
import * as ParseResult from "@effect/schema/ParseResult"

const NormalizedUrlString: S.Schema<string, string> = S.string.pipe(
  S.filter((value) => {
    try {
      return new URL(value).toString() === value
    } catch (_) {
      return false
    }
  })
)

const NormalizeUrlString: S.Schema<string, string> = S.transformOrFail(
  S.string,
  NormalizedUrlString,
  (value, _, ast) =>
    ParseResult.try({
      try: () => new URL(value).toString(),
      catch: (err) =>
        ParseResult.parseError([ParseResult.type(ast, value, err instanceof Error ? err.message : undefined)])
    }),
  ParseResult.succeed
)
```

nikgraf: Thanks a lot! Helps to clarify a lot how to use effect/schema

I pushed a new version which has two examples:
1. one only doing simple validation
2. your normalization example using `transformOrFail`

I tested both on https://stackblitz.com/edit/effect-ts-examples-q6pwrk?file=package.json,src%2Findex.ts

packages/schema/README.md
```diff
@@ -2785,6 +2785,65 @@ import * as S from "@effect/schema/Schema";
...
+            err instanceof Error ? err.message : undefined,
+          ),
+        ]),
+    }),
+  ParseResult.succeed,
+);
+
+const parser = S.parseSync(NormalizeUrlString);
+const validUrlString = parser("https://www.effect.website");
+// validUrlString is will be "https://www.effect.website/"
```
gcanti, line 2844: typo

gcanti (approved):

packages/schema/README.md
nikgraf: force pushed the type fix

PR #1858, opened 2024-01-04
Fix platform exports

fubhy:


fubhy:
@tim-smart The `/platform` package never had the default entrypoint. Was that intentional?

tim-smart:
Yeah it was removed as the server modules were ending up in browser bundles.

fubhy:
@tim-smart this should be better then

fubhy:
Hi Datner!

tim-smart:
The Http modules were excluded from the index.ts module as we have the /HttpClient and /HttpServer modules which re-export everything that is needed.

Also we would probably want to prefix everything if we were to add it to index (HttpApp, HttpHeaders etc)

fubhy:
> The Http modules were excluded from the index.ts module as we have the /HttpClient and /HttpServer modules which re-export everything that is needed.
> 
> Also we would probably want to prefix everything if we were to add it to index (HttpApp, HttpHeaders etc)

Aaah, sorry... So you are saying that this was intentional and this PR should be closed? Sorry I didn't understand you comment at first ;-)

tim-smart:
At this stage I will close this. We will be evaluating a POC restructuring of platform soon, and we can re-evaluate the http modules at the same time.

Issue #1825, opened 2023-11-22
JSON Schema to Schema converter


huypham50:
### What is the problem this feature would solve?

Discussed on discord

```typescript
import * as S from "@effect/schema/Schema";
import * as JSONSchema from "@effect/schema/JSONSchema";

// I want to construct a schema
const schema = S.struct({...})
const jsonSchema = JSONSchema.to(UserSchema);

// I want to store the serialized version of the schema (json schema) in a remote database
await orm.insertRow({..., jsonSchema})

// I want to retrieve the row, deserialize, and keep working with it
const storedJsonSchema = await orm.getRow(...) // same as jsonSchema
const originalSchema = JSONSchema.deserialize(storedJsonSchema) // same as schema - looking for this
```

### What is the feature you are proposing to solve the problem?

Would love a `deserialize` method to convert a json schema back to an effect schema. Thanks for the great work!

### What alternatives have you considered?

Plain json schema + ajv

PR #1864, opened 2024-01-05
json schema decoding and AST to TS codegen

leonitousconforti:
Implements json schema decoding and AST to TS codegen. For the json schema decoding, there were two consumer use cases I wanted to cover with this:

1. Singular json schema object or json schema root object decoded to Effect schema - <https://github.com/Effect-TS/effect/issues/1825>
2. Json schema root object decoded to multiple Effect schemas - for decoding things like OpenApi or swagger documents 

And I determined there were three 'decoding' cases that would have to be dealt with in order to implement both consumer use cases:

1. JsonSchema object (not a JsonSchema root object) -> `Schema.Schema<unknown, unknown>`
2. JsonSchema root object with top level definition -> `Schema.Schema<unknown, unknown>`
3. JsonSchema root object without top level definition -> `Record<string, Schema.Schema<unknown, unknown>>`

I have implemented the outlined consumer use cases with two public functions

```js
const decodeSingleSchema = (schema: JsonSchema7Root | JsonSchema7) => Schema.Schema<unknown, unknown>
const decodeMultiSchema = (schema: JsonSchema7Root) => Record<string, Schema.Schema<unknown, unknown>>
```

For the AST to TS codegen, the main consumer case I see for it right now is generating Effect Schemas from decoded json schemas but I also think it might open doors to useful refactoring tools? I'm not entirely sure how VS Code refactoring works but I'm imagining an extension where it can read your schema as an AST, modify it slightly, and then regenerate you a new schema.

If you want a __TLDR__ of what is below: 
:sparkles: JsonSchema decoding upholds the "round-trip" property except for when your effect schema has a refinement or a jsonSchema annotation.
:speech_balloon: There is no way I am aware of to retain all the information about your TS code in the AST, so the codegen is slightly opinionated. I'd like to know if y'all share the same opinions that I do about structuring schemas
:bug: I still have a couple of FIXMEs that I need clarification on 
:construction: and a few todo's to finish up (refactoring to make things more readable)
:rocket: but the core functionality appears to be working.
:rotating_light: I used `@effect/cli` and `@effect/platform-node` for the schema gen cli tool which introduces cyclic package dependencies (will require updating the publishing workflow I think or maybe it should live somewhere else)
<br></br>

For the public API, I really didn't like the idea of one function that returned `Schema.Schema<unknown, unknown> | Record<string, Schema.Schema<unknown, unknown>>` and while it might technically be possible to encapsulate all three of the decoding cases into one function signature that narrowed the return type, I don't see any way that it would work unless the input is a string literal which I don't think is very likely based on the two consumer use cases outlined above. 

I personally don't like the idea of have to check the result to see if its either a schema or a collection of schemas because if you expected to get a single schema back but instead got a collection, that is an error with your json schema (it couldn't parse the top level component) and should be an error. Currently, `decodeSingleSchema` will throw an error if it is given a json schema without a top level component. If others feel like there should be one `decode` function that returns `Schema.Schema<unknown, unknown> | Record<string, Schema.Schema<unknown, unknown>>` we can certainly add it.

All the previous tests still pass, and I enable the part of the json schema tests where it checks jsonSchema->effectSchema->jsonSchema (Keeping Encode and Decode in Sync) is the exact same. The only time this property doesn't hold is when you start with an effect schema and your effect schema contains refinement node / jsonSchema annotation, then its possible that when going from effect schema -> json schema -> effect schema to lose information.

My initial implementation was a two pass approach. In the first pass, it parsed all the definitions (top level and non top level) but didn't set any references. Instead, references are converted to unique symbols. Then, in the post processing step, it mapped over every node in the AST updated unique symbol nodes in place to be suspends with identifiers. This required updating the AST in place (which is readonly) so it meant I had to get a little naughty with the AST. 

My second implementation makes use of the fact that in JS "Primitives are passed by value but Objects are passed by 'copy of a reference"'. So I can make one object before I start parsing that has keys for all the referenced schemas, then I can update those keys whenever they finish parsing. The suspends can be creating during the first pass of the json schema and while some of the referenced keys might resolve to undefined at the time of parsing that schema, by the time all schemas are done being parsed it is guaranteed that all referenced schemas will be resolved. 

Q: Why so convoluted? Why do I think we need to either update the AST in place or do this wired reference passing?
A: For mutually referencing schemas

I also built a small CLI tool. My dream would be to be able to run something like `npx @effect/schema gen <path-to-openapi-document>` and to codegen an effect schema from an openapi/swagger/jsonSchema document. I used the effect cli and platform-node packages for this which introduces cyclic package dependencies into the effect monorepo. Since @effect/cli and @effect/platform-node are only dev dependencies, I think the new publishing workflow wouldn't need to change too much but when publishing @effect/schema you would publish with the most recent versions from npm instead of the local versions.

I am also considering making the decoder interface similar to what the `Pretty` schema module does so that a consumer could specify their own decoder. Say you were using this for number: 

```js
const JsonNumber = Schema.number.pipe(
  Schema.filter((n) => !Number.isNaN(n) && Number.isFinite(n), {
    jsonSchema: { type: "number" }
  })
)
```

it would get encoded to just `{ type: number }` but then you could specify a custom decoder that adds the filter back. Let me know if you think that is a good idea.


packages/schema/src/JSONSchema.ts
```diff
@@ -526,3 +651,279 @@ const go = (ast: AST.AST, $defs: Record<string, JsonSchema7>): JsonSchema7 => {
...
+    // ---------------------------------------------
+    Match.when(isJsonSchema7Object, ({ additionalProperties, patternProperties, properties, required }) => {
+      const fields = Object.entries(properties).map(([name, property]) => {
+        const isRequired = required.includes(name)
+        const decodedProperty = decodeWithReferences(property, references)
+        const withOptional = isRequired
+          ? Schema.required(decodedProperty)
+          : Schema.optional(decodedProperty, { exact: true }) // FIXME: How do I know if this is "exact" / will it always be "exact" here?
+        return [name, withOptional] as const
+      })
```
leonitousconforti, line 787: @gcanti When decoding an object, I check the `required` array to see if it is required/optional, but how do I know the property is exact or not? Will it always be exact when decoding?

gcanti (commented):
@leonitousconforti this shuld be an external tool, like `json-schema-to-zod` for [zod](https://github.com/colinhacks/zod?tab=readme-ov-file#x-to-zod)

leonitousconforti:
> @leonitousconforti this shuld be an external tool, like `json-schema-to-zod` for [zod](https://github.com/colinhacks/zod?tab=readme-ov-file#x-to-zod)

Sounds good. I'll publish it as my own package then

successkrisz:
@leonitousconforti did you end up publishing this? 

I think for "openapi first" development it would be incredibly useful to be able to generate @effect/schema types as well as @effect HTTP clients from the openapi specs 🤩

mkrause:
I published a code gen library for OpenAPI v3.1 to Effect Schema: https://www.npmjs.com/package/openapi-to-effect

PR #1871, opened 2024-01-08
Add `Trie` data structure

SandroMaglione:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

Added `Trie` data structure. 

A `Trie` can be used to to efficiently store and locate keys within a set.
- Search hits take time proportional to the length of the search key
- Search misses involve examining only a few characters

This PR adds a `Trie` module implemented as a **Ternary Search Tree**. A Ternary Search Tree is more space-efficient and allows to use an arbitrary `string` as key (instead of a limited alphabet).

### Questions
- While still being a `Trie`, the new module is technically a Ternary Search Trie (TST). Is the name `Trie` okay?
- Are there some guidelines to provide benchmarks for a new data structure?

### TODO

- [x] More documentation and examples
- Add some more common methods
  - [x] `has`
  - [x] `insertMany`
  - [x] `removeMany`
  - [x] `map`
  - [x] `forEach`
  - [x] `reduce`
  - [x] `isEmpty`
  - [x] `filterMap`
  - [x] `filter` 
  - [x] `compact`
  - [x] `modify`


mikearnaldi:
A few questions:
- why is key fixed to be `string`?
- how does this compare to `HashMap`?

SandroMaglione:
> why is key fixed to be `string`?

Most of the main usecases of Tries have to do with strings, for example for full-text search, autocomplete (predictive text), spell checking. Some of these usecases take advantage of prefix search (`keysWithPrefix`).

Prefix search is also the main feature that makes a Trie more suited than an hash table for certain usecases.

That being said, I think that technically there should be no issue in making the key also a generic `K`. Let me know if we want to support this.

***

> how does this compare to `HashMap`?

Trie are better suited for autocomplete, text search, spell checking, and generally when working with string and prefixes:
- A Trie may use less space since it stores common prefixes in the same nodes (If there are many common prefixes, the space they require is shared). This is the case for dictionaries of words, which makes a Trie ideal for storing and retrieving words from a large list (i.e. searching a list of words inside a text)
- The structure of a Trie allows for quick look up of prefixes of keys and enumerate all entries with a given prefix (`longestPrefixOf` and `keysWithPrefix`)
- A Trie has O(n) lookup time where n is the size of the key, or even less than n time on search misses (predictable). An hash table requires the full string to be scanned to compute the hash (less predictable). In the best case is O(1), but this is data dependent.

mikearnaldi:
> > why is key fixed to be `string`?
> 
> Most of the main usecases of Tries have to do with strings, for example for full-text search, autocomplete (predictive text), spell checking. Some of these usecases take advantage of prefix search (`keysWithPrefix`).
> 
> Prefix search is also the main feature that makes a Trie more suited than an hash table for certain usecases.
> 
> That being said, I think that technically there should be no issue in making the key also a generic `K`. Let me know if we want to support this.
> 
> > how does this compare to `HashMap`?
> 
> Trie are better suited for autocomplete, text search, spell checking, and generally when working with string and prefixes:
> 
> * A Trie may use less space since it stores common prefixes in the same nodes (If there are many common prefixes, the space they require is shared). This is the case for dictionaries of words, which makes a Trie ideal for storing and retrieving words from a large list (i.e. searching a list of words inside a text)
> * The structure of a Trie allows for quick look up of prefixes of keys and enumerate all entries with a given prefix (`longestPrefixOf` and `keysWithPrefix`)
> * A Trie has O(n) lookup time where n is the size of the key, or even less than n time on search misses (predictable). An hash table requires the full string to be scanned to compute the hash (less predictable). In the best case is O(1), but this is data dependent.

Thanks, would such `L` be fully generic or a subtype of `string`, if the latter (which I assume given prefix definition, etc) then this is already OK

packages/effect/src/Trie.ts
```diff
@@ -0,0 +1,198 @@
...
+ * A Trie.
+ *
+ * @since 2.0.0
+ * @category models
+ */
+export interface Trie<out Value> extends Iterable<[string, Value]>, Equal, Pipeable, Inspectable {
+  readonly [TypeId]: {
+    readonly _Key: Types.Invariant<string>
+    readonly _Value: Types.Covariant<Value>
+  }
```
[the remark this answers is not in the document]
SandroMaglione: I was looking at `RedBlackTree` and in it value is Covariant. How is this different from `RedBlackTree`? 

https://github.com/Effect-TS/effect/blob/a904a739459bfd0fa7844b00b902d2fa984fb014/packages/effect/src/RedBlackTree.ts#L34-L39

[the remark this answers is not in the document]
mikearnaldi: probably it is broken in RedBlackTree too. Also I notice you're defining variance for the Key but in Trie you don't have a generic key, it is fixed to `string` so defining a variance entry doesn't change anything

SandroMaglione:
@mikearnaldi I completed the API implementation. I am now going to update and expand the docs and then the PR is ready.

packages/effect/src/Trie.ts
```diff
@@ -0,0 +1,312 @@
...
+ *
+ * @since 2.0.0
+ * @category constructors
+ */
+export const make: <Entries extends Array<readonly [string, any]>>(
+  ...entries: Entries
+) => Trie<Entries[number] extends readonly [any, infer V] ? V : never> = TR.make
+
+/**
+ * Creates an empty `Trie`.
```
fubhy, line 58: Wrong description.

packages/effect/src/Trie.ts
```diff
@@ -0,0 +1,312 @@
...
+const TypeId: unique symbol = TR.TrieTypeId as TypeId
+
+/**
+ * @since 2.0.0
+ * @category symbol
+ */
+export type TypeId = typeof TypeId
+
+/**
+ * A Trie.
```
fubhy, line 20: Redundant.

packages/effect/src/Trie.ts
```diff
@@ -0,0 +1,312 @@
...
+ */
+export const unsafeGet: {
+  (key: string): <V>(self: Trie<V>) => V
+  <V>(self: Trie<V>, key: string): V
+} = TR.unsafeGet
+
+/**
+ * Remove the entry for the specified key in the `Trie`.
+ *
+ * @since 2.0.0
```
fubhy, line 211: What about a dedicated docs category for mutations?

packages/effect/src/internal/trie/node.ts
```diff
@@ -0,0 +1,11 @@
...
+export class Node<out V> {
+  constructor(
+    public key: string,
+    public count: number,
+    public value: V | undefined = undefined,
+    public left: Node<V> | undefined = undefined,
+    public mid: Node<V> | undefined = undefined,
+    public right: Node<V> | undefined = undefined
+  ) {}
+}
```
fubhy, line 11: Not sure how much this needs to be optimized for performance, but if this is just a value bag, you don't need a class for it. Especially not if it's only used internally. A plain object & interface should give you significantly better performance (easy to proof that with a benchmark).

Considering the simplicity of this, you might just want to create these plain objects inline even (a `make` function wouldn't really add any value imho).

packages/effect/src/internal/trie.ts
```diff
@@ -0,0 +1,713 @@
...
+>(
+  2,
+  <V>(self: TR.Trie<V>, key: string) => {
+    let n: Node.Node<V> | undefined = (self as TrieImpl<V>)._root
+    if (n === undefined || key.length === 0) return self
+
+    const count = n.count - 1
+    // -1:left | 0:mid | 1:right
+    const d_stack: Array<Ordering.Ordering> = []
+    const n_stack: Array<Node.Node<V>> = []
```
fubhy, line 440: Just regarding style: We don't use snake case anywhere else in the code base.

fubhy (changes requested):

packages/effect/src/internal/trie/node.ts
fubhy: And when I say "significant", this is what I mean :-)

https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJweXREb012SlFaQXNvbGtrdVlka2MiLCJjb2RlIjoiZnVuY3Rpb24gUG9pbnQoeCwgeSkge1xuICByZXR1cm4geyB4LCB5IH1cbn1cblxuZm9yIChsZXQgaSA9IDA7IGkgPD0gMTAwMDA7IGkrKykge1xuICBQb2ludCgxMjMsIDQ1Nilcbn0iLCJuYW1lIjoiT2JqZWN0IExpdGVyYWwifSx7ImlkIjoiS21SMjdLeWxlR2pQMHlDb0lrMDQ5IiwiY29kZSI6ImNsYXNzIFBvaW50IHtcbiAgY29uc3RydWN0b3IoeCwgeSkge1xuICAgIHRoaXMueCA9IHhcbiAgICB0aGlzLnkgPSB5XG4gIH1cbn1cblxuZm9yIChsZXQgaSA9IDA7IGkgPD0gMTAwMDA7IGkrKykge1xuICBuZXcgUG9pbnQoMTIzLCA0NTYpXG59IiwibmFtZSI6IkNsYXNzIn1dLCJjb25maWciOnsibmFtZSI6Ik9iamVjdCBMaXRlcmFsIHZzLiBDbGFzcyIsInBhcmFsbGVsIjp0cnVlLCJnbG9iYWxUZXN0Q29uZmlnIjp7ImRlcGVuZGVuY2llcyI6W119LCJkYXRhQ29kZSI6IiJ9fQ==

fubhy: object-literal-vs -class-(10 000-iterations)

packages/effect/src/internal/trie.ts
SandroMaglione: I borrowed this naming from `RedBlackTree`. Should we update there too?

https://github.com/Effect-TS/effect/blob/8eec87e311ce55281a98517e6df0ef103b43e8a8/packages/effect/src/internal/redBlackTree.ts#L278-L279

SandroMaglione:
@fubhy @mikearnaldi I added all the documentation and completed the implementation. The PR is ready for a complete review.

fubhy (approved):
LGTM

I'll have a think about what we can do to make the `codegen` step for the barrel files a bit ... less easily skippable.

SandroMaglione:
@fubhy LGTM as well 👍

fubhy:
@SandroMaglione Please also check out https://github.com/Effect-TS/effect/pull/1871/commits/338e1a743f8c482dcac2789972b9875e56d1bff3

I would propose to simplify this to plain object literlas. That way we also don't have to initialize this internal node representation with a bunch of `undefined` properties in some places. 

I think we should also patch this in `RedBlackTree` (the `Node` value bag).

The absolute performance gain is negligible but in relative terms it's a lot and that makes a difference imho. And since there's no DX downgrade by doing this (imho) we should do it. Not sure what Mike things.

SandroMaglione:
> I think we should also patch this in RedBlackTree (the Node value bag)

I think this should be part of another PR. I also opened [another issue](https://github.com/Effect-TS/effect/issues/1889) related to `RedBlackTree`, we could refactor `RedBlackTree` from there

fubhy:
> > I think we should also patch this in RedBlackTree (the Node value bag)
> 
> I think this should be part of another PR. I also opened [another issue](https://github.com/Effect-TS/effect/issues/1889) related to `RedBlackTree`, we could refactor `RedBlackTree` from there

Oh for sure. That's what I meant :-)

PR #1884, opened 2024-01-09
Add `Cron` module and `Schedule.cron` constructor.

fubhy:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This adds new module `Cron` for parsing cron expressions. It also adds a new `Schedule.cron` constructor that allows the creation of schedules like this:

```ts
// At 04:30 on day-of-month 5 and 15 and on Wednesday.
Schedule.cron("30 4 5,15 * WED")
```

### DIFF
--- a/packages/effect/test/Cron.test.ts
+++ b/packages/effect/test/Cron.test.ts
@@ -1,0 +1,97 @@
+import { assertFalse, assertTrue, deepStrictEqual } from "effect-test/util"
+import * as Cron from "effect/Cron"
+import * as Either from "effect/Either"
+import * as Equal from "effect/Equal"
+import { identity } from "effect/Function"
+import { describe, it } from "vitest"
+
+const parse = (input: string) => Either.getOrThrowWith(Cron.parse(input), identity)
+const match = (input: string, date: Date) => Cron.match(parse(input), date)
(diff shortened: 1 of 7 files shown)


packages/effect/test/Cron.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    const after = new Date("2024-01-04 16:21:00")
+    deepStrictEqual(next("5 0 8 2 *", after), new Date("2024-02-08 00:05:00"))
+    deepStrictEqual(next("15 14 1 * *", after), new Date("2024-02-01 14:15:00"))
+    deepStrictEqual(next("23 0-20/2 * * 0", after), new Date("2024-01-07 00:23:00"))
+    deepStrictEqual(next("5 4 * * SUN", after), new Date("2024-01-07 04:05:00"))
+    deepStrictEqual(next("5 4 * DEC SUN", after), new Date("2024-12-01 04:05:00"))
+  })
+
+  it("sequence", () => {
+    const start = new Date("2024-01-01 00:00:00")
```
fubhy, line 81: @mikearnaldi set fixed start time for assertion.

I still think we should still consider adding support for user-defined timezones, but that was not the cause of the test fail. That can be a follow-up imho.

PR #1891, opened 2024-01-10
Struct: add `get`, closes #1890

### LINKED ISSUE #1890: Implement Struct.get

Similar to `Struct.pick` it would be nice to have a util function that can extract the value of a single property within the given object. This is useful when iterating over an array of objects in order to extract just one property, for example:

```ts
const users = [ { name: 'Bob', age: 42 }, { name: 'Alice', age: 21 } ];
const getName = Struct.get('name');
const allNames = users.map(getName) // [ 'Bob', 'Alice' ]
```

The implementation could look like this

```ts
function get<T extends string>(key: T): <S extends Record<T, unknown>>(s: S) => S[T] {
    return (s) => {
        return s[key];
    };
}
```

Example Usage:

```ts
const getAge = get('age');

type User = {
    age: number;
    name: string;
};
const user: User = { name: 'Bob', age: 42 };
const result = getAge(user); // number

type JobDescription = {
    title: string;
};
const job: JobDescription = { title: 'Sales-Manager' };
const result2= getAge(job); // type error
```

_No response_

gcanti:
## Type

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #1890
- Closes #1890

### DIFF
--- a/packages/effect/dtslint/Struct.ts
+++ b/packages/effect/dtslint/Struct.ts
@@ -1,8 +1,157 @@
 import { pipe } from "effect/Function"
 import * as S from "effect/Struct"
 
+const asym = Symbol.for("effect/dtslint/a")
+const bsym = Symbol.for("effect/dtslint/b")
+const csym = Symbol.for("effect/dtslint/c")
+const dsym = Symbol.for("effect/dtslint/d")
+
+declare const stringNumberRecord: Record<string, number>
+declare const symbolNumberRecord: Record<symbol, number>
+declare const numberNumberRecord: Record<number, number>
+
+const stringStruct = { a: "a", b: 1, c: true }
+const symbolStruct = { [asym]: "a", [bsym]: 1, [csym]: true }
+const numberStruct = { 1: "a", 2: 1, 3: true }
+
+// -------------------------------------------------------------------------------------
+// evolve
+// -------------------------------------------------------------------------------------
+
 // $ExpectType { a: boolean; }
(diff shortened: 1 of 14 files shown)


packages/effect/dtslint/Struct.ts
```diff
@@ -1,8 +1,157 @@
...
+const dsym = Symbol.for("effect/dtslint/d")
+
+declare const stringNumberRecord: Record<string, number>
+declare const symbolNumberRecord: Record<symbol, number>
+declare const numberNumberRecord: Record<number, number>
+
+const stringStruct = { a: "a", b: 1, c: true }
+const symbolStruct = { [asym]: "a", [bsym]: 1, [csym]: true }
+const numberStruct = { 1: "a", 2: 1, 3: true }
+
```
lo1tuma, line 16: I think two more interesting test cases are:

* Using template strings in index signatures:
    ```ts
    declare const templateStringNumberRecord: Record<`a${string}`, number>
    
    // $ExpectType number | undefined
    pipe(templateStringNumberRecord, S.get("a"))
    ```
* Using an intersection of a struct and record:
     ```ts
    declare const structRecordIntersection: Record<string, number> & { a: number };
    
    // $ExpectType number
    pipe(structRecordIntersection, S.get("a"))

    // @ts-expect-error
    pipe(structRecordIntersection, S.get("b"))
    ```

PR #1892, opened 2024-01-10
Schema: fix ParseIssue.actual on transformation

matheuspuel:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

It was showing the partially transformed value instead of the original value.


gcanti (approved):
good catch, thanks @matheuspuel

PR #1896, opened 2024-01-10
feat: Add immediate:boolean to RunForkOptions

TylorS:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- Adds `immediate: boolean` to `RunForkOptions` to enable creating a Fiber which uses `fiberRuntime.start()` or `fiberRuntime.resume()`

- Adds `RunCallbackOptions<E, A>` which extends `RunForkOptions` with an `onExit` callback, such that runCallback has access to the same fork options as runFork.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -4749,7 +4749,7 @@ export const runFork: <E, A>(effect: Effect<never, E, A>) => Fiber.RuntimeFiber<
  */
 export const runCallback: <E, A>(
   effect: Effect<never, E, A>,
-  onExit?: (exit: Exit.Exit<E, A>) => void
+  options?: Runtime.RunCallbackOptions<E, A> | undefined
 ) => Runtime.Cancel<E, A> = _runtime.unsafeRunEffect
 
 /**
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -4741,7 +4741,10 @@ export const makeSemaphore: (permits: number) => Effect<never, never, Semaphore>
  * @since 2.0.0
  * @category execution
  */
-export const runFork: <E, A>(effect: Effect<never, E, A>) => Fiber.RuntimeFiber<E, A> = _runtime.unsafeForkEffect
+export const runFork: <E, A>(
+  effect: Effect<never, E, A>,
+  options?: Runtime.RunForkOptions
+) => Fiber.RuntimeFiber<E, A> = _runtime.unsafeForkEffect
 
 /**
  * @since 2.0.0
--- a/packages/effect/src/internal/runtime.ts
+++ b/packages/effect/src/internal/runtime.ts
@@ -1,3 +1,4 @@
+import { equals } from "effect/Equal"
 import type * as Cause from "../Cause.js"
 import * as Context from "../Context.js"
 import type * as Effect from "../Effect.js"
@@ -60,7 +61,16 @@ export const unsafeFork = <R>(runtime: Runtime.Runtime<R>) =>
   if (options?.scope) {
     effect = core.flatMap(
       _scope.fork(options.scope, executionStrategy.sequential),
-      (closeableScope) => core.onExit(effect, (exit) => _scope.close(closeableScope, exit))
+      (closeableScope) =>
+        core.zipRight(
+          core.scopeAddFinalizer(
+            closeableScope,
+            core.fiberIdWith((id) =>
+              equals(id, fiberRuntime.id()) ? core.unit : core.interruptAsFiber(fiberRuntime, id)
+            )
+          ),
+          core.onExit(effect, (exit) => _scope.close(closeableScope, exit))
+        )
     )
   }
 
--- a/packages/effect/src/internal/runtime.ts
+++ b/packages/effect/src/internal/runtime.ts
@@ -15,8 +15,10 @@ import type * as ReadonlyArray from "../ReadonlyArray.js"
 import type * as Runtime from "../Runtime.js"
 import type * as RuntimeFlags from "../RuntimeFlags.js"
 import * as _scheduler from "../Scheduler.js"
+import * as _scope from "../Scope.js"
 import * as InternalCause from "./cause.js"
 import * as core from "./core.js"
+import * as executionStrategy from "./executionStrategy.js"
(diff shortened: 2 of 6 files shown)


mattiamanzati:
Bonus, could we please have this also on unsafeRunCallback? It is useful to have the callback eventually be called synchronously if the effect is sync as well

TylorS:
@mattiamanzati Updated in the latest pushes

mattiamanzati:
@TylorS Thanks 🙏
Can't wait for this PR to be accepted and released!

mikearnaldi:
> @mattiamanzati I should be able to add the `immediate` flag there as well. I'm going to remove the `Scope` option though, as it could lead to memory leaks if the surrounding scope is open indefinitely as the Fiber is never removed from the Scope upon exiting

Why would the fiber not being removed upon exit? that should be the behaviour

TylorS:
@mikearnaldi When you do `Scope.addFinalizer(scope, Fiber.interrupt(fiber))`, there is no corresponding API for removing the finalizer from the Scope after the Fiber exits. So if the scope it is being added to lives indefinitely (i.e. the scope of a React component's event listeners), the Fibers will be unable to be garbage collected due to being kept within the open Scope.

mikearnaldi:
> @mikearnaldi When you do `Scope.addFinalizer(scope, Fiber.interrupt(fiber))`, there is no corresponding API for removing the finalizer from the Scope after the Fiber exits. So if the scope it is being added to lives indefinitely (i.e. the scope of a React component's event listeners), the Fibers will be unable to be garbage collected due to being kept within the open Scope.

you're right, I will investigate if adding an api to remove a finaliser may be useful

TylorS:
In the same vein, I'm not positive that `forkIn` or `forkScoped` are entirely safe. They do utilize forked Scope's which allows the Fiber to be garbage collected after completion, but there's a (albeit smaller) memory leak of child scopes if the part scope is kept open indefinitely.

mikearnaldi:
```ts
/** @internal */
export const forkIn = dual<
  (scope: Scope.Scope) => <R, E, A>(self: Effect.Effect<R, E, A>) => Effect.Effect<R, never, Fiber.RuntimeFiber<E, A>>,
  <R, E, A>(self: Effect.Effect<R, E, A>, scope: Scope.Scope) => Effect.Effect<R, never, Fiber.RuntimeFiber<E, A>>
>(
  2,
  (self, scope) =>
    core.uninterruptibleMask((restore) =>
      core.flatMap(scope.fork(executionStrategy.sequential), (child) =>
        pipe(
          restore(self),
          core.onExit((exit) => child.close(exit)),
          fiberRuntime.forkDaemon,
          core.tap((fiber) =>
            child.addFinalizer(() =>
              core.fiberIdWith((fiberId) =>
                Equal.equals(fiberId, fiber.id()) ?
                  core.unit :
                  core.asUnit(core.interruptFiber(fiber))
              )
            )
          )
        ))
    )
)
```

This is safe, it forks a separate scope and that scope is closed as the fiber ends

TylorS:
But closing that child/forked Scope doesn't remove its reference as a Finalizer in the parent Scope when it is closed

TylorS:
I take that back, I found the code that keeps it safe here -https://github.com/Effect-TS/effect/blob/main/packages%2Feffect%2Fsrc%2Finternal%2FfiberRuntime.ts

packages/effect/src/Effect.ts
```diff
@@ -4749,7 +4749,7 @@ export const runFork: <E, A>(effect: Effect<never, E, A>) => Fiber.RuntimeFiber<
  */
 export const runCallback: <E, A>(
   effect: Effect<never, E, A>,
-  onExit?: (exit: Exit.Exit<E, A>) => void
+  options?: Runtime.RunCallbackOptions<E, A> | undefined
```
tim-smart, line 4752: Will need to add a minor changeset for this.

tim-smart (changes requested):

packages/effect/src/Effect.ts
TylorS: Updated

tim-smart (approved):

tim-smart:
> Bonus, could we please have this also on unsafeRunCallback? It is useful to have the callback eventually be called synchronously if the effect is sync as well

The current behaviour is to start immediately, so you will see no change with this PR.

mattiamanzati:
> > Bonus, could we please have this also on unsafeRunCallback? It is useful to have the callback eventually be called synchronously if the effect is sync as well
> 
> The current behaviour is to start immediately, so you will see no change with this PR.

Uhm but run callback relies on run fork, so it should be affected as well right?

TylorS:
This PR enables non-immediate starts through `immediate:false` as the default is `true` to avoid breaking changes

mattiamanzati:
> enables

Ohh, I though of it in the completely opposite direction.

mikearnaldi:
Can we add back Scope in the style used by forkIn? also it is worth updating the main runFork method in Effect.ts to take the RunOptions currently only Runtime exposes it

TylorS:
@mikearnaldi PR Updated with Scope [here](https://github.com/Effect-TS/effect/pull/1896/commits/e00349277787453ad0e4425425b7657b3183ad0e) and the additional options to Effect.runFork [here](https://github.com/Effect-TS/effect/pull/1896/commits/ec58bab4ee5f6efefe6eb680653f70980b772801)

packages/effect/src/internal/runtime.ts
```diff
@@ -55,6 +55,15 @@ export const unsafeFork = <R>(runtime: Runtime.Runtime<R>) =>
     runtime.runtimeFlags
   )
 
+  let effect: Effect.Effect<R, E, A> = self
+
+  if (options?.scope) {
```
tim-smart, line 60: Need to add the finalizer :)

TylorS: 🤦 , updated https://github.com/Effect-TS/effect/pull/1896/commits/8215ce67a8a1b0c22e4de4fdac53bbd7e98ea058

tim-smart:
@TylorS if you rebase that should fix CI

TylorS:
Rebased

PR #1900, opened 2024-01-11
Annotations for transform nodes

leonitousconforti:
## Type

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I would like to add annotations directly to transform nodes while creating them. Is there a reason these function signatures aren't present other than there weren't needed? Does it go against some rule about adding annotations directly to transform nodes? 

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/brave-yaks-marry.md
+++ b/.changeset/brave-yaks-marry.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+annotations for transform nodes
--- a/packages/schema/src/Schema.ts
+++ b/packages/schema/src/Schema.ts
@@ -1455,6 +1455,7 @@ export const transform: {
       annotations
     )
 )
+
 /**
  * Creates a new `Schema` which transforms literal values.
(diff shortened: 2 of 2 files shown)


gcanti:
> Is there a reason these function signatures aren't present other than there weren't needed?

to keep them as simple as possible basically, you can always add annotations with `.pipe(Schema.annotations(...))`

mikearnaldi:
> > Is there a reason these function signatures aren't present other than there weren't needed?
> 
> to keep them as simple as possible basically, you can always add annotations with `.pipe(Schema.annotations(...))`

I can see the usefulness of doing it in one shot, what I don't particularly like is making the annotations mandatory, maybe an optional field in the options?

mikearnaldi:
After discussion with @gcanti we agreed that allowing annotations of transform feels arbitrary, better to have the standard `.pipe(Schema.annotations({...}))`

PR #1903, opened 2024-01-11
convert value bag classes to object literals

fubhy:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Class instance initialization comes with a "substantial" (relatively speaking) overhead compared to raw object literals. They should therefore be avoided for plain value bags. 

## Related

- Related Issue #1871

### DIFF
--- a/packages/effect/src/MutableList.ts
+++ b/packages/effect/src/MutableList.ts
@@ -79,14 +79,16 @@ interface MutableListImpl<A> extends MutableList<A> {
 interface LinkedListNode<T> {
   removed: boolean
   value: T
-  prev?: LinkedListNode<T> | undefined
-  next?: LinkedListNode<T> | undefined
+  prev: LinkedListNode<T> | undefined
+  next: LinkedListNode<T> | undefined
 }
 
 /** @internal */
 const makeNode = <T>(value: T): LinkedListNode<T> => ({
   value,
-  removed: false
+  removed: false,
+  prev: undefined,
+  next: undefined
 })
 
 /**
--- a/packages/effect/src/MutableList.ts
+++ b/packages/effect/src/MutableList.ts
@@ -76,12 +76,18 @@ interface MutableListImpl<A> extends MutableList<A> {
 }
 
 /** @internal */
-class LinkedListNode<T> {
-  removed = false
-  prev: LinkedListNode<T> | undefined = undefined
-  next: LinkedListNode<T> | undefined = undefined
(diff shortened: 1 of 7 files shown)


tim-smart:
What is the perf gain in doing this approx.?

mikearnaldi:
> What is the perf gain in doing this approx.?

https://github.com/Effect-TS/effect/pull/1871#discussion_r1447212356

tim-smart:
> [#1871 (comment)](https://github.com/Effect-TS/effect/pull/1871#discussion_r1447212356)

Nice :)

fubhy:
8c3gpf

packages/effect/src/MutableList.ts
```diff
@@ -76,13 +76,19 @@ interface MutableListImpl<A> extends MutableList<A> {
...
 /** @internal */
-class LinkedListNode<T> {
-  removed = false
-  prev: LinkedListNode<T> | undefined = undefined
-  next: LinkedListNode<T> | undefined = undefined
-  constructor(readonly value: T) {}
+interface LinkedListNode<T> {
+  removed: boolean
+  value: T
+  prev?: LinkedListNode<T> | undefined
```
TylorS, line 82: I'd suggest making these `prev` and `next` values required, though still `| undefined`, to avoid assigning these fields in different orders over time, leading to de-optimizations

fubhy: @TylorS Thanks for the suggestion. Can you clarify that for me though? How would assigning it to `undefined` just to preserve assignment order change the performance characteristics?

TylorS: In the engines, they optimize object creation through the use of "hidden classes", and they utilize order to improve the speculation of inlining optimizations. I find this blog a pretty good primer on object poly/megamorphism (also about a tool that helps spot these) https://devblogs.microsoft.com/typescript/introducing-deopt-explorer/

fubhy: TIL, thanks!

TylorS: No problem! I kinda think of the object's memory being a lot like that of an `Array`, it's optimal to know that property "foo" is always at index 0, and not sometimes at index 1 (or 2 or whatever). The hidden classes essentially map your user-land keys, like "foo", into spaces in memory at particular indexes with other information like the amount of memory that it will likely need allocated

PR #1906, opened 2024-01-11
Schema: show error message when throw ParseError

matheuspuel:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When throw ParseError it was printing
```
ParseError:
    at Object.parseError...
```
now it prints
```
ParseError: <TreeFormatter message>
    at Object.parseError...
```

### DIFF
--- a/packages/schema/test/ParseResult.test.ts
+++ b/packages/schema/test/ParseResult.test.ts
@@ -43,6 +43,10 @@ describe("ParseResult", () => {
       )
   })
 
+  it.only("Error.stack", () => {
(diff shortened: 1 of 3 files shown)


packages/schema/test/ParseResult.test.ts
```diff
@@ -43,6 +43,10 @@ describe("ParseResult", () => {
       )
   })
 
+  it.only("Error.stack", () => {
```
gcanti, line 46: there's a `it.only` here

gcanti (changes requested):

gcanti (approved):

PR #1922, opened 2024-01-15
add context tracking to Schema, closes #1873, #1954, #1956

### LINKED ISSUE #1873: Add context tracking to Schema

Tracking context usage in schema, currently it's only possible to effects that don't access context (or hack it).

Adding either:
- a single parameter `R` that tracks requirements of both encoding and decoding
- two different parameters `REnc` and `RDec` to track requirements of encoding and decoding independently

Note that for the unsafe parse functions (sync) we should only allow schemas with `never` in the context positions

_No response_

gcanti:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Closes #1873


packages/schema/src/Serializable.ts
```diff
@@ -93,15 +93,17 @@ export const exitSchema = <IE, E, IA, A>(
  * @since 1.0.0
  * @category model
  */
-export interface SerializableWithResult<IS, S, IE, E, IA, A> extends Serializable<IS, S>, WithResult<IE, E, IA, A> {}
+export interface SerializableWithResult<R, IS, S, IE, E, IA, A>
```
tim-smart, line 96: I think a separate R channel for the success/failure schema's should be added.

packages/schema/src/Schema.ts
```diff
@@ -4405,15 +4413,15 @@ type MissingSelfGeneric<Usage extends string, Params extends string = ""> =
  * @category classes
  * @since 1.0.0
  */
-export interface Class<I, A, C, Self, Inherited = Data.Case> extends Schema<I, Self> {
+export interface Class<I, A, C, Self, Inherited> extends Schema<never, I, Self> {
```
tim-smart, line 4416: This should likely be `Class<R, I, A, C, Self, Inherited> extends Schema<R, I, Self>`

packages/schema/src/Schema.ts
```diff
@@ -4468,33 +4476,35 @@ export interface Class<I, A, C, Self, Inherited = Data.Case> extends Schema<I, S
  * @since 1.0.0
  */
 export const Class = <Self>() =>
-<Fields extends StructFields>(
+<Fields extends StructFields<never>>(
```
tim-smart, line 4479: `StructFields<any>`

packages/schema/src/Schema.ts
```diff
@@ -4468,33 +4476,35 @@ export interface Class<I, A, C, Self, Inherited = Data.Case> extends Schema<I, S
...
   fields: Fields
 ): [unknown] extends [Self] ? MissingSelfGeneric<"Class">
   : Class<
     Simplify<FromStruct<Fields>>,
```
tim-smart, line 4483: Extract the `R` here from the struct fields.

PR #1935, opened 2024-01-16
make Effect.tap safer

matheuspuel:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Effect.tap` discards the effect result, so a function that returns a simple value is useless, e.g.:
```typescript
Effect.tap(Effect.succeed(1), () => 2)
```
Disallowing this would make the type checker notify of some mistakes. For example, in complex pipelines it could take some time to understand why the following effect is not being called:
```typescript
const log1 = () => Console.log(1)
Effect.tap(Effect.succeed(1), () => log1) // forgot to call the function
```

It's not a problem in `Effect.andThen` because it returns the value and the error would be spotted on the next step.

The only use case I found is to quickly introspect some value: `Effect.tap((v) => console.log(v))`, but it's temporary and could be done in many other ways.

Besides that, passing a Promise without a wrapping function is probably a mistake. This will run without even using `Effect.run*`:
```typescript
Effect.tap(Effect.unit, new Promise(() => console.log('immediate')))
```

This Promise won't resolve again:
```typescript
pipe(
  Effect.unit,
  Effect.tap(new Promise(res => res(console.log('only once')))),
  Effect.repeatN(1),
  Effect.runPromise,
)
```


tim-smart:
A resolved promise will immediately call the .then callback no matter how many times you attach it. But enforcing them to be lazy isn't a bad idea.

packages/effect/src/internal/core.ts
```diff
@@ -1136,47 +1136,54 @@ export const sync = <A>(evaluate: LazyArg<A>): Effect.Effect<never, never, A> =>
 /* @internal */
 export const tap = dual<
   {
-    <A, X>(
-      f: (a: NoInfer<A>) => X
+    <A, R1, E1>(
+      f: (a: NoInfer<A>) => Effect.Effect<R1, E1, unknown>
```
tim-smart, line 1140: Instead of `unknown`, using a type hole is safer (you can just add an unused generic like `X` or `_`)

packages/effect/dtslint/Effect.ts
```diff
@@ -14,6 +14,28 @@ declare const numberRecord: Record<string, Effect.Effect<"dep-4", "err-4", numbe
...
+Effect.tap(string, number)
+
+// $ExpectType Effect<"dep-1", "err-1" | UnknownException, string>
+Effect.tap(string, async (
+  // $ExpectType string
+  _a
+) => console.log(""))
+
+// @ts-expect-error
+Effect.tap(string, () => 1)
```
tim-smart, line 33: This should be valid code, as we want to allow `Effect.tap((_) => console.log(_))` / `Effect.tap((_) => map.delete(_))` / `Effect.tap((_) => postMessage(_))` / whatever (perform sync side effects without having to wrap with Effect.sync)

This was the primary reason we changed the .tap signature.

tim-smart (changes requested):

packages/effect/dtslint/Effect.ts
matheuspuel: Hmm, now I see the use case for it.
Unfortunately, I won't be able to restrict mistakes like this:
```typescript
const log1 = () => Console.log(1)
Effect.tap(Effect.succeed(1), () => log1) // forgot to call the function
```
But ok, I can live with that.

packages/effect/src/internal/core.ts
matheuspuel: I'm curious about this. How is it safer? Any examples?

mikearnaldi: It improves inference, otherwise TS may mistakenly consider it as a candidate for inference in generic contexts

tim-smart:
Could you bump this to a minor release? It is a breaking change :)

packages/effect/dtslint/Effect.ts
```diff
@@ -14,6 +14,34 @@ declare const numberRecord: Record<string, Effect.Effect<"dep-4", "err-4", numbe
 // forEach
```
tim-smart, line 14:
```suggestion
// tap
```

tim-smart (changes requested):
I think any changes to .tap should be mirrored in .andThen, as they are sibling apis.

matheuspuel:
For `Effect.andThen`, instead of forbidding passing an unwrapped Promise, I think it's more appropriate to just return it like any other non-Effect value.

tim-smart:
I'm not convinced on having Promise's handled in different ways depending on how they are supplied.
I starting to think we should just leave the signature as they currently are?

I'll cc @mikearnaldi here as it would be nice to have one more opinion from a maintainer.

mikearnaldi:
> I'm not convinced on Promise's being handled in different ways depending on how they are supplied. I starting to think we should just leave the signature as they currently are?
> 
> I'll cc @mikearnaldi here as it would be nice to have one more opinion from a maintainer.

not sure there is a lot of value in trying to make this safer imho

mikearnaldi:
After a good amount of thoughts I confirm the opinion, it doesn't add a significant level of safety and it risks prohibiting valid cases

PR #1947, opened 2024-01-18
feat: Effect.some

sukovanej:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I was missing the counterpart of `Effect.none`.


mikearnaldi:
The reason it is not there is because `Option<A>` is already an effect, so you can just use `Effect.flatten`

sukovanej:
Ok, makes sense.

PR #1957, opened 2024-01-21
Effect: cache the `FiberId` hash in the constructor

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/internal/fiberId.ts
+++ b/packages/effect/src/internal/fiberId.ts
@@ -109,7 +109,7 @@ class Composite implements FiberId.Composite {
     readonly right: FiberId.FiberId
   ) {
     this._hash = pipe(
-      Hash.string(FiberIdSymbolKey + this._tag),
+      Hash.string(`${FiberIdSymbolKey}-${this._tag}`),
       Hash.combine(Hash.hash(this.left)),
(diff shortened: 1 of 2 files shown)


packages/effect/src/internal/fiberId.ts
```diff
@@ -66,18 +67,20 @@ class None implements FiberId.None {
...
-  ) {}
-  [Hash.symbol](): number {
-    return pipe(
-      Hash.hash(FiberIdSymbolKey),
-      Hash.combine(Hash.hash(this._tag)),
-      Hash.combine(Hash.hash(this.id)),
-      Hash.combine(Hash.hash(this.startTimeMillis))
+  ) {
+    this._hash = pipe(
+      Hash.string(FiberIdSymbolKey + this._tag),
```
tim-smart, line 76: I think we could just use

```ts
Hash.string(`${FiberIdSymbolKey}-${this._tag}-${this.id}-${this.startTimeMillis}`)
```

IMax153: Works for me 👍

PR #1958, opened 2024-01-22
Fix signatures related to predicates, closes #1916

### LINKED ISSUE #1916: From Discord: Possible Salad in Effect.filterXXX Functions
# Summary
In the above conversation, there is a discussion about the correctness of the types in the `Effect.filterXXX` functions, specifically `filterOrElse` and `filterOrFail`. The concern raised is that the current types may allow calling the filter function with values that are not defined in its domain, leading to potential runtime errors.

The suggestion is to flip the extends constraint in the type signatures of these functions to ensure that the filter function is called with values that are "broader" than the filtered values. This would help prevent calling the filter function with incompatible values and reduce the risk of runtime errors.

Key takeaways from the conversation:
1. There is a concern about the correctness of the types in the `Effect.filterXXX` functions.
2. The current types may allow calling the filter function with values that are not defined in its domain, potentially leading to runtime errors.
3. Flipping the extends constraint in the type signatures of these functions could help prevent such issues.
4. The intention is to provide feedback to benefit the community and improve the library.

# Discord thread

https://discord.com/channels/795981131316985866/1195374461772382308

gcanti:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Discord: https://discord.com/channels/795981131316985866/1199039549062512640/1199060809679130644
- Closes #1916

### DIFF
--- a/.changeset/fresh-tigers-tie.md
+++ b/.changeset/fresh-tigers-tie.md
@@ -1,0 +1,8 @@
+---
+"effect": patch
+---
+
+Fix filters signature:
+
+- `Either`
+  - `filterOrLeft`
--- a/.changeset/fresh-tigers-tie.md
+++ b/.changeset/fresh-tigers-tie.md
@@ -2,7 +2,4 @@
 "effect": patch
 ---
 
-Fix filters signature:
-
-- `Either`
-  - `filterOrLeft`
+Fix signatures related to predicates, closes #1916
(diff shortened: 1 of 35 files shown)


tim-smart:
Did you also want to do the Effect ones at the same time? Otherwise I will look doing the same for those.

gcanti:
@tim-smart after many trials and errors I think I've found a better approach that uses `NoInfer`

tim-smart:
Using NoInfer seems like a good option :)

PR #1986, opened 2024-01-26
Add Either.tryPromise

bjacobso:
Add a helper for streamlined wrapping of promise in an Either. I'd like to introduce Effect into our codebase, without going all in on the Effect type yet. I think this will help illustrate the value of using things like `Either`, `Data.TaggedError`, `Match` and `Schema` without needing to adopt fully effectual code paths.

Feedback welcome if there is another approach or a way to use an effect in a similar manner

```typescript
// Example
function getUser(id: number): Promise<Either.Either<UserNotFound, User>> {
  return Either.tryPromise({
    try: async () => {
       const result = await db.sql("...");
       // process result
    },
    catch: (error) => new UserNotFound(...)
  });
}
```

### DIFF
--- a/.changeset/calm-dodos-jump.md
+++ b/.changeset/calm-dodos-jump.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add Either.tryPromise
--- a/packages/effect/src/Either.ts
+++ b/packages/effect/src/Either.ts
@@ -187,6 +187,40 @@ export {
   try_ as try
 }
 
+const tryPromise_: {
+  <A, E>(
+    options: {
+      readonly try: LazyArg<Promise<A>>
+      readonly catch: (error: unknown) => E
+    }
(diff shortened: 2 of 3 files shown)


patroza:
I believe the absence of anything async/Promise related in the Either module is not by coincidence.
Either is a result/output type, where as Effect (Task/TaskEither equivalent in fp-ts) models operations.
Of course for a softer landing you could always put the proposed utility somewhere in user app land.

patroza:
As to using an Effect in similar manner:
`<E, A>(eff: Effect.Effect<never, E, A>) => Effect.runPromise(Effect.either(eff))`

or if you have a Runtime with some Context prepared:
```ts
const runPromise = Runtime.runPromise(rt)
<E, A>(eff: Effect.Effect<typeof rt["context"], E, A>) => runPromise(Effect.either(eff))`

PR #1995, opened 2024-01-26
use TimeoutException for timeout

sukovanej:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I was staring at a failing test shouting `NoSuchElementException` at me and it took me some time to realise it is actually a timeout. 

While I somewhat get the point that being able to easily convert such an error to `Option` using `Effect.optionFromOptional` is convenient, I believe it is quite confusing. In such a case using explicit `Effect.catchTag("TimeoutException", () => Effect.succeedNone)` and communicating clearly what's happening to a reader of my code is IMHO better.

### DIFF
--- a/packages/effect/test/Effect/promise.test.ts
+++ b/packages/effect/test/Effect/promise.test.ts
@@ -17,7 +17,7 @@ describe("Effect", () => {
     })
     const program = effect.pipe(
       Effect.timeout("10 millis"),
-      Effect.optionFromOptional
+      Effect.option
     )
     const exit = await Effect.runPromiseExit(program)
     expect(exit._tag).toBe("Success")
--- a/.changeset/violet-doors-share.md
+++ b/.changeset/violet-doors-share.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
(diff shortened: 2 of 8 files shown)


packages/effect/test/Effect/promise.test.ts
```diff
@@ -17,7 +17,7 @@ describe("Effect", () => {
     })
     const program = effect.pipe(
       Effect.timeout("10 millis"),
```
IMax153, line 19: For all tests / library code that was previously using `Effect.timeout(...).pipe(Effect.optionFromOptional)`, we should be explicit about using `Effect.asSome` for the happy path.

Similar to what was done in the `NodeClient.test.ts` module.

IMax153 (changes requested):

IMax153 (approved):
I personally like this change. Though it limits the use of `optionFromOptional` for handling `timeout` errors, I feel as though the explicitness is much better.

Will wait for @mikearnaldi and @tim-smart to also weigh in.

PR #2003, opened 2024-01-29
Fix CLI readme

wmaurer:
* add missing code to example 'final CLI application' so that it compiles
* removed `slice(2)` from `process.argv`
* removed no longer correct explanation about slicing `process.argv`


IMax153:
@wmaurer - thank you very much for helping us keep the docs up-to-date :) I'll merge as soon as CI succeeds.

IMax153 (approved):

PR #2009, opened 2024-01-30
add ReadonlyRecord key type

tim-smart:


packages/effect/src/ReadonlyRecord.ts
```diff
@@ -275,6 +275,44 @@ export const get: {
   <A>(self: ReadonlyRecord<A>, key: string): Option.Option<A> => has(self, key) ? Option.some(self[key]) : Option.none()
 )
 
+/**
+ * Apply a function to the element at the specified key, creating a new record,
+ * or return `None` if the key doesn't exist.
```
gcanti, line 280: "or return **a copy of the input** if the key doesn't exist"

packages/effect/src/ReadonlyRecord.ts
```diff
@@ -275,6 +275,44 @@ export const get: {
...
+ * @since 2.0.0
+ */
+export const modify: {
+  <A, B>(key: string, f: (a: A) => B): (self: ReadonlyRecord<A>) => Record<string, A | B>
+  <A, B>(self: ReadonlyRecord<A>, key: string, f: (a: A) => B): Record<string, A | B>
+} = dual(
+  3,
+  <A, B>(self: ReadonlyRecord<A>, key: string, f: (a: A) => B): Record<string, A | B> => {
+    if (!has(self, key)) {
+      return self
```
gcanti, line 310: should return `{ ...self }` as the return type is mutable

packages/effect/src/ReadonlyRecord.ts
```diff
@@ -360,6 +396,9 @@ export const remove: {
   (key: string): <A>(self: ReadonlyRecord<A>) => Record<string, A>
   <A>(self: ReadonlyRecord<A>, key: string): Record<string, A>
 } = dual(2, <A>(self: ReadonlyRecord<A>, key: string): Record<string, A> => {
+  if (!has(self, key)) {
+    return self
```
gcanti, line 400: we can't return the input, the return type is mutable

packages/effect/src/ReadonlyRecord.ts
```diff
@@ -795,23 +832,22 @@ export const upsert: {
...
   <B>(key: string, value: B): <A>(self: ReadonlyRecord<A>) => Record<string, A | B>
   <A, B>(self: ReadonlyRecord<A>, key: string, value: B): Record<string, A | B>
 } = dual(3, <A, B>(self: ReadonlyRecord<A>, key: string, value: B): Record<string, A | B> => {
-  const out: Record<string, A | B> = { ...self }
   if (has(self, key)) {
-    out[key] = value
+    return { ...self, [key]: value }
   }
-  return out
+  return self
```
gcanti, line 850: idem

packages/effect/src/ReadonlyRecord.ts
```diff
@@ -948,10 +984,10 @@ export const union: {
     combine: (selfValue: A, thatValue: A) => A
   ): Record<string, A> => {
     if (isEmptyRecord(self)) {
-      return { ...that }
+      return that
```
gcanti, line 987: idem

packages/effect/src/ReadonlyRecord.ts
```diff
@@ -948,10 +984,10 @@ export const union: {
...
     }
     if (isEmptyRecord(that)) {
-      return { ...self }
+      return self
```
gcanti, line 990: idem

packages/effect/src/ReadonlyRecord.ts
```diff
@@ -1020,10 +1056,10 @@ export const difference: {
   <A>(self: ReadonlyRecord<A>, that: ReadonlyRecord<A>): Record<string, A>
 } = dual(2, <A>(self: ReadonlyRecord<A>, that: ReadonlyRecord<A>): Record<string, A> => {
   if (isEmptyRecord(self)) {
-    return { ...that }
+    return that
```
gcanti, line 1059: idem

packages/effect/src/ReadonlyRecord.ts
```diff
@@ -1020,10 +1056,10 @@ export const difference: {
...
   }
   if (isEmptyRecord(that)) {
-    return { ...self }
+    return self
```
gcanti, line 1062: idem

gcanti (changes requested):
I like the renaming 👍

packages/effect/src/ReadonlyRecord.ts
tim-smart: The performance side of my brain is going crazy haha. But I guess it is either change the return type or return copies.

tim-smart: I quite like the idea of a `.toRecord` function that converts a ReadonlyRecord to a Record by returning a copy.

tim-smart: (you could then use ReadonlyRecord for the return types)

gcanti: I'm not sure, AFAIK returning a mutable type is the standard practice we follow (for example, in ReadonlyArray)

mikearnaldi: Imho if one looks for perf those modules shouldn't be used, those are rather convenience utils. Returning ReadonlyRecord without copy is unsafe because the input may be mutable, readonly doesn't guarantee immutability it only says that something only performs read ops but it says nothing about the source (eg `Record<string, X>` can be assigned to `ReadonlyRecord<string, X>`)

.changeset/four-turtles-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
+---
+
+add key type to ReadonlyRecord
```
gcanti, line 5: if we add the K parameter then I would also try to handle symbols as keys (which is something I've wanted to do for a long time)

tim-smart: I actually did add that then reverted it.

Will add it back later.

packages/effect/src/ReadonlyRecord.ts
```diff
@@ -164,7 +168,9 @@ export const fromIterableBy = <A>(items: Iterable<A>, f: (a: A) => string): Reco
  * @category conversions
  * @since 2.0.0
  */
-export const fromEntries: <A>(self: Iterable<readonly [string, A]>) => Record<string, A> = fromIterableWith(identity)
+export const fromEntries: <K extends string, A>(self: Iterable<readonly [K, A]>) => Record<K, A> = fromIterableWith(
```
gcanti, line 171: all the `from*` APIs 
- `fromIterableWith`
- `fromIterable`
- `fromIterableBy`
- `fromEntries`
are unsafe now (`Record<K, A>` where `K` is a union of literals, let's say "a" | "b", is equivalente to a struct `{ a: value1, b: value2 }`).

Example with `fromEntries`:

```ts
const myiterable: Iterable<["a" | "b", string]> = []
// const myrecord: Record<"a" | "b", string>
const myrecord = fromEntries(myiterable)
myrecord.a.trim() // throws
```

gcanti:
Given the increasing number of changes, could we divide this PR into smaller, more manageable parts?

Like a PR containing:

- add ReadonlyRecord.modify
- rename ReadonlyRecord.upsert to .set
- rename ReadonlyRecord.update to .replace

and a PR trying to add the K parameter (which instead is a tough nut to crack)

IMax153:
@tim-smart - I think we have a bunch of conflicts.

tim-smart:
> @tim-smart - I think we have a bunch of conflicts.

I think the recent updates to the minor release queue made it go a bit weird.

gcanti:
what about

```ts
export const empty = (): Record<never, never> => ({})
```

it would prevent the last assignement

```ts
// empty

export const empty1: Record<string, number> = ReadonlyRecord.empty()
export const empty2: Record<symbol, number> = ReadonlyRecord.empty()
// @ts-expect-error
export const empty3: Record<"a", number> = ReadonlyRecord.empty()
```

tim-smart:
> what about

I have made never the defaults, and allowed them to be set. Seems like the most practical approach.

gcanti:
RE: NonLiteralKey

1) I'm puzzled by the complexity of the definition, is there something wrong with the naive definition:

```ts
export type NonLiteralKey<K extends string | symbol> = K extends string ? string : symbol
```

all type level tests seem to be green

```ts
type A1 = ReadonlyRecord.NonLiteralKey<never> // never
type A2 = ReadonlyRecord.NonLiteralKey<string> // string
type A3 = ReadonlyRecord.NonLiteralKey<symbol> // symbol
type A4 = ReadonlyRecord.NonLiteralKey<`a${string}`> // string
type A5 = ReadonlyRecord.NonLiteralKey<"a"> // string
type A6 = ReadonlyRecord.NonLiteralKey<string | symbol> // string | symbol
```

2) I wonder if there is a way to detect template literals so that for example

`NonLiteralKey<`a${string}`>` is still `a${string}` instead of `string`

packages/effect/src/ReadonlyRecord.ts
```diff
@@ -17,16 +17,37 @@ import type { NoInfer } from "./Types.js"
...
+    : K1 & K2
 }
 
 /**
  * @category type lambdas
  * @since 2.0.0
  */
 export interface ReadonlyRecordTypeLambda extends TypeLambda {
-  readonly type: ReadonlyRecord<this["Target"]>
+  readonly type: ReadonlyRecord<string | symbol, this["Target"]>
```
gcanti, line 50: I'm afraid `string | symbol` will make the instances in `typeclass` less useful, for example

```ts
import * as ReadonlyRecordInstances from "@effect/typeclass/data/ReadonlyRecord"

const map = ReadonlyRecordInstances.Covariant.map
declare const r: Record<string, number>
// const result: ReadonlyRecord<string | symbol, number>
const result = map(r, (n) => n)
```

Maybe we could make `ReadonlyRecordTypeLambda` somehow parametric, I'll try to make some attempts

tim-smart:
I think the naive type should work. I can't remember the issues I was facing earlier.

For detecting template literals, I wonder if we use the `length` property and check if it is finite.

gcanti (approved):

PR #2023, opened 2024-01-31
feat: Make Equal an implicit trait, namely remove the type Data.Data

### LINKED ISSUE #1962: ReadonlyArray refinement tacid usage, odd behavior or limitation?
Given some `A[]`, using `ReadonlyArray.isNonEmptyArray` will correctly change it to `[A, ...A[]]`
But if the array is Data, the behavior gets some odd behaviors:

```ts
type A = { _tag: "A" }
declare const a: Data.Data<readonly A[]> // or readonly A[] & Equal.Equal

if (ReadonlyArray.isNonEmptyReadonlyArray(a)) {
	a // -> readonly A[] & Equal.Equal & readonly [A, ...A[]]
}

Effect.succeed(a).pipe(
	Effect.filterOrFail(ReadonlyArray.isNonEmptyReadonlyArray, () => new Error()),
	Effect.map(_ => _) // -> readonly A[] & Equal
)

Effect.succeed(a).pipe(
	Effect.filterOfFail(ReadonlyArray.isNonEmptyReadonlyArray<A>, () => new Error()),
	Effect.map(_ => _) // -> readonly [A, ...A[]], works but removes the Equal trait
)
```

Is there anything we can do to solve this?

mikearnaldi:
With this change we remove the `Data.Data` type and we make `Equal.Equal` & `Hash.Hash` an implicit traits.

The main reason is that `Data.Data<A>` was structurally equivalent to `A & Equal.Equal` but extending `Equal.Equal` doesn't mean that the equality is implemented by-value, so the type was simply adding noice without gaining any level of safety.

The module `Data` remains unchanged at the value level, all the functions priorly available are supposed to work in exactly the same manner.

At the type level instead the functions return `Readonly` variants, so for example we have:

```ts
import { Data } from "effect";

const obj = Data.struct({
  a: 0,
  b: 1,
});
```

will have the `obj` typed as:

```ts
declare const obj: {
  readonly a: number;
  readonly b: number;
};
```

closes #1962

### DIFF
--- a/.changeset/mighty-donuts-behave.md
+++ b/.changeset/mighty-donuts-behave.md
@@ -1,0 +1,33 @@
+---
+"@effect/platform": minor
+"effect": minor
+"@effect/schema": minor
+"@effect/cli": minor
+"@effect/rpc": minor
+---
+
+With this change we remove the `Data.Data` type and we make `Equal.Equal` & `Hash.Hash` implicit traits.
+
+The main reason is that `Data.Data<A>` was structurally equivalent to `A & Equal.Equal` but extending `Equal.Equal` doesn't mean that the equality is implemented by-value, so the type was simply adding noise without gaining any level of safety.
+
+The module `Data` remains unchanged at the value level, all the functions previously available are supposed to work in exactly the same manner.
+
+At the type level instead the functions return `Readonly` variants, so for example we have:
+
+```ts
+import { Data } from "effect";
+
+const obj = Data.struct({
+  a: 0,
+  b: 1,
+});
+```
+
+will have the `obj` typed as:
+
+```ts
+declare const obj: {
+  readonly a: number;
+  readonly b: number;
+};
+```
--- a/packages/cli/src/Options.ts
+++ b/packages/cli/src/Options.ts
@@ -177,13 +177,13 @@ export const choice: <A extends string, C extends NonEmptyReadonlyArray<A>>(
  *
  * export type Animal = Dog | Cat
  *
- * export interface Dog extends Data.Case {
+ * export interface Dog {
  *   readonly _tag: "Dog"
  * }
  *
  * export const Dog = Data.tagged<Dog>("Dog")
  *
- * export interface Cat extends Data.Case {
+ * export interface Cat {
  *   readonly _tag: "Cat"
  * }
  *
--- a/packages/effect/dtslint/Data.ts
+++ b/packages/effect/dtslint/Data.ts
@@ -10,13 +10,13 @@ declare const mutableArray: Array<string>
 // struct
 // -------------------------------------------------------------------------------------
 
-// $ExpectType Data<{ readonly a: string; }>
+// $ExpectType { readonly a: string; }
 const struct1 = Data.struct(mutableStruct)
 
 // @ts-expect-error
 struct1.a = "a"
 
-// $ExpectType Data<{ readonly a: string; }>
+// $ExpectType { readonly a: string; }
 const struct2 = Data.struct(readonlyStruct)
 
 // @ts-expect-error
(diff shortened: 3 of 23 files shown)


.changeset/mighty-donuts-behave.md
```diff
@@ -0,0 +1,33 @@
...
+"@effect/schema": minor
+"@effect/cli": minor
+"@effect/rpc": minor
+---
+
+With this change we remove the `Data.Data` type and we make `Equal.Equal` & `Hash.Hash` an implicit traits.
+
+The main reason is that `Data.Data<A>` was structurally equivalent to `A & Equal.Equal` but extending `Equal.Equal` doesn't mean that the equality is implemented by-value, so the type was simply adding noice without gaining any level of safety.
+
+The module `Data` remains unchanged at the value level, all the functions priorly available are supposed to work in exactly the same manner.
```
fubhy, line 13:
```suggestion
The module `Data` remains unchanged at the value level, all the functions previously available are supposed to work in exactly the same manner.
```

fubhy: Uuuh, Mr. Fancy Pants ... "priorly"  :-P

fubhy (changes requested):

datner:
There is a pitfall in this though, users might unwittingly squander the `Data` behavior and be non the wiser
```ts
pipe(
  ReadonlyArray.makeBy(10, () => new A()),
  Data.array,
  Effect.forEach(_ => Effect.as(notifyMikeAboutA(_), _))
  Effect.filterOrDie(
    Equal.equals(Data.array(ReadonlyArray.makeBy(10, () => new A()))),
    () => "w00t"
  )
) // -> it dies with w00t
```

.changeset/mighty-donuts-behave.md
datner: let the man priorize as he pleases

.changeset/mighty-donuts-behave.md
```diff
@@ -0,0 +1,33 @@
...
+"@effect/platform": minor
+"effect": minor
+"@effect/schema": minor
+"@effect/cli": minor
+"@effect/rpc": minor
+---
+
+With this change we remove the `Data.Data` type and we make `Equal.Equal` & `Hash.Hash` an implicit traits.
+
+The main reason is that `Data.Data<A>` was structurally equivalent to `A & Equal.Equal` but extending `Equal.Equal` doesn't mean that the equality is implemented by-value, so the type was simply adding noice without gaining any level of safety.
```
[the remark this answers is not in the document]
datner: noice 👌🏻

mikearnaldi:
> There is a pitfall in this though, users might unwittingly squander the `Data` behavior and be non the wiser
> 
> ```ts
> pipe(
>   ReadonlyArray.makeBy(10, () => new A()),
>   Data.array,
>   Effect.forEach(_ => Effect.as(notifyMikeAboutA(_), _))
>   Effect.filterOrDie(
>     Equal.equals(Data.array(ReadonlyArray.makeBy(10, () => new A()))),
>     () => "w00t"
>   )
> ) // -> it dies with w00t
> ```

That's correct as forEach returns a new copy of the array which is mutable

.changeset/mighty-donuts-behave.md
mikearnaldi: priorly is proper english but whatever

.changeset/mighty-donuts-behave.md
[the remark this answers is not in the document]
mikearnaldi: LOL

.changeset/mighty-donuts-behave.md
fubhy: It is. But not for numbnuts like me. I want my sentences simple and easy to grok.

datner:
> That's correct as forEach returns a new copy of the array which is mutable, I guess it is a trade-off, we could also make Data.array idempotent, so that if you want to be really sure it compares correctly you can just always call it

I can't come up with a solution for this pitfall, I guess we will just have to make sure to note it in the docs.

IMax153 (approved):

datner:
@mikearnaldi hmmm the bot (and @/gcanti 😃 ) is quite noisy here, is this a one-time or will all updates be this noisy until they publish?

PR #2025, opened 2024-02-01
remove re-exports for platform packages

tim-smart:
- add node-shared package
- remove re-exports for platform packages


mikearnaldi:
Currently naming is like `KeyValueStoreNode, NodeContext, RuntimeNode`, maybe it is more consistent to prefix everything with `Node` instead?

tim-smart:
> Currently naming is like `KeyValueStoreNode, NodeContext, RuntimeNode`, maybe it is more consistent to prefix everything with `Node` instead?

Moved to prefixes :)

PR #2028, opened 2024-02-02
feat(breaking): Add ClassTag, improve Effect.serviceConstants, require string identifier for all tags

mikearnaldi:
With this change we now require a string key to be provided for all tags and renames the dear old `Tag` to `GenericTag`, so when previously you could do:

```ts
import { Effect, Context } from "effect";
interface Service {
  readonly _: unique symbol;
}
const Service = Context.Tag<
  Service,
  {
    number: Effect.Effect<never, never, number>;
  }
>();
```

you are now mandated to do:

```ts
import { Effect, Context } from "effect";
interface Service {
  readonly _: unique symbol;
}
const Service = Context.GenericTag<
  Service,
  {
    number: Effect.Effect<never, never, number>;
  }
>("Service");
```

This makes by default all tags globals and ensures better debuggaility when unexpected errors arise.

Furthermore we introduce a new way of constructing tags that should be considered the new default:

```ts
import { Effect, Context } from "effect";
class Service extends Context.Tag("Service")<
  Service,
  {
    number: Effect.Effect<never, never, number>;
  }
>() {}

const program = Effect.flatMap(Service, ({ number }) => number).pipe(
  Effect.flatMap((_) => Effect.log(`number: ${_}`))
);
```

this will use "Service" as the key and will create automatically an opaque identifier (the class) to be used at the type level, it does something similar to the above in a single shot.


patroza:
Love the class. But then go all the way; allow to take the name of the class?

mikearnaldi:
As a follow up I propose renaming `Tag` to `GenericTag` and moving `ClassTag` to be `Tag`

patroza:
> As a follow up I propose renaming `Tag` to `GenericTag` and moving `ClassTag` to be `Tag`

Some more ideas to infer the shape from the make effect; https://github.com/effect-ts-app/libs/blob/34c0458a3e9e5a42b723260802785ce369ee07d5/packages/prelude/_src/service.ts#L72

mikearnaldi:
> Love the class. But then go all the way; allow to take the name of the class?

Not sure I understand your suggestion, please provide a code snippet

patroza:
> Not sure I understand your suggestion, please provide a code snippet

```ts
  return class extends (TagClass as any) {
    static get key(): string {
      return (this as any).name
    }
  } as any
  ```f

IMax153:
@mikearnaldi - big fan of this - simplifies a lot of boilerplate around `Tag`. I also agree with `Tag` -> `GenericTag` and `ClassTag` -> `Tag`.

 Nice work 👍

tim-smart:
Looks good! Agree with the renaming too.

Quick fix to ensure the definition stack is set correctly: #2029

PR #2031, opened 2024-02-03
`RateLimiter` port from Rezilience

hsubra89:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This patch ports `RateLimiter` from [Rezilience](https://github.com/svroonland/rezilience/blob/master/rezilience/shared/src/main/scala/nl/vroste/rezilience/RateLimiter.scala) including all related tests. No changes or alterations have been made so far. 

**NOTE**: I'm still figuring my way around the code base so please bear with me. 

I understand that the intent is to update the `make` interface to support Persistence in the future. I'm looking for some directional input on what that might look like and i'll be happy to make those changes :) .( I noticed the experimental persistence module in the codebase )

I also deliberately haven't added the `changeset` yet. I assume it'll be a minor release if this patch makes it in; but for now, I just want to make a patch that works :) 

## Related
- Related Issue #1685


packages/effect/src/internal/rateLimiter.ts
```diff
@@ -0,0 +1,74 @@
+// Direct port of https://github.com/svroonland/rezilience/blob/master/rezilience/shared/src/main/scala/nl/vroste/rezilience/RateLimiter.scala
+import type { Duration } from "effect"
+import { Chunk, Deferred, Effect, pipe, Queue, Ref, Stream } from "effect"
+
+// Find the next power of 2 greater than or equal to n.
+/** @internal */
+const nextPow2 = (n: number): number => {
```
hsubra89, line 7: This method is duplicated and also exists in `pubsub.ts`. What would be best location for utility functions like this one?

packages/effect/src/RateLimiter.ts
```diff
@@ -0,0 +1,11 @@
+import type { Effect } from "./Effect.js"
```
hsubra89, line 1: I wasn't quite sure which package to put this under, so its in the `effect` package namespace for now.

fubhy: Yeah that's correct

hsubra89:
btw, I should state that `RateLimiter` on its own might not be sufficient. Composition with `BulkHead` would be required for a few of the cases that come to mind.

packages/effect/src/internal/rateLimiter.ts
mikearnaldi: a file in `internal` probably is best

packages/effect/src/internal/rateLimiter.ts
```diff
@@ -0,0 +1,74 @@
+// Direct port of https://github.com/svroonland/rezilience/blob/master/rezilience/shared/src/main/scala/nl/vroste/rezilience/RateLimiter.scala
+import type { Duration } from "effect"
+import { Chunk, Deferred, Effect, pipe, Queue, Ref, Stream } from "effect"
```
mikearnaldi, line 3: in library imports should come from individual files as this would prevent tree shaking in bundlers like esbuild

packages/effect/src/internal/rateLimiter.ts
```diff
@@ -0,0 +1,74 @@
...
+// Find the next power of 2 greater than or equal to n.
+/** @internal */
+const nextPow2 = (n: number): number => {
+  const nextPow = Math.ceil(Math.log(n) / Math.log(2))
+  return Math.max(Math.pow(2, nextPow), 2)
+}
+
+/** @internal */
+export const make = (max: number, interval: Duration.DurationInput) => {
+  return Effect.gen(function*(_outer) {
```
mikearnaldi, line 14: no need to use `outer / inner` JS has shadowing, you can simply use a `$`

mikearnaldi:
Re persistence, I think we can tackle this at a later point, the idea would be to have a separate `makePersistent` which implements the same interface (just a function taking effect in this case), not yet sure how we would do that but at that point we can decide how much to extract and share between `make` vs `makePersistent`

packages/effect/src/internal/rateLimiter.ts
hsubra89: Done.

packages/effect/src/internal/rateLimiter.ts
hsubra89: Done.

packages/effect/src/internal/rateLimiter.ts
hsubra89: Done.

packages/effect/src/RateLimiter.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Note that only the moment of starting the effect is rate limited: the number of concurrent executions is not bounded.
+ *
+ * Calls are queued up in an unbounded queue until capacity becomes available.
+ */
+export interface RateLimiter {
+  <R, E, A>(task: Effect<R, E, A>): Effect<R, E, A>
+}
+
+export const make = (max: number, interval: Duration.DurationInput): Effect<Scope, never, RateLimiter> =>
```
IMax153, line 17:
```suggestion
export const make = (limit: number, window: Duration.DurationInput): Effect<Scope, never, RateLimiter> =>
```

Small suggestion to make the constructor arguments a bit more semantic.

packages/effect/src/internal/rateLimiter.ts
```diff
@@ -0,0 +1,78 @@
...
+import type { DurationInput } from "../Duration.js"
+import * as Effect from "../Effect.js"
+import { pipe } from "../Function.js"
+import * as Queue from "../Queue.js"
+import * as Ref from "../Ref.js"
+import * as Stream from "../Stream.js"
+import { nextPow2 } from "./nextPow2.js"
+
+/** @internal */
+export const make = (max: number, interval: DurationInput) => {
```
IMax153, line 17:
```suggestion
export const make = (limit: number, window: DurationInput) => {
```

IMax153 (changes requested):

packages/effect/src/RateLimiter.ts
hsubra89: Done :)

packages/effect/src/RateLimiter.ts
```diff
@@ -0,0 +1,18 @@
+import type { Effect } from "./Effect.js"
+import type { Duration } from "./index.js"
```
IMax153, line 2:
```suggestion
import type { Duration } from "./Duration.js"
```

packages/effect/src/RateLimiter.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Note that only the moment of starting the effect is rate limited: the number of concurrent executions is not bounded.
+ *
+ * Calls are queued up in an unbounded queue until capacity becomes available.
+ */
+export interface RateLimiter {
+  <R, E, A>(task: Effect<R, E, A>): Effect<R, E, A>
+}
+
+export const make = (limit: number, window: Duration.DurationInput): Effect<Scope, never, RateLimiter> =>
+  internal.make(max, interval)
```
IMax153, line 18: @hsubra89 Thanks for making the change, but I don't think the changes in the argument names were propagated to the actual implementation.

IMax153 (changes requested):

packages/effect/src/RateLimiter.ts
hsubra89: LOL, I just blindly committed using the Github UI and didn't re-check since I was mobile. Apologies. Will fix :(

hsubra89: Fixed properly this time.

packages/effect/src/RateLimiter.ts
hsubra89: Done. 

I assume this has already been considered, but I wonder if there's a lint rule which we can use to define these import rules.

packages/effect/src/internal/rateLimiter.ts
```diff
@@ -0,0 +1,78 @@
...
+        }),
+        Stream.mapEffect(([_interrupted, eff]) => eff, { concurrency: "unbounded", unordered: true }),
+        Stream.runDrain,
+        Effect.forkScoped
+      )
+    )
+
+    const apply = <R, E, A>(task: Effect.Effect<R, E, A>) =>
+      Effect.gen(function*($) {
+        const start = yield* $(Deferred.make<never, void>())
```
IMax153, line 44: I think we're seeing all the type errors in CI because we've re-ordered the type parameters of many of the data types within Effect (cc @mikearnaldi @tim-smart). It's a massive breaking change that we've actually developed the Effect [`@effect/codemod`](https://github.com/Effect-TS/codemod) library to fix these issues automatically.

You can either use the codemod lib to fix these changes, or you can fix the types manually, and since we re-ordered the types you should now be able to do:

```ts
Effect<void>
```

```ts
Deferred.make<void>()
```

IMax153 (changes requested):

packages/effect/src/internal/rateLimiter.ts
hsubra89: Done.

IMax153 (changes requested):
Thanks @hsubra89 - just a few last comments for a bit of polish and you'll have my 👍.

packages/effect/src/RateLimiter.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Limits the number of calls to a resource to a maximum amount in some interval using the token bucket algorithm.
+ *
+ * Note that only the moment of starting the effect is rate limited: the number of concurrent executions is not bounded.
+ *
+ * Calls are queued up in an unbounded queue until capacity becomes available.
+ */
+export interface RateLimiter {
+  <R, E, A>(task: Effect<A, E, R>): Effect<A, E, R>
```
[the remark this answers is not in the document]
hsubra89: I did the migration by hand and clearly missed a few spots. Thank you. 

Done :)

mikearnaldi:
I guess this should be rebase on next-minor

packages/effect/src/internal/rateLimiter.ts
```diff
@@ -0,0 +1,78 @@
...
+          cost: Chunk.size,
+          units: limit
+        }),
+        Stream.mapEffect(([_interrupted, eff]) => eff, { concurrency: "unbounded", unordered: true }),
+        Stream.runDrain,
+        Effect.forkScoped
+      )
+    )
+
+    const apply = <R, E, A>(task: Effect.Effect<R, E, A>) =>
```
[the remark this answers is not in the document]
hsubra89: Done

hsubra89:
> I guess this should be rebase on next-minor

Done :)

PR #2032, opened 2024-02-03
fix: Schema.omit on a union

sukovanej:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I included a test example how we discovered the problem. 

Result without the change:

```
  Object {
    "_id": "Either",
    "_tag": "Right",
    "right": Object {
-     "b": Object {
-       "_id": "Option",
-       "_tag": "Some",
-       "value": "a",
-     },
+     "b": "a",
    },
  }
```

### DIFF
--- a/.changeset/witty-wombats-dream.md
+++ b/.changeset/witty-wombats-dream.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+Fix `Schema.omit` on a union schema.
--- a/packages/schema/src/Schema.ts
+++ b/packages/schema/src/Schema.ts
@@ -1167,26 +1167,32 @@ export const omit = <A, Keys extends ReadonlyArray<keyof A>>(...keys: Keys) =>
 <R, I extends { [K in keyof A]?: any }>(
   self: Schema<R, I, A>
 ): Schema<R, Simplify<Omit<I, Keys[number]>>, Simplify<Omit<A, Keys[number]>>> => {
-  const ast = self.ast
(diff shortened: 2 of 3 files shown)


sukovanej:
@gcanti could you please take a look? 

I'm aware this is not precise from the type stand-point as the type-level `Omit` is not distributive. Still, this feels like the correct and expected behaviour in the runtime so I'd definitely vote for not disabling this case.

gcanti:
Sorry for responding only now (I've been super busy with the latest release). Looks like it's working as intended though, since `schema2: S.Schema<{}, {}, never>`.

sukovanej:
> Sorry for responding only now (I've been super busy with the latest release)
I was checking the minor branch periodically and saw you producing commits like crazy. So no worry, I completely understood what's happening 😄 

> Looks like it's working as intended though, since schema2: S.Schema<{}, {}, never>.
I meant it in a way that `Omit` not distributing on a union is causing confusion a lot (based on stackoverflow and github  discussions)

gcanti:
Yes, I understand that it can be a source of confusion, but the principle is to keep the runtime and type-level as aligned as possible. So, either we maintain the current signature of `omit` and its corresponding implementation (which I prefer), or if we change the implementation, the signature also needs to be changed.

mikearnaldi:
I think we should keep alignment and eventually do a distributiveOmit version that can distribute

sukovanej:
Hey, that makes sense to me. Thank you. Closing.

PR #2033, opened 2024-02-04
Enhancement: Add toJSON for Secret

rehos:
## Type

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Since Secrets are "special" strings it would be nice if JSON.stringify transforms the Secret back into a "redacted" string.

However the current behavior is:
```
const s = Secret.fromString('SECRET')
console.log(JSON.stringify(s))
// Output: {}
```

This PR changes the behavior to:
```
const s = Secret.fromString('SECRET')
console.log(JSON.stringify(s))
// Output: "<redacted>"
```

## Related

n/a

### DIFF
--- a/.changeset/twenty-clouds-behave.md
+++ b/.changeset/twenty-clouds-behave.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add toJSON for Secret
--- a/packages/effect/src/internal/secret.ts
+++ b/packages/effect/src/internal/secret.ts
@@ -40,6 +40,12 @@ export const make = (bytes: Array<number>): Secret.Secret => {
       return "Secret(<redacted>)"
     }
   })
+  Object.defineProperty(secret, "toJSON", {
+    enumerable: false,
+    value() {
+      return "<redacted>"
+    }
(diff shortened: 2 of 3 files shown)


IMax153:
@mikearnaldi @rehos - maybe we should consider unifying the `console.log` output as well to `"<redacted>"`? I don't know that having `Secret(<redacted>)` is really that useful.

rehos:
> @mikearnaldi @rehos - maybe we should consider unifying the `console.log` output as well to `"<redacted>"`? I don't know that having `Secret(<redacted>)` is really that useful.

I thought about that but it could potentially break tests if people are using `toString()` in their unit tests. But I'm happy to change the `toString` behavior too if everybody agrees.

tim-smart:
> @mikearnaldi @rehos - maybe we should consider unifying the `console.log` output as well to `"<redacted>"`? I don't know that having `Secret(<redacted>)` is really that useful.

I think we should keep the `Secret`wrapper, as it denotes that it is an instance of Secret.

PR #2034, opened 2024-02-04
change Effect type parameters order

gcanti:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### DIFF
--- a/.changeset/fast-socks-bake.md
+++ b/.changeset/fast-socks-bake.md
@@ -1,0 +1,21 @@
+---
+"@effect/platform-node-shared": minor
+"@effect/platform-browser": minor
+"@effect/opentelemetry": minor
+"@effect/platform-node": minor
+"@effect/rpc-http-node": minor
+"@effect/experimental": minor
+"@effect/platform-bun": minor
+"@effect/printer-ansi": minor
+"@effect/rpc-workers": minor
+"@effect/rpc-nextjs": minor
+"@effect/platform": minor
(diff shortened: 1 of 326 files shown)


bwbuchanan:
Should this have been a major release item? It breaks essentially all code that explicitly references the `Effect` type. In one of my codebases, I have more than 1500 type signatures that now need to be updated.

Also, the documentation at https://effect.website/docs/essentials/effect-type still references the old type parameter order.

mikearnaldi:
> Should this have been a major release item? It breaks essentially all code that explicitly references the `Effect` type. In one of my codebases, I have more than 1500 type signatures that now need to be updated.
> 
> Also, the documentation at https://effect.website/docs/essentials/effect-type still references the old type parameter order.

Up until version 3.0 minors are considering breaking, that's the release policy we set for to exit from `next` targets.

There is a codemod available to help you migrate the code: `npx @effect/codemod minor-2.3 src/**/*.ts`.

You may want to join discord to be up to date and follow the release notes

bwbuchanan:
@mikearnaldi Thanks for the clarification and the pointer to the codemod, which handled about 99.5% of the transformation.

mikearnaldi:
> @mikearnaldi Thanks for the clarification and the pointer to the codemod, which handled about 99.5% of the transformation.

Nice! thanks for letting us know it worked :)

PR #2047, opened 2024-02-05
make array types in cli more permissive

tim-smart:


IMax153 (approved):
LGTM! Thanks @tim-smart - meant to do this a while ago 😅

PR #2057, opened 2024-02-06
Fix for possible stack overflow issues due to Array.push in combination with spread operator

### LINKED ISSUE #2048: ReadOnly Array flatMap stack overflow

2.2.3

```
import { pipe } from 'effect/Function';
import * as RA from 'effect/ReadonlyArray';

const arr: number[][] = [[]];
for (let i = 0; i < 135000; i++) {
  arr[0].push(i);
}

const flattened = pipe(
  arr,
  RA.flatMap((x) => x),
);

console.log(flattened.length);
```

135000

### What do you see instead?

RangeError: Maximum call stack size exceeded
    at ....../node_modules/.pnpm/effect@2.2.3/node_modules/effect/dist/cjs/ReadonlyArray.js:1046:9

### Additional information

https://stackoverflow.com/questions/61740599/rangeerror-maximum-call-stack-size-exceeded-with-array-push

joepjoosten:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Closes #2048

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -40,7 +40,7 @@ module.exports = {
     "no-unused-vars": "off",
     "no-restricted-syntax": ["error", {
         "selector": "CallExpression[callee.property.name='push'] > SpreadElement.arguments",
-        "message": "Do use spread arguments in Array.push"
+        "message": "Do not use spread arguments in Array.push"
     }],
     "prefer-rest-params": "off",
     "prefer-spread": "off",
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -38,6 +38,10 @@ module.exports = {
     "prefer-destructuring": "off",
     "sort-imports": "off",
     "no-unused-vars": "off",
+    "no-restricted-syntax": ["error", {
+        "selector": "CallExpression[callee.property.name='push'] > SpreadElement.arguments",
+        "message": "Do use spread arguments in Array.push"
+    }],
     "prefer-rest-params": "off",
     "prefer-spread": "off",
     "import/first": "error",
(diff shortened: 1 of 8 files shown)


mikearnaldi:
Looks good, would you mind adding a changeset with patch level and a description of the change pls, you can generate one using `pnpm changeset` and following the wizard

.eslintrc.cjs
```diff
@@ -38,6 +38,10 @@ module.exports = {
     "prefer-destructuring": "off",
     "sort-imports": "off",
     "no-unused-vars": "off",
+    "no-restricted-syntax": ["error", {
+        "selector": "CallExpression[callee.property.name='push'] > SpreadElement.arguments",
+        "message": "Do use spread arguments in Array.push"
```
tatchi, line 43:
```suggestion
        "message": "Do not use spread arguments in Array.push"
```

mikearnaldi:
The changeset is set at minor, should be set at patch

mikearnaldi (approved):

mikearnaldi:
Seems like linter suggests two other places where spread is used

tim-smart:
Does `array.push.apply(array, otherArray)` have the same issue?

tim-smart:
> Does `array.push.apply(array, otherArray)` have the same issue?

The answer is yes.

joepjoosten:
> Seems like linter suggests two other places where spread is used

When running lint locally, it doesn't complain, and i can't find the push it is complaining about...? Strange?

mikearnaldi:
Very strange...

PR #2067, opened 2024-02-07
add releaseAll api to Semaphore

tim-smart:


packages/effect/src/internal/effect/circular.ts
```diff
@@ -73,11 +75,17 @@ class Semaphore {
...
+        while (this.waiters.length > 0) {
+          if (this.waiters[0]() === false) {
+            break
+          }
+        }
       }, fiber.getFiberRef(core.currentSchedulingPriority))
       return core.unit
     })
 
+  readonly releaseAll: Effect.Effect<never, never, void> = core.suspend(() => this.release(this.taken))
```
tim-smart, line 87: Will need to move this to make it atomic

packages/effect/src/internal/effect/circular.ts
```diff
@@ -35,7 +35,7 @@ import * as supervisor from "../supervisor.js"
 
 /** @internal */
 class Semaphore {
-  public waiters = new Array<() => void>()
+  public waiters = new Array<() => boolean>()
```
tim-smart, line 38: Benchmarks show using a Set scales much better than an Array for this.

tim-smart: Insertion order is preserved too, so it is safe to use.

PR #2080, opened 2024-02-08
Add functional analogue of `satisfies` operator

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Add functional analogue of `satisfies` operator.
I believe this is a convenient operator to use in the pipe chain to localize type errors closer to their source.

### DIFF
--- a/.changeset/mean-dancers-judge.md
+++ b/.changeset/mean-dancers-judge.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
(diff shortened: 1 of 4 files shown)


mikearnaldi:
can you please add the example also in the changeset?

KhraksMamtsov:
AHAHA)) It seems that the universe is against this idea

PR #2084, opened 2024-02-09
simplify RateLimiter implementation using semaphore

tim-smart:


packages/effect/src/internal/rateLimiter.ts
```diff
@@ -1,79 +1,30 @@
...
-
-import * as Chunk from "../Chunk.js"
-import * as Deferred from "../Deferred.js"
 import type { DurationInput } from "../Duration.js"
 import * as Effect from "../Effect.js"
-import { pipe } from "../Function.js"
-import * as Queue from "../Queue.js"
-import * as Ref from "../Ref.js"
-import * as Stream from "../Stream.js"
-import { nextPow2 } from "./nextPow2.js"
```
fubhy, line 14: Could also delete this file again and move that back into `pubsub.ts`.

tim-smart: Moved to the Number module.

PR #2090, opened 2024-02-10
`RateLimiter.withCost` combinator + Rate Limiter composition tests

hsubra89:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This patch updates the `RateLimiter` module to include another constructor which includes a sub-constructor which allows us to specify a cost while maintaining composability with other `RateLimiter` instances. This enables the use of this module to rate limit for the following cases : 
- Accessing resources with an "API Credit" limit over a time window.

- With composition, accounting for multiple "credit" limits. ( eg: [Github API](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits) ) 

- Accounting for different "costs" per api per provider within a single credit limit. 
  - Many providers have different API costs for different endpoints within their ecosystem.

## Usage Example
```ts
import { Effect, RateLimiter } from "effect";
import { compose } from "effect/Function";

const program = Effect.scoped(
  Effect.gen(function* ($) {
    // Create a rate limiter that has an hourly limit of 1000 credits
    const rateLimiter = yield* $(RateLimiter.make(1000, "1 hours"));
    // Query API costs 1 credit per call ( 1 is the default cost )
    const queryAPIRL = compose(rateLimiter, RateLimiter.withCost(1));
    // Mutation API costs 5 credits per call
    const mutationAPIRL = compose(rateLimiter, RateLimiter.withCost(5));
    // ...
    // Use the pre-defined rate limiters
    yield* $(queryAPIRL(Effect.log("Sample Query")));
    yield* $(mutationAPIRL(Effect.log("Sample Mutation")));

    // Or set a cost on-the-fly
    const history = yield* $(
      rateLimiter(Effect.log("Another query with a different cost")).pipe(
        RateLimiter.withCost(3)
      )
    );
  })
);
```

### DIFF
--- a/packages/effect/src/internal/rateLimiter.ts
+++ b/packages/effect/src/internal/rateLimiter.ts
@@ -7,36 +7,8 @@ import type { Scope } from "../Scope.js"
 import * as SynchronizedRef from "../SynchronizedRef.js"
 
 /** @internal */
-export const makeWithCost = (limit: number, window: DurationInput): Effect.Effect<
-  (cost: number) => RateLimiter,
-  never,
-  Scope
-> =>
-  Effect.gen(function*(_) {
-    const scope = yield* _(Effect.scope)
-    const semaphore = yield* _(Effect.makeSemaphore(limit))
-    const ref = yield* _(SynchronizedRef.make(false))
-    const reset = SynchronizedRef.updateEffect(
-      ref,
-      (running) =>
-        running ? Effect.succeed(true) : Effect.sleep(window).pipe(
-          Effect.zipRight(SynchronizedRef.set(ref, false)),
-          Effect.zipRight(semaphore.releaseAll),
-          Effect.forkIn(scope),
-          Effect.interruptible,
-          Effect.as(true)
-        )
-    )
-
-    return (cost: number) => (effect) => {
-      const take = Effect.zipRight(semaphore.take(cost), reset)
-      return Effect.zipRight(take, effect)
-    }
-  })
-
-/** @internal */
-const defaultCost = globalValue(
-  Symbol.for("effect/RateLimiter/defaultCost"),
+const currentCost = globalValue(
+  Symbol.for("effect/RateLimiter/currentCost"),
   () => FiberRef.unsafeMake(1)
 )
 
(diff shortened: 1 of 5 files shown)


IMax153:
@hsubra89 - apologies for the back-and-forth with the `RateLimiter`. We discovered the updated implementation was utilizing the fixed-window algorithm for token replenishment instead of the token-bucket algorithm. 

The fix is in #2097. Please update this MR once the bugfix is merged.

hsubra89:
> @hsubra89 - apologies for the back-and-forth with the `RateLimiter`. We discovered the updated implementation was utilizing the fixed-window algorithm for token replenishment instead of the token-bucket algorithm. 
> 
> 
> 
> The fix is in #2097. Please update this MR once the bugfix is merged.

No problem. I did notice that when the implementation was changed but didn't look into it too deeply.

hsubra89:
Depends on #2097

packages/effect/src/internal/rateLimiter.ts
```diff
@@ -1,9 +1,45 @@
 import type { DurationInput } from "../Duration.js"
 import * as Effect from "../Effect.js"
+import * as FiberRef from "../FiberRef.js"
+import { globalValue } from "../GlobalValue.js"
 import type { RateLimiter } from "../RateLimiter.js"
 import type { Scope } from "../Scope.js"
 import * as SynchronizedRef from "../SynchronizedRef.js"
 
+/** @internal */
+export const makeWithCost = (limit: number, window: DurationInput): Effect.Effect<
```
tim-smart, line 10: This can be removed now?

packages/effect/src/internal/rateLimiter.ts
```diff
@@ -25,6 +61,12 @@ export const make = (limit: number, window: DurationInput): Effect.Effect<
...
-    const take = Effect.zipRight(semaphore.take(1), reset)
+
+    const cost = FiberRef.get(defaultCost).pipe(Effect.flatMap(semaphore.take))
+    const take = Effect.zipRight(cost, reset)
+
     return (effect) => Effect.zipRight(take, effect)
   })
+
+/** @internal */
+export const withCost = (cost: number) => Effect.locallyWith(defaultCost, () => cost)
```
tim-smart, line 72:
```suggestion
export const withCost = (cost: number) => Effect.locally(defaultCost, cost)
```

tim-smart (changes requested):
I think we should merge this first, as the other PR will need to add more tests to make sure `cost` is well supported.

IMax153 (changes requested):

packages/effect/src/internal/rateLimiter.ts
hsubra89: Sorry, that was dumb. Removed.

packages/effect/src/internal/rateLimiter.ts
```diff
@@ -1,9 +1,45 @@
...
+    )
+
+    return (cost: number) => (effect) => {
+      const take = Effect.zipRight(semaphore.take(cost), reset)
+      return Effect.zipRight(take, effect)
+    }
+  })
+
+/** @internal */
+const defaultCost = globalValue(
```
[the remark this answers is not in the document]
hsubra89: Done.

tim-smart (approved):

PR #2094, opened 2024-02-10
revert some type param adjustments in FiberSet

tim-smart:


mikearnaldi:
`R` should have a `never` default like normal runtime, this looks breaking though so maybe `minor` is better

PR #2097, opened 2024-02-10
Ensure `RateLimiter` utilizes the token-bucket algorithm

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/RateLimiter.ts
+++ b/packages/effect/src/RateLimiter.ts
@@ -22,14 +22,47 @@ export interface RateLimiter {
 }
 
 /**
- * Constructs a new `RateLimiter` which will utilize the token-bucket algorithm
- * to limit requests. The specified number of `tokens` is immediately available
- * after constructing the `RateLimiter`.
- *
- * **NOTE**: The maximum number of requests (specified by `tokens`) will be
- * spread out over the provided `interval`. For example, creating a `RateLimiter`
- * with `10` `tokens` and an `interval` of `1 seconds` will mean that `1` request
- * can be made every `100 millis` if no tokens are available.
+ * @since 2.0.0
+ */
+export declare namespace RateLimiter {
+  /**
+   * @since 2.0.0
+   * @category models
+   */
+  export interface Options {
+    /**
+     * The maximum number of requests that should be allowed.
+     */
+    readonly limit: number
+    /**
+     * The interval to utilize for rate-limiting requests. The semantics of the
+     * specified `interval` vary depending on the chosen `algorithm`:
+     *
+     * `token-bucket`: The maximum number of requests will be spread out over
+     * the provided interval if no tokens are available.
+     *
+     * For example, for a `RateLimiter` using the `token-bucket` algorithm with
+     * a `limit` of `10` and an `interval` of `1 seconds`, `1` request can be
+     * made every `100 millis`.
+     *
+     * `fixed-window`: The maximum number of requests will be reset during each
+     * interval. For example, for a `RateLimiter` using the `fixed-window`
+     * algorithm with a `limit` of `10` and an `interval` of `1 seconds`, a
+     * maximum of `10` requests can be made each second.
+     */
+    readonly interval: DurationInput
+    /**
+     * The algorithm to utilize for rate-limiting requests.
+     *
+     * Defaults to `token-bucket`.
+     */
+    readonly algorithm?: "fixed-window" | "token-bucket"
+  }
+}
+
+/**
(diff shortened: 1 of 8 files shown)


packages/effect/src/RateLimiter.ts
```diff
@@ -13,9 +15,11 @@ import * as internal from "./internal/rateLimiter.js"
 import type { Scope } from "./Scope.js"
 
 /**
- * Limits the number of calls to a resource to a maximum amount in some interval using the token bucket algorithm.
+ * Limits the number of calls to a resource to a maximum amount in some interval
+ * using the token-bucket algorithm.
```
tim-smart, line 19: This description is out of date now.

packages/effect/test/RateLimiter.test.ts
```diff
@@ -1,223 +1,271 @@
...
-import * as RateLimiter from "effect/RateLimiter"
 import { assert, describe } from "vitest"
 
-describe("RateLimiterSpec", () => {
-  it.effect("execute up to max calls immediately", () => {
+describe("RateLimiter", () => {
+  describe("fixed-window", () => {
+    RateLimiterTestSuite("fixed-window")
+
+    it.effect("uses the token-bucket algorithm for token replenishment", () =>
```
tim-smart, line 9: Should move this to the token bucket block.

packages/effect/test/RateLimiter.test.ts
```diff
@@ -1,223 +1,271 @@
...
+        // Use up all of the available tokens
+        yield* _(Effect.forEach(ReadonlyArray.range(1, 10), () => limit(Effect.unit)))
+
+        // Make an additional request when there are no tokens available
+        yield* _(
+          limit(Effect.unit),
+          Effect.zipRight(Deferred.succeed(deferred, void 0)),
+          Effect.fork
+        )
+        assert.isFalse(yield* _(Deferred.isDone(deferred)))
```
tim-smart, line 30: You will technically need a yieldNow before this, as forks don't execute immediately.

tim-smart (changes requested):

packages/effect/src/RateLimiter.ts
```diff
@@ -27,11 +31,49 @@ export interface RateLimiter {
...
+    /**
+     * The interval to utilize for rate-limiting requests. The semantics of the
+     * specified `interval` vary depending on the chosen `algorithm`:
+     *
+     * `token-bucket`: The maximum number of requests will be spread out over
+     * the provided interval if no tokens are available.
+     *
+     * For example, for a `RateLimiter` using the `token-bucket` algorithm with
+     * a `limit` of `10` and an `interval` of `1 seconds`, `1` request can be
+     * made every `100 millis`.
```
hsubra89, line 55:
```suggestion
     * For example, for a `RateLimiter` using the `token-bucket` algorithm with
     * a `limit` of `10` and an `interval` of `1 seconds`, `1` request can be
     * made every `100 millis` when the token bucket is empty.
```

packages/effect/src/internal/rateLimiter.ts
```diff
@@ -1,14 +1,53 @@
...
+    const millisPerToken = Math.ceil(Duration.toMillis(window) / limit)
+    const queue = yield* _(Queue.bounded<void>(limit - 1))
+    yield* _(
+      queue.take,
+      Effect.zipRight(Effect.sleep(millisPerToken)),
+      Effect.forever,
+      Effect.forkScoped,
+      Effect.interruptible
+    )
+    return (effect) => Effect.zipRight(queue.offer(void 0), effect)
```
hsubra89, line 44: It took me a while to understand this, but this is certainly very elegant :)

packages/effect/test/RateLimiter.test.ts
IMax153: Good call - fixed 👍

packages/effect/test/RateLimiter.test.ts
IMax153: That was dumb - thanks 👍

packages/effect/src/RateLimiter.ts
IMax153: Good call - fixed 👍

IMax153:
@tim-smart @mikearnaldi - since we're now updating the public signature of `RateLimiter.make`, should this be a minor?

tim-smart (approved):

PR #2099, opened 2024-02-11
optimize Effect.zip{Left,Right}

tim-smart:


mikearnaldi:
LGTM but we should use andThen more often that is directly using flatMap

tim-smart:
> LGTM but we should use andThen more often that is directly using flatMap

Hah yes, need to force the habit.

PR #2102, opened 2024-02-11
Swap Either type params

mikearnaldi:
Swap type params of Either from `Either<E, A>` to `Either<R, L = never>`.

Along the same line of the other changes this allows to shorten the most common types such as:

```ts
import { Either } from "effect";

const right: Either.Either<string> = Either.right("ok");
```

While it was confusing to use `Either<A, E>` I don't find it particularly confusing to use `Either<R, L>`, the type reads like "it's either a right or a left"

### DIFF
--- a/.changeset/cuddly-parrots-know.md
+++ b/.changeset/cuddly-parrots-know.md
@@ -1,0 +1,23 @@
+---
+"effect": minor
+---
+
+Consolidate `Effect.asyncOption`, `Effect.asyncEither`, `Stream.asyncOption`, `Stream.asyncEither`, and `Stream.asyncInterrupt`
+
+This PR removes `Effect.asyncOption` and `Effect.asyncEither` as their behavior can be entirely implemented with the new signature of `Effect.async`, which optionally returns a cleanup `Effect` from the registration callback.
+
+```ts
+declare const async: <A, E = never, R = never>(
+  register: (callback: (_: Effect<A, E, R>) => void, signal: AbortSignal) => void | Effect<void, never, R>,
+  blockingOn?: FiberId
+) => Effect<A, E, R>
+```
+
+Additionally, this PR removes `Stream.asyncOption`, `Stream.asyncEither`, and `Stream.asyncInterrupt` as their behavior can be entirely implemented with the new signature of `Stream.async`, which can optionally return a cleanup `Effect` from the registration callback.
+
+```ts
+declare const async: <A, E = never, R = never>(
+  register: (emit: Emit<R, E, A, void>) => Effect<void, never, R> | void,
+  outputBuffer?: number
+) => Stream<A, E, R>
+```
--- a/.changeset/olive-pants-sparkle.md
+++ b/.changeset/olive-pants-sparkle.md
@@ -1,0 +1,17 @@
+---
+"@effect/platform-node-shared": minor
+"@effect/typeclass": minor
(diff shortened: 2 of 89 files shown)


ethanniser:
**I think this a good change to keep it in line with all of the other changed types**

But I am curious why the decision to only modify the type parameter order, and not the runtime semantics?

Surely the function named `Either.right` would result in a `Either` with the type of the right?

```ts
const right: Either.Either<string, never> = Either.right("ok"); // right = left
const left: Either.Either<never, number> = Either.left(5); // left = right
```

because `Either` is "just a tagged union" there shouldn't be any reason that 'right' belongs to the first type parameter anymore than left

making them match just feels nicer?

of course it makes it a bit of a bigger migration, but its already a breaking change

mikearnaldi:
> **I think this a good change to keep it in line with all of the other changed types**
> 
> But I am curious why the decision to only modify the type parameter order, and not the runtime semantics?
> 
> Surely the function named `Either.right` would result in a `Either` with the type of the right?
> 
> ```ts
> const right: Either.Either<string, never> = Either.right("ok"); // right = left
> const left: Either.Either<never, number> = Either.left(5); // left = right
> ```
> 
> because `Either` is "just a tagged union" there shouldn't be any reason that 'right' belongs to the first type parameter any more than left
> 
> making them match just feels nicer?
> 
> of course it makes it a bit of a bigger migration, but its already a breaking change

Users complained that "right" feels like "success" and "left" doesn't

patroza:
gj for swapping to more semantic L and R.
there's still the obvious swapping in the generics: Right is on the Left and Left is on the Right.
but that's alright

fubhy (approved):

jessekelly881:
+1 for changing the name of `left`, `right`. Rust uses Ok() and Err(). Maybe that's a good alternative name.

mikearnaldi:
> +1 for changing the name of `left`, `right`. Rust uses Ok() and Err(). Maybe those are good alternative names.

Ok / Err would suggest primary usage as a Result, Either is used in Effect as a disjoined union given that for results we use Exit

jessekelly881:
Hmm. Yeah, I get your point. Although there are instances where it's implicitly treated as a Result. E.g. Effect.flatMap(either).

mikearnaldi:
> Hmm. Yeah, I get your point. Although there are instances where it's implicitly treated as a Result. E.g. Effect.flatMap(either).

That's for convenience nothing more

datner:
I think it should be noted in the docs, otherwise we will have fp devotees knocking at the door

mikearnaldi:
> I think it should be noted in the docs, otherwise we will have fp devotees knocking at the door

they can keep believing their cult and use fp-ts v2 :) jokes aside, yes everything needs to be documented

samhh:
I'm late to the party on this, but does this not make any prospective migration from fp-ts more difficult? If so, is it really worth it?

mikearnaldi:
> I'm late to the party on this, but does this not make any prospective migration from fp-ts more difficult? If so, is it really worth it?

We are not optimizing for fp-ts migration, we are optimizing for newcomers and effect users.o

The migration from fp-ts is a much larger topic, in fp-ts Either was used as a Result, in Effect results are expressed as Exit and also 90% of the direct Either usages should probably instead be straight Effect usages.

PR #2103, opened 2024-02-11
improve: Expand Either and Option `andThen` to support the `map` case like Effects' `andThen`

patroza:


packages/effect/src/Either.ts
```diff
@@ -568,12 +568,28 @@ export const flatMap: {
...
   <E1, A, E2, B>(self: Either<E1, A>, f: (a: A) => Either<E2, B>): Either<E1 | E2, B>
   <E1, A, E2, B>(self: Either<E1, A>, f: Either<E2, B>): Either<E1 | E2, B>
+  <E1, A, B>(self: Either<E1, A>, f: (a: A) => B): Either<E1, B>
+  <E1, A, B>(self: Either<E1, A>, f: B): Either<E1, B>
 } = dual(
   2,
   <E1, A, E2, B>(self: Either<E1, A>, f: (a: A) => Either<E2, B> | Either<E2, B>): Either<E1 | E2, B> =>
-    isFunction(f) ? flatMap(self, f) : flatMap(self, () => f)
+    flatMap(self, (a) => {
+      if (isFunction(f)) {
```
tim-smart, line 581: `const b = isFunction(f) ? f(a) : f` is a bit cleaner.

tim-smart (changes requested):

packages/effect/src/Either.ts
patroza: yea it's done

patroza: gotcha, done

tim-smart (approved):

PR #2104, opened 2024-02-11
Move `sqlfx/sql` into the `@effect/sql` package 

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #


packages/sql/src/Client.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * @since 1.0.0
+ * @since 2.0.0
```
tim-smart, line 2: You can leave this as `1.0.0`, as this isn't part of the `effect` package.

IMax153: done 👍

packages/sql/src/Transform.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @category transform
+ * @since 1.0.0
+ */
+export const pascalToSnake = (x: string) => (x.slice(0, 1) + x.slice(1).replace(/([A-Z])/g, "_$1")).toLowerCase()
+
+/**
+ * @category transform
+ * @since 1.0.0
+ */
+export const kebabToSnake = (x: string) => x.replace(/-/g, "_")
```
fubhy, line 56: Can we move this to the `effect/String` module?

fubhy (changes requested):

PR #2113, opened 2024-02-12
disable instanceof check for encoding schema classes

tim-smart:


packages/schema/src/Schema.ts
```diff
@@ -4922,13 +4922,21 @@ const makeClass = <A, I, R>(
...
-        pretty: () => (self: any) => `${self.constructor.name}(${pretty(self)})`,
-        arbitrary: () => (fc: any) => arb(fc).map((props: any) => new this(props))
-      })
+      const declaration: Schema<any, any, never> = declare(
+        [],
+        () => (input, _, ast) =>
+          input instanceof this
+            ? ParseResult.succeed(input)
+            : ParseResult.fail(ParseResult.type(ast, input)),
+        () => (input) => ParseResult.succeed(input),
```
gcanti, line 4931: It should be fine in any case since `selfSchema` still performs validation

gcanti (approved):

PR #2114, opened 2024-02-13
Fix the return order of batched requests

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/late-clocks-retire.md
+++ b/.changeset/late-clocks-retire.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix the ordering of results returned from batched requests
--- a/packages/effect/src/internal/blockedRequests.ts
+++ b/packages/effect/src/internal/blockedRequests.ts
@@ -385,7 +385,7 @@ export const parallelCollectionCombine = (
       key,
       Option.match(HashMap.get(map, key), {
         onNone: () => value,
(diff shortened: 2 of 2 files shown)


mikearnaldi:
I don't think I understand the problem and by consequence the solution, would you mind writing a test?

mikearnaldi:
> While it works and looks good to me, I'm not sure why the behaviour changed after Mike's request refactor.
> 
> I'll defer to @mikearnaldi to review for this one :)

I didn't touch that function

IMax153:
> > While it works and looks good to me, I'm not sure why the behaviour changed after Mike's request refactor.
> > I'll defer to @mikearnaldi to review for this one :)
> 
> I didn't touch that function

I was pretty convinced this was my error in the original port :) We just probably never noticed the issue before.

IMax153:
The issue basically was - if you send a set of batched requests that were received as follows:

```
Request 1 -> Request 2 -> Request 3
```

the results were being returned as:

```
Result 3 -> Result 2 -> Result 1
```

We tracked down the inadvertent reversal of results to this method which was combining results in reverse order instead of the correct order.

I'll write a test in a separate PR - apologies for merging before seeing your messages.

PR #2122, opened 2024-02-13
Add a way to redact HTTP headers

thewilkybarkid:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a way to redact specified HTTP headers by turning them into secrets. This changes the type from `Headers` to a `ReadonlyRecord<string | Secret>`.

This doesn't solve the problem of accidentally logging the whole `Headers` object, but the change to use secrets for certain/any header is fairly explosive. This does make it easier to redact headers manually, though, as opposed to https://github.com/PREreview/orcid-sync/blob/b4b92c6996ab7f0716c0f8615e038253f98b290c/src/index.ts#L14 (where keys are case-sensitive etc).

## Related

https://discord.com/channels/795981131316985866/1206639016464613486/1206720315565015040

### DIFF
--- a/packages/platform/src/Http/Headers.ts
+++ b/packages/platform/src/Http/Headers.ts
@@ -147,12 +147,9 @@ export const remove: {
  * @since 1.0.0
  */
 export const redact: {
-  (key: string | Array<string>): (self: Headers) => ReadonlyRecord.ReadonlyRecord<string | Secret.Secret>
-  (self: Headers, key: string | Array<string>): ReadonlyRecord.ReadonlyRecord<string | Secret.Secret>
+  (key: string): (self: Headers) => Record<string, string | Secret.Secret>
+  (self: Headers, key: string): Record<string, string | Secret.Secret>
 } = dual<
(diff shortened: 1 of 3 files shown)


tim-smart (changes requested):

tim-smart (approved):
🚀

packages/platform/src/Http/Headers.ts
```diff
@@ -141,3 +142,14 @@ export const remove: {
   (key: string) => (self: Headers) => Headers,
   (self: Headers, key: string) => Headers
 >(2, (self, key) => ReadonlyRecord.remove(self, key.toLowerCase()) as Headers)
+
+/**
+ * @since 1.0.0
+ */
+export const redact: {
+  (key: string): (self: Headers) => Record<string, string | Secret.Secret>
```
tim-smart, line 150: I think we should support a ReadonlyArray, as you can't chain this function currently

PR #2128, opened 2024-02-13
allow passing structs when encoding schema classes

tim-smart:


packages/schema/src/Schema.ts
```diff
@@ -4921,10 +4921,23 @@ const makeClass = <A, I, R>(
 
     static get ast() {
       const toSchema = to(selfSchema)
+      const encode = Parser.encode<any, any, never>(toSchema)
```
gcanti, line 4924: I think that this is `Parser.encodeUnknown`

tim-smart: 👍🏻

PR #2130, opened 2024-02-14
Improve Struct `pick` to support optional properties

evelant:
Makes `pick` behave more consistently. Allow picking optional properties and have them as optional in the result.

~~dtslint complains about this change, I'm not sure how to interpret the results however, happy to make any necessary changes~~

### DIFF
--- a/.changeset/dirty-badgers-add.md
+++ b/.changeset/dirty-badgers-add.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Improve types of Struct `pick` to support optional properties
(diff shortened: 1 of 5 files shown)


evelant:
Fixed dtslint errors

gcanti (commented):
The proposed change will modify the runtime behavior. While I'm ok with this change, it's advisable to include it in the next minor release.

Additionally, if we modify the behavior, we can utilize the built-in `Pick` type, eliminating the necessity for additional type helpers to obtain the optional properties.

I've submitted a PR to the `next-minor` branch, along with some tests (https://github.com/Effect-TS/effect/pull/2161).

evelant:
👍 Thanks for improving on my work, I don't know why I didn't think to just use `Pick` instead of adding a helper.

narley:
Thank you both for giving some tlc on this issue.

PR #2136, opened 2024-02-14
POC for schema class dual package hazard

tim-smart:


mikearnaldi:
Wondering if the identifier should be part of the type

patroza:
> Wondering if the identifier should be part of the type

you're killing me :D
you would basically break assignability and inheritance, perhaps depending on how you implement it.

tim-smart:
I don't think we need to add it to the type. I think just using class to get the free opaque identifier is a valid use case, so allowing structures of the same shape to be encoded is nice.

patroza:
> I don't think we need to add it to the type. I think just using class to get the free opaque identifier is a valid use case, so allowing structures of the same shape to be encoded is nice.

sometimes when there are 2 opposing forces pulling a thing into 2 directions, it may be really 2 things in disguise.
I am not opposed to supporting both, but indeed would prefer not to kill the convenient opaque identifier (and a constructor) case (everyone knows it's my primary).

tim-smart:
> sometimes when there are 2 opposing forces pulling a thing into 2 directions

I don't that is the case here. dual package hazard and encoding assignability are two separate issues in my opinion.

Hiding the identifier or exposing it in the types are both perfectly valid options.

patroza:
> Hiding the identifier or exposing it in the types are both perfectly valid options, but the latter will break the encoding assignability you are after.

Isn’t that what I’m saying? If there’s a valid use case for in the type, and a valid use case for it not in the type, we could have a strict and non strict class, for example.

tim-smart:
I would rather have one class api that can used everywhere, which is why I don't think we should add the symbol to the type.

We can solve both the encoding issue (already done) and the dual package issue with one api.

gcanti:
Closing in favour of https://github.com/Effect-TS/effect/pull/2172

PR #2157, opened 2024-02-15
feat(either): add do notation to Either

successkrisz:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add the following `Do notation` methods to `Either`:

- Either.Do
- Either.bindTo
- Either.bind
- Either.let

Convention is following how it is implemented for `Effect`s

### DIFF
--- a/packages/effect/src/Either.ts
+++ b/packages/effect/src/Either.ts
@@ -710,3 +710,94 @@ export const gen: Gen.Gen<EitherTypeLambda, Gen.Adapter<EitherTypeLambda>> = (f)
     return right(state.value)
   }
 }
+
+// -------------------------------------------------------------------------------------
+// do notation
+// -------------------------------------------------------------------------------------
+
+/**
+ * @since 2.4.0
+ * @category models
+ */
+export type MergeRecord<K, H> = {
+  [k in keyof K | keyof H]: k extends keyof K ? K[k]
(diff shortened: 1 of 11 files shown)


mikearnaldi:
can you please rebase on `next-minor`, type parameters of Either have been swapped

successkrisz:
Rebased and updated 👍

packages/effect/src/Either.ts
```diff
@@ -710,3 +710,94 @@ export const gen: Gen.Gen<EitherTypeLambda, Gen.Adapter<EitherTypeLambda>> = (f)
...
+
+// -------------------------------------------------------------------------------------
+// do notation
+// -------------------------------------------------------------------------------------
+
+/**
+ * @since 2.4.0
+ * @category models
+ */
+export type MergeRecord<K, H> = {
```
gcanti, line 722: I noticed that there's already a `MergeRecord` in `Effect.ts`. I think we can relocate this helper to `Types.ts` and import it from there

successkrisz: That certainly sounds like a good idea 👍

successkrisz: Done ✅

PR #2163, opened 2024-02-16
Fix order of interruptibility in experimental data loader

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/quick-pens-cough.md
+++ b/.changeset/quick-pens-cough.md
@@ -1,0 +1,5 @@
+---
+"@effect/experimental": patch
+---
+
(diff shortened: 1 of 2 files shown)


tim-smart:
I'll merge this because it feels more "correct", but the order doesn't actually matter as far as I am aware. This is still a clean exit:

```ts

Effect.acquireRelease(
  Effect.never.pipe(Effect.interruptible, Effect.forkScoped),
  () => Effect.unit,
).pipe(Effect.scoped, Effect.runFork)
```

PR #2167, opened 2024-02-16
add Hash.cached for caching hashes

tim-smart:


mikearnaldi:
The fastest access would be if `cache` directly mutates the object and writes a property `Hash.symbol` instead of the getter

tim-smart:
> Why should we use `cachedMethod` in favour of `cached` ?

It had a slight performance advantage, but it was quite small so I have consolidated the apis.

PR #2168, opened 2024-02-16
fix: Schema.filter error reporting

sukovanej:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When I have schema like `MySchema.pipe(Schema.filter(..., { message: (value) => 'filter message' }))` and the decoding fails because of `MySchema` and not because of the filter, it currently reports the `'filter message'` anyway.

Additionally, the `value` within `{ message: (value) => 'filter message' }` is typed as `A` of the `MySchema` which is not correct because it receives whatever is the input value because the parsing of `MySchema` might have failed.

### DIFF
--- a/.changeset/quiet-bobcats-walk.md
+++ b/.changeset/quiet-bobcats-walk.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+Propagate an original error of schema with `Schema.filter`.
--- a/packages/schema/src/TreeFormatter.ts
+++ b/packages/schema/src/TreeFormatter.ts
@@ -114,7 +114,7 @@ const getParseIsssueMessage = (
 /** @internal */
 export const getRefinementMessage = (e: ParseResult.Refinement, actual: unknown): Option.Option<string> => {
   if (e.kind === "From") {
-    return getParseIsssueMessage(e.error, () => getMessage(e.ast, actual))
+    return getParseIsssueMessage(e.error, () => getMessage(e.ast.from, actual))
   }
   return getMessage(e.ast, actual)
 }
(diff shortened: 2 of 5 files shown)


gcanti (commented):
> it currently reports the 'filter message' anyway

Yes, that's the intended behavior. Users expect that if they provide a custom message, it will override any default message. Indeed, `AST.MessageAnnotation<A>` is not correct; it shouldn't be parametric, but simply `(u: unknown) => string`.

sukovanej:
Hmm 🤔 , I understand the override is useful in case of a filter over "small" schemas (like constrained numbers, strings, etc). But in case of the schema being constrained describes a more complicated structure and I'm imposing an additional constraint on the structure then I'd really want to have separate validation errors depending on what part of the schema ast failed.

The current behaviour gave us hard times recently in our team. We have a filter handling a dependency between some fields within quite a complex object and we were staring at the error message wondering how could the filter condition not be satisfied and we needed to do a laborious investigation to find out there was an empty string deep down in the object that should have been non-empty.

gcanti:
Yes, I understand the use case, but we need to establish a clear rule to manage custom messages so that it's clear what happens. At the moment, the rule is simple: the custom message set by the user closest to the source of the error is used; otherwise, a default message is generated.

```ts
const schema = S.string.pipe(
  S.message(() => "invalid string"),
  S.nonEmpty({ message: () => "should be non empty" })
)
await Util.expectDecodeUnknownFailure(schema, null, `invalid string`) // because of S.message(() => "invalid string")
await Util.expectDecodeUnknownFailure(schema, "", `should be non empty`) // because of message: () => "should be non empty"
```

Issue #2199, opened 2024-02-22
Expose `ConfigError` not as a Union but wrapped as Class


mathiasmaerker:
### What version of Effect is running?

2.4

### What steps can reproduce the bug?

Sitting in the beginner workshop, we just spotted this not so nice behaviour
![image](https://github.com/Effect-TS/effect/assets/4051929/9b32c011-1e22-444f-8640-9ca7ad56731e)

### What is the expected behavior?

`ConfigError`should be matchable but only the union elements are catchable

### What do you see instead?

```typescript
.pipe(Effect.catchTag('ConfigError', () => Effect.succeed('bar')))
```

jessekelly881:
Relevant: https://github.com/Effect-TS/effect/issues/1589 (specifically, the overuse of `_tag`)

tim-smart:
Fixed in https://github.com/Effect-TS/effect/pull/2206

PR #2200, opened 2024-02-22
Fix ConfigError tag bug and make ConfigError a TaggedError

jessekelly881:
Fix for #2199. Replaces the use of `_tag` for discriminating sub-types(e.g. Add, Or, etc.) with a `ConfigError.Discriminant` symbol and reserves `_tag` for the top level type(i.e. "ConfigError"). Also, converts ConfigError to a TaggedError.

### DIFF
--- a/packages/effect/src/ConfigError.ts
+++ b/packages/effect/src/ConfigError.ts
@@ -17,6 +17,18 @@ export const ConfigErrorTypeId: unique symbol = internal.ConfigErrorTypeId
 export type ConfigErrorTypeId = typeof ConfigErrorTypeId
 
 /**
+ * @since 2.0.0
+ * @category symbols
+ */
+export const Discriminant: unique symbol = internal.Discriminant
+
+/**
+ * @since 2.0.0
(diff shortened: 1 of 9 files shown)


packages/effect/src/ConfigError.ts
```diff
@@ -2,8 +2,42 @@
  * @since 2.0.0
  */
 import type * as Cause from "./Cause.js"
+import type { And, InvalidData, MissingData, Or, SourceUnavailable, Unsupported } from "./internal/configError.js"
 import * as internal from "./internal/configError.js"
 
+export {
```
mikearnaldi, line 8: We can't re-export from internal without explicit typing, files in internal don't appear in the compiled `.d.ts`

mikearnaldi:
closing in favour of https://github.com/Effect-TS/effect/pull/2206

jessekelly881:
@mikearnaldi Do you think it makes sense to convert ConfigError sub-types to TaggedError though? If so, I'll create a separate PR.

PR #2201, opened 2024-02-23
feat: added Arbitrary.sample

jessekelly881:
Exposes Arbitrary.sample helper


gcanti:
Thanks @jessekelly881 but it doesn't add much to the module, it's just a repetition of a fast-check API. I would keep the API surface as small as possible.

PR #2204, opened 2024-02-24
remove flake utils

ricardo-valero:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Removing the `flake-utils` dependency from `flake.nix`

## Related

- Related Issue #
- Closes #

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,23 +1,5 @@
 {
   "nodes": {
-    "flake-utils": {
-      "inputs": {
-        "systems": "systems"
-      },
-      "locked": {
-        "lastModified": 1694529238,
-        "narHash": "sha256-zsNZZGTGnMOf9YpHKJqMSsa0dXbfmxeoJ7xHlrt+xmY=",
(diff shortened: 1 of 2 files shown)


IMax153:
@ricardo-valero - alright, I think that's a reasonable enough justification. I would suggest abstracting `nixpkgs.lib.genAttrs systems` into a helper function:

```nix
forAllSystems = nixpkgs.lib.genAttrs systems;
# Later ...
# forAllSystems (system: ...)
```

Also, please be aware that `import nixpkgs { inherit system; }` vs `nixpkgs.legacyPackages.${system}` has some subtle but non-trivial differences, which you can read about in this [discussion](https://discourse.nixos.org/t/using-nixpkgs-legacypackages-system-vs-import/17462). 

In this flake, the difference does not matter, but something to be aware of when using this pattern.

IMax153 (approved):

PR #2209, opened 2024-02-25
Schema: add pickLiteral to Schema literal

### LINKED ISSUE #2192: Picking literals in Schema

I would like to make reusable Schemas, let's consider the following example:
I have an Entity A and an Entity B. I have a property "ABC" that represent a union of strings in EntityA. 

```ts
const ABC = Schema.literal("a", "b", "c")

const EntityA = Schema.struct({
  abc: ABC,
});

const EntityB = Schema.struct({
  ab: Schema.literal("a", "b"),
});
```

**Problem:** With the following, if the model change in EntityA, it will not affect EntityB. And the source of truth that represent ABC, must live in EntityA. This can cause errors, if we do not remember to change manually the "ab" prop in EntityB accordingly to what has been change in EntityA. 
 
**Opportunity:** To keep both Entities in a relationship, I would like to reuse the ABC schema, and narrow down to just "a" or "b". 

An Expected API could look like this:
```ts
const ABC = Schema.literal("a", "b", "c");
const AB = ABC.pipe(Schema.pickLiteral("a", "b"));
```

```ts
const ABC = Schema.struct({
  a: Schema.literal("a"),
  b: Schema.literal("b"),
  c: Schema.literal("c"),
});

const AB = Schema.union(
  ABC.pipe(Schema.pluck("a", { transformation: false })),
  ABC.pipe(Schema.pluck("b", { transformation: false }))
);
```

steffanek:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add `pickLiteral` to Schema so that we can pick values from a Schema literal.

## Related

- Related Issue #2192 
- Closes #2192

### DIFF
--- a/packages/schema/README.md
+++ b/packages/schema/README.md
@@ -1130,6 +1130,40 @@ S.literal(2n); // bigint literal
 S.literal(true);
 ```
 
+We can also `pickLiteral` from a literal schema.
+
+```ts
+import * as S from "@effect/schema/Schema";
+
+S.literal("a", "b", "c").pipe(S.pickLiteral("a", "b")); //same as S.literal("a", "b")
+```
+
+Sometimes we need to reuse a schema literal in other parts of the code. Let's see an example:
+
+```ts
+import * as S from "@effect/schema/Schema";
+
+const FruitId = S.number;
+// the source of truth regarding the Fruit category
+const FruitCategory = S.literal("sweet", "citrus", "tropical");
+
+const Fruit = S.struct({
+  id: FruitId,
+  category: FruitCategory,
+});
+
+// Here, we want to reuse our FruitCategory definition to create a subtype of Fruit
+const SweetAndCitrusFruit = S.struct({
+  fruitId: FruitId,
+  category: FruitCategory.pipe(S.pickLiteral("sweet", "citrus")),
+  /*
+    By using pickLiteral from the FruitCategory, we ensure that the values selected
+    are those defined in the category definition above.
+    If we remove "sweet" from the FruitCategory definition, TypeScript will notify us.
+    */
+});
+```
+
 ## Template literals
 
 The `templateLiteral` combinator allows you to create a schema for a TypeScript template literal type.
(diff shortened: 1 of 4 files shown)


packages/schema/README.md
```diff
@@ -1130,6 +1130,14 @@ S.literal(2n); // bigint literal
 S.literal(true);
 ```
 
+We can also `pickLiteral` from a literal schema.
```
gcanti, line 1133: I would add an explanation of the use case, that is, the motivation for why a user should use `S.literal("a", "b", "c").pipe(S.pickLiteral("a", "b"))` instead of simply using `S.literal("a", "b")`, it's not at all obvious at first glance.

steffanek: Proposal: (I would add the following bloc under the current version)
Sometimes we need to reuse schemas in other parts of the code. Let's see an example:
```ts
const FruitId = Schema.number;
// the source of truth regarding the Fruit category
const FruitCategory = Schema.literal("sweet", "citrus", "tropical"); 

class Fruit extends Schema.Class<Fruit>()({
  id: FruitId,
  category: FruitCategory,
}) {}

//Here we want to reuse are FruitCategory definition to create a sub type of the Fruit.
class SweetAndCitrusFruit extends Schema.Class<SweetAndCitrusFruit>()(
  {
    fruitId: FruitId,
    category: FruitCategory.pipe(Schema.pickLiteral("sweet", "citrus")), 
    /* 
    By using pickLiteral from the FruitCategory, we ensure that the values selected 
    are those defined in the category definition above. 
    If we remove "sweet" from the FruitCategory definition, TypeScript will notify us.
    */
  }
) {}
```
ps. the **`Literals`** section is located at the beginning of the `README.md`, should we consider using `Schema.Class`, `Schema.struct` or both in the proposal above?

gcanti: LGTM 

(only a typo in "//Here we want to reuse **are** FruitCategory definition to create a sub type of the Fruit", I'd propose "Here, we want to reuse our FruitCategory definition to create a subtype of Fruit")

p.s.
let's use `Schema.struct`

PR #2211, opened 2024-02-25
add ManagedRuntime module, to make incremental adoption easier

tim-smart:


mikearnaldi:
not sure if this helps adoption or just induces users into a bad pattern, namely using `using` would lead to scope being open and closed on demand instead of having the layer survive for the program lifetime

tim-smart:
> not sure if this helps adoption or just induces users into a bad pattern, namely using `using` would lead to scope being open and closed on demand instead of having the layer survive for the program lifetime

Having it implement `Disposable` is definitely up for debate. The main thing I wanted to tackle was providing a way to run effects in non-effect land in a nice way without having to juggle a Scope, Runtime and also memoizing the runtime.

PR #2214, opened 2024-02-25
Change `Effect/Stream.acquireRelease` function signatures in favor of using an object with named functions

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Using an object improves readability. 
Brings the signature of a function to the same pattern as used in `Effect.tryPromise({ try, catch })`

```ts
// before: 
 const acquireDBConnection = Effect.acquireRelease(
   Effect.sync(() => createConnection('mysql://...')),
   (connection) => Effect.sync(() => connection.end(() => {})),
 )

// after:
 const acquireDBConnection = Effect.acquireRelease({
   acquire: Effect.sync(() => createConnection('mysql://...')),
   release: (connection) => Effect.sync(() => connection.end(() => {})),
 })
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/olive-chefs-yawn.md
+++ b/.changeset/olive-chefs-yawn.md
@@ -1,0 +1,12 @@
+---
+"effect": minor
+"@effect/platform-node-shared": patch
+"@effect/platform-browser": patch
+"@effect/opentelemetry": patch
+"@effect/platform-node": patch
+"@effect/experimental": patch
+"@effect/platform": patch
+"@effect/cli": patch
+---
+
+change `Effect.acquireRelease`, `Effect.acquireReleaseInterruptible`, `Stream.acquireRelease` function signatures in favor of using an object with named functions
--- a/packages/cli/test/services/MockTerminal.ts
+++ b/packages/cli/test/services/MockTerminal.ts
@@ -40,10 +40,10 @@ export const MockTerminal = Context.GenericTag<Terminal.Terminal, MockTerminal>(
 // =============================================================================
 
 export const make = Effect.gen(function*(_) {
-  const queue = yield* _(Effect.acquireRelease(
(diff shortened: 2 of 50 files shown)


KhraksMamtsov:
I've developed codemod also. so if this PR has value, I'll do a PR in codemode if needed.

mikearnaldi:
It used to be like that but unfortunately inference is problematic, given that `release` depends on the type of `acquire`, maybe we can get around it with the new `NoInfer`? Cc @tim-smart @IMax153 @gcanti

KhraksMamtsov:
I've used custom `NoInfer<>`, or you mean ts-5.4 intrinsic?

mikearnaldi:
> I've used custom `NoInfer<>`, or you mean ts-5.4 intrinsic?

Custom one, was wondering though if `NoInfer` is the reason we could not make it work before

KhraksMamtsov:
I think it still has problems( I had to rewrite https://github.com/Effect-TS/effect/pull/2214/files#diff-4ba919f54512656f08fc7467137253953332e37248856cd471f5995f9f5616edR3048 because of inferring problems
I've tried 5.4.0-beta and it doesn't work as I expected( but it is beta

mikearnaldi:
Well if it doesn't work it doesn't work

KhraksMamtsov:
please close the PR if you think it is necessary

mikearnaldi:
I will check it better when I am back Tuesday, I hoped it was working

mikearnaldi:
MMM this seems to be a TS bug imho, if we change the `acquire` signature to be a function like `() => ...` it works fine

mikearnaldi:
I mean next major

PR #2216, opened 2024-02-26
Default `batching` to `"inherit"` if not specified in `Effect.forEach`

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

During the advanced workshop, it was noted by @tim-smart  and I that `Effect.forEach` does not inherit the regional value for request batching. 

This PR defaults `batching` to `"inherit"` if not specified in `Effect.forEach`, so the following code:

```ts
Effect.forEach([1, 2, 3], myRequest)
```

becomes equivalent to:

```ts
Effect.forEach([1, 2, 3], myRequest, { batching: "inherit" })
```

whereas previously, it would have been equivalent to:

```ts
Effect.forEach([1, 2, 3], myRequest, { batching: false })
```

cc-ing @mikearnaldi in case this was an intentional decision. If so, I will open a separate PR to document this behavior.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/spotty-cameras-call.md
+++ b/.changeset/spotty-cameras-call.md
@@ -1,0 +1,23 @@
+---
+"effect": patch
+---
+
+Default `batching` to `"inherit"` if not specified in `Effect.forEach`
+
+With this change, the following:
+
+```ts
+Effect.forEach([1, 2, 3], myRequest)
+```
+
+is equivalent to:
+
+```ts
+Effect.forEach([1, 2, 3], myRequest, { batching: "inherit" })
+```
+
+Previously, it would have been equivalent to:
+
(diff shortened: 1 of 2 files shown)


mikearnaldi:
Batching has to be opt-in as having effects with requests mixed with effects without requests leads to requests never being processed. All settings of concurrency are opt-in when not specified

tim-smart:
We could set the FiberRef to false by default?

mikearnaldi:
No because setting it to true would have impact on all children which is potentially unsafe. All concurrency and batching have to default to false to be safe

PR #2225, opened 2024-02-29
Add Effect.Tag to simplify access to service

mikearnaldi:


packages/effect/src/Effect.ts
```diff
@@ -5198,3 +5197,53 @@ export const fromNullable: <A>(value: A) => Effect<NonNullable<A>, Cause.NoSuchE
...
+    Object.defineProperty(TagClass, "stack", {
+      get() {
+        return creationError.stack
+      }
+    })
+    // @ts-ignore
+    TagClass["$"] = new Proxy({}, {
+      get(_target: any, prop: any, _receiver) {
+        // @ts-expect-error
+        return core.andThen(TagClass, (s) => s[prop])
```
patroza, line 5235: maybe cache?

mikearnaldi: caching proxy leads to Max Call Stack

packages/effect/src/Effect.ts
```diff
@@ -5198,3 +5197,53 @@ export const fromNullable: <A>(value: A) => Effect<NonNullable<A>, Cause.NoSuchE
...
+      }
+    })
+    const done = new Proxy(TagClass, {
+      get(_target: any, prop: any, _receiver) {
+        if (prop in TagClass) {
+          // @ts-expect-error
+          return TagClass[prop]
+        }
+        // @ts-expect-error
+        return (...args: Array<any>) => core.andThen(TagClass, (s: any) => s[prop](...args))
```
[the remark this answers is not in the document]
mikearnaldi: same as below

packages/effect/src/Effect.ts
patroza: I mean like this https://github.com/Effect-TS/effect/pull/2230/files

patroza:
@mikearnaldi is it intentional that the `$` shape is not filtering out functions? This way there are dupes with different signatures.

packages/effect/src/Effect.ts
mikearnaldi: I think this would be slower, not sure, the issue caching would try to prevent is proxy access but proxy is accessed anyway, for the rest a cache lookup or a single function call have comparable overhead

patroza: my point was more like garbage, creating a new function instead of lookup and return old. at least for the function case. but sounds like micro optimisations, especially considering all the functions we create and destroy on our side of the pond :D

mikearnaldi:
> @mikearnaldi is it intentional that the `$` shape is not filtering out functions? This way there are dupes with different signatures.
> 
> like this would do: `[k in keyof Shape as Shape[k] extends (...args: Array<any>) => any ? never : k]` (inverse of the function one)

It is intentional, if the signatures are not the same they are not dupes. Every member is available to be accessed as a constant

patroza:
> It is intentional, if the signatures are not the same they are not dupes. Every member is available to be accessed as a constant
but

mikearnaldi:
> > It is intentional, if the signatures are not the same they are not dupes. Every member is available to be accessed as a constant
> 
> but Screenshot 2024-03-01 at 16 11 19

to access a function and use it later for example

mikearnaldi:
it's of arguable importance though

patroza:
> it's of arguable importance though

yea, if you don't access it for use, it's kinda better to consume the service

patroza:
@mikearnaldi 
`readonly doStuff: (...strings: readonly string[]) => Effect.Effect<string>`
and 
`readonly doStuff: (...strings: ReadonlyArray<string>) => Effect.Effect<string>`

infer to `never`

mikearnaldi:
@patroza seems like a TS bug:

https://www.typescriptlang.org/play?#code/C4TwDgpgBAGlC8UAKBDATigthYE0GcAeACgDpz0BzfALigCUIUATAewDsAbEAQTQxCF8wNAEt2lAHwBKBJKgA3VqOaSAUAHoNUKAD0A-EA

.changeset/brave-carrots-rest.md
```diff
@@ -0,0 +1,26 @@
+---
+"effect": patch
+---
+
+Add Effect.Tag to simplify access to service.
+
+This change allows to define tags in the following mannar:
```
patroza, line 7: manner

PR #2242, opened 2024-03-01
Idea: stop gh test suite after first test fails

jessekelly881:
Stop running test suite when the first test has failed. The down side is that you have to run the test suite locally to see all failures but that should be done anyway..


mikearnaldi:
It's helpful to see all tests that fail

PR #2247, opened 2024-03-04
xstate integration

tim-smart:


thanhnguyen2187:
Hi @tim-smart! Sorry for digging this up, but I'm curious why was was this discontinued? I've been using XState for quite a while, and feel like it would be an excellent companion to Effect (XState as the "big picture" state management, and Effect as the more granular function replacement), and would love to see those two go together

fubhy:
Some state machine work has since happened here (experimental): https://github.com/Effect-TS/effect/blob/main/packages/experimental/test/Machine.test.ts

tim-smart:
Xstate and Effect didn't play together too well, as xstate doesn't have typed errors, and you have to juggle an Effect runtime if you want to use your services.

You could use xstate with a ManagedRuntime, but I haven't explored it.

PR #2252, opened 2024-03-05
CLI: Print unknown argument errors

ccntrq:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Minor improvement

## Description
All other validation errors are handled and printed in cliApp.ts:83. These errors were previously not printed and silently returned to the user on the error channel. We should either print all validation errors or handle all of them silently and leave printing/reporting to the caller. This commit implements the former option.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/selfish-kangaroos-sparkle.md
+++ b/.changeset/selfish-kangaroos-sparkle.md
@@ -1,0 +1,5 @@
+---
+"@effect/cli": minor
+---
+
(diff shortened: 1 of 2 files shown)


IMax153 (approved):
LGTM! Thank you @ccntrq 👍

PR #2258, opened 2024-03-06
Prototype deep equality

### LINKED ISSUE #2251: From Discord: Issue with Request.tagged and Equal.equals for non-primitive parameters
# Summary
In the above conversation, the user raised an issue with using `Request.tagged` and `Equal.equals` for requests that contain non-primitive parameters. It was observed that two requests with equal values are not considered equal according to `Equal.equals` if they contain non-primitive parameters. This can lead to caching not working for these types of requests.

To address this issue, there are a few best practices to consider:

1. Use the Data module for internal objects: By using the Data module for non-primitive parameters, you can ensure that they are compared by value rather than by reference. This can help in achieving the desired behavior for caching.

2. Implement custom Equals for Request objects: Another approach is to manually implement the Equals function for Request objects. This allows you to customize how the objects are compared, taking into account the specific properties that should be considered for equality.

3. Consider comparing specific properties: In some cases, comparing specific properties, such as an "id" field, may be sufficient for determining equality. This can be a simpler approach, especially for entities where the unique identifier is the most important factor.

Overall, the key takeaway is that when dealing with requests that have non-primitive parameters, it's important to ensure that the equality comparison is implemented correctly to enable caching and other desired behaviors.

# Discord thread

https://discord.com/channels/795981131316985866/1214495832246853682

mikearnaldi:
If we decide that this goes in the right direction we'll have to extend the rest of constructors to allow deep comparison.

cc @tim-smart for further testing & potential api extensions

linked discord discussion: https://discord.com/channels/795981131316985866/795983589644304396/1214495832246853682

closes #2251

### DIFF
--- a/packages/effect/src/internal/data.ts
+++ b/packages/effect/src/internal/data.ts
@@ -1,4 +1,5 @@
 import * as Equal from "../Equal.js"
+import { pipe } from "../Function.js"
 import * as Hash from "../Hash.js"
 import type * as Types from "../Types.js"
 import { StructuralPrototype } from "./effectable.js"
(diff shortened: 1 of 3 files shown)


packages/effect/src/internal/data.ts
```diff
@@ -34,3 +35,108 @@ export const Structural: new<A>(
...
+        h ^= pipe(Hash.string(keys[i]! as string), Hash.combine(deepHash((value as any)[keys[i]!])))
+      }
+      return Hash.optimize(h)
+    }
+  }
+  return Hash.hash(value)
+}
+
+function deepComp(this: any, that: any) {
+  if (typeof that === typeof this && that !== null && this[deepSymbol] === that[deepSymbol]) {
```
tim-smart, line 107:
```suggestion
  if (typeof that === typeof this && that !== null && this[deepSymbol] && that[deepSymbol]) {
```

Only if both have deep equality enabled?

tim-smart (commented):
I wonder if we could do it regionally:

```ts
const Thing = Data.withDeepEquality(() => Data.tagged<Thing>())
```

Adding an options param everywhere feels a bit clunky.

tim-smart:
Tried out the regional idea: https://github.com/Effect-TS/effect/pull/2259/files

mikearnaldi:
I feel like regional would be far too verbose and it's not a solution for classes as you can't creste a class inside the region

packages/effect/src/internal/data.ts
mikearnaldi: yes, equality should be symmetric no?

tim-smart: Just wasn't sure if setting deep to false on both should pass the condition.

mikearnaldi:
@tim-smart I think we need to make a choice here, either we add the param to the api or we decide it's not important. WDYT? cc @IMax153 @gcanti @fubhy

gcanti:
against adding params

fubhy:
Both seem a bit awkward tbh. Also, apologies in advance as I'm probably missing something here, but why would you ever want to use the `Data.proxy()` and then set `deep` to false? Isn't the whole point of it to support deep equality?

Also, the name isn't great. The consumers shouldn't care that it uses a proxy. The name should reflect what it does, not how.

tim-smart:
I think deep equality is something worth pursuing, as a beginner will likely expect Equal.equals to also check the `metadata` object:

```ts
import { Data } from "effect"

class User extends Data.Class<{ name: string; metadata: { foo: string } }> {}
```

But I wonder if we tackle it in the constructors rather than in the actual Equal implementation. I.e. ensure any nested object implement Equal & Hash.

mikearnaldi:
```ts
import { Data } from "effect"

class User extends Data.Class<{ name: string; metadata: { foo: string } }>({deep: true}) {}
```

I don't see any single reason not to add parameters.

Constructors don't exist for plain objects for example `Data.tagged` and in classes it would require typing parameters which would look horrendous.

The decision here is:
1) we care about deep equal => we add params
2) we don't care about deep equals => everything remains the same

If anyone has concrete concerns about adding params that are not "I am against" please explain with examples and reasons

gcanti:
If the issue is unexpected behavior, adding an opt-in parameter `{ deep: true }` doesn't solve the problem, it should be the default behavior.

fubhy:
> Having an object / array already constructed and wanting to have a view (proxy) which compares by value instead of by reference

Yes, I get that, but my question was why would you ever want that then to NOT perform deep comparison. As Giulio also said, it should at least be deep by default then. I'm just additionally questioning if you'd ever want it to NOT perform deep comparison, hence wondering if we can just drop that option entirely.

mikearnaldi:
Note: the issue is not unexpected behaviour, it's absolutely expected not to have deep comparison if you use mutable data such as nested objects. But that doesn't mean the user wouldn't want deep comparison, this is about making it easier to do it instead of having to declare nested classes

fubhy:
Could this be an option too?

```ts
import { Data } from "effect"

class User extends Data.Class<{ 
  name: string;
  metadata: { foo: string }; 
  a: { b: MutableStuff } }
({
  a: {
    b: MyCustomCmpFn // Explicitly override the default behavior here.
  }
}) {}
> ```

mikearnaldi:
> Could this be an option too?
> 
> ```ts
> import { Data } from "effect"
> 
> class User extends Data.Class<{ 
>   name: string;
>   metadata: { foo: string }; 
>   a: { b: MutableStuff } }
> ({
>   a: {
>     b: MyCustomCmpFn // Explicitly override the default behavior here.
>   }
> }) {}
> ```

it's not only a comparator function but also a hashing implementation, not sure if it would be possible, also I would have no idea how to write the type of the parameter

mikearnaldi:
Alternative idea: we could use decorators instead of parameters for classes

I think we need to approach this in steps:

1) provide a good way of creating deeply compared data classes, ideally via a decorator for backward compatibility

2) long term re-evaluate deep equality by default with the caveat of not deeply comparing custom typed but only plain structures, this if it ever lands will be material for 4.0 after a good period of testing

tim-smart:
I think decorators could be nice, you could use them for both classes and the other apis from what I recall.

I remember trying to use them when trying out apis for making Hash lazy, but ran into some issues with tsconfig.

fubhy:
Closing this PR because it has been marked as no longer actionable. If the change is still needed, please open a fresh PR with current context.

PR #2262, opened 2024-03-08
Don't log an empty message

thewilkybarkid:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/spicy-brooms-look.md
+++ b/.changeset/spicy-brooms-look.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+Don't log an empty message when responding to a request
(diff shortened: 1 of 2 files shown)


tim-smart:
The empty message was intentional, as all the information is in the annotations. Was it breaking something for you?

thewilkybarkid:
It's not breaking anything, it was just a little harder when eyeballing a stream of log messages. (Though this was the only case where the message was empty, so easier once I worked that out. 🙃)

mikearnaldi:
> The empty message was intentional, as all the information is in the annotations. Was it breaking something for you?

Any reason why a log message would be bad?

tim-smart:
Not bad, but my only nit pick is that it makes the stream of logs quite repetitive, as the message adds no value.

mikearnaldi:
Are we logging purely annotations?

PR #2264, opened 2024-03-08
fix: tap and andThen fallthrough function

patroza:
fixes for example:
```ts
declare const foo: (x: number) => Effect.Effect<void>;
Effect.succeed(true).pipe(Effect.tap(foo)); // no error
Effect.succeed(true).pipe(Effect.tap(x => foo(x))); // error
```

### DIFF
--- a/.changeset/angry-shirts-repair.md
+++ b/.changeset/angry-shirts-repair.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+catch unmatched function fallthrough in andThen
--- a/.changeset/angry-shirts-repair.md
+++ b/.changeset/angry-shirts-repair.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-catch unmatched function fallthrough in andThen
+fix: unmatched function fallthrough in `andThen` and `tap`
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -54,7 +54,7 @@ import * as Scheduler from "./Scheduler.js"
 import type * as Scope from "./Scope.js"
 import type * as Supervisor from "./Supervisor.js"
 import type * as Tracer from "./Tracer.js"
-import type { Concurrency, Covariant, MergeRecord, NoInfer } from "./Types.js"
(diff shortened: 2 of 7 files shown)


mikearnaldi:
Mmm what about Effects that are functions? Like the proxy ones from the new tag

patroza:
> Mmm what about Effects that are functions? Like the proxy ones from the new tag

What about them? It they match the supported function signature, then they’re golden. Otherwise they are I believe rightfully rejected.

mikearnaldi:
I think if an arg is both a function and an effect we have to force `() =>` because the impl will take a choice, either there is an `isFunction` or `isEffect`. Quite a conondrum...

patroza:
> I think if an arg is both a function and an effect we have to force `() =>` because the impl will take a choice, either there is an `isFunction` or `isEffect`. Quite a conondrum...

I see, seems like a related topic, how is it affected by this change?

mikearnaldi:
It's not affected but related, with the intro of effect.tag we can't
basically distinguish functions from effects, it is an issue for tacit
usage

patroza:
> It's not affected but related, with the intro of effect.tag we can't basically distinguish functions from effects, it is an issue for tacit usage

alright, would at least like to get the bug fix merged so that users stop getting bit by it.
In the meantime we can look into it.
From my gut feeling though; isn’t the proxy hiding the internals on access? On phone. Would need some test cases I guess

tim-smart (approved):

tim-smart:
I think there are a few other places where we assume an Effect isn't a function, will need to hunt them down.

patroza:
> It's not affected but related, with the intro of effect.tag we can't basically distinguish functions from effects, it is an issue for tacit usage

I see now yes, that's certainly 'fun' :D

patroza:
> It's not affected but related, with the intro of effect.tag we can't basically distinguish functions from effects, it is an issue for tacit usage

How about https://github.com/Effect-TS/effect/pull/2266/files#diff-1eaca66f74be66210f25595750619630e4d14443b5ddaa1f2f8a265256432778R5292

PR #2266, opened 2024-03-08
proposal for Effect.Tag conundrum

patroza:
(I left out the changeset on purpose; what is the policy for changesets for fixes for unreleased changes)

### DIFF
--- a/.changeset/tasty-seals-film.md
+++ b/.changeset/tasty-seals-film.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+fix Effect.Tag generated proxy functions to work with andThen/tap, or others that do function/isEffect checks
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -5289,7 +5289,15 @@ export const Tag: <const Id extends string>(id: Id) => <Self, Type>() =>
           return cache.get(prop)
         }
         // @ts-expect-error
-        const fn = (...args: Array<any>) => core.andThen(TagClass, (s: any) => s[prop](...args))
+        const fn = (...args: Array<any>) =>
+          core.andThen(TagClass, (s: any) => {
+            if (typeof s[prop] === "function") {
+              cache.set(prop, (...args: Array<any>) => core.andThen(TagClass, (s: any) => s[prop](...args)))
+              return s[prop](...args)
+            }
+            cache.set(prop, core.andThen(TagClass, (s) => s[prop]))
+            return s[prop]
+          })
         // @ts-expect-error
         const cn = core.andThen(TagClass, (s) => s[prop])
         Object.assign(fn, cn)
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -5288,13 +5288,15 @@ export const Tag: <const Id extends string>(id: Id) => <Self, Type>() =>
         if (cache.has(prop)) {
           return cache.get(prop)
         }
-        // @ts-expect-error
         const fn = (...args: Array<any>) =>
+          // @ts-expect-error
           core.andThen(TagClass, (s: any) => {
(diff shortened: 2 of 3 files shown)


mikearnaldi:
would be a patch, not sure I understand the fix though

patroza:
> would be a patch, not sure I understand the fix though

andThen & friends detect a function, while underlying it's an effect.
it's going to call that function, which in turn can really check if the underlying is a function to be called, or an effect.
solving the disambiguity. once it has been able to check if its a function to call, or an effect to return, it caches that for the next calls.

mikearnaldi:
> > not sure I understand the fix though
> 
> andThen & friends detect a function, while underlying it may be an Effect (aka the conundrum). it's going to call that function, which in turn can really check if the underlying service member is a function to be called, or an Effect to return. solving the disambiguity. once it has been able to check if its a function to call, or an Effect to return, it caches that knowledge for the next calls so that you only pay the price once.
> 
> the requirement for this to work consistently, is that at every place we want to deal with both functions and effects, we must first check if it's a function, only then if it's an Effect. as luck so has it (or well, it must be done that way anyway), that appears to be the case already.
> 
> > would be a patch
> 
> isn't that weird, to have a changelog entry for an unreleased feature?

Not really weird if we link changelogs to PRs. RE the fix, I am kind of unsure, there would be a mismatch between type and what happens behind the scenes, this would work with `andThen` but it feels like it's made as a special case for that.

cc @tim-smart

tim-smart:
I propose a TagClassTypeId that can be checked at runtime?

tim-smart:
There is one other case in cli that needs addressing:

image

Also, instead of using a map you could put the cached props back on the class itself:

image

tim-smart:
Hmm I think the only way to ensure functions are functions, and effects are effects is to separate the access like the previous revision.

patroza:
> Hmm I think the only way to ensure functions are functions, and effects are effects is to separate the access like the previous revision.

What’s wrong with my solution?

tim-smart:
Ah yeah thinking about it some more your solution will work :)

Could also re-use the `cn` variable for the constant case.

patroza:
> Ah yeah thinking about it some more your solution will work :)
> 
> Could also re-use the `cn` variable for the constant case.

Awesome. Yea sure, who doesn’t like free stuff :)

tim-smart:
The one case where it will still fail is if something expects a union of a function they doesn't return an Effect, or a plain Effect. I.e `LazyArg<A> | Effect<A>`

patroza:
> The one case where it will still fail is if something expects a union of a function that doesn't return an Effect, or a plain Effect. I.e `LazyArg<A> | Effect<A>`

Got a failing test?

patroza:
> > Hmm I think the only way to ensure functions are functions, and effects are effects is to separate the access like the previous revision.
> 
> The issue is that it's valid to have a function be an effect, it's a dsl that we used in the past too if I recall in platform, so this issue is kind of orthogonal to tag

Only in implementation or also on the type level? If also on type level it should be fine as long as andThen and friends have the same bias type wise as implementation wise

mikearnaldi:
> > > Hmm I think the only way to ensure functions are functions, and effects are effects is to separate the access like the previous revision.
> > 
> > 
> > The issue is that it's valid to have a function be an effect, it's a dsl that we used in the past too if I recall in platform, so this issue is kind of orthogonal to tag
> 
> Only in implementation or also on the type level? If also on type level it should be fine as long as andThen and friends have the same bias type wise as implementation wise

No not necessarily, the function will return something different and expect potentially different arguments. It's not safe to assume andThen behaviour

mikearnaldi:
After a good amount of thinking it I think it is safe to discriminate between `Function` and `NotFunction` so this looks like it is working. The only troubles I can see are in tacit usage cases which we are advising against anyway

PR #2269, opened 2024-03-10
add PlatformLogger module

### LINKED ISSUE #2210: Platform/FileSystem: Add a default file system Logger instance

Logging to a file is common enough that it might be useful to have a default file system logger built on top of Platform/FileSystem. Format could be equivalent to Logger.defaultLogger. 

```ts
import { fileSystemLogger } from "@effect/platform/FileSystem"

const layer = Logger.addScoped(fileSystemLogger("./test.log", { flag: "a+" }))
```

_No response_

jessekelly881:
Adds a default file logger based on @effect/platform/FileSystem. Closes #2210

I'm not what the best approach is for adding tests for this until we have an in memory file system we can test against(writing to a test file and re-reading it I suppose)..  Also, I'm not 100% sure if using stringLogger is correct or if fmtLogger would be preferred(or maybe both?)


packages/platform/src/internal/fileSystem.ts
```diff
@@ -124,3 +127,32 @@ const stream = (file: File, {
...
+            new TextEncoder().encode(Chunk.join(chunk, "\n") + "\n")
+          )),
+        scope
+      )
+    )
+
+    yield* _(Effect.yieldNow())
+
+    return Logger.make((options) => {
+      const formatted = Logger.stringLogger.log(options)
```
tim-smart, line 155: I think the logger should be passed in rather than assume the format.

And maybe move to a PlatformLogger module.

tim-smart (changes requested):

packages/platform/src/internal/fileSystem.ts
jessekelly881: Updated. I kind of like having a `format?: (_: Logger.Logger.Options<unknown>) => string` more than passing in a Logger though with the option to pass in for example `{ format: Logger.stringLogger.log }`. Its more flexible.

tim-smart: I generally prefer composition over options if possible. I had a play around and I think something like this is nice: https://github.com/Effect-TS/effect/commit/ed0d513fb5c8f14d588553cbe8197e684a69e60b#diff-8f9519c9651eac0a7c450530c155b62fe4f248e54db12f7ac0c62fb90b1be189R5

tim-smart: Also found a bug that needs addressing - if the program shuts down logs can be dropped. You can see the test case in the branch above.

jessekelly881: Ah yeah, that's definitely better. Composition is a lot nicer in this case.

tim-smart: > Also found a bug that needs addressing - if the program shuts down logs can be dropped. You can see the test case in the branch above.

Switching to a runtime based approach fixes the issue: https://github.com/Effect-TS/effect/compare/file-logger-composition#diff-a5d9b91049b63b5c55abf4414eea8eacd153c692a23a3e03e9bfb11b4d7ef686R36

jessekelly881: Oh nice. It seems to work well. It's even simpler w/o the queue. I wonder what would happen if logs came in faster than they could be written though.

tim-smart (approved):

jessekelly881:
@tim-smart committed the log.txt file :P

PR #2277, opened 2024-03-11
refactor: Unify ReadonlyArray.isNonEmptyArray and isNonEmptyReadonlyArray as isNonEmpty

patroza:
addresses: https://discord.com/channels/795981131316985866/1216002676064784455/1216358773258457118


patroza:
So this is a problem with lifting, like Option.liftPredicate, and probably the reason why it was implemented separately.

PR #2280, opened 2024-03-11
added support for PromiseLike

jessekelly881:
Extends Promise apis to use PromiseLike instead.


mikearnaldi:
What's the point of `resolve`, can't we simply call `then`? also please don't use Merge Commits, it is safer to rebase the branch

packages/effect/src/internal/core-effect.ts
```diff
@@ -1171,17 +1171,16 @@ export const patchFiberRefs = (patch: FiberRefsPatch.FiberRefsPatch): Effect.Eff
   updateFiberRefs((fiberId, fiberRefs) => pipe(patch, fiberRefsPatch.patch(fiberId, fiberRefs)))
 
 /* @internal */
-export const promise = <A>(evaluate: (signal: AbortSignal) => Promise<A>): Effect.Effect<A> =>
+export const promise = <A>(evaluate: (signal: AbortSignal) => PromiseLike<A>): Effect.Effect<A> =>
   evaluate.length >= 1
     ? core.async((resolve, signal) => {
-      evaluate(signal)
+      Promise.resolve(evaluate(signal))
```
tim-smart, line 1177: Update everything to use .then? Then we don't need to wrap with Promise.resolve. Would be nice to handle everything consistently.

jessekelly881: https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExcTRsb2I4eWxpZzlvM2tjYTB4N2tmYmFpbXN1bWVmcjcxZjZ3dGU0aSZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/2scLMiWJNWhMkyjF4z/giphy.gif

:( Updated..

tim-smart: lol

tim-smart (approved):
:shipit:

PR #2285, opened 2024-03-12
add support for AbortSignal's to runPromise

tim-smart:


packages/effect/src/internal/runtime.ts
```diff
@@ -257,30 +257,39 @@ export const unsafeRunSyncExit =
...
       const op = fastPath(effect)
       if (op) {
         resolve(op)
       }
-      unsafeFork(runtime)(effect).addObserver((exit) => {
+      const fiber = unsafeFork(runtime)(effect)
+      fiber.addObserver((exit) => {
         resolve(exit)
       })
+      options?.signal?.addEventListener("abort", () => {
```
tim-smart, line 290: Add check to see if signal is already aborted.

PR #2290, opened 2024-03-12
Remove function renaming from internals, introduce new cutpoint strategy

mikearnaldi:


tim-smart (approved):
Looks good, and improves perf for sync effect decently:

image

Issue #1997, opened 2024-01-28
Schema: default class field constructors


patroza:
### What is the problem this feature would solve?

I often find myself needing to provide default values at construction, but not at parsing.
At parsing time, a missing value is often regarded as a bug; something in the storage or in flight is wrong, or perhaps there’s some backwards compatibility to deal with.

```ts
class Person extends S.Class<Person>()({
  name: S.string,
  createdAt: S.Date,
  posts: S.array(S.string)
}) {}
const person = new Person({ name: "Patrick", createdAt: new Date(), posts: [] })
```

### What is the feature you are proposing to solve the problem?

Add a combinator to enhance property descriptors for fields in class schemas, so that constructor defaults can be expressed.
```ts
class Person extends S.Class<Person>()({
  name: S.string,
  createdAt: S.Date.pipe(S.withDefaultMake(() => new Date())),
  posts: S.array(S.string).pipe(S.withDefaultMake(() => []))
}) {}
const person = new Person({ name: "Patrick" })
```

Naive implementation;
- combinator: https://github.com/patroza/effect/pull/1/commits/f8a0564f749cfc73f1aa529e1b167472225cfc54#diff-7da43f2b7b3dea1554f2627482ba41e4f885c908848b72704d9887ffcc065153R4675

- constructor: https://github.com/patroza/effect/pull/1/commits/f8a0564f749cfc73f1aa529e1b167472225cfc54#diff-7da43f2b7b3dea1554f2627482ba41e4f885c908848b72704d9887ffcc065153R4557

### What alternatives have you considered?

```ts
class Person extends S.Class<Person>()({
  name: S.string,
  createdAt: S.Date,
  posts: S.array(S.string)
}) {
  constructor(props: Omit<Person, "createdAt" | "posts" | keyof Data.Case> & { createdAt?: Date, posts?: string[] }, disableValidation?: boolean) {
    super({ createdAt: props.createdAt ?? new Date(), posts: props.date ?? [], ...posts} , disableValidation)
  }
}
```
but it is obviously tedious and it gets even more hairy if possibly all values have defaults and you would like to support `new Person()`.
It's also not re-usable.

mikearnaldi:
@gcanti what's your thought about this? we mentioned a few times the possibility of adding a generic to `Schema` to represent the type from construction, this can be useful in many other cases such as for example branded types & refinements, being able to have a validating constructor kind of completes the idea of defining types (data models) through schema.

[...]

PR #2319, opened 2024-03-14
feat: add `withDefaultConstructor`

patroza:
#1997

- the defaults are carried with on the fields so that they can be picked and transferred to build other classes

```ts
    it("supports defaults", () => {
      class A extends S.Class<A>("A")({
        a: S.string,
        b: S.number.pipe(S.withDefaultConstructor(() => 1)),
        c: S.number.pipe(S.withDefaultConstructor(() => 1), S.fromKey("d"))
      }) {}
      expect({ ...new A({ a: "abc" }) }).toStrictEqual({ a: "abc", b: 1, c: 1 })
      expect({ ...new A({ a: "abc", b: 2, c: 2 }) }).toStrictEqual({ a: "abc", b: 2, c: 2 })
    })
```
this is where defaults really matter 90% of my usecases.

### DIFF
--- a/.changeset/fair-items-tell.md
+++ b/.changeset/fair-items-tell.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+feat: add `withDefaultConstructor`
--- a/packages/schema/dtslint/Context.ts
+++ b/packages/schema/dtslint/Context.ts
@@ -164,21 +164,21 @@ S.NonEmptyArray(aContext)
 // propertySignatureDeclaration
 // ---------------------------------------------
 
-// $ExpectType PropertySignature<":", string, never, ":", string, "aContext">
+// $ExpectType PropertySignature<":", string, never, ":", string, false, "aContext">
 S.propertySignature(aContext)
 
 // ---------------------------------------------
 // optionalToRequired
 // ---------------------------------------------
 
-// $ExpectType PropertySignature<":", string, never, "?:", string, "aContext">
+// $ExpectType PropertySignature<":", string, never, "?:", string, false, "aContext">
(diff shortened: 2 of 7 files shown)


steffanek:
@patroza would be nice to have 2 types of default:
- One that can be overwritten by the caller, like you did,
- but also another one that can't be overwritten.
Would that make sense?

patroza:
> @patroza would be nice to have 2 types of default:
> 
> * One that can be overwritten by the caller, like you did,
> * but also another one that can't be overwritten.
> 
> Would that make sense?

If you want a constant just put it on the class as a prop?

steffanek:
> > @patroza would be nice to have 2 types of default:
> > 
> > * One that can be overwritten by the caller, like you did,
> > * but also another one that can't be overwritten.
> > 
> > Would that make sense?
> 
> If you want a constant just put it on the class as a prop?

Not really, cuz the constant would be available for the entire class and I only need it for the construction.

patroza:
> Not really, cuz the constant would be available for the entire class and I only need it for the construction.

I'm very confused. maybe you can paint the use case with example code.

steffanek:
> > Not really, cuz the constant would be available for the entire class and I only need it for the construction.
> 
> I'm very confused. maybe you can paint the use case with example code.

I like to have an attribute to represent the Event, (ex: "PersonCreated" for a DomainStartEvent). It's only needed for the construction, but I don't want the caller to be able to change it.

patroza:
@steffanek use `TaggedClass`

steffanek:
> @steffanek use `TaggedClass` ?

nah I don't want the event to be an object by itself. I come up with a different domain modeling approach.

gcanti:
Closing in favour of https://github.com/Effect-TS/effect/pull/2634

PR #2335, opened 2024-03-15
added Number.negate, Schema.negate

jessekelly881:
Adds Number.negate. Schema.negate

### DIFF
--- a/packages/schema/src/Schema.ts
+++ b/packages/schema/src/Schema.ts
@@ -3779,21 +3779,24 @@ export const clamp =
     )
 
 /**
+ * @category api interface
+ * @since 1.0.0
+ */
+export interface Negate extends Annotable<Negate, number> {}
+
+/**
  * Negates a `Number`.
  *
  * @category number transformations
  * @since 1.0.0
  */
-export const negate = <A extends number, I, R>(
-  self: Schema<A, I, R>
-): Schema<A, I, R> =>
-  transform(
-    self,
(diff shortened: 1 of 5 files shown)


packages/schema/src/Schema.ts
```diff
@@ -3778,6 +3778,23 @@ export const clamp =
...
+ * @category number transformations
+ * @since 1.0.0
+ */
+export const negate = <A extends number, I, R>(
+  self: Schema<A, I, R>
+): Schema<A, I, R> =>
+  transform(
+    self,
+    typeSchema(self),
+    (self) => N.negate(self),
```
gcanti, line 3793: this could simply be `N.negate`

gcanti: The `A extends ...` pattern is useful when defining filters, this could simply be:
```ts
export const Negate = S.transform(S.number, S.number, N.negate, N.negate)
```

However, I wonder whether it's worth adding this kind of schemas. There are **many** functions from which you can derive a transformation (`negate`, `Math.abs`, `increment`, etc.), but we can't include all of them.

jessekelly881: The reason I thought it would be useful in this case is that there is already a bigDecimalNegate and negate is bidireccional negate(negate(a)) = a. But you're right, there are too many fns that could possibly be transformations. I'm not sure where we should draw the line.

PR #2370, opened 2024-03-20
Schema template literal regex escape special characters

leonitousconforti:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Literals inside template literals were not escaped, so when creating the regex for the overall template literal it might not find a match correctly. This solution is based on [Lodash's escapeRegExp](https://lodash.com/docs#escapeRegExp).

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/angry-coins-notice.md
+++ b/.changeset/angry-coins-notice.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+Template literal regex escape special characters
--- a/packages/schema/src/AST.ts
+++ b/packages/schema/src/AST.ts
@@ -1819,15 +1819,21 @@ export const keyof = (ast: AST): AST => Union.unify(_keyof(ast))
  * @since 1.0.0
  */
 export const getTemplateLiteralRegExp = (ast: TemplateLiteral): RegExp => {
-  let pattern = `^${ast.head}`
+  const reRegExpChar = /[\\^$.*+?()[\]{}|]/g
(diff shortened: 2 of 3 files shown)


gcanti (commented):
Thanks @leonitousconforti maybe we could extract a `escapeSpecialChars` function, WDYT?

```ts
const specialCharsRegex = /[.*+?^${}()|[\]\\]/g

const escapeSpecialChars = (s: string): string =>
  specialCharsRegex.test(s) ?
    s.replace(specialCharsRegex, "\\$&") // $& means the whole matched string
    : s

const STRING_KEYWORD_PATTERN = ".*"
const NUMBER_KEYWORD_PATTERN = "[+-]?\\d*\\.?\\d+(?:[Ee][+-]?\\d+)?"

/**
 * @since 1.0.0
 */
export const getTemplateLiteralRegExp = (ast: TemplateLiteral): RegExp => {
  let pattern = `^${escapeSpecialChars(ast.head)}`

  for (const span of ast.spans) {
    if (isStringKeyword(span.type)) {
      pattern += STRING_KEYWORD_PATTERN
    } else if (isNumberKeyword(span.type)) {
      pattern += NUMBER_KEYWORD_PATTERN
    }
    pattern += escapeSpecialChars(span.literal)
  }

  pattern += "$"
  return new RegExp(pattern)
}
```

gcanti (approved):

PR #2385, opened 2024-03-22
update typescript to 5.4

tim-smart:


tim-smart:
@gcanti Looks like there is an issue with Schema.tuple

gcanti:
@tim-smart wow... it seems that there has been a significant change in the behavior of the built-in type `Required` when it comes to tuples

from

```ts
// ts 5.3
type A = readonly [string, ...number[], boolean]
type B = Required<A> // readonly [string, ...(number | boolean)[], number | boolean]
```

to

```ts
// ts 5.4
type A = readonly [string, ...number[], boolean]
type B = Required<A> // readonly [string, ...number[], boolean]
```

tim-smart:
The typescript 5.3 behaviour seems strange, unless I'm missing something.

gcanti:
The 5.4 behaviour seems definitely more correct, probably a bugfix, all previous versions behaved differently, and I had to implement the "wrong" behaviour to keep runtime and type-level aligned. I think we should change the runtime to align with version 5.4

gcanti (approved):

tim-smart:
This will mean mis-alignment for previous typescript versions? I wonder if this should be in the next minor release.

gcanti:
yes, but in any case, we have a mis-alignment, either with the previous versions or with version 5.4. However, the behaviour of version 5.4 is the "correct" one, so I would prioritize it and treat it as a bugfix

PR #2392, opened 2024-03-22
passthrough vitest context to test factory

AMar4enko:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This will passthrough vitest context into test factory function, so test-scoped assertions can be used:
```
it.effect(`Global expect can't be used here`, ({ expect }) =>  Effect.gen(function* ($) {
   expect.assertions(1)
   // Code that does condition-based assertions
})
```
If using expect.assertions from global vitest scope, test result above might be interfered by other tests, using expect.assertions, running simultaneously

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/tender-cooks-scream.md
+++ b/.changeset/tender-cooks-scream.md
@@ -1,0 +1,5 @@
+---
+"@effect/vitest": patch
+---
+
+Passthrough vitest context to test factory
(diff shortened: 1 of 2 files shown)


mikearnaldi:
it's unfortunately not safe to discriminate `Effect` and function

PR #2396, opened 2024-03-23
fix: Schema.extend with nested union

sukovanej:
Started failing probably after the union flattening removal.

### DIFF
--- a/packages/schema/src/Schema.ts
+++ b/packages/schema/src/Schema.ts
@@ -2335,7 +2335,11 @@ const intersectUnionMembers = (
   AST.Union.make(
     xs.flatMap((x) => {
       return ys.map((y) => {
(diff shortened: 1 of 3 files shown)


gcanti (approved):

packages/schema/src/Schema.ts
```diff
@@ -2335,7 +2335,11 @@ const intersectUnionMembers = (
   AST.Union.make(
     xs.flatMap((x) => {
       return ys.map((y) => {
-        if (AST.isTypeLiteral(x)) {
+        if (AST.isUnion(x)) {
+          return intersectUnionMembers(x.types, [y], path)
```
gcanti, line 2339: Notice that this will create sub-unions, we could un-nest them, see https://github.com/Effect-TS/effect/pull/2398

PR #2407, opened 2024-03-25
Add Config.duration

thewilkybarkid:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I'd like to have an envvar containing a value like `10 seconds` turned into a `Duration`.

(A `DurationInput` refinement would work too.)

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Duration.ts
+++ b/packages/effect/src/Duration.ts
@@ -131,6 +131,57 @@ export const decode = (input: DurationInput): Duration => {
   throw new Error("Invalid duration input")
 }
 
+/**
+ * @since 2.5.0
+ */
+export const decodeUnknown = (input: unknown): Option.Option<Duration> => {
+  if (isDuration(input)) {
+    return Option.some(input)
+  } else if (isNumber(input)) {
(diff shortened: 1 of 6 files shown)


packages/effect/src/Duration.ts
```diff
@@ -131,6 +131,57 @@ export const decode = (input: DurationInput): Duration => {
...
+    return Option.some(input)
+  } else if (isNumber(input)) {
+    return Option.some(millis(input))
+  } else if (isBigInt(input)) {
+    return Option.some(nanos(input))
+  } else if (Array.isArray(input)) {
+    if (input.length === 2 && isNumber(input[0]) && isNumber(input[1])) {
+      return Option.some(nanos(BigInt(input[0]) * bigint1e9 + BigInt(input[1])))
+    }
+  } else if (typeof input === "string") {
```
tim-smart, line 148: I think we can just add this check to `decode`, then use Option.liftThrowable.

PR #2415, opened 2024-03-26
add Iterable module

tim-smart:
I guess the question is, if this isn't needed due to Stream.


mikearnaldi:
They kind of solve independent problems so I'd be in favour of adding the module

PR #2430, opened 2024-03-28
Compare with zod

### LINKED ISSUE #2406: Comparison with zod
### What is the type of issue?

Something else

### What is the issue?

Would be helpful for newbies like me to document the differences between effect schema and zod (perhaps others too); [this comparison page](https://zod.dev/?id=comparison) could be a useful example. I can make a PR if someone helps me understand.

### Where did you find it?

_No response_

rosano:
I might have put this at the beginning of the README, but to be safer opted for just before the 'Documentation' heading.

## Type

- [x] Documentation Update

## Description

To help newbies like me understand the differences between effect schema and similar projects. [This comparison page](https://zod.dev/?id=comparison) could be a useful example for a more complete comparison.

## Related

Closes #2406.

### DIFF
--- a/packages/schema/README.md
+++ b/packages/schema/README.md
@@ -5432,6 +5432,19 @@ const result = S.decodeUnknownEither(Name)("").pipe(
 console.log(result); // => { _id: 'Either', _tag: 'Left', left: 'should be non empty' }
 ```
 
+# Comparison
+
+## Zod
+
+Feature-wise, `schema` can do practically everything that `zod` can do.
+
+The main differences are:
+
+1.  `schema` transformations are bidirectional, so it not only decodes like `zod` but also encodes.
+2.  `schema` is integrated with `Effect` and inherits some benefits from it (such as dependency tracking in transformations).
+3.  `schema` is highly customizable through annotations, allowing users to attach meta-information.
+4.  `schema` uses a functional programming style with combinators and transformations (while `zod` provides a chainable API).
+
 # Documentation
 
 - [API Reference](https://effect-ts.github.io/effect/docs/schema)


KhraksMamtsov:
Feature-wise, `zod` can do 50% that `schema` can do.

 with the previous wording, it seems that the scheme is in a weak position, but this is not the case.

packages/schema/README.md
```diff
@@ -5432,6 +5432,19 @@ const result = S.decodeUnknownEither(Name)("").pipe(
...
+## Zod
+
+Feature-wise, `schema` can do practically everything that `zod` can do.
+
+The main differences are:
+
+1.  `schema` transformations are bidirectional, so it not only decodes like `zod` but also encodes.
+2.  `schema` is integrated with `Effect` and inherits some benefits from it (such as dependency tracking in transformations).
+3.  `schema` is highly customizable through annotations, allowing users to attach meta-information.
+4.  `schema` uses a functional programming style with combinators and transformations (while `zod` provides a chainable API).
```
vecerek, line 5446: Avoid mentioning FP at any cost, IMHO 😅 

```suggestion
4.  `schema` provides many combinators and transformations making it highly composable (while `zod` provides a chainable API).
```

vecerek: But with my suggestion, I'm not sure how composability contrasts with the chainable API because you can compose stuff through chaining, too.

The downside is that chainable APIs are not tree-shakable because the methods are attached to the objects prototype itself. `schema` may result in much smaller bundle size compared to `zod`, which is something frontenders will appreciate. But this needs to be verified in practice, I haven't done that.

mikearnaldi: Given that schema will include the Effect runtime the smaller bundle size will only materialize in extreme use cases,  wouldn't make such argument. What we can safely say is that chainable APIs can't be extended due to lack of extension functions in JS / TS, pipeable APIs can be extended as they are just composition of functions

PR #2436, opened 2024-03-29
preserve jsdoc and cmd-click

### LINKED ISSUE #2431: From Discord: JSDoc and click-navigation lost when fields defined in Schema Struct or Class
# Summary
The conversation revolves around the challenge of preserving TypeScript (TS) comments and documentation (JSDoc) visibility when transitioning from a plain interface to a schema definition using the Effect-TS ecosystem. The participants discuss their experiences and attempts to maintain the visibility of comments in Visual Studio Code (VSCode) when using schemas for type definitions.

### Key Points:

1. **Initial Problem**: When defining configurations using a plain interface, VSCode successfully displays comments and documentation. However, this functionality seems to be lost when the configuration is rewritten using the `S.struct` schema approach.

2. **Manual Interface Definition**: One solution explored is manually defining the interface with comments and then creating a corresponding schema. This approach retains the documentation visibility in VSCode but requires additional effort to maintain both the interface and the schema.

3. **Effect-TS Schema Limitations**: It's noted that in the legacy Effect-TS schema, it was possible to hover over a property name to see its JSDoc and click on a property to jump to its definition within the schema. This functionality appears to be missing in the current version of Effect-TS Schema.

4. **Technical Investigation**: The conversation delves into the technical aspects of why the documentation visibility issue occurs. It's suggested that the problem might lie in how TypeScript maps the input to the type, particularly in the `Struct` namespace's `Type` type definition. Modifying this definition to simplify the mapping between fields and their types is proposed as a potential solution.

5. **Proposed Solution**: A code snippet is provided that modifies the `Type` type definition to potentially restore the JSDoc visibility and navigation functionality. This solution involves using `UnionToIntersection` and adjusting the mapping to handle optional properties correctly. However, it's acknowledged that this solution is incomplete and does not fully address the handling of TypeTokens.

### Key Takeaways:

- Transitioning from plain interfaces to schema definitions in Effect-TS can lead to the loss of documentation visibility in IDEs like VSCode.
- Manually defining interfaces alongside schemas is a workaround but may not be ideal due to the maintenance overhead.
- The issue may stem from how TypeScript handles type mappings in schemas, particularly with optional properties and TypeTokens.
- A potential solution involves modifying the internal type mappings within the Effect-TS Schema to better preserve the link between properties and their documentation.
- Further investigation and development are needed to fully resolve the issue and restore the desired functionality.

# Discord thread

https://discord.com/channels/795981131316985866/1222950531807576136

https://github.com/Effect-TS/effect/assets/42661/8b7d57e5-8939-4b46-a782-02d66afebfdf

patroza:
fixes #2431

### DIFF
--- a/packages/schema/dtslint/Schema.ts
+++ b/packages/schema/dtslint/Schema.ts
@@ -1685,7 +1685,7 @@ hole<ConstructorParameters<typeof VoidTaggedClass>>()
 export const StructTypeTest1 = <S extends S.Schema.Any>(
   input: S.Struct.Type<{ s: S }>
 ) => {
-  input // $ExpectType UnionToIntersection<S extends PropertySignature<"?:", any, PropertyKey, Token, any, unknown> | PropertySignature<"?:", any, PropertyKey, Token, never, unknown> | PropertySignature<"?:", never, PropertyKey, Token, any, unknown> | PropertySignature<"?:", never, PropertyKey, Token, never, unknown> ? { readonly s?: Type<S>; } : { readonly s: Type<S>; }>
+  input // $ExpectType Type<{ s: S }>
 }
 
 // $ExpectType {}
--- a/packages/schema/dtslint/Schema.ts
+++ b/packages/schema/dtslint/Schema.ts
@@ -949,7 +949,7 @@ S.required(
   S.struct({ a: S.optional(S.string, { exact: true }), b: S.optional(S.number, { exact: true }) })
 )
 
-// $ExpectType Schema<{ readonly b: number; readonly a: string; readonly c: number; }, { readonly b: string; readonly a: string; readonly c: string; }, never>
+// $ExpectType Schema<{ readonly a: string; readonly b: number; readonly c: number; }, { readonly b: string; readonly a: string; readonly c: string; }, never>
 S.required(
   S.struct({
     a: S.optional(S.string, { exact: true }),
@@ -1685,7 +1685,7 @@ hole<ConstructorParameters<typeof VoidTaggedClass>>()
 export const StructTypeTest1 = <S extends S.Schema.Any>(
   input: S.Struct.Type<{ s: S }>
 ) => {
-  input // $ExpectType Type<{ s: S; }>
+  input // $ExpectType UnionToIntersection<S extends PropertySignature<"?:", any, PropertyKey, Token, any, unknown> | PropertySignature<"?:", any, PropertyKey, Token, never, unknown> | PropertySignature<"?:", never, PropertyKey, Token, any, unknown> | PropertySignature<"?:", never, PropertyKey, Token, never, unknown> ? { readonly s?: Type<S>; } : { readonly s: Type<S>; }>
 }
 
 // $ExpectType {}
--- a/packages/schema/dtslint/Schema.ts
+++ b/packages/schema/dtslint/Schema.ts
@@ -1685,7 +1685,7 @@ hole<ConstructorParameters<typeof VoidTaggedClass>>()
 export const StructTypeTest1 = <S extends S.Schema.Any>(
   input: S.Struct.Type<{ s: S }>
 ) => {
-  input // $ExpectType Type<{ s: S }>
+  input // $ExpectType Type<{ s: S; }>
 }
 
 // $ExpectType {}
--- a/packages/schema/src/Schema.ts
+++ b/packages/schema/src/Schema.ts
@@ -1821,7 +1821,11 @@ export declare namespace Struct {
    */
   export type Type<F extends Fields> = Types.UnionToIntersection<
     {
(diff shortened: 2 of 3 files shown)


packages/schema/src/Schema.ts
```diff
@@ -1819,16 +1819,30 @@ export declare namespace Struct {
   /**
    * @since 1.0.0
    */
-  export type Type<F extends Fields> =
-    & { readonly [K in Exclude<keyof F, TypeTokenKeys<F>>]: Schema.Type<F[K]> }
-    & { readonly [K in TypeTokenKeys<F>]?: Schema.Type<F[K]> }
```
Andarist, line 1824: are those 2 just selecting/filtering keys from `F`? could they be replaced with `as`?

Andarist: Ill put up a PR soon with a different change in an attempt to address this

mikearnaldi: You mean https://github.com/Effect-TS/effect/pull/2435 ?

packages/schema/dtslint/Schema.ts
```diff
@@ -1685,7 +1685,7 @@ hole<ConstructorParameters<typeof VoidTaggedClass>>()
 export const StructTypeTest1 = <S extends S.Schema.Any>(
   input: S.Struct.Type<{ s: S }>
 ) => {
-  input // $ExpectType Type<{ s: S; }>
+  input // $ExpectType UnionToIntersection<S extends PropertySignature<"?:", any, PropertyKey, Token, any, unknown> | PropertySignature<"?:", any, PropertyKey, Token, never, unknown> | PropertySignature<"?:", never, PropertyKey, Token, any, unknown> | PropertySignature<"?:", never, PropertyKey, Token, never, unknown> ? { readonly s?: Type<S>; } : { readonly s: Type<S>; }>
```
patroza, line 1688: Only point of contention I think

packages/schema/src/Schema.ts
Andarist: Yeah, exactly. This should work fine here and is the most straightforward for the compiler to handle. I see that u had to had to suppress one error but that's likely fine (it's in an internal implementation anyway. I could investigate why this fails at some other time.

mikearnaldi: While I agree it would make it lose order of keys, not that it matters too much

mikearnaldi: Would be curious to know why it fails indeed

Andarist: order of keys? what do u mean?

mikearnaldi: it would put all optional keys at the end of the object, so `{ a: S.number, b: S.optional(S.number), c: S.number }` would become in type: `{ a: number; c: number; b?: number }` instead of `{ a: number; b?: number; c: number }`

patroza: This was already the case though, but I kinda do like that the new solution fixes that

mikearnaldi: I remember this was the reason I didn't go for `as` in the old schema

Andarist: I guess I just wouldn't be bothered with the key order here ;p You could try to preserve the order using something like this: [TS playground](https://www.typescriptlang.org/play?exactOptionalPropertyTypes=true#code/PTAEAEFMA8EMGMAuB5ADoglgewHawDYAKATlqpMYgJ4AqV5AzgFyiLECukAUNeaAEqQAjuwzFIAExJkGoALygA5ACNFXHvUigAIlgDKidgDMjAdQyIAFgeNGAPDQB880AG8uoUAG0A0qAw4oADWkFRYRqA0ALoA-CxeNL5RUQDcoCCslhiyAO5Y7PgSoJawAG5aiFigkPgYALYBsIhaAAYAPqDsOBKQRgGSLaBGpHWZWljKiLD9RbzcAL6gAGRuHulgYeyg8LCB8Fh1qOzNY6C1DIig4acAROIiYpI3oKikqLISVHgNO-j4VGtfP5AoIHuIpG8GFEWF0gjgsDkcCkuPNkRo+ABZKg2EwudyeYAAKkJoDyOVAhOAa1gLBw7DqygoyIJxNAdXY8EsFKpnmULAuxACAHNmelWbAcpAGActJS1vAWMosFh8JBdsjUeo5gIpQVLgpdDizBZrIYTHYsUbHGjtYIGHrpKgXHa9V4bsobqktZpQABhA5HZoOZwKVzePwBYKha7RFiJHxRUCLFaw+GIm0+gBijXwLvw+r9AeOkDshrNxqsRot2PLjmt6gyFmKsFkuC0MDYsG2ZTVl0sFAqlnyQssABpQOitNm8EQ3i5pwQ84g3R6vUA)

gcanti (approved):

PR #2438, opened 2024-03-29
Support Heterogeneous Effects in Effect Iterable apis

mikearnaldi:


mikearnaldi:
@fubhy any other signature you think we should support heterogeneous types?

fubhy:
> @fubhy any other signature you think we should support heterogeneous types?

My gut feeling is that ~all~ most of the `Effect.*` functions that work on iterables of Effect should work with heterogenous types. What's your thinking?

PR #2450, opened 2024-03-30
Platform: Auto-instrument HTTP client

### LINKED ISSUE #2449: Auto-instrument HTTP client

Currently, traces are missing HTTP client spans. With default instrumentation of the HTTP clients (the same way as the HTTP server requests are instrumented), the library users won't have to do this themselves.

Auto-instrumentation of all HTTP client requests.

_No response_

vecerek:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Auto-instruments all outbound HTTP requests similarly to how the HTTP middleware instruments all inbound HTTP requests.

## Related

- Related Issue #
- Closes #2449

### DIFF
--- a/packages/platform/src/internal/http/client.ts
+++ b/packages/platform/src/internal/http/client.ts
@@ -3,7 +3,9 @@ import type * as ParseResult from "@effect/schema/ParseResult"
 import * as Schema from "@effect/schema/Schema"
 import * as Context from "effect/Context"
 import * as Effect from "effect/Effect"
-import { dual, pipe } from "effect/Function"
+import * as FiberRef from "effect/FiberRef"
+import { constFalse, dual, identity, pipe } from "effect/Function"
+import { globalValue } from "effect/GlobalValue"
 import * as Layer from "effect/Layer"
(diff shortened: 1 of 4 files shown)


packages/platform/src/internal/http/client.ts
```diff
@@ -118,7 +140,13 @@ export const fetch = (options?: RequestInit): Client.Client.Default =>
                       reason: "Transport",
                       error: _
                     })
-                })
+                }).pipe(
+                  disabledWhen(request as any) ?
+                    identity :
+                    Effect.withSpan(`http ${request.method}`, {
```
vecerek, line 146: One thing to think about here: the name of the span is the same for both inbound and outbound HTTP requests. Should we change them to `http.server` and `http.client`, respectively, so they can be distinguished?

vecerek: cc @tim-smart

tim-smart: Went with `http.client`

Issue #2345, opened 2024-03-16
Effect.if evaluates onTrue when condition is false


alex-dixon:
### What version of Effect is running?

2.2.3

### What steps can reproduce the bug?

```typescript
import { Effect } from 'effect'

const f = (s: string) => Effect.succeed(s)

const foo: { bar?: { baz: string } } = {}

Effect.runSync(
  Effect.if('bar' in foo, {
    onTrue: f(foo.bar!.baz),
    onFalse: Effect.fail('oops'),
  })
)
```

### What is the expected behavior?

onTrue does not evaluate when truth test is false

### What do you see instead?

```
/repro/scratch.ts:9
    onTrue: f(foo.bar!.baz),
                       ^
TypeError: Cannot read properties of undefined (reading 'baz')
    at Object.<anonymous> (/repro/scratch.ts:9:24)
    at Module._compile (node:internal/modules/cjs/loader:1368:14)
    at Module.m._compile (/repro/node_modules/ts-node/src/index.ts:1618:23)
```

tim-smart:
You will need to wrap it with Effect.suspend

mikearnaldi:
> You will need to wrap it with Effect.suspend

might be worth considering if conditional branches should be lazy, while the fact that the branch would evaluate is pretty obvious from JS eager evaluation semantics the golden path should be not to evaluate the branch

tim-smart:
Closed by #2451

PR #2451, opened 2024-03-31
use LazyArg for Effect.if branches

tim-smart:


.changeset/three-poems-beg.md
```diff
@@ -0,0 +1,23 @@
+---
+"effect": minor
+---
+
+use LazyArg for Effect.if branches
```
vecerek, line 5: Just out of curiosity, what's the rationale behind this change?

mikearnaldi: See https://github.com/Effect-TS/effect/issues/2345

PR #2452, opened 2024-03-31
support variadic arguments in Effect.log

tim-smart:


mikearnaldi:
not sure about the interface of logger containing a cause imho, the idea would be to make the default loggers be:

```ts
/**
 * @since 2.0.0
 * @category constructors
 */
export const defaultLogger: Logger<unknown[], void> = fiberRuntime.defaultLogger
```

and be able to pass a variadic number of arguments of which some can be a Cause

PR #2455, opened 2024-04-02
move non-streaming rpc apis to seperate module

tim-smart:


mikearnaldi:
why?

tim-smart:
> why?

I felt the *Effect naming was a bit too ambiguous, so seeing if I prefer *NonStreaming. Feels too long though, so still thinking about it.

PR #2466, opened 2024-04-03
make Logger.structuredLogger default for browser

jessekelly881:
Logger.structuredLogger makes a better default for the browser. I wonder if maybe this is too opinionated though.

### DIFF
--- a/packages/platform-browser/src/internal/runtime.ts
+++ b/packages/platform-browser/src/internal/runtime.ts
@@ -2,14 +2,28 @@ import type { RunMain } from "@effect/platform/Runtime"
 import * as Cause from "effect/Cause"
 import * as Effect from "effect/Effect"
 import * as Logger from "effect/Logger"
+import * as Layer from "effect/Layer"
+import * as HashSet from "effect/HashSet"
+import * as FiberRef from "effect/FiberRef"
+
+/** @internal */
+const useStructuredLogger = Layer.fiberRefLocallyScopedWith(FiberRef.currentLoggers, (loggers) =>  {
+  if(HashSet.has(loggers, Logger.defaultLogger)) {
+    const set = HashSet.remove(loggers, Logger.defaultLogger);
+    return HashSet.add(set, Logger.structuredLogger)
+  }
+
+  return loggers;
+})
+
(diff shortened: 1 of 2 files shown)


tim-smart:
I'm thinking it is a bit too opinionated as it is, as Logger.replace adds the logger regardless (even if the default logger has already been replaced).

I think if we only replace the logger if the default one is present then it might be ok.

tim-smart:
> Oh I wasn't aware of that. Replacing only the default logger and not if it was overridden was my intention. Hmm. What's the correct way to do that then? By adding a `Logger.get` and `Logger.Equivalence` and testing `Logger.get(Logger.defaultLogger)` against `Logger.defaultLogger` to see if it has been overwritten?

Layer.locallyWith that uses a HashSet.has check before replacing the default logger.

jessekelly881:
Something like this? This is the same as doing Logger.replace though, no? I'm probably missing something jaja

```ts
  const setDefaultLogger = Layer.fiberRefLocallyScopedWith(FiberRef.currentLoggers, (loggers) =>  {
    if(HashSet.has(loggers, Logger.defaultLogger)) {
      return loggers;
    }
    const set = HashSet.remove(Logger.defaultLogger);
    return HashSet.add(set as any, Logger.structuredLogger)
  })
```

packages/platform-browser/src/internal/runtime.ts
```diff
@@ -2,14 +2,28 @@ import type { RunMain } from "@effect/platform/Runtime"
...
+
 
 /** @internal */
 export const runMain: RunMain = (
   effect,
   options
 ) => {
-  const _effect = Effect.provide(effect, Logger.structured)
-
+  const _effect = Effect.provide(effect, useStructuredLogger)
```
tim-smart, line 25: I think instead of using a Layer, we should just use `Effect.locallyWith`

jessekelly881: Updated

tim-smart: I'll hand it over to eslint for the rest of the review :)

PR #2472, opened 2024-04-04
add Readable & Subscribable modules

tim-smart:


jessekelly881:
`Subscribable.unwrap :: Effect<Subscribable> -> Subscribable`?

mikearnaldi:
Maybe a `Readable` with the `get` and `Subscribable` as a sub-type of `Readable`?

tim-smart:
I was thinking about it, but Readable just seems like `Effect<A>`.

mikearnaldi:
> I was thinking about it, but Readable just seems like `Effect<A>`.

if we want Ref & co to implement it it needs to be a `{ get }`

jessekelly881:
Effect could probably implement Readable as well. `get get(){ return this }` (so could Option and Either but there are probably circularity issues)

tim-smart:
> Effect could probably implement Readable as well. `get get(){ return this }` (same with Option and Either). I wonder if there might not be circularity issues though..

Hmm doesn't make too much sense to me, as Readable / Ref is supposed to represent a handle to some mutable value. Effect represents a computation.

PR #2477, opened 2024-04-05
add PlatformConfigProvider module

tim-smart:


packages/platform/src/PlatformConfigProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+              Effect.catchIf(isPlatformError, handlePlatformError(pathSegments))
+            ),
+          patch: PathPatch.empty
+        })
+      )
+    })
+  )
+
+/**
+ * Add the file tree ConfigProvider to the environment, falling back to the default ConfigProvider.
```
vecerek, line 83: Just a nit: I think reading configuration from `process.env` is a more common scenario. If the config is not found there, that's when it should fall back to the file system, not the other way around. Checking if the value exists in `process.env` is also faster than checking the file system for the existence of a file.

tim-smart: Yeah I think we can do that.

PR #2494, opened 2024-04-09
Add Duration.divide

thewilkybarkid:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds `divide` to the `Duration` module (it already has `times`).

~Unlike `Number.divide` and `BigInt.divide`, this doesn't have safe/unsafe variants.~

## Related

- Related Issue #
- Closes #


thewilkybarkid:
Looking to replace https://github.com/PREreview/coar-notify/blob/958faa16e9bf2fa4aee536c1d66d7ba4d5c36c89/src/BullMq.ts#L135-L136.

Could be achieved with

```ts
Duration.times(value, 0.5)
```

but a division API would be more straightforward.

jessekelly881:
Maybe also a `Duration.map`? 

```ts
Duration.map("30 seconds", {
  onMillis: (val: number) => val / 2,
  onNanos: (val: bigint) => val / 2n
})
```

The problem with `Duration.times` is that the Duration could be a number of ns(bigint) or ms(number) and any fn like Duration.times would have to handle both cases.

packages/effect/src/Duration.ts
```diff
@@ -515,6 +515,29 @@ export const clamp: {
...
+  (by: number): (self: DurationInput) => Duration
+  (self: DurationInput, by: number): Duration
+} = dual(
+  2,
+  (self: DurationInput, by: number): Duration =>
+    match(self, {
+      onMillis: (millis) => make(millis / by),
+      onNanos: (nanos) => {
+        if (isNaN(by) || by < 0 || Object.is(by, -0)) {
+          return make(0)
```
tim-smart, line 532: You can use the `zero` and `infinity` constants here

tim-smart:
Hmm I think we might need a safe variant, and call this unsafeDivide.

thewilkybarkid:
@tim-smart I've had a stab at that. I've followed the `Number`/`BigInt` modules in making it different between what returns a `None` and what throws.

packages/effect/src/Duration.ts
```diff
@@ -515,6 +515,59 @@ export const clamp: {
...
+ * @since 2.4.19
+ * @category math
+ */
+export const divide: {
+  (by: number): (self: DurationInput) => Option.Option<Duration>
+  (self: DurationInput, by: number): Option.Option<Duration>
+} = dual(
+  2,
+  (self: DurationInput, by: number): Option.Option<Duration> =>
+    match(self, {
```
tim-smart, line 528: You can simplify this by wrapping unsafeDivide with a try catch.

thewilkybarkid: The results are a bit different (https://github.com/Effect-TS/effect/pull/2494#issuecomment-2047454353).

packages/effect/test/Duration.test.ts
```diff
@@ -162,6 +162,48 @@ describe("Duration", () => {
     }))
   })
 
+  it("divide", () => {
+    expect(Duration.divide(Duration.minutes(1), 2)).toEqual(Option.some(Duration.seconds(30)))
+    expect(Duration.divide(Duration.seconds(1), 3)).toEqual(Option.some(Duration.nanos(333333333n)))
+    expect(Duration.divide(Duration.nanos(2n), 2)).toEqual(Option.some(Duration.nanos(1n)))
+    expect(Duration.divide(Duration.nanos(1n), 3)).toEqual(Option.some(Duration.zero))
```
thewilkybarkid, line 169: Seems like a _very_-edge case, but I wonder if this should be a `none`?

tim-smart: I think this is ok.

tim-smart (approved):

Issue #2417, opened 2024-03-26
Schema: consolidate schema names


mikearnaldi:
We currently have:

```ts
Schema.number
```

and

```ts
Schema.NumberFromString
```

We should decide on a single casing across the package

cc @fubhy

gcanti:
**Proposal**: I think the most readable casing is the same as that used in TypeScript, so for example, all schemas related to the TypeScript type `number` should start with "number", all schemas related to the type `Date` should start with "Date", etc...

Regarding the type constructors, the same applies with the modification that `<` and `>` are replaced by `(` and `)`.

Example

```ts
import * as S from "@effect/schema/Schema"

const schema = S.asSchema(S.struct({
  a: S.string, // good
  b: S.number, // good
  c: S.bigint, // good,
  d: S.bigintFromSelf, // good,
  e: S.Date, // good,

  f: S.NumberFromString, // bad, should be `numberFromString`,
  g: S.array(S.string), // bad, should be `S.ReadonlyArray(S.string)`,
  h: S.record(S.string, S.number), // bad, should be S.Record(S.string, S.number)
  i: S.either({ left: S.string, right: S.number }), // bad, should be S.Either(S.number, S.string)
  l: S.option(S.string), // bad, should be S.Option(S.string)
  m: S.chunk(S.number) // bad, should be S.Chunk(S.string)
}))
```

mikearnaldi:
not sure I'd follow what TS does in this case, it doesn't really make sense it seems more a historical inconsistency

mikearnaldi:
Proposal:

```ts
import * as S from "@effect/schema/Schema"

const schema = S.asSchema(S.Struct({
  a: S.String,
  b: S.Number,
  c: S.BigInt,
  d: S.BigIntFromSelf,
  e: S.Date,

  f: S.NumberFromString,
  g: S.Array(S.String),
  h: S.Record(S.String, S.Number),
  i: S.Either({ left: S.String, right: S.Number }),
  l: S.Option(S.String),
  m: S.Chunk(S.Number)
}))
```

This way an UpperCamelCase term (constant) denotes a schema

mikearnaldi:
> Can we also fix the schema names that currently use `Bigint` to all be `BigInt` :-) (sorry, my OCD ...)
> 
> I'm talking about `clampBigint` => `clampBigInt` etc.

yes pls :)

patroza:
[...]

PR #2497, opened 2024-04-09
Schema: consolidate schema names

gcanti:
RE https://github.com/Effect-TS/effect/issues/2417


.changeset/old-rats-remember.md
```diff
@@ -32,3 +32,107 @@ ArrayFormatter
...
+- rename `hashMap` to `HashMap`
+  - rename `key`, `value` arguments to `Key`, `Value`
+- rename `listFromSelf` to `ListFromSelf`
+- rename `list` to `List`
+- rename `sortedSetFromSelf` to `SortedSetFromSelf`
+- rename `sortedSet` to `SortedSet`
+
+?
+
+- `encodedSchema`
```
TylorS, line 118: Just my 2 cents, PascalCase for all of the data types, to mirror TypeScript types conventions is a win and they are what I've always wanted as they will never shadow any runtime variables by most JS/TS conventions being camelCase. They're also nouns in my mind, they are definitely "things", like `User` or `Profile`.

I think the verbs listed below continue to make sense as camelCase as they represent "doing" something. If we were to follow my logic though, some of this list below aren't really verbs, so prefixing them may make sense to make them verbs to represent the active transformations or getters e.g. `encodedSchema` might make more sense as `getEncodedSchema` or similar. `partial` isn't really a verb either, so maybe `asPartial` or `toPartial`.

I'll try to come back and make more explicit suggestions when I have more time to write something if it'd be helpful

fubhy: Prefixing everything with `to`, `get` and `as` would be weird imho. Schemas are fundamentally a declarative API and `get` doesn't work well within that context imho.

gcanti: Note: I'll go through and make changes following these simple rules:
- (constructors) everything that builds a brand new Schema -> PascalCase (TypeSchema, EncodedSchema, Partial, Omit, Tuple)
- (modifiers) CamelCase (filer, annotations, message, asSchema)
- (PropertySignature constructors) CamelCase (optional, optionalToRequired, propertySignature, fromKey)

so at least we have a more consistent foundation to reason about

fubhy (approved):
As discussed: Big thumbs up from me. Just needs the adjustments in README, test names & test files accordingly, etc.

gcanti:
> test names & test files accordingly

no rush for these, I'll handle it in another PR

gcanti:
Note: here's a codemod for this new minor release: https://github.com/Effect-TS/codemod/pull/21

.changeset/old-rats-remember.md
TylorS: @fubhy FWIW I believe declarative APIs should read as close to human language as possible such that the user doesn't need to understand the inner-workings to know what it means, a mirror to the technical goals of a declarative API - to hide the inner workings, to focus on "what" and not "how". The API should convey all of its meaning using natural language, something we all share by default. You and I don't communicate what we do or did without a verb, why should our code be any different?

mikearnaldi:
```ts
type PartialPerson = Partial<Person>
```

This is plain TS, it isn't `type PartialPerson = GetPartial<Person>`

The distinction in this case is that some types are finished and some other types require parameters to be provided

TylorS: I'd use your own words from the corresponding issue here - https://github.com/Effect-TS/effect/issues/2417#issuecomment-2025964530
> not sure I'd follow what TS does in this case, it doesn't really make sense it seems more a historical inconsistency

patroza:
can't say i'm a fan of:
```
- rename `pluck` to `Pluck`
- rename `pick` to `Pick`
- rename `omit` to `Omit`
- rename `encodedSchema` to `EncodedSchema`
- rename `typeSchema` to `TypeSchema`
- rename `pickLiteral` to `PickLiteral`
- rename `fromBrand` to `FromBrand`
- rename `partial` to `Partial`
- rename `required` to `Required`
- rename `mutable` to `Mutable`
- rename `split` to `Split`
- rename `transformLiteral` to `TransformLiteral`
- rename `transformLiterals` to `TransformLiterals`
- rename `attachPropertySignature` to `AttachPropertySignature`
- rename `compose` to `Compose`
- rename `parseJson` to `ParseJson`
``

patroza:
in terms of consistency, I think `Array` should be `ReadonlyArray` and `NonEmptyArray` should be `NonEmptyReadonlyArray`

.changeset/old-rats-remember.md
mikearnaldi: That's not just typescript it's standard of type constructors, `Option<A>` not `GetOption<A>`

patroza:
I like the distinction of an accompanying type.
I would say PascalCase for all schemas and container types, and for nothing else. Certainly not for compose, or even function option argument keys (right/left, key/value) - there's no reason here to break camelCase convention.

I like tim's definition:

> I think the line is anything that directly represents a type.
> S.Struct -> a struct
> S.Exit -> a Exit
> S.compose -> doesn't represent a type, so in the combinator bin

.changeset/old-rats-remember.md
TylorS: There's a difference in a type constructor and a transformation though. In case of `Option<A>` is placed _within_ another type with semantics it should be `Option<A>`, like a value-level constructor function/class would. It created a "thing", a noun with semantics of its own. 

`Partial<A>` takes some `A` and computes an output, like a value-level function would. It's literally a for-loop at the type-level.

datner: @TylorS 

> the technical goals of a declarative API - to hide the inner workings, to focus on "what" and not "how".

and

> `Partial<A>` takes some `A` and computes an output, like a value-level function would. It's literally a for-loop at the type-level. [therefore it should be `getPartial`]

are contradictory statements 🤔 
_What_ it is, is a `Partial<User>`, _how_ it does it is by computing it and `GetPartial<User>` (read: get partial from user)

I agree with the overall sentiment, and there is an obvious distinction between `Partial<A>` and an `Option<A>`, but I believe that the difference is heuristic and not semantic

TylorS: "What" I want is to compute a partial from some type, "how" is through mapped types, but as a user I only care "what" I'm getting back, and not the syntax or choices made to compute the "what". In the case of `type PartialUser = Partial<User>`, which it is to me at least, is `PartialUser`, what was returned. `Partial<User>` itself is a description of a specific computation.

I think this is entirely symmetrical with how declarative functions work. We just use `Effect.andThen`, but we shouldn't need to know how it works beneath the hood. "What" we want to do is chain together a sequence of Effects, "how" it does it is quite a few steps starting with checks on all the options, and then building the correct sequences of Effects.

PR #2498, opened 2024-04-09
Added {Readable,Subscribable}.unwrap

jessekelly881:
I found myself needing these.

### DIFF
--- a/packages/effect/src/Readable.ts
+++ b/packages/effect/src/Readable.ts
@@ -85,6 +85,8 @@ export const mapEffect: {
  * @category constructors
  */
 export const unwrap = <A, E, R, E1, R1>(
-  effect: Effect<Readable<A, E, R>, E1, R1>): Readable<A, E | E1, R | R1> => make(
-    core.flatMap(effect, ({ get }) => get),
+  effect: Effect<Readable<A, E, R>, E1, R1>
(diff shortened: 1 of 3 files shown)


packages/effect/src/Readable.ts
```diff
@@ -79,3 +79,14 @@ export const mapEffect: {
...
+
+/**
+ * @since 2.0.0
+ * @category constructors
+ */
+export const unwrap = <A, E, R, E1, R1>(
+  effect: Effect<Readable<A, E, R>, E1, R1>
+): Readable<A, E | E1, R | R1> =>
+  make(
+    core.flatMap(effect, ({ get }) => get)
```
tim-smart, line 91: Avoid destructuring in case `get` is dependant on `this`.

tim-smart (changes requested):

packages/effect/src/Readable.ts
jessekelly881: Ah gotcha.

tim-smart (commented):
Just need a changeset now

jessekelly881:
I left it out because the Subscribable and Readable modules haven't been published yet. I'll add one though.

tim-smart (approved):

PR #2509, opened 2024-04-12
WIP: added Subscribable.fromStream

jessekelly881:
## Type

- [x] Feature

## Description

Adds Subscribable.fromStream. The idea is that the Subscribable keeps track of the last value of the Stream(or the defaultValue if the stream hasn't yet emitted a value). I honestly doubt that this is the best implementation for this fn though.. :(

### DIFF
--- a/.changeset/bright-drinks-run.md
+++ b/.changeset/bright-drinks-run.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+added Subscribable.fromStream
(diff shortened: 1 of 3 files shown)


tim-smart:
What's the use case for this?

jessekelly881:
I'm using this for periodically polling an api w/ `Option.none` as the default.

jessekelly881:
I'm using this for periodically polling an api w/ `Option.none` as the default.

tim-smart:
I think in that case using a SubscriptionRef is a better fit.

PR #2511, opened 2024-04-13
make `AST.pick` correctly handle key renames

gcanti:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Closes #2510

### DIFF
--- a/.changeset/swift-bottles-kick.md
+++ b/.changeset/swift-bottles-kick.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": minor
+---
+
+make `AST.pick` correctly handle key renames
--- a/packages/schema/dtslint/Schema.ts
+++ b/packages/schema/dtslint/Schema.ts
@@ -834,6 +834,9 @@ S.struct({ a: S.string.pipe(S.optional({ exact: true, nullable: true, as: "Optio
 // pick
 // ---------------------------------------------
 
+// @ts-expect-error
+pipe(S.struct({ a: S.propertySignature(S.number).pipe(S.fromKey("c")) }), S.pick("a"))
+
 // $ExpectType Schema<{ readonly a: string; readonly b: number; }, { readonly a: string; readonly b: number; }, never>
 pipe(S.struct({ a: S.string, b: S.number, c: S.boolean }), S.pick("a", "b"))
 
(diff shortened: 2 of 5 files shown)


nspaeth:
It looks like using the new name for the key is explicitly dis-allowed.  Is the reasoning for this that there is no way to determine which key the new key was renamed from in the type signature?

I think that in almost every case, if a developer has chosen to rename a key they expect to refer to it that way from then on, so this is really confusing. if after seeing that it doesn't typecheck and they have to use the old name, I expect that they are likely to be suspicious that it is a bug rather than an intentional descision because it is so counter-intuitive.

```typescript
const someSchema = S.struct({
  a: S.propertySignature(S.string).pipe(S.fromKey('b')),
  b: S.number,
})

const schema = pipe(
  someSchema,
  S.omit('a') // <-- Seems like the expected usage
)
```

gcanti:
this schema is illegal and raises a `Duplicate property signature "b"` error
```ts
const someSchema = S.struct({
  a: S.propertySignature(S.string).pipe(S.fromKey("b")),
  b: S.number
})
```

gcanti:
let's change `S.fromKey("b")` to `S.fromKey("c")`

```ts

const opaque = S.struct({
  a: S.propertySignature(S.string).pipe(S.fromKey("c")),
  b: S.number
})

/*
const schema: S.Schema<{
    readonly a: string;
    readonly b: number;
}, {
    readonly c: string;
    readonly b: number;
}, never>
*/
const schema = S.asSchema(opaque)

const p = schema.pipe(S.omit("a"))
```

what do you expect `p` to look like?

nspaeth:
> let's change `S.fromKey("b")` to `S.fromKey("c")`
Yes, sorry. Made an error when simplifying from a local test.

> what do you expect `p` to look like?
I think I'd expect:
```typescript
const p = schema.pipe(S.omit("a"))
type p = S.Schema<{ readonly b: number; }, { readonly b: number; }, never>

const q = schema.pipe(S.pick("a"))
type q = S.Schema<{ readonly a: number; }, { readonly c: number; }, never>
```

gcanti:
That would be possible at runtime (since the underlying `ast` still has all the information), but it's a problem at the type-level as there's nothing anymore linking the `a` prop with the `c` prop.

nspaeth:
I wonder if it is possible to create some kind of sentinal type:

```typescript
type opaque = S.Schema<{
  readonly a: string;
  readonly b: number;
}, {
  readonly c: Renamed<'c', 'a'>;
  readonly b: number;
}, never>


// S.Schema.Encoded<opaque>
type opaqueEncoded = {
  readonly c: string;
  readonly b: number;
}

// S.Schema.Type<opaque>
type opaqueEncoded = {
  readonly a: string;
  readonly b: number;
}
```

gcanti:
yeah it might be possible, but it complicates things and I'm not sure it's worth it, when working with structs I would suggest the following workaround
```ts
import { omit, pick } from "effect/Struct"

const schema = S.struct({
  a: S.propertySignature(S.string).pipe(S.fromKey("c")),
  b: S.number
})

const p = S.struct(omit(schema.fields, "a"))

const q = S.struct(pick(schema.fields, "a"))
```

PR #2513, opened 2024-04-14
added Stream.fromEventListener

jessekelly881:
Added a helper to convert `EventTarget.addEventListener` to a stream and added Platform specific implementations. 

```ts
export const isOnline = Stream.merge(
	fromEventListenerWindow("online").pipe(Stream.map(() => true)),
	fromEventListenerWindow("offline").pipe(Stream.map(() => false))
);


```

### DIFF
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -4555,3 +4555,13 @@ export const decodeText: {
  * @category encoding
  */
 export const encodeText: <E, R>(self: Stream<string, E, R>) => Stream<Uint8Array, E, R> = internal.encodeText
+
+/**
+ * Creates a `Stream` using addEventListener.
+ * @since 1.0.0
+ */
+export const fromEventListener: (
+  target: EventTarget,
+  type: string,
+  options?: boolean | Omit<AddEventListenerOptions, "signal">
+) => Stream<Event> = internal.fromEventListener
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -4560,7 +4560,7 @@ export const encodeText: <E, R>(self: Stream<string, E, R>) => Stream<Uint8Array
  * Creates a `Stream` using addEventListener.
  * @since 3.1.0
  */
-export const fromEventListener: <A extends Event = Event>(
+export const fromEventListener: <A = Event>(
   target: EventTarget,
   type: string,
   options?: boolean | Omit<AddEventListenerOptions, "signal">
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -4558,10 +4558,10 @@ export const encodeText: <E, R>(self: Stream<string, E, R>) => Stream<Uint8Array
 
 /**
  * Creates a `Stream` using addEventListener.
- * @since 1.0.0
+ * @since 3.1.0
  */
-export const fromEventListener: (
+export const fromEventListener: <A extends Event = Event>(
   target: EventTarget,
   type: string,
   options?: boolean | Omit<AddEventListenerOptions, "signal">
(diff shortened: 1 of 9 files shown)


packages/effect/src/internal/stream.ts
```diff
@@ -8045,3 +8045,17 @@ export const encodeText = <E, R>(self: Stream.Stream<string, E, R>): Stream.Stre
...
+export const fromEventListener = (
+  target: EventTarget,
+  type: string,
+  options?: boolean | Omit<AddEventListenerOptions, "signal">
+): Stream.Stream<Event> =>
+  _async<Event>((emit) => {
+    function cb(e: Event) {
+      emit.single(e)
+    }
+    target.addEventListener(type, cb, options)
```
tim-smart, line 8059: Might be a good idea to batch the events:

```suggestion
    let batch: Array<Event> = []
    let taskRunning = false
    function cb(e: Event) {
      batch.push(e)
      if (!taskRunning) {
        taskRunning = true
        queueMicrotask(() => {
          const events = batch
          batch = []
          taskRunning = false
          emit.chunk(Chunk.unsafeFromArray(events))
        })
      }
    }
    target.addEventListener(type, cb, options)
```

jessekelly881: Hmm. Testing this, it doesn't seem to be actually batching the events. It's just emitting a new chunk for each event.

jessekelly881: Do you think it would make sense to use a `window:Duration` for batching?

tim-smart: No I think this is fine :)

packages/effect/src/Stream.ts
```diff
@@ -4555,3 +4555,13 @@ export const decodeText: {
  * @category encoding
  */
 export const encodeText: <E, R>(self: Stream<string, E, R>) => Stream<Uint8Array, E, R> = internal.encodeText
+
+/**
+ * Creates a `Stream` using addEventListener.
+ * @since 1.0.0
```
tim-smart, line 4561: `@since 3.1.0`

It will also require a minor changeset for "effect".

tim-smart (changes requested):

packages/effect/src/internal/stream.ts
```diff
@@ -8045,3 +8045,28 @@ export const encodeText = <E, R>(self: Stream.Stream<string, E, R>): Stream.Stre
     const encoder = new TextEncoder()
     return map(self, (s) => encoder.encode(s))
   })
+
+/** @internal */
+export const fromEventListener = (
```
tim-smart, line 8050: I think we can add a generic here for overriding the event type, with a default of `Event`.

tim-smart: You can then use that to replace the as any casts in platform.

jessekelly881:
Updated! :)

packages/effect/src/internal/stream.ts
jessekelly881: Ah yeah, that's better. :)

packages/effect/src/Stream.ts
```diff
@@ -4558,10 +4558,10 @@ export const encodeText: <E, R>(self: Stream<string, E, R>) => Stream<Uint8Array
 
 /**
  * Creates a `Stream` using addEventListener.
- * @since 1.0.0
+ * @since 3.1.0
  */
-export const fromEventListener: (
+export const fromEventListener: <A extends Event = Event>(
```
tim-smart, line 4563:
```suggestion
export const fromEventListener: <A = Event>(
```

Just in case someone wants to use something non-standard.

tim-smart:
You can run `pnpm codegen` to fix the build issue.

tim-smart (approved):

PR #2535, opened 2024-04-16
Use tsconfig.build.json between packages

fubhy:


packages/platform/src/Template.ts
```diff
@@ -2,6 +2,7 @@
  * @since 1.0.0
  */
 import * as Effect from "effect/Effect"
+// TODO: This should not be importing internal types.
 import type * as core from "effect/internal/core"
```
fubhy, line 6: Importing internal types here currently only works because we are using the `src` tsconfigs within the monorepo between projects atm. This is... fine but not great and not very portable. It could also lead to problems for consumers if we ever accidently expose any of these types.

I'd suggest we don't do this and lock down our types also between projects within the monorepo.

fubhy: ./cc @tim-smart

tim-smart: Feel free to replace the `as core.Primitive` with `as any`

PR #2546, opened 2024-04-17
merge revert runtime flag operations to prevent interruption gaps

tim-smart:


mikearnaldi:
Looks good, maybe instead of using Mutable we can change the type in RevertFlags

PR #2547, opened 2024-04-17
Add Effect.makeAbortSignal

jessekelly881:
Exports existing makeAbortSignal

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -5392,3 +5392,27 @@ export const Tag: <const Id extends string>(id: Id) => <Self, Type>() =>
     })
     return done
   }
+  
+/**
+ * Creates an `AbortSignal`.
+ * @since 1.0.0
+ */
+export const makeAbortSignal: Effect<AbortSignal> = map(
+  tap(
+    sync(() => new AbortController()), 
+    controller => onInterrupt(() => sync(() => controller.abort())
+  )),
+  (_) => _.signal
+) 
+
+/**
+ * Creates an `AbortSignal` that is automatically aborted when the Scope is closed.
+ * @since 1.0.0
+ */
+export const makeAbortSignalScoped: Effect<AbortSignal, never, Scope.Scope> = map(
+  acquireRelease(
+    sync(() => new AbortController()),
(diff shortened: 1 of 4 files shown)


packages/effect/src/Effect.ts
```diff
@@ -5392,3 +5392,27 @@ export const Tag: <const Id extends string>(id: Id) => <Self, Type>() =>
...
   }
+
+/**
+ * Creates an `AbortSignal`.
+ * @since 3.1.0
+ */
+export const makeAbortSignal: Effect<AbortSignal> = map(
+  tap(
+    sync(() => new AbortController()),
+    (controller) => onInterrupt(() => sync(() => controller.abort()))
```
tim-smart, line 5403: This doesn't make sense, as the onInterrupt effect is only applied regionally / to the effect it wraps.

packages/effect/src/Effect.ts
```diff
@@ -5392,3 +5392,27 @@ export const Tag: <const Id extends string>(id: Id) => <Self, Type>() =>
...
+  ),
+  (_) => _.signal
+)
+
+/**
+ * Creates an `AbortSignal` that is automatically aborted when the Scope is closed.
+ * @since 3.1.0
+ */
+export const makeAbortSignalScoped: Effect<AbortSignal, never, Scope.Scope> = map(
```
tim-smart, line 5412: Hmm will think about it. There aren't many cases where this is useful - usually the signals passed to tryPromise / async are all you need.

packages/effect/src/Effect.ts
jessekelly881: Oh interesting. Ok. I'll remove this.

packages/effect/src/Effect.ts
jessekelly881: Typically async and  promise are enough. But there have been a few circumstances where I've needed a signal inside of an Effect.gen for example and using async/promise didn't make sense.

tim-smart:
Considering this would no longer be used internally, I don't think we need this addition.

If you feel strongly about adding it, feel free to re-open to discuss.

PR #2551, opened 2024-04-17
Array: refactor `fromIterable` to accept an optional mapping function…

### LINKED ISSUE #2538: Array fromIterable mapfn

Without that, people must write `globalThis.Array` or rename Effect import to use Array.from with mapfn.

Array.from has a second mapfn arg, and because Effect Array.fromIterable uses native Array.from, it should mimic its API imho.

_No response_

gcanti:
… for transformation, closes #2538

## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Closes #2538

### DIFF
--- a/packages/effect/src/Array.ts
+++ b/packages/effect/src/Array.ts
@@ -119,8 +119,11 @@ export const replicate: {
  * @category constructors
  * @since 2.0.0
  */
-export const fromIterable = <A>(collection: Iterable<A>): Array<A> =>
-  Array.isArray(collection) ? collection : Array.from(collection)
+export function fromIterable<A, B>(collection: Iterable<A>, f: (a: A, i: number) => B): Array<B>
+export function fromIterable<A>(collection: Iterable<A>): Array<A>
+export function fromIterable<A>(collection: Iterable<A>, f?: (a: A, i: number) => A): Array<A> {
+  return f ? Array.from(collection, f) : Array.isArray(collection) ? collection : Array.from(collection)
+}
 
 /**
  * Takes a record and returns an array of tuples containing its keys and values.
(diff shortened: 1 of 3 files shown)


packages/effect/src/Array.ts
```diff
@@ -119,8 +119,11 @@ export const replicate: {
  * @category constructors
  * @since 2.0.0
  */
-export const fromIterable = <A>(collection: Iterable<A>): Array<A> =>
-  Array.isArray(collection) ? collection : Array.from(collection)
+export function fromIterable<A, B>(collection: Iterable<A>, f: (a: A, i: number) => B): Array<B>
+export function fromIterable<A>(collection: Iterable<A>): Array<A>
+export function fromIterable<A>(collection: Iterable<A>, f?: (a: A, i: number) => A): Array<A> {
+  return f ? Array.from(collection, f) : Array.isArray(collection) ? collection : Array.from(collection)
+}
```
fubhy, line 126: This is imho a breaking change because it add a new position parameter that can cause problems with tacit usage.

fubhy (changes requested):

packages/effect/src/Array.ts
gcanti: Yes, I haven't changed the test to show that it's a breaking change, I think we should close this and tag the corresponding issue with a "next-major" label.

fubhy: Agreed

mikearnaldi: I don't think we generally support tacit usage

PR #2555, opened 2024-04-18
prevent use of `Array` as import name to solve bundler issues

tim-smart:


steida:
@tim-smart Which bundlers? Should I rename Array in Evolu, too? Maybe it should be documented. Thank you.

mikearnaldi:
> @tim-smart Which bundlers? Should I rename Array in Evolu, too? Maybe it should be documented. Thank you.

Metro, so react-native

PR #2573, opened 2024-04-19
Prefix `pnpm` calls in root `package.json` scripts with `corepack`

nickrttn:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

While working on #2572, I ran into errors running several scripts in the root `package.json`:

```sh
$ pnpm dtslint

> @ dtslint /Users/nick/dev/effect
> pnpm --recursive --parallel run dtslint

sh: pnpm: command not found
ELIFECYCLE  Command failed.
```

I fixed the issue by prefixing `pnpm` calls in `package.json` scripts with `corepack ...` to make sure `corepack` resolves `pnpm`'s binary. 

This is likely a quirk of my environment which I manage using Nix + Home Manager, but the changes potentially help others as well.

Then, I locked the version of `pnpm` specified in `packageManager` to the exact hash of the version, to make sure everyone's `corepack` uses that exact version. Set/upgrade using `corepack use pnpm@<semver>`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,26 +1,26 @@
 {
   "private": true,
   "type": "module",
-  "packageManager": "pnpm@8.15.1",
+  "packageManager": "pnpm@8.15.1+sha256.245fe901f8e7fa8782d7f17d32b6a83995e2ae03984cb5b62b8949bfdc27c7b5",
   "workspaces": [
     "packages/*"
   ],
   "scripts": {
     "clean": "node scripts/clean.mjs",
-    "codegen": "pnpm --recursive --parallel run codegen",
-    "build": "tsc -b tsconfig.build.json && pnpm --recursive --parallel run build",
+    "codegen": "corepack pnpm --recursive --parallel run codegen",
+    "build": "tsc -b tsconfig.build.json && corepack pnpm --recursive --parallel run build",
     "circular": "node scripts/circular.mjs",
     "test": "vitest",
     "coverage": "vitest --coverage",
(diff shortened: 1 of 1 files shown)


fubhy:
Most (all?!) of the ppl on the team use Nix & Home Manager too but (to my knowledge) noone is currently experiencing this issue. Note that we are using `direnv` to set up a repository specific environment within this repo. With that, `pnpm` should be directly available and this change /should/ not be needed. Are you using `direnv`?

nickrttn:
I was not using `direnv`. I just enabled it and it works as expected now. Perhaps a note in the README would be useful?

fubhy:
> I was not using `direnv`. I just enabled it and it works as expected now. Perhaps a note in the README would be useful?

Yeah that makes sense. A quick addition to our CONTRIBUTING.md for existing Nix users. Basically not prompting every user to adopt Nix first but inviting existing Nix users to utilize our `direnv` setup. Do you want to add sth. like that here? https://github.com/EFfect-TS/.github

nickrttn:
Effect-TS/.github#4 does that work?

Does locking `pnpm` to a specific hash make sense to you or should we close this PR?

fubhy:
Yep! Asked for a review from @IMax153. Let's close this imho, I don't see a reason to lock the pnpm version like that.

fubhy:
Thanks!

PR #2597, opened 2024-04-22
add SpanKind to Tracer spans

patroza:


packages/effect/src/Tracer.ts
```diff
@@ -80,6 +81,25 @@ export interface ExternalSpan {
   readonly context: Context.Context<never>
 }
 
+/**
+ * @since 3.0.5
```
tim-smart, line 85:
```suggestion
 * @since 3.1.0
```

packages/experimental/src/DevTools/Client.ts
```diff
@@ -164,7 +164,7 @@ export const makeTracer: Effect.Effect<Tracer.Tracer, never, Client> = Effect.ge
 
   return Tracer.make({
     span(name, parent, context, links, startTime) {
-      const span = currentTracer.span(name, parent, context, links, startTime)
+      const span = currentTracer.span(name, parent, context, links, startTime, "internal")
```
tim-smart, line 167: This needs to use the span kind passed to it.

packages/opentelemetry/src/internal/tracer.ts
```diff
@@ -112,15 +122,16 @@ export const Tracer = Context.GenericTag<OtelApi.Tracer>("@effect/opentelemetry/
...
       return new OtelSpan(
         OtelApi.context,
         tracer,
         name,
         parent,
         context,
         links,
-        startTime
+        startTime,
+        kind ?? "internal"
```
tim-smart, line 134:
```suggestion
        kind
```

packages/effect/src/internal/tracer.ts
```diff
@@ -39,6 +39,7 @@ export class NativeSpan implements Tracer.Span {
   readonly spanId: string
   readonly traceId: string = "native"
   readonly sampled = true
+  readonly kind = "internal"
```
tim-smart, line 42: This can be moved to the constructor as a parameter.

tim-smart (changes requested):

tim-smart (approved):

PR #2598, opened 2024-04-22
schema next minor

gcanti:


packages/schema/src/Schema.ts
```diff
@@ -1466,7 +1438,8 @@ const propertySignatureAnnotations_ = (
         ast.type,
         ast.isOptional,
         ast.isReadonly,
-        { ...ast.annotations, ...annotations }
+        { ...ast.annotations, ...annotations },
+        undefined
```
patroza, line 1442: why not take the existing default? (there are 2 more occurrences of this)

packages/schema/src/Schema.ts
```diff
@@ -6219,11 +6204,13 @@ export interface Class<Self, Fields extends Struct.Fields, A, I, R, C, Inherited
     disableValidation?: boolean | undefined
```
patroza, line 6204: line 6203
> props: keyof C extends never ? void | {} : C,

props with defaults need to be subtracted, because you can create without props when the required ones have defaults.

packages/schema/src/Schema.ts
```diff
@@ -2023,111 +2086,136 @@ export interface TypeLiteral<
...
+    }
+
+    static fields = { ...fields }
+
+    static records = [...records] as Records
+
+    static make(
+      props: Types.Simplify<TypeLiteral.Constructor<Fields, Records>>
+    ): Types.Simplify<TypeLiteral.Type<Fields, Records>> {
+      return ParseResult.validateSync(this)({ ...getFieldsDefaults(fields), ...props as any })
```
patroza, line 2217: better, but I wonder how expensive this is getting for calls that do set them,
like they're lazy but are activated every time a new instance is created, regardless if they will be discarded immediately after. for Date objects it may become expensive.

gcanti:
Closing in favour of https://github.com/Effect-TS/effect/pull/2633

PR #2601, opened 2024-04-23
feat: multiple span processors and metric readers

patroza:


packages/opentelemetry/test/Tracer.test.ts
```diff
@@ -11,7 +11,7 @@ const TracingLive = NodeSdk.layer(Effect.sync(() => ({
   resource: {
     serviceName: "test"
   },
-  spanProcessor: new SimpleSpanProcessor(new InMemorySpanExporter())
+  spanProcessors: [new SimpleSpanProcessor(new InMemorySpanExporter())] as const
```
mikearnaldi, line 14: is this `as const` needed? it would be bad for a user to have to use `as const`

patroza: sadly it is needed in the Effect case, it works fine without effect:
```ts
const TracingLive = NodeSdk.layer(() => ({
  resource: {
    serviceName: "test"
  },
  spanProcessors: [new SimpleSpanProcessor(new InMemorySpanExporter())]
})))
```

the alternative is an explicit annotation;
```ts
const TracingLive = NodeSdk.layer(Effect.sync((): NodeSdk.Configuration => ({
  resource: {
    serviceName: "test"
  },
  spanProcessors: [new SimpleSpanProcessor(new InMemorySpanExporter())]
})))
```

we could also switch to normal array, we just need to then not only check for undefined, but also if the array is empty.

patroza: I've made some improvements, think it's best of both worlds.

tim-smart:
Made backwards compatible here: #2623

PR #2602, opened 2024-04-24
Repro for variance change

mikearnaldi:


packages/effect/src/Effect.ts
```diff
@@ -94,6 +94,14 @@ export interface Effect<out A, out E = never, out R = never> extends Effect.Vari
   readonly [Unify.typeSymbol]?: unknown
   readonly [Unify.unifySymbol]?: EffectUnify<this>
   readonly [Unify.ignoreSymbol]?: EffectUnifyIgnore
+
+  [Symbol.iterator](): EffectGenerator<Effect<A, E, R>>
+}
```
TylorS, line 99: I'm not at a laptop to test, but maybe `this` in place of `Effect<A, E, R>` would work?

mikearnaldi:
```ts
interface EffectGenerator<Eff extends Effect<any, any, any>> {
  next(): IteratorResult<Eff, Effect.Success<Eff>>
  return(value: Effect.Success<Eff>): IteratorResult<Eff, Effect.Success<Eff>>
  throw(e: any): IteratorResult<Eff, Effect.Success<Eff>>
  [Symbol.iterator](): EffectGenerator<Eff>
}
```

This is nuts but it may actually work...

packages/effect/src/Effect.ts
tim-smart: I tried both and it didn't made a difference 🤷🏼‍♂️

enricopolanski:
What a massive, massive PR. Kudos.

fubhy:
> /snapshot

Branch was already deleted here. Either need to do that in the `Version Packages` PR (after merge) or do it here before merging :-)

PR #2607, opened 2024-04-24
Propagete _tag to static side

mikearnaldi:
we might also want to extend variants such as Schema.TaggedError etc

### DIFF
--- a/packages/effect/src/Data.ts
+++ b/packages/effect/src/Data.ts
@@ -441,11 +441,15 @@ export const Error: new<A extends Record<string, any> = {}>(
  * @since 2.0.0
  * @category constructors
  */
-export const TaggedError = <Tag extends string>(tag: Tag): new<A extends Record<string, any> = {}>(
-  args: Types.Equals<A, {}> extends true ? void
-    : { readonly [P in keyof A as P extends "_tag" ? never : P]: A[P] }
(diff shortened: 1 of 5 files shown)


packages/effect/src/Data.ts
```diff
@@ -441,11 +441,15 @@ export const Error: new<A extends Record<string, any> = {}>(
...
-) => Cause.YieldableError & { readonly _tag: Tag } & Readonly<A> => {
+export const TaggedError = <Tag extends string>(tag: Tag): {
+  readonly _tag: Tag
+  new<A extends Record<string, any> = {}>(
+    args: Types.Equals<A, {}> extends true ? void
+      : { readonly [P in keyof A as P extends "_tag" ? never : P]: A[P] }
+  ): Cause.YieldableError & { readonly _tag: Tag } & Readonly<A>
+} => {
   class Base extends Error<{}> {
+    static _tag = tag
```
patroza, line 452: readonly ?

mikearnaldi: type is not exposed, it's the internal impl

patroza: I know, but look below: `readonly _tag = tag`
anyway it's a non issue as you said.

datner:
Any updates on this?

PR #2608, opened 2024-04-24
feat: foreach preserve non emptyness.

patroza:
refs #2606

### DIFF
--- a/packages/effect/dtslint/Effect.ts
+++ b/packages/effect/dtslint/Effect.ts
@@ -100,20 +100,20 @@ pipe(
 )
 
 // -------------------------------------------------------------------------------------
-// forEach - tuple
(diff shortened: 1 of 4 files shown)


packages/effect/dtslint/Effect.ts
```diff
@@ -4,6 +4,7 @@ import { pipe } from "effect/Function"
 import * as Option from "effect/Option"
 import * as Predicate from "effect/Predicate"
 import * as Schedule from "effect/Schedule"
+import type { NonEmptyReadonlyArray } from "../src/Array.js"
```
tim-smart, line 7:
```suggestion
import type { NonEmptyReadonlyArray } from "effect/Array"
```

PR #2610, opened 2024-04-24
rename and export platform file schemas

wmaurer:
Making `Http.multipart.fileSchema` externally available could be useful to users of `@effect/platform`

For example, if it were exported, I could create the following Schema: 

```ts
const SingleFileSchema = Schema.transform(
  Http.multipart.filesSchema.pipe(Schema.itemsCount(1)),
  Http.multipart.fileSchema,
  {
    decode: ([file]) => file,
    encode: (file) => [file]
  }
)
```
WDYT @tim-smart?

Second question: do you think it would be valuable to add `SingleFileSchema` into the `Http.multipart` module?

### DIFF
--- a/.changeset/hungry-ghosts-kiss.md
+++ b/.changeset/hungry-ghosts-kiss.md
@@ -1,0 +1,9 @@
+---
+"@effect/platform-node": minor
+"@effect/platform-bun": minor
+"@effect/platform": minor
+---
+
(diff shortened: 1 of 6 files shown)


tim-smart:
Yes I think a single file schema can be added here. Maybe capitalize them too and move to a minor version changeset.

wmaurer:
@tim-smart thanks for the feedback. I think it's good to go now.

PR #2618, opened 2024-04-25
add Effect.annotateLogsScoped

tim-smart:


packages/effect/src/Effect.ts
```diff
@@ -4437,6 +4437,28 @@ export const annotateLogs: {
   <A, E, R>(effect: Effect<A, E, R>, values: Record<string, unknown>): Effect<A, E, R>
 } = effect.annotateLogs
 
+/**
+ * Annotates each log with the specified log annotation(s), until the Scope is closed.
+ *
+ * @since 2.0.0
```
patroza, line 4443: Since when?

PR #2619, opened 2024-04-25
add Effect.scopedAsDisposable, for integrating Scope with the `using` keyword

tim-smart:


mikearnaldi:
maybe just `Effect.disposable` ?

mikearnaldi:
Ah but it doesn't wait for cleanup to complete, it uses runFork so this won't work as it would leak resources

mikearnaldi:
wondering if there is a way to let this "yield" from the dispose

tim-smart:
> wondering if there is a way to let this "yield" from the dispose

You can only have async finalizers if you use async / await unfortunately.

PR #2628, opened 2024-04-26
feat: `Types.DeepMutable`

juliusmarminge:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Creates a `DeepMutable` type helper that works recursively
```ts
// with old type
type ShallowMutable = Types.Mutable<{ 
  readonly a: string; 
  readonly b: readonly string[]
}> 
// { a: string; b: readonly string[] }

// with new deep helper
type DeepMutableStruct = Types.DeepMutable<{ 
  readonly a: string; 
  readonly b: readonly string[]
}> 
// { a: string; b: string[] }
```

## Related

- Related Issue #
- Closes #


packages/effect/src/Types.ts
```diff
@@ -170,6 +170,27 @@ export type Mutable<T> = {
...
+ *   readonly a: string;
+ *   readonly b: readonly string[]
+ * }>
+ * // { a: string; b: string[] }
+ *
+ * @since 3.1.0
+ * @category types
+ */
+export type DeepMutable<T> = {
+  -readonly [P in keyof T]: T[P] extends Array<infer U> ? Mutable<U>
```
tim-smart, line 189: This will only go 2 levels deep :) You will have to recurse into array and object members to get a true "deep" transformation.

TylorS: `T[P] extends Array<infer U> ? Mutable<U>` seems wrong as well, it's unnesting an array to its value and should be 

```ts
T[P] extends ReadonlyArray<infer U> ? Array<DeepMutable<U>>
````

because `ReadonlyArray` is not as subtype of `Array` and this will not match them. The example is matching on `extends object` instead, and it also shouldn't lose the array it is wrapped within.

I'd also say that this is missing clauses for `ReadonlySet`, and `ReadonlyMap` to cover all the native JS collection types for this to cut out my own copies of this type.

At the end of it, it'd probably look something like this 

```ts
type DeepMutable<T> = T extends ReadonlyMap<infer K, infer V> ? Map<DeepMutable<K>, DeepMutable<V>>
  : T extends ReadonlySet<infer V> ? Set<DeepMutable<V>>
  : T extends ReadonlyArray<infer V> ? Array<DeepMutable<V>>
  : [keyof T] extends [never] ? T
  : { -readonly [K in keyof T]: DeepMutable<T[K]> }
```

[TypeScript Playgroud](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAIhFgLIFdgEMBGAbCAeAKgHxQC8U+UEAHsBAHYAmAzlAEoRoMD2dWIiaMLgCWdAGYQATlADSAGiiiJ0gGqEAUFCgB+KAKFwEKdNjwzCCw0lSYcuNRq0AucpRr1mbDt14gAyhDAIuJSUGqaOlABQVbGtngOES4U1LSMLOycPHwAgpKSaCDBymGOkXkFRbE2pvaEZS4A2gDWECBcYuQAum5pno10EABuUl0RuvhJUADeUAC0kt7ZIFCNMop0UK3tnfhdLtUmdvhrXcQAvgDc6uoA9LdQAHIQTLQMUItZvlBcGABWEAAxsB1KBIOQ0ABzSEQBgqNBYZAvXA5YhkaYRT4+PhQAD66EhLleklEkOuWixyygQwRSJcmWxIAqhRRhGuVxu9y8Xz40VB4GgOVI3MZ0Vws0p3xpiIgyShMLhtORdGQAFsMFILmz+eCcnFTMLDvFWdc7g8Gct9DroAAhYUW3z6Ajy2HwmVMXDE0kWEXLMX4F2K924FXqzX1a5g236nCG+DWI54G3as1PF5vZkga1QADC9qWvkzzuhrqVHoDJaDSPLgbd1dwGC4XBwaDo9Qj2ZzMegZCNtRzKa5MC4LzoAHJgFA0GIJMCoGASarhMBhCMmNmAGJN4WNCJ9uxeuiQix7+PdkNqjWSE9afd4RvNjhtuSnow1OwYYSQ0TAE9da5AA)

juliusmarminge:
Made it fully recursive and also added tests. Should be good to go now

tim-smart (approved):
If you run `pnpm changeset` and add a minor changeset, we can get this merged!

Thanks.

PR #2633, opened 2024-04-26
Schema next minor 2

gcanti:


gcanti:
https://github.com/Effect-TS/effect/pull/2634

PR #2635, opened 2024-04-26
typeclass - record - create SemigroupUnion, MonoidUnion and SemigroupIntersection

FredericEspiau:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Creates `SemigroupUnion`, `MonoidUnion` and `SemigroupIntersection` for `Record`

I didn't create anything for `difference` as I didn't see the point
I didn't ceate a `Monoid` for `intersection` as I can't think of a logical empty element (I could do it with some kind of Proxy maybe ?)

## Related

- Related Issue #
- Closes #


gcanti (commented):
I'd suggest adding a `get` prefix to align it with the current convention, for example, `getSemigroupUnion`.

I wonder how useful an instance for `ReadonlyRecord<string | symbol, A>` is. Generally, it's much more common to work with `ReadonlyRecord<string, A>`. Furthermore, the current implementation does not work with records with symbolic keys (`Record.union` only works with keys of type string).

PR #2646, opened 2024-04-28
feat: @effect/vitest custom testers

sukovanej:
The MVP of https://github.com/Effect-TS/effect/pull/2629

### DIFF
--- a/.changeset/flat-lobsters-hunt.md
+++ b/.changeset/flat-lobsters-hunt.md
@@ -1,0 +1,5 @@
+---
+"@effect/vitest": minor
+---
+
+Introduce `addEqualityTesters` function which adds a custom vitest tester handling equality of `Option`, `Either` and `Exit` effect types.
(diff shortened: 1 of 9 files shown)


tim-smart (commented):
I wonder if it would be better to add a tester that checks for `Equal.isEqual` and uses `Equal.equals` if true.

sukovanej:
> I wonder if it would be better to add a tester that checks for Equal.isEqual and uses Equal.equals if true.
I'm exploring this option, the problem is the fallback on strict equality for objects not implementing `Equal`, e.g. chunk or list containing arbitrary objects.

mikearnaldi:
Wondering if we can make equal act structurally for plain objects with a regional flag

sukovanej:
While making the equals a little bit smarter by moving closer to the comparison by value absolutely feels like a good move, IMO we should still not break the equality logic of the vitest. Vitest (and I assume generally any testing lib) can have its own comparison logic and most importantly a mechanism for injecting a custom testers (users of Vitest can introduce their own equality for specific cases), therefore the best way to approach this seems to be to introduce an equality tester checking effect's data types "structurally" and than giving the control back to the equality logic of the testing lib for other nested values.

PR #2649, opened 2024-04-28
remove generator boxing by using a Proxy

tim-smart:


mikearnaldi:
Why would a proxy provide less boxing? Usually proxy is slower

tim-smart:
> Why would a proxy provide less boxing? Usually proxy is slower

By boxing, I mean returning a boxed type from the iterator (YieldWrap).

Perf wise, this performs better on jsc (bun) but worse on v8 (node) - so not sure it is worth it.

The only benefit is future proofing ourselves, if ts ever decides to let you define the behaviour of plain `yield`.

Node:
Screenshot 2024-04-29 at 9 46 03 AM

Bun:
Screenshot 2024-04-29 at 9 46 55 AM

tim-smart:
Bun:
image

Node:
image

mikearnaldi:
Not sure I understand how this prevents a loop though

tim-smart:
> Not sure I understand how this prevents a loop though

The iterators return a copy of the original Effect but with an empty Symbol.iterator.

tim-smart:
It is good to know it is possible though, if the day ever comes.

PR #2653, opened 2024-04-29
Consider type of failure in Cause equality

mikearnaldi:


sukovanej:
this one will still fail

```ts
expect(
  Equal.equals(
    Cause.parallel(Cause.fail("fail1"), Cause.die("fail2")),
    Cause.parallel(Cause.fail("fail2"), Cause.die("fail1"))
  )
).toBe(false)
```

PR #2654, opened 2024-04-29
Actually fix Cause equality

mikearnaldi:


mikearnaldi:
@sukovanej would you mind a review pls

PR #2655, opened 2024-04-29
fix typo in Effect.never documentation

btoo:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update


btoo:
> `Effect` might be better as well.

it doesn't seem like the rest of the documentation follows this convention, though I would support a follow-up PR that implements this suggestion

tim-smart (approved):
❤️

PR #2658, opened 2024-04-30
Chunk difference methods

RareBodhi:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a set of `difference` operations directly on the Chunk method. I also have a zipped stream use case, and can imagine other use cases getting chunk values from Map's etc. where a `differenceOption` function would be useful.

My initial implementation enforced the first "comparator" Chunk to be a `Option<Chunk<A>>` however I opted to separate the methods.

Please let me know if this is in accordance with the envisaged interface, or if you'd rather not split the option handling. For now my use case has less boilerplate when the first Chunk is Optional.

### DIFF
--- a/packages/effect/src/Chunk.ts
+++ b/packages/effect/src/Chunk.ts
@@ -1417,3 +1417,23 @@ export const difference: {
   2,
   <A>(self: Chunk<A>, that: Chunk<A>): Chunk<A> => unsafeFromArray(RA.difference(that, self))
 )
+
+/**
+ * Creates a `Chunk` of values not included in the other given `Chunk`.
+ * The order and references of result values are determined by the first `Chunk`.
+ *
+ * The first comparator `Chunk` may be wrapped in an `Option`. A `None` value is
+ * treated as an empty `Chunk` during comparison.
+ *
+ * This is primarily for convenience on zip operations, gets, or similar.
+ *
+ * @since 3.2.0
+ */
+export const differenceOption: {
+  <A>(that: Chunk<A>): (self: Option<Chunk<A>>) => Chunk<A>
+  <A>(self: Option<Chunk<A>>, that: Chunk<A>): Chunk<A>
+} = dual(
+  2,
+  <A>(self: Option<Chunk<A>>, that: Chunk<A>): Chunk<A> =>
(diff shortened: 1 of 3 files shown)


packages/effect/src/Chunk.ts
```diff
@@ -1387,3 +1387,53 @@ export const reduceRight: {
...
+ * @since 3.2.0
+ */
+export const differenceOption: {
+  <A>(that: Chunk<A>): (self: Option<Chunk<A>>) => Chunk<A>
+  <A>(self: Option<Chunk<A>>, that: Chunk<A>): Chunk<A>
+} = dual(
+  2,
+  <A>(self: Option<Chunk<A>>, that: Chunk<A>): Chunk<A> =>
+    unsafeFromArray(RA.difference(that, O.getOrElse(self, empty)))
+)
```
fubhy, line 1439: I'm always on the fence when it comes to adding additional permutations / combinations like these that can be easily composed in userland. It always raises the question "where do we stop?". E.g. here we could also add a variant for `Either` and `Exit`, etc.

Even though that is also true for the other two, I can see the value in those, but I would leave this one out.

fubhy (changes requested):

packages/effect/src/Chunk.ts
```diff
@@ -1387,3 +1387,33 @@ export const reduceRight: {
   <B, A>(b: B, f: (b: B, a: A, i: number) => B): (self: Chunk<A>) => B
   <A, B>(self: Chunk<A>, b: B, f: (b: B, a: A, i: number) => B): B
 } = RA.reduceRight
+
+/**
+ * Creates a `Chunk` of values not included in the other given `Chunk` using the provided `isEquivalent` function.
+ * The order and references of result values are determined by the first `Chunk`.
+ *
+ * @since 2.0.0
```
tim-smart, line 1395:
```suggestion
 * @since 3.2.0
```

tim-smart (changes requested):
Could you add a minor changeset?

tim-smart:
I think you forgot to add the file 😅

PR #2660, opened 2024-04-30
Replace Order {  min, max } with { gt, gte, lt, lte }

jessekelly881:
`Order.between({ minimum, maximum })` is not as flexable or as clear as `Order.between({ gt, gte, lt, lte })`. I'm just curious what people think before doing a complete refactor.


jessekelly881:
Closing

PR #2663, opened 2024-04-30
Schema / Improve DX for JSON Schema annotations by providing IDEs autocomplete

suddenlyGiovanni:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

DX improvements for JSON Schema annotations

- Add optional `format` property to `JsonSchema7String`.
- Enhance IDE autocompletion for `JSONSchemaAnnotation`

The `format` property on the string JSONSchema instances, as part of the spec, allows semantic validation for a range of values like dates, email addresses, IP Addresses, hostnames, etc., which conforms to RFC or external specifications.

The extension of the (AST) definition for `JSONSchemaAnnotation` allows for better autocompletion support in IDEs.
Note that these modifications do not have any impact on the type system.

TL;DR 👇:

```ts
const stringDate =
  <A extends string>(annotations?: S.Annotations.Filter<A>) =>
    <I, R>(self: S.Schema<A, I, R>) => {
      return self.pipe(
        S.filter(
          (maybeStringDate): maybeStringDate is A =>
            Number.isNaN(Date.parse(maybeStringDate))
              ? false
              : new Date(maybeStringDate).toISOString().slice(0, 10) === maybeStringDate,
          {
            description: 'a string that is a valid YYYY-MM-DD date',
            message: issue => `expected a sting date 'YYYY-MM-DD', got '${issue.actual}'`,
            jsonSchema: {
              // 🪄 now the IDE can autocomplete the jsonSchema... 🧚
              minLength: 10,
              maxLength: 10,
              // 🪄 it also knows about `format` property and its possible values `builtins & string` ... 🧚
              format: 'date',
              ...annotations?.jsonSchema
            },
            ...(annotations ? omit(annotations, 'jsonSchema') : {}),
          },
        ),
      )
    }

// as well as...
const Url = Schema.String.annotation({
  // [...]
  jsonSchema: {      // <- IDE autocompletion support
    format: 'uri',   // <- format values autocomplete
    maxLength: 2000,
    minLength: 1,
    //               ..etc
  },
})


```

https://github.com/Effect-TS/effect/assets/15946771/f28ad833-75e6-4ae6-bd18-cf4c1306ce18

https://github.com/Effect-TS/effect/assets/15946771/f8da8144-4be9-41fd-932f-577fb8afcc07


gcanti (commented):
In theory, the best type to use would be the one that describes the entire JSON Schema specification. The `JsonSchema7` type represents only a portion and was defined 1) to aid implementation and 2) suggest the form of the generated schema (excluding any custom annotations that the user may add)

suddenlyGiovanni:
> In theory, the best type to use would be the one that describes the entire JSON Schema specification. The `JsonSchema7` type represents only a portion and was defined 1) to aid implementation and 2) suggest the form of the generated schema (excluding any custom annotations that the user may add)

Are you referring to the `JsonSchema7Root`? 

```ts
// packages/schema/src/JSONSchema.ts

/**
 * @category model
 * @since 1.0.0
 */
export type JsonSchema7Root = JsonSchema7 & {
  $schema?: string
  $defs?: Record<string, JsonSchema7>
}
```


if so, the only one is located here...

```diff
// packages/schema/src/AST.ts

/**
 * @category annotations
 * @since 1.0.0
 */
export type JSONSchemaAnnotation =
  | object
- | (object & JsonSchema7)
+ | (object & JsonSchema7Root)

```

Is this correct 🤔?

gcanti:
> Are you referring to the JsonSchema7Root?

No, I mean a type that actually describes the JSON Schema specification, something like https://www.npmjs.com/package/@types/json-schema.

suddenlyGiovanni:
> > Are you referring to the JsonSchema7Root?
> 
> No, I mean a type that actually describes the JSON Schema specification, something like https://www.npmjs.com/package/@types/json-schema.

🤦 Ahh... now I get it. Thx 🙇.

But before I start adding any external dependencies to the project, I'd like to know the project policy on such a topic.

Naively, I'm leaning toward not adding any and instead copy-pasta 🍝 what is needed to a new module/file after deep review. This approach would mean that we would be responsible for its maintenance. But it could be argued that 
- JSONSchema v7 is set in stone and, therefore, it would not change
- It is only about type definition and no code
- It seems to be quite a core component/responsibility of the Schema package itself.

That said, these are only my considerations; happy to hear what's your take before jumping to the implementation.

gcanti:
Yeah, the reason I ended up leaving it as a generic `object` is that: 1) I didn't want to maintain the types, 2) I didn't want to add another dependency, 3) some validators like `ajv` support non-standard extensions, so I wouldn't have easily covered those anyway. In the end, I'd leave the responsibility to the end user who can always import the types from a package of their choice and type the annotations as they wish, but I'll think about it, thanks for the PR.

suddenlyGiovanni:
Had some spare time, and my OCD started to kick in, so I gave it a spin...

I have ported over the `JSONSchema7` interface from `@types/json-schema` and reverted the changes that were previously relaying on the old implementation.

This is the DX in the IDEs. 

https://github.com/Effect-TS/effect/assets/15946771/97b699c8-cda6-447b-b888-48aeb2c17558

PS1: Don't want to force it, just thought was a nice experience improvement. Please do whatever you see fit about it, no hard feelings, no ego 😉.
PS2: Might need to double-check the type inference as the schema definitions are recursive.  
PS3: Didn't know where to put the new interface, ended up putting it in the internals folder.
PS4: Used both types and interfaces. The main entry type is an opaque interface composed of multiple different ts naked types. might not be the best decision.

suddenlyGiovanni:
Yep, it sounds reasonable. 👍🏻 
Do you need support updating the docs? 
Should I convert this PR by dropping the code changes and adding the suggestion to the readme, OR would you prefer to close it as it stands and handle the suggestions in a separate PR?

🙇🏻 thx you for the review.

suddenlyGiovanni:
> Yes thanks, if you feel like creating a new PR for the documentation, I appreciate it.

Done it with #2726; Closing this one.

See you there 😉

PR #2673, opened 2024-05-01
capture stack trace for tracing spans

tim-smart:


mikearnaldi:
we should probably have a way of disabling capture

tim-smart:
You can set `captureStackTrace` to false to disable it.

mikearnaldi:
```
Error: ok
    at /Users/michaelarnaldi/repositories/effect/packages/effect/test/Effect/cause-rendering.test.ts:13:25
    at Generator.next (<anonymous>)
    at /Users/michaelarnaldi/repositories/effect/packages/effect/src/internal/core-effect.ts:789:28
    at spanA
        at /Users/michaelarnaldi/repositories/effect/packages/effect/test/Effect/cause-rendering.test.ts:12:18
    at spanB
        at /Users/michaelarnaldi/repositories/effect/packages/effect/test/Effect/cause-rendering.test.ts:11:16
```

This format while nice may break parsers, such as vitest, probably better to use:

```
Error: ok
    at /Users/michaelarnaldi/repositories/effect/packages/effect/test/Effect/cause-rendering.test.ts:13:25
    at Generator.next (<anonymous>)
    at /Users/michaelarnaldi/repositories/effect/packages/effect/src/internal/core-effect.ts:789:28
    at spanA (/Users/michaelarnaldi/repositories/effect/packages/effect/test/Effect/cause-rendering.test.ts:12:18)
    at spanB (/Users/michaelarnaldi/repositories/effect/packages/effect/test/Effect/cause-rendering.test.ts:11:16)
```

mikearnaldi:
> You can set `captureStackTrace` to false to disable it.

true, I didn't see it in the functionWithSpan equivalent

mikearnaldi:
Screenshot 2024-05-01 at 10 00 07

options should probably be optional

mikearnaldi:
Something is odd in the filtering algo, trying to hide the internal call point with custom, leads to this:

Screenshot 2024-05-01 at 10 10 14

mikearnaldi:
I mean trying to remove that line with internal:

Screenshot 2024-05-01 at 10 11 30

mikearnaldi:
Looks like there is an issue:

Screenshot 2024-05-02 at 09 56 29

mikearnaldi:
Fixed, the error rendering is still broken though:

Screenshot 2024-05-02 at 09 59 01

This contains a lot of internals and the last line format is incorrect:

`at myError (<anonymous> (/Users/michaelarnaldi/repositories/effect/scratchpad/ok.ts:13:19))`

Should be instead:

```
(FiberFailure) TooLarge
    at /Users/michaelarnaldi/repositories/effect/scratchpad/ok.ts:6:45
    at myError (/Users/michaelarnaldi/repositories/effect/scratchpad/ok.ts:13:19)
```

to be aligned with standard stack traces

mikearnaldi:
even more standard would be:

```
(FiberFailure) TooLarge
    at <anonymous> (/Users/michaelarnaldi/repositories/effect/scratchpad/ok.ts:6:45)
    at myError (/Users/michaelarnaldi/repositories/effect/scratchpad/ok.ts:13:19)
 ```

**Comment by tim-smart** [comment]
Better now:
image

I added `EffectPrimitive.commit` as a cutpoint. Could also change YieldableError to only capture 1 frame, but then it wouldn't work with other types of Error instances.

**Comment by mikearnaldi** [comment]
> Better now: image
> 
> I added `EffectPrimitive.commit` as a cutpoint. Could also change YieldableError to only capture 1 frame, but then it wouldn't work with other types of Error instances.

Not sure about `EffectPrimitive.commit` as cutpoint given that `EffectPrimitive` could be minified (class names are not guaranteed to remain), that's why `custom` uses `effect_instruction_i` as a cutpoint

**Comment by mikearnaldi** [comment]
Why not using the custom trick? We could have commit call internally `effect_instruction_i` and it would cut

**Comment by mikearnaldi** [comment]
If we allowed breaking we could just rename commit, this is kind of an in between till 4.0

**Comment by tim-smart** [comment]
For now I have added an intermediary step that adds a cut point. We could look to remove it in 4.0.

File: .changeset/lovely-frogs-scream.md [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+---
+
+add Effect.functionWithSpan
+
+Allows you to define an effectful function that is wrapped with a span.
+
+```ts
+import { Effect } from "effect"
+
+const getTodo = Effect.functionWithSpan({
```
patroza, line 12: why not `functionWithSpan(body, options)` ?

tim-smart: I tried that first, but it doesn't format very well with dprint.

PR #2674, opened 2024-05-01
add Effect.genFn, for turning a Effect.gen with arguments into a function

tim-smart:
Not so sure I like it though, as it kills composability (you can't just tack on a `.pipe()` after the definition.

### DIFF
--- a/.changeset/soft-pillows-divide.md
+++ b/.changeset/soft-pillows-divide.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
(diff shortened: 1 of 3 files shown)


mikearnaldi:
Also it feels very ad-hoc, why genFn and not any other constructor? imho not worth

datner:
+1
Besides it looking cooler than `const f = () => .gen` theres no clear reason for `const f = .genFn`.

mikearnaldi:
I guess than we all agree that this isn't a good choice, we might re-evaluate at a different point

PR #2688, opened 2024-05-02
fix: vitest custom tester with toMatchObject

sukovanej:
Currently, the usage of the custom tester breaks `expect(...).toMatchObject(...)`. Failing example:

```ts
    expect({ a: 1, b: 2 }).toMatchObject({ a: 1 })
```

The reason is that `toMatchObject` in vitest is implemented as an equality check with `subsetEquality` as a last custom tester. Our custom tester currently doesn't have any skip logic to allow continuation of the testers pipeline it's part of. Another problematic thing is that the `toMatchObject` is not commutative, and it uses equality which is commutative. Therefore in equal impl of Option, Either, ... the matching breaks when the arguments are passed to `Equal.equals` in an opposite order.

### DIFF
--- a/packages/schema/test/Schema/brand.test.ts
+++ b/packages/schema/test/Schema/brand.test.ts
@@ -110,6 +110,7 @@ describe("brand", () => {
       IntegerFromString(1.1)
       assert.fail("expected `IntegerFromString(1.1)` to throw an error")
     } catch (e) {
+      const ast = S.typeSchema(IntegerFromString).ast as AST.Refinement
       expect(e).toStrictEqual(
         Brand.error(
           `IntegerFromString
@@ -117,10 +118,10 @@ describe("brand", () => {
    └─ Expected IntegerFromString (an integer), actual 1.1`,
           ParseResult.parseError(
             new ParseResult.Refinement(
-              IntegerFromString.ast as any,
+              ast,
               1.1,
               "Predicate",
-              new ParseResult.Type(IntegerFromString.ast, 1.1)
+              new ParseResult.Type(ast, 1.1)
             )
           )
         )
--- a/.changeset/forty-lizards-tap.md
+++ b/.changeset/forty-lizards-tap.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Call Equal.equals internally in order inputs were passed.
--- a/.changeset/shiny-crabs-bake.md
+++ b/.changeset/shiny-crabs-bake.md
@@ -1,0 +1,5 @@
+---
+"@effect/vitest": patch
+---
+
+Fix usage of `toMatchObject`.
--- a/packages/effect/src/internal/either.ts
+++ b/packages/effect/src/internal/either.ts
@@ -34,7 +34,7 @@ const RightProto = Object.assign(Object.create(CommonProto), {
   _tag: "Right",
   _op: "Right",
   [Equal.symbol]<L, R>(this: Either.Right<L, R>, that: unknown): boolean {
-    return isEither(that) && isRight(that) && Equal.equals(that.right, this.right)
+    return isEither(that) && isRight(that) && Equal.equals(this.right, that.right)
   },
   [Hash.symbol]<L, R>(this: Either.Right<L, R>) {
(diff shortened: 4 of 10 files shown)


mikearnaldi:
Not sure why but this makes some tests fail, also not sure about the early escape of Equal

packages/schema/test/Schema/brand.test.ts
```diff
@@ -117,7 +117,10 @@ describe("brand", () => {
    └─ Expected IntegerFromString (an integer), actual 1.1`,
           ParseResult.parseError(
             new ParseResult.Refinement(
-              IntegerFromString.ast as any,
+              // FIXME: the `from` is actually a Number AST instead of a Transformation
+              // expected IntegerFromString.ast as any
+              // console.log(e[0].meta.error.ast)
+              Object.assign(IntegerFromString.ast, { from: S.Number.ast }) as any,
```
sukovanej, line 123: @gcanti could please check locally to make sure I'm not hallucinating? The `e[0].meta.error.ast` really seems to be a refinement from a number instead of a refinement from the `S.NumberFromString` transformation.

sukovanej:
> Not sure why but this makes some tests fail, also not sure about the early escape of Equal
It really looks like the assertion in the failing test is incorrect. I'm not sure why it'd occur just now tho.

packages/schema/test/Schema/brand.test.ts
gcanti: the test is wrong, IMO should be
```ts
  it("the constructor should throw on invalid values", () => {
    const IntegerFromString = S.NumberFromString.pipe(
      S.int({ identifier: "IntegerFromString" }),
      S.brand("Int")
    )
    expect(IntegerFromString(1)).toEqual(1)
    try {
      IntegerFromString(1.1)
      assert.fail("expected `IntegerFromString(1.1)` to throw an error")
    } catch (e: any) {
      const ast = S.typeSchema(IntegerFromString).ast as AST.Refinement
      expect(e).toStrictEqual(
        Brand.error(
          `IntegerFromString
└─ Predicate refinement failure
   └─ Expected IntegerFromString (an integer), actual 1.1`,
          ParseResult.parseError(new ParseResult.Refinement(ast, 1.1, "Predicate", new ParseResult.Type(ast, 1.1)))
        )
      )
    }
  })
```

sukovanej: perfect, that passes, thanks 👍 (updated)

sukovanej:
> also not sure about the early escape of Equal

Me neither, but I'm not aware of any better solution. Also, I have another example that will fail even with the proposed change.

```ts
  it("Data.struct", () => {
    const alice = Data.struct({ name: "Alice", age: 30 })

    expect(alice).toMatchObject(Data.struct({ name: "Alice" }))
  })
```

mikearnaldi:
> > also not sure about the early escape of Equal
> 
> Me neither, but I'm not aware of any better solution. Also, I have another example that will fail even with the proposed change.
> 
> ```ts
>   it("Data.struct", () => {
>     const alice = Data.struct({ name: "Alice", age: 30 })
> 
>     expect(alice).toMatchObject(Data.struct({ name: "Alice" }))
>   })
> ```

Is that because of the symbols?

sukovanej:
> > > also not sure about the early escape of Equal
> > 
> > 
> > Me neither, but I'm not aware of any better solution. Also, I have another example that will fail even with the proposed change.
> > ```ts
> >   it("Data.struct", () => {
> >     const alice = Data.struct({ name: "Alice", age: 30 })
> > 
> >     expect(alice).toMatchObject(Data.struct({ name: "Alice" }))
> >   })
> > ```
> 
> Is that because of the symbols?

It's because of hash check in `compareBoth`.

mikearnaldi:
Hash checks are nullified when structural is enabled, they are all 0

sukovanej:
I debugged that locally, the circuit shorts [in here](https://github.com/Effect-TS/effect/blob/main/packages/effect/src/Equal.ts#L47). Even if the hash check wasn't there it would return false on comparison inside of the equal of the `StructuralPrototype`.

PR #2691, opened 2024-05-03
Browser Geolocation API

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Wrap Geolocation API with Effect
https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API

But I have some questions:
- Can I use `Effect.Tag` for Geolocation Tag? Or general practice to use `GenericTag`?
- Should native error be wrapped with `PlatformError`? and if so, should the original error be reflected at the type level?
- Is it necessary to accumulate events as it is done in `Stream.fromEventListener`?

## Related

- Related Issue #
- Closes #


packages/platform-browser/src/internal/geolocation.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { Geolocation } from "../Geolocation.js"
+
+/** @internal */
+export const tag = Context.GenericTag<Geolocation>("@effect/platform-browser/Geolocation")
+
+/** @internal */
+export const layer = Layer.succeed(
+  tag,
+  {
+    getCurrentPosition: (options?: PositionOptions) =>
```
tim-smart, line 16: I wonder if we use the watchPosition api for both, as it can be cancelled 🤔

packages/platform-browser/src/internal/geolocation.ts
```diff
@@ -0,0 +1,35 @@
...
+          (position) => resume(Effect.succeed(position)),
+          (error) => resume(Effect.fail(error)),
+          options
+        )
+      ),
+    watchPosition: (options?: PositionOptions) =>
+      Stream.async<GeolocationPosition, GeolocationPositionError>((emit) => {
+        const handlerId = navigator.geolocation.watchPosition(
+          (position) => emit(Effect.succeed(Chunk.of(position))),
+          (error) => emit(Effect.fail(Option.some(error))),
```
tim-smart, line 28: If this errors, is the watcher also cancelled?

KhraksMamtsov: I can't say for sure - but I think unsubscribing happens automatically because an error callback is optional

KhraksMamtsov: after a little research, I realized that it is worth unsubscribing only after an `PermissionDeniedError`
the other two errors do not affect the receipt of events with coordinates

PR #2712, opened 2024-05-07
sql: use constants from otel semantic-conventions instead of hardcoded values

vecerek:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This change uses the exported constants from `@opentelemetry/semantic-conventions` as span attribute names instead of hard-coding the values.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/sql-mssql/package.json
+++ b/packages/sql-mssql/package.json
@@ -48,6 +48,7 @@
     "effect": "workspace:^"
   },
   "dependencies": {
+    "@opentelemetry/semantic-conventions": "^1.24.1",
     "tedious": "^18.2.0"
   }
 }
--- a/.changeset/eight-tigers-shout.md
+++ b/.changeset/eight-tigers-shout.md
@@ -1,0 +1,12 @@
+---
+"@effect/sql-sqlite-react-native": patch
+"@effect/sql-sqlite-node": patch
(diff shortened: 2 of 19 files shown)


packages/sql-mssql/package.json
```diff
@@ -40,11 +40,13 @@
   "devDependencies": {
     "@effect/platform": "workspace:^",
     "@effect/sql": "workspace:^",
+    "@opentelemetry/semantic-conventions": "^1.24.1",
     "effect": "workspace:^"
   },
   "peerDependencies": {
     "@effect/platform": "workspace:^",
     "@effect/sql": "workspace:^",
+    "@opentelemetry/semantic-conventions": "^1.22",
```
tim-smart, line 49: I think we can make it a normal dependency, as all the usage is internal.

tim-smart (changes requested):

packages/sql-mssql/package.json
vecerek: Thanks! Addressed in https://github.com/Effect-TS/effect/pull/2712/commits/170080dbaa4aebfef4aad3b934213f48edd95085

tim-smart (approved):

PR #2725, opened 2024-05-10
add TSubscriptionRef

evelant:
## Type

- [ ] Refactor
- [ X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add TSubscriptionRef for STM

## Related

- Related Issue #
- Closes #


jessekelly881:
Would it make sense for this to implement the Subscribable trait?

evelant:
@mikearnaldi would love to get your feedback on this when you've got a moment. I think it's good to go but this is my first stab at adding a new module and I may have missed things.

evelant:
removed some comments and debug stuff I accidentally left in internal

mikearnaldi:
Cc @tim-smart for a second review, I think I am fine with it albeit have some doubts about mixing stream with stm

evelant:
I'm curious what doubts you have about mixing stream and stm? The use case leading me to implement this is when I need the synchronization capabilities of stm but also need the value-over-time for things like rendering UI.

mikearnaldi:
> I'm curious what doubts you have about mixing stream and stm? The use case leading me to implement this is when I need the synchronization capabilities of stm but also need the value-over-time for things like rendering UI.

Namely subscribing gives you back a non transactional primitive (Stream) I think I'd prefer if subscribing gave you a TDequeue like TPubSub does:

```ts
export const subscribe: <A>(self: TPubSub<A>) => STM.STM<TQueue.TDequeue<A>>
```

evelant:
Hmm interesting. I'm not sure what benefit that would offer. I can't think of a use case when I'd want a TDequeue instead of a Stream. Did you have something in mind?

mikearnaldi:
> Hmm interesting. I'm not sure what benefit that would offer. I can't think of a use case right now when I'd want a TDequeue instead of a Stream. Did you have something in mind?

The TQueue interface has all methods returning STM so they can be used transactionally. It's strictly more powerful as you can derive a Stream from a TQueue but not the opposite

packages/effect/src/TQueue.ts
```diff
@@ -262,7 +262,7 @@ export const isFull: <A>(self: TQueue<A>) => STM.STM<boolean> = internal.isFull
  * @since 2.0.0
  * @category getters
  */
-export const isShutdown: <A>(self: TQueue<A>) => STM.STM<boolean> = internal.isShutdown
+export const isShutdown: <A>(self: TDequeue<A> | TEnqueue<A>) => STM.STM<boolean> = internal.isShutdown
```
tim-smart, line 265: Can these remain as `TQueue`?

packages/effect/src/TSubscriptionRef.ts
```diff
@@ -0,0 +1,171 @@
...
+export interface TSubscriptionRef<in out A> extends TSubscriptionRef.Variance<A>, Pipeable {
+  /** @internal */
+  readonly ref: TRef.TRef<A>
+  /** @internal */
+  readonly pubsub: TPubSub.TPubSub<A>
+  /**
+   * A stream containing the current value of the `Ref` as well as all changes
+   * to that value.
+   */
+  readonly changes: Stream.Stream<A>
```
tim-smart, line 41: I think this should be a `STM<TQueue<A>>`, and maybe also add a `changesScoped: Effect<TQueue<A>, never, Scope>` and `changesStream`.

The scoped and stream variants could just be module functions.

packages/effect/src/TSubscriptionRef.ts
```diff
@@ -0,0 +1,171 @@
+/**
+ * @since 2.0.0
```
tim-smart, line 2: `since` tags should now be `3.7.0`

tim-smart (changes requested):

packages/effect/src/TQueue.ts
evelant: I'll have to check, IIRC it doesn't typecheck as TQueue but I don't remember why

evelant:
I'll update the PR with the requested changes later this week, thanks for the feedback!

packages/effect/src/TSubscriptionRef.ts
evelant: Sorry for taking so long to get to this. I'm having a bit of trouble figuring out how to do this with a TQueue. The issue is that in order to provide a TQueue that has the current value of the ref and any changes to it I'd need some way to merge queues. Like this for example does not work:

```ts
 get changes(): STM.STM<TQueue.TQueue<A>>{
    return STM.flatMap(TQueue.unbounded<A>(), queue => pipe(
      STM.flatMap(TRef.get(this.ref), a => TQueue.offer(queue, a)),
      STM.flatMap(() => TPubSub.subscribe(this.pubsub)),
      STM.flatMap(dequeue => pipe(
          TQueue.poll(dequeue),
          STM.tap(a => Option.isNone(a) ? TQueue.shutdown(dequeue) : TQueue.offer(queue, a.value)),
      )),
      STM.as(queue)
    )
```

I don't see how I can provide a TQueue with the current value _and_ any changes from the queue given by `TPubSub.subscribe`. Any thoughts or pointers?

mikearnaldi: changes doesn't contain the current value usually

evelant: @mikearnaldi it does with SubscriptionRef which is why I thought we probably wanted the same behavior here, from SubscriptionRef.ts:

```ts
 /**
   * A stream containing the current value of the `Ref` as well as all changes
   * to that value.
   */
  readonly changes: Stream.Stream<A>
```

mikearnaldi: then probably just implementing the TDequeue interface with a custom impl

mikearnaldi: I think thanks to STM transactional behaviour we can actually merge queues: https://github.com/Effect-TS/effect/pull/2725/commits/91e6a5552fe1e917db730c6ed1285e949f2ad4ea this would need a bit of testing though

packages/effect/src/TQueue.ts
mikearnaldi: actually this is better typed as `TDequeue<A> | TEnqueue<A>`, given that such functionality (together with others, updated in https://github.com/Effect-TS/effect/pull/2725/commits/bd7d59669fef016236e831b2acb758da4bc19d0d) are shared between `TEnqueue` and `TDequeue` so requiring a `TQueue` would be too restrictive

mikearnaldi:
@tim-smart I think this is now in a good state, pls have a final review, feel free to merge directly is all looks good

evelant:
@mikearnaldi thanks for finishing this up!

packages/effect/src/TSubscriptionRef.ts
```diff
@@ -0,0 +1,192 @@
...
+ * @category mutations
+ */
+export const changesScoped: <A>(self: TSubscriptionRef<A>) => Effect.Effect<TQueue.TDequeue<A>, never, Scope.Scope> =
+  internal.changesScoped
+
+/**
+ * @since 3.10.0
+ * @category mutations
+ */
+export const changesStream: <A>(self: TSubscriptionRef<A>) => Stream.Stream<A, never, never> = internal.changesStream
```
tim-smart, line 186:
```suggestion
export const changesStream: <A>(self: TSubscriptionRef<A>) => Stream.Stream<A> = internal.changesStream
```

packages/effect/src/internal/stream.ts
```diff
@@ -3133,6 +3135,31 @@ export const fromPubSub: {
...
+  }): Effect.Effect<Stream.Stream<A>, never, Scope.Scope>
+  <A>(pubsub: TPubSub.TPubSub<A>, options?: {
+    readonly scoped?: false | undefined
+    readonly shutdown?: boolean | undefined
+  }): Stream.Stream<A>
+} = (pubsub, options): any => {
+  if (options?.scoped) {
+    const effect = Effect.map(
+      TPubSub.subscribeScoped(pubsub),
+      (queue) => fromTQueue(queue, { shutdown: true })
```
tim-smart, line 3152: This seems strange to me, as the `Scope` should be responsible for shutting down the subscription.

mikearnaldi: good catch, the full options story didn't quite make sense, cleared up in https://github.com/Effect-TS/effect/pull/2725/commits/55dcb28e25e473eb172d20d43b817f4c0aa7d7f4

tim-smart (approved):

PR #2726, opened 2024-05-10
Schema / Document how to introduce `jsonSchema` annotation type constraints in user-space.

suddenlyGiovanni:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

This PR tries to address the points discussed in #2663;

### Context

The outcome of the discussion and decision in n2663 is that `@effect/schema` will not extend nor maintain its internal typings for `JSONSchema7`. Instead, it will push such responsibility to the user space; the reasons are sound, and you can explore them by following the discussion.

However, it has also been decided to update the docs to showcase one possible technique and provide an example of how to achieve the same results that n2663 aimed to solve (better type safety and improved DX).

### Changes

The `README.md` file in the schema package has been updated to provide detailed guidance on managing type constraints when dealing with non-standard extensions via the `jsonSchema` property. The update includes a code example demonstrating how to apply type constraints using the `@types/json-schema` package for TypeScript definitions.

## Related

- Closes #2663


gcanti (changes requested):
@suddenlyGiovanni thanks! (no need for a changeset in this case, you can remove it)

gcanti (approved):

PR #2737, opened 2024-05-13
feat: added KeyValueStore.layerStorage

jessekelly881:
added platform `KeyValueStore.layerStorage` to create a KeyValueStore from an instance of the Storage type(e.g. SessionStorage and LocalStorage). 

Thoughts on the pathOrDescriptor arg? I'm not sure if that's a good idea.


tim-smart:
> Thoughts on the pathOrDescriptor arg? I'm not sure if that's a good idea.

You can just use the `key` as the path.

packages/platform/src/KeyValueStore.ts
```diff
@@ -205,3 +205,13 @@ export const layerSchema: <A, I, R>(
...
 } = internal.layerSchema
+
+/**
+ * Creates an KeyValueStorage from an instance of the `Storage` api.
+ *
+ * @since 1.0.0
+ * @category layers
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API
+ */
+export const layerStorage: (pathOrDescriptor: string, storage: Storage) => Layer.Layer<KeyValueStore, never, never> =
```
tim-smart, line 216:
```suggestion
export const layerStorage: (storage: Storage) => Layer.Layer<KeyValueStore> =
```

jessekelly881: I actually had changed this but I forgot to git add it before committing lol

PR #2744, opened 2024-05-14
make `Array.separate`, `Array.getRights`, `Array.getLefts`, `Array.getSomes` heterogeneous

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #


gcanti (changes requested):
I simplified the signatures (https://github.com/Effect-TS/effect/pull/2823) and added a few dtslint tests (https://github.com/Effect-TS/effect/pull/2825).

I would extend the same support to:

- getRights
- getLefts
- getSomes

Additionally, I would add the following dtslint tests (also for the other 3) to showcase the new capabilities:

```typescript
// $ExpectType [(string | Date)[], (number | boolean)[]]
Array.separate(hole<Iterable<Either.Either<number, string> | Either.Either<boolean, Date>>>())

// $ExpectType [(string | Date)[], (number | boolean)[]]
Array.separate(hole<Iterable<Either.Either<number, string>> | Iterable<Either.Either<boolean, Date>>>())
```

packages/effect/src/Array.ts
```diff
@@ -2254,7 +2254,14 @@ export const partitionMap: {
  * @category filtering
  * @since 2.0.0
  */
-export const getSomes: <A>(self: Iterable<Option<A>>) => Array<A> = filterMap(identity)
+
+export const getSomes: <
```
gcanti, line 2258: does this work as well?

```ts
export const getSomes: <T extends Iterable<Option<X>>, X = any>(
  self: T
) => Array<Option.Value<ReadonlyArray.Infer<T>>> = filterMap(identity as any)
```

KhraksMamtsov: yes

gcanti (approved):

gcanti:
@KhraksMamtsov thanks!

PR #2747, opened 2024-05-14
Improve Cause Rendering (Vitest)

mikearnaldi:
Screenshot 2024-05-15 at 13 47 40


tim-smart:
This will likely create some merge conflicts with next-minor (with the span tracing changes), but I can resolve those 👍🏼

packages/effect/src/internal/cause.ts
```diff
@@ -990,10 +993,18 @@ class PrettyError extends globalThis.Error implements Cause.PrettyError {
...
+      Object.keys(originalError).forEach((key) => {
+        if (!(key in this)) {
+          // @ts-expect-error
+          this[key] = originalError[key]
+        }
+      })
+    }
   }
 
   get span(): Span | undefined {
```
tim-smart, line 1006: We could just make this a property I guess. Then we don't need `this.#cause`.

mikearnaldi: made the change but apparently there seems to be an issue, I pushed a failing test where spans are not propagated not sure why, I'm going to take a flight will check when I am back if you don't figure it out before

mikearnaldi:
@tim-smart why did we end up with unrelated commits in this branch? Seems like a rebase issue

tim-smart:
> @tim-smart why did we end up with unrelated commits in this branch? Seems like a rebase issue 

Not sure. I ran another rebase against next-minor and it was clean. GitHub must struggle when the base branch also rebases.

PR #2759, opened 2024-05-16
Test build output before publishing

fubhy:
@tim-smart How about this? It's not a silver bullet, but it'd help detect potential issues with the build output. Since this only runs on changeset publish, it'd also not slow down our pipeline elsewhere.


fubhy:
This assumes that some tests import from the root entrypoint. Which, I must admit, aren't that many.

Another option could be to generate snapshot tests for the exported symbols of all public modules. Basically just "dialing in" the public interface in a generated test case.

PR #2772, opened 2024-05-18
Add onNoneEncoding to schema.optional

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This adds `onNoneEncoding` option to `schema.optiona` when used with `{as: "Option"}`. I took inspiration from `OptionFromNullishOr`

Without `onNoneEncoding`, the prop gets removed from the containing Struct
```typescript
const schema = S.Struct({
  a: S.optional(S.String, { as: "Option")
})
S.encode(schema)({ a: O.none() }) // => {}
```

With `onNoneEncoding`, the prop is kept with the value being the value of `onNoneEncoding`
```typescript
const schema = S.Struct({
  a: S.optional(S.String, { as: "Option", onNoneEncoding: undefined })
})
S.encode(schema)({ a: O.none() }) // => { a: undefined }
```
## Related

N/A

### DIFF
--- a/packages/schema/src/Schema.ts
+++ b/packages/schema/src/Schema.ts
@@ -1871,7 +1871,14 @@ export const optional: {
       readonly as: "Option"
       readonly default?: never
       readonly exact?: true
-      readonly nullable?: true
+      readonly nullable?: never
+      readonly onNoneEncoding?: undefined
+    } | {
+      readonly as: "Option"
+      readonly default?: never
+      readonly exact?: true
+      readonly nullable: true
+      readonly onNoneEncoding?: null | undefined
     } | undefined
   >(
     options?: Options
@@ -1889,7 +1896,7 @@ export const optional: {
       | (Types.Has<Options, "as"> extends true ? option_.Option<A> : A)
       | (Types.Has<Options, "as" | "default" | "exact"> extends true ? never : undefined),
       never,
-      "?:",
+      Types.Has<Options, "onNoneEncoding"> extends true ? ":" : "?:",
       | I
       | (Types.Has<Options, "nullable"> extends true ? null : never)
       | (Types.Has<Options, "exact"> extends true ? never : undefined),
@@ -1914,7 +1921,14 @@ export const optional: {
       readonly as: "Option"
       readonly default?: never
       readonly exact?: true
-      readonly nullable?: true
+      readonly nullable?: never
+      readonly onNoneEncoding?: undefined
+    } | {
+      readonly as: "Option"
+      readonly default?: never
+      readonly exact?: true
+      readonly nullable: true
+      readonly onNoneEncoding?: null | undefined
     } | undefined
   >(
     schema: Schema<A, I, R>,
@@ -1933,7 +1947,7 @@ export const optional: {
       | (Types.Has<Options, "as"> extends true ? option_.Option<A> : A)
       | (Types.Has<Options, "as" | "default" | "exact"> extends true ? never : undefined),
       never,
-      "?:",
+      Types.Has<Options, "onNoneEncoding"> extends true ? ":" : "?:",
       | I
       | (Types.Has<Options, "nullable"> extends true ? null : never)
       | (Types.Has<Options, "exact"> extends true ? never : undefined),
--- a/packages/schema/src/Schema.ts
+++ b/packages/schema/src/Schema.ts
@@ -1872,13 +1872,13 @@ export const optional: {
       readonly default?: never
       readonly exact?: never
       readonly nullable?: never
-      readonly onNoneEncoding?: undefined
+      readonly onNoneEncoding?: option_.Option<undefined>
     } | {
       readonly as: "Option"
       readonly default?: never
       readonly exact?: never
       readonly nullable: true
-      readonly onNoneEncoding?: null | undefined
+      readonly onNoneEncoding?: option_.Option<null | undefined>
     } | {
       readonly as: "Option"
       readonly default?: never
@@ -1890,7 +1890,7 @@ export const optional: {
       readonly default?: never
       readonly exact: true
       readonly nullable: true
-      readonly onNoneEncoding?: null
+      readonly onNoneEncoding?: option_.Option<null>
     } | undefined
   >(
     options?: Options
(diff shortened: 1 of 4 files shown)


packages/schema/src/Schema.ts
```diff
@@ -1870,8 +1870,27 @@ export const optional: {
...
+      readonly default?: never
+      readonly exact: true
+      readonly nullable?: never
+      readonly onNoneEncoding?: never
+    } | {
+      readonly as: "Option"
+      readonly default?: never
+      readonly exact: true
+      readonly nullable: true
+      readonly onNoneEncoding?: null
```
vinassefranche, line 1893: The types are more complex, mostly because of the exact case. I feel like it does not make sense to allow `onNoneEncoding` for the exact case when `nullable` is not true.
It could be challenged to add this option for the exact case too.

packages/schema/src/Schema.ts
```diff
@@ -2035,13 +2082,29 @@ export const optional: {
...
       } else {
         return optionalToRequired(
           UndefinedOr(schema),
           OptionFromSelf(typeSchema(schema)),
-          { decode: option_.filter(Predicate.isNotUndefined<A | undefined>), encode: identity }
+          {
+            decode: option_.filter(Predicate.isNotUndefined<A | undefined>),
+            encode: (t) => {
+              if ("onNoneEncoding" in options) {
+                return option_.orElse(t, () => option_.some(options.onNoneEncoding as undefined))
```
vinassefranche, line 2103: This `as` is safe, because the types of option prevents setting `onNoneEncoding` to null when `nullable` is not set.
I don't know the policy for the `as`. Tell me if you think it's wrong

vinassefranche:
If this is accepted, we should update the documentation. Tell me if I should do it in this P.R.

gcanti:
I think it would be better if `onNoneEncoding` had a more explicit type like `() => Option.Option<X>`, where `X` can be `undefined` or `null | undefined`, `None` indicates removal, while `Some` indicates retention (default `None`).

packages/schema/src/Schema.ts
```diff
@@ -1889,7 +1908,7 @@ export const optional: {
       | (Types.Has<Options, "as"> extends true ? option_.Option<A> : A)
       | (Types.Has<Options, "as" | "default" | "exact"> extends true ? never : undefined),
       never,
-      "?:",
+      Types.Has<Options, "onNoneEncoding"> extends true ? ":" : "?:",
```
gcanti, line 1911: This doesn't seem correct, it should always be `"?:"` since you are declaring an optional property

vinassefranche: I understand. I wanted to type properly the output of the encode but ot seems like I can't do that without impacting the input type of `S.decode(schema)`, right?
Is it ok to keep `"?:"` even though we know that the encode output will always be present? Or you think it's a bad idea to add this `onNoneEncoding`?

vinassefranche:
> I think it would be better if `onNoneEncoding` had a more explicit type like `() => Option.Option<X>`, where `X` can be `undefined` or `null | undefined`, `None` indicates removal, while `Some` indicates retention (default `None`).

I did not see this comment before I answered the other.
I understand what you mean and indeed think it can be great. I'll try to update my P.R. in that way and ping you when it's done. Thanks for your input

vinassefranche:
> I think it would be better if `onNoneEncoding` had a more explicit type like `() => Option.Option<X>`, where `X` can be `undefined` or `null | undefined`, `None` indicates removal, while `Some` indicates retention (default `None`).

@gcanti why a function `() => Option.Option<X>` instead of simply `Option.Option<X>`? As it's something passed on the schema creation and it won't take the real value as an argument, I don't see the value of the function

vinassefranche:
> > I think it would be better if `onNoneEncoding` had a more explicit type like `() => Option.Option<X>`, where `X` can be `undefined` or `null | undefined`, `None` indicates removal, while `Some` indicates retention (default `None`).
> 
> @gcanti why a function `() => Option.Option<X>` instead of simply `Option.Option<X>`? As it's something passed on the schema creation and it won't take the real value as an argument, I don't see the value of the function

I've implemented the version without the function so you can check if I understood well. Tell me if the global idea is good and if you think it would be better with a function.

gcanti:
Mainly for consistency with our typical practices in Effect, although I now realize that `onNoneEncoding` in `OptionFromNullishOr` is not lazy

packages/schema/src/Schema.ts
gcanti: > I can't do that without affecting the input type of `S.decode(schema)`, correct?

right

> Is it okay to keep `"?:"` even though we know that the encoded output will always be present?

yes because the intention here is to define an optional property within the `I` type of the schema

packages/schema/src/Schema.ts
```diff
@@ -1933,7 +1971,7 @@ export const optional: {
...
```
gcanti, line 1974: the same applies here, it should be `"?:"` as before

vinassefranche: My bad, sorry! I'm fixing that quickly
Why not having define types that are re-used in both overloads? It feels a bit painful to have to update both every time but there might be some reasons I don't know of.

vinassefranche:
> Mainly for consistency with our typical practices in Effect, although I now realize that `onNoneEncoding` in `OptionFromNullishOr` is not lazy

I did the change! I was taking inspiration from `OptionFromNullishOr` :-)

gcanti:
@vinassefranche I'd propose a couple of improvements WDYT? https://github.com/Effect-TS/effect/commits/2772-local/

vinassefranche:
> @vinassefranche I'd propose a couple of improvements WDYT? https://github.com/Effect-TS/effect/commits/2772-local/

@gcanti Great! I wanted to add types tests but did not know where to add them and forgot to ask.
I'd also add these tests:
```
// @ts-expect-error
S.optional(S.String, { as: "Option", onNoneEncoding: () => Option.some(null) })

// @ts-expect-error
S.String.pipe(S.optional({ as: "Option", onNoneEncoding: () => Option.some(null) }))
```

How do I integrate your commits in my branch. Can you do it on your own? Do I have to cherry-pick them?

gcanti:
If you're okay with the changes, I think I can push to your branch. Wait, let me try...

gcanti:
lol no, I can't do it. Please just copy the changes if you're okay with them

mikearnaldi:
> lol no, I can't do it. Please just copy the changes if you're okay with them

you should be able to, repo has the setting to allow it

gcanti:
ok, this time it seems to have worked

vinassefranche:
> ok, this time it seems to have worked

Great! Should I push my additional type tests?

packages/schema/src/Schema.ts
gcanti: yeah, now that the type is quite big, it makes even more sense to define the type only once

PR #2785, opened 2024-05-20
For `Data.TaggedEnum` added capitalized constructor naming convention.

KhraksMamtsov:
https://discord.com/channels/795981131316985866/920065989557747723/1242059983999467520
For `Data.TaggedEnum` added capitalized constructor naming convention.
Helper functions `$is` and `$match` have been renamed to `is` and `match`.
image

## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/happy-jeans-invent.md
+++ b/.changeset/happy-jeans-invent.md
@@ -1,0 +1,6 @@
+---
+"effect": major
+---
+
+For `Data.TaggedEnum` added capitalized constructor naming convention.
+Helper functions `$is` and `$match` have been renamed to `is` and `match`.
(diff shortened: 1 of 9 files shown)


tim-smart:
Maybe instead of enforcing capitalisation, we just reserve the `is` and `match` keywords?

If a user wants lower case tags, then I don't think we should prevent that.

KhraksMamtsov:
with this restriction, we reserve space for helpers that may appear in the future. otherwise, it will be a breaking change every time.

PR #2788, opened 2024-05-20
refactor: deduplicate schema optional types

vinassefranche:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Followup of this [comment](https://github.com/Effect-TS/effect/pull/2772#discussion_r1605990786) on a previous P.R.

I did not add any changeset as it's supposed to be an internal refactor but, as shown in the update of the types' tests, there's a small potential external change so tell me if you think I should add one.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/schema/dtslint/Schema.ts
+++ b/packages/schema/dtslint/Schema.ts
@@ -704,7 +704,7 @@ S.Struct({
 })
 
 // $ExpectType Struct<{ a: PropertySignature<":", "a" | "b", never, "?:", "a" | "b", true, never>; }>
-S.Struct({ a: S.optional(S.Literal("a", "b"), { default: () => "a", exact: true }) })
+S.Struct({ a: S.optional(S.Literal("a", "b"), { default: () => "a" as const, exact: true }) })
 
 // $ExpectType Schema<{ readonly a: "a" | "b"; }, { readonly a?: "a" | "b"; }, never>
 S.asSchema(S.Struct({ a: S.Literal("a", "b").pipe(S.optional({ default: () => "a" as const, exact: true })) }))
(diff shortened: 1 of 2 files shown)


packages/schema/src/Schema.ts
```diff
@@ -1850,134 +1850,85 @@ export const optionalToOptional = <FA, FI, FR, TA, TI, TR>(
     )
   )
 
+type OptionalOptions<A> = {
```
vinassefranche, line 1853: Not a huge fan of the name. Tell me if you have a better idea.

packages/schema/dtslint/Schema.ts
```diff
@@ -704,7 +704,7 @@ S.Struct({
 })
 
 // $ExpectType Struct<{ a: PropertySignature<":", "a" | "b", never, "?:", "a" | "b", true, never>; }>
-S.Struct({ a: S.optional(S.Literal("a", "b"), { default: () => "a", exact: true }) })
+S.Struct({ a: S.optional(S.Literal("a", "b"), { default: () => "a" as const, exact: true }) })
```
vinassefranche, line 707: Not completely sure why it was needed but as it was done for the other overload, I thought it was fine to add `as const`

gcanti: I'd revert this change, it's good to know that `as const` is not needed

vinassefranche: It was actually needed with my change. If I revert it, I have an error: 
image
Any idea where this could come from?
I don't have the issue if I only extract `OptionalOptions`. It comes when I extract OptionalReturnType.

gcanti: ah, sorry, I read it wrong. I thought it was done for consistency, not because it was causing a new error. ok, let me see what could be causing it...

vinassefranche: I tries a few `in` or `out` in the `OptionalOptions` type but without success

gcanti: The issue seems related to `default` being a lazy arg `() => A`
I have an idea to solve the issue, but it involves quite a refactoring. I'll send another PR to replace this one

gcanti:
Closing in favour of https://github.com/Effect-TS/effect/pull/2794

PR #2794, opened 2024-05-21
add `optional` and `optionalWithOptions` API interfaces

gcanti:
RE https://github.com/Effect-TS/effect/pull/2788

/cc @vinassefranche

### DIFF
--- a/.changeset/real-ears-hide.md
+++ b/.changeset/real-ears-hide.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+add `optional` and `optionalWithOptions` API interfaces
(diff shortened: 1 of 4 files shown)


gcanti:
@vinassefranche this also resolves the issue with the old `as const`s, not just the new ones 🙌

vinassefranche (approved):
Amazing! Is it mostly the use of `LazyArg` that fixes the issue or separating into `optional` and `optionalWithOptions` ?(or both were needed?)

gcanti:
No the use of `LazyArg` is merely cosmetic. Adding these API interfaces was something I had wanted to do for a while, and in doing so, I realized that the issue disappeared. TypeScript is weird.

PR #2795, opened 2024-05-21
Harmonize Schema.OptionFromNullishOr signature

vinassefranche:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This harmonizes the second argument of `Schema.OptionFromNullishOr` with the rest of the code base:
- Make the second argument a named argument
- Make onNoneEncoding a lasy argument

This follows [this discussion](https://discordapp.com/channels/795981131316985866/1242189617352937524/1242365827395747943) and this [P.R. comment](https://github.com/Effect-TS/effect/pull/2772#issuecomment-2118882488).

## Related

N/A

### DIFF
--- a/.changeset/fifty-beans-hope.md
+++ b/.changeset/fifty-beans-hope.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": minor
+---
+
+Harmonize OptionFromNullishOr signature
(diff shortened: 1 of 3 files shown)


gcanti:
It seems a bit strange to me since `onNoneEncoding` would be the only option; it's a required argument, not one among many options.

vinassefranche:
> It seems a bit strange to me since `onNoneEncoding` would be the only option; it's a required argument, not one among many options.

I get what you mean. I agree options is not a good name in any case.
If it was me, I think I would have done something like `Schema.OptionFromNullishOr({schema, onNoneEncoding: null})` but the second overload would be very different so it might not be great.

I still think that reading `Schema.OptionFromNullishOr(schema, null)` is not an amazing developer experience. If you are not used to the function, you wonder what that `null` is for.
Any other idea to improve this?

PR #2797, opened 2024-05-21
Improve internalization of functions to clean stack traces

mikearnaldi:
Should resolve the following output:

https://effect.website/play#bdCxboMwEAbgV7FugghcICEJlSp1yRipO2awzEFo4WwZ0whFvHtFIBJDJw//99/ZzsG0cqytHqgM41N8TLLsnJ6S/QGCPAcj1Y+skX/3miCA9XgIYkyAGw0KeGcCOl0OLQoIlqBEg1QiqQb7GTz9HHxiVaFyb6aVrtK2W9oRT/c8jtf6fy4kXeILH7L/ca9u2MmXOp74eYMWs4R7nvDtANer8C6dui3xkSc82sajwV7Zxqz1lKc8Ckv85UmUHKI0zgTMdhI0QRHk0MmGuOsh2FQhgKYz2jr2YJfnXdjEKqu7zd0ECVKaeseM1bWVHftYLa+RvGog5RpNO89ffnRssC13L1LJpvUI7+xirbaegOv1KsD3BU3+PHlldqAvq7umR2/d4nM1P96bV+sWeatrXxAURfEH

### DIFF
--- a/packages/effect/src/Utils.ts
+++ b/packages/effect/src/Utils.ts
@@ -784,3 +784,23 @@ export const structuralRegion = <A>(body: () => A, tester?: (a: unknown, b: unkn
     structuralRegionState.tester = currentTester
   }
 }
+
+const tracingFunction = (name: string) => {
+  const internalCall = <A>(body: () => A): A => body()
+  Object.defineProperty(internalCall, "name", { value: name })
+  return internalCall
+}
+
+/**
+ * @since 3.2.2
+ * @status experimental
+ * @category tracing
+ */
+export const internalCall = tracingFunction("effect_internal_function")
+
+/**
+ * @since 3.2.2
+ * @status experimental
+ * @category tracing
+ */
+export const internalGeneratorCall = tracingFunction("effect_internal_generator")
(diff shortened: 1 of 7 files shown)


packages/effect/src/Utils.ts
```diff
@@ -784,3 +784,23 @@ export const structuralRegion = <A>(body: () => A, tester?: (a: unknown, b: unkn
     structuralRegionState.tester = currentTester
   }
 }
+
+const tracingFunction = (name: string) => {
+  const internalCall = <A>(body: () => A): A => body()
```
tim-smart, line 789: Maybe we can use an object to prevent the rename perf hit:

```ts
const obj = {
  [name]() { return body() }
}
return obj[name]
```

mikearnaldi: Rename is done once at module load, not at function call so not sure the cost has an impact. The issue with the object method is that apparently it isn't preserved in some cases, I.e. in the playground it doesn't work (check the repro)

tim-smart (approved):

PR #2799, opened 2024-05-21
Add Date module

jdharrisnz:
## Type

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add Date module.

## Related

N/A

## Questions from me

Hi, my first time contributing to this! Hopefully this module is valuable and well-written.

I had some questions about code style and documentation.

First - `@since` tags - who adds these? I don't think I can do it since I don't know which version this might be added to.

Second - Predicate-based dual api functions. I've never seen these in Effect, but there are a few functions that offer optional arguments with default values, such as the week functions and their `dowOffset`. There's no difference between these usages:
```typescript
pipe(Date.create(), Date.toWeekNumber)
pipe(Date.create(), Date.toWeekNumber(1))
```
...and TypeScript has no problem figuring out which overload to use, but it's a bit confusing to read.

Third - Is there some requirement for test coverage? I've added tests for everything that might be considered complex, but a lot of functions are also extremely trivial.

Fourth - I've added a lot in here and surely cover all the basics, but there's surely also more inspiration I could take from projects like `date-fns`. How comprehensive is something like this expected to be before it's accepted?

Thanks!

### DIFF
--- a/.changeset/cold-dingos-thank.md
+++ b/.changeset/cold-dingos-thank.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add Date module
--- a/packages/effect/src/Date.ts
+++ b/packages/effect/src/Date.ts
@@ -1,0 +1,1936 @@
+import type { Branded } from "./Brand.js"
+import * as Duration from "./Duration.js"
+import * as Either from "./Either.js"
+import * as Equivalence from "./Equivalence.js"
+import { constant, dual, flow, pipe, unsafeCoerce } from "./Function.js"
+import { getUTCWeekNumBaseYear, getWeekNumBaseYear, isDateFirstArg, isDateFirstTwoArgs } from "./internal/date.js"
+import * as Option from "./Option.js"
+import * as Order from "./Order.js"
+import * as String from "./String.js"
+
+import type { LazyArg } from "./Function.js"
+import type * as Predicate from "./Predicate.js"
+/** @category instances */
+export { Date as Equivalence } from "./Equivalence.js"
+
+/** @category instances */
+export { Date as Order } from "./Order.js"
+
+/** @category instances */
+export { isDate } from "./Predicate.js"
+
+/**
+ * Branded type representing the number of milliseconds elapsed since midnight, January 1, 1970 Universal Coordinated Time (UTC).
+ *
+ * @category models
+ */
+export type Milliseconds = Branded<number, "Milliseconds">
+
+/**
+ * Branded type representing a date string that can be safely parsed to a `Date`.
+ *
+ * @category models
+ */
+export type DateString = Branded<string, "DateString">
+
+/**
+ * Convert a `number` to the `Milliseconds` branded type.
+ *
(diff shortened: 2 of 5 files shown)


tim-smart:
Hi, thanks for the PR :) Tackling a `Date` module is a brave first contribution lol.

> First - `@since` tags - who adds these? I don't think I can do it since I don't know which version this might be added to.

You can add `@since 3.3.0` for the time being.

> Second - Predicate-based dual api functions. I've never seen these in Effect, but there are a few functions that offer optional arguments with default values, such as the week functions and their `dowOffset`. There's no difference between these usages:

I think what you have there is fine.
 
> Third - Is there some requirement for test coverage? I've added tests for everything that might be considered complex, but a lot of functions are also extremely trivial.

It will be reviewed on a case by case basis. If we feel there are tests missing during review, we will let you know.

> Fourth - I've added a lot in here and surely cover all the basics, but there's surely also more inspiration I could take from projects like `date-fns`. How comprehensive is something like this expected to be before it's accepted?

We will give your PR a proper review soon. A `Date` module has the potential to get quite complex (especially if you want to add timezone support etc), so sticking to the basics for now is a good choice.

.changeset/cold-dingos-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
```
tim-smart, line 2:
```suggestion
"effect": minor
```

jdharrisnz:
Hi @tim-smart , is there anything I need to do to move this along, or still just waiting for code review?

I see one of the checks above failed but I'm not sure what causes it.

tim-smart:
> Hi @tim-smart , is there anything I need to do to move this along, or still just waiting for code review?
> 
> I see one of the GitHub checks fails but I'm not sure what causes it.

Sorry for the neglect, but I think we just need to set aside some time to think about how we want to approach dates in effect.

Maybe we could introduce a `UtcDate` module first, which has APIs that only deal with UTC. Then afterwards look at how to add timezone support.

js Date also has so many footguns it might be better to side step it and create a more type safe data structure.

tim-smart:
I had a play around with the domain here: https://github.com/Effect-TS/effect/pull/3296

Not sure about the direction yet, but I think putting up some guard rails is a good approach.

jdharrisnz:
Thanks Tim. Agreed on shipping only UTC functions - there's a lot to keep in mind with the local stuff 😅

Your domain looks like a great start. Do you want me to move over the UTC getters, setters, and math functions from this PR into that? (Not sure how this fits with the WithZone part of it though - is that basically just an offset on top of UTC?)

tim-smart:
I think for time zones we can lean on the `Intl` APIs.

Yeah feel free to pitch in.

For the math functions it could be quite nice to have something like:

```ts
DateTime.add(self, 6, "months")
DateTime.add(self, Duration)
```

Then we can just "do the right thing" to prevent issues when crossing over daylight savings.

PR #2805, opened 2024-05-22
fix internal cutpoint name preservation

tim-smart:
This was the only thing that worked across these environments:

Playground:
image

tsx esm:
image

tsx cjs:
image

Browser:
image


tim-smart:
And bun gives us absolutely nothing to work with...
image

PR #2812, opened 2024-05-22
add Array.ensure api

datner:
yum yum, I love getting garbage from my backend

- test(Array): fix typos
- chore: update readme to be more accurate
- feat(Array): add cast

### DIFF
--- a/.changeset/warm-pants-yawn.md
+++ b/.changeset/warm-pants-yawn.md
@@ -2,13 +2,14 @@
 "effect": minor
 ---
 
-add `cast` util for Array, used to normalize `A | Array<A>`
+add `ensure` util for Array, used to normalize `A | ReadonlyArray<A>`
 
 ```ts
-import { cast } from "effect/Array"
+import { ensure } from "effect/Array"
 
+// lets say you are not 100% sure if it's a member or a collection
 declare const someValue: {foo: string} | Array<{foo: string}>
 
 // $ExpectType ({ foo: string })[]
-const normalized = cast(someValue)
+const normalized = ensure(someValue)
 ```
--- a/.changeset/warm-pants-yawn.md
+++ b/.changeset/warm-pants-yawn.md
@@ -1,0 +1,14 @@
+---
+"effect": minor
+---
+
+add `cast` util for Array, used to normalize `A | Array<A>`
+
+```ts
+import { cast } from "effect/Array"
+
+declare const someValue: {foo: string} | Array<{foo: string}>
+
+// $ExpectType ({ foo: string })[]
+const normalized = cast(someValue)
+```
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ Welcome to Effect, a powerful TypeScript framework that provides a fully-fledged
   // ...
   "compilerOptions": {
     // ...
-    "strict": true
-  }
+    "strict": true,
+  },
 }
 ```
 
--- a/packages/effect/dtslint/Array.ts
+++ b/packages/effect/dtslint/Array.ts
@@ -633,26 +633,26 @@ pipe(
 )
 
 // -------------------------------------------------------------------------------------
-// cast
+// ensure
 // -------------------------------------------------------------------------------------
 
 // $ExpectType unknown[]
-Array.cast(unknownValue)
+Array.ensure(unknownValue)
 
 // $ExpectType number[]
-Array.cast(primitiveNumber)
+Array.ensure(primitiveNumber)
 
 // $ExpectType (string | number)[]
-Array.cast(primitiveNumberOrString)
+Array.ensure(primitiveNumberOrString)
 
 // $ExpectType number[]
-Array.cast(numbers)
+Array.ensure(numbers)
 
 // $ExpectType string[]
-Array.cast(strings)
+Array.ensure(strings)
 
 // $ExpectType (string | number)[]
-Array.cast(numbersOrStrings)
+Array.ensure(numbersOrStrings)
 
 // -------------------------------------------------------------------------------------
 // span
--- a/packages/effect/dtslint/Array.ts
+++ b/packages/effect/dtslint/Array.ts
@@ -19,7 +19,7 @@ declare const iterStrings: Iterable<string>
 declare const numbersOrStrings: Array<number | string>
 
 declare const primitiveNumber: number
-declare const primitiveNumberOrString: string | number
+declare const primitiveNumerOrString: string | number
 declare const predicateNumbersOrStrings: Predicate.Predicate<number | string>
 
 declare const unknownValue: unknown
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ We welcome contributions via pull requests! Here are some guidelines to help you
 
 1. Fork the repository and clone it to your local machine.
 2. Create a new branch for your changes: `git checkout -b my-new-feature`.
-3. Ensure you have the required dependencies installed by running: `pnpm install` (assuming pnpm version `8.x`).
+3. Ensure you have the required dependencies installed by running: `pnpm install` (assuming pnpm version `9.x`).
```
datner, line 44: package json demands a 9.x pnpm

README.md
```diff
@@ -51,7 +51,7 @@ We welcome contributions via pull requests! Here are some guidelines to help you
    - `pnpm lint`: Check for code style adherence (if you happen to encounter any errors during this process, you can use `pnpm lint-fix` to automatically fix some of these style issues).
    - `pnpm dtslint`: Run type-level tests.
    - `pnpm docgen`: Check the integrity of the generated documentation.
-6. Create a changeset for your changes: before committing your changes, create a changeset to document the modifications. This helps in tracking and communicating the changes effectively. To create a changeset, run the following command: `pnpm changeset`. Always choose the `patch` option when prompted (please note that we are currently in pre-release mode).
```
datner, line 54: we are no longer in pre-release so yeah?

packages/effect/dtslint/Array.ts
datner, line 1: I added a couple tests and fixed a type and the lint-fix reformatted the entire document, idk mang

packages/effect/src/Array.ts
```diff
@@ -145,6 +145,21 @@ export const replicate: {
...
+ * import { cast } from "effect/Array"
+ *
+ * assert.deepStrictEqual(cast("a"), ["a"])
+ * assert.deepStrictEqual(cast(["a"]), ["a"])
+ * assert.deepStrictEqual(cast(["a", "b", "c"]), ["a", "b", "c"])
+ *
+ * @category constructors
+ * @since 2.0.0
+ */
+export const cast = <A>(self: Array<A> | A): Array<A> => Array.isArray(self) ? self : [self]
```
datner, line 161: did not use `of` because `cast([])` is a valid value, so I can't promise `[A, ...A[]]`

packages/effect/src/Array.ts
```diff
@@ -145,6 +145,21 @@ export const replicate: {
...
+ *
+ * @example
+ * import { cast } from "effect/Array"
+ *
+ * assert.deepStrictEqual(cast("a"), ["a"])
+ * assert.deepStrictEqual(cast(["a"]), ["a"])
+ * assert.deepStrictEqual(cast(["a", "b", "c"]), ["a", "b", "c"])
+ *
+ * @category constructors
+ * @since 2.0.0
```
tim-smart, line 159:
```suggestion
 * @since 3.3.0
```

.changeset/warm-pants-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
+---
+
+add `cast` util for Array
```
tim-smart, line 5: If you could add a quick code example that would be appreciated :)

gcanti:
How about using "ensure" instead of "cast"?
Shouldn't it take `Iterable<A>` as input like other APIs in the module?

datner:
@gcanti 
> How about using "ensure" instead of "cast"? Shouldn't it take `Iterable<A>` as input like other APIs in the module?

hmm `ensure` sounds like it would make sure that some condition is fulfilled, like `every`. I used `cast` because of [`castArray`](https://lodash.info/doc/castArray). Regarding `Iterable` I thought that accepting `A | Iterable` would make the `fromIterable` util redundant, since it would be `fromIterable` + wrapping. What do you think?

gcanti:
> it would make sure that some condition is fulfilled

Yes, the idea I've formed about the goal of this function is to "normalize" or "ensure" that you always have an array as output (which is the fulfilled condition)

the issue with `Array<A> | A` is that I cannot pass a readonly array

```typescript
declare const readonlyNumbers: ReadonlyArray<number>;
const x = cast(readonlyNumbers); // error
```

jessekelly881:
+1 A Schema for this might be nice as well. :) `Schema.CastArray(A) :: Schema<readonly A[], A | A[]>`

datner:
@gcanti There are a few questions that limit me adding the `A | Iterable<A>` behavior..
for `number | string | Iterable<string>`, should the outcome be:
1. `Array<number> | Array<string>`
2. `Array<number | string>`
3. `Array<number | string | Iterable<string>>`

for `Record<string, any>`, should the outcome be:
1. `Array<[string,any]>` - transforming the record to an array
2. `Array<Record<string,any>>`

and what would that mean for `Record<string, any> | Iterable<Record<string, any>>`?

Once we go beyond `Array` to the conceptual `Iterable` it becomes weird...


Also.. `<A>(self: A | Iterable<A>)` does not behave like `<A>(self: A | Array<A>)` for some reason, which requires multiple signatures to correctly type

gcanti:
Ok, then my suggestion would be:
- Change from `Array<A> | A` to `ReadonlyArray<A> | A` to support readonly arrays
- Use `normalize` or `ensure` as more descriptive names
- Add a test

datner:
@gcanti I have no idea why theres a conflict with `Schema`, I don't touch that code

datner:
A bit of git magic later-- and presto! 😄

packages/effect/src/Array.ts
```diff
@@ -145,6 +145,24 @@ export const replicate: {
...
+ * assert.deepStrictEqual(Array.ensure(["a"]), ["a"])
+ * assert.deepStrictEqual(Array.ensure(["a", "b", "c"]), ["a", "b", "c"])
+ *
+ * const set = new Set([1, 2, 3])
+ * assert.deepStrictEqual(Array.ensure(set), [1, 2, 3])
+ *
+ * @category constructors
+ * @since 3.3.0
+ */
+export const ensure = <A>(self: ReadonlyArray<A> | A): Array<A> => Array.isArray(self) ? self : of(self)
```
gcanti, line 164:
```suggestion
export const ensure = <A>(self: ReadonlyArray<A> | A): Array<A> => Array.isArray(self) ? self.slice() : of(self)
```

datner: `fromIterable` does not clone the array, and if it should then we should remove it since it would `===` the vanilla `Array.from` lol
```ts
export const fromIterable: <A>(self: Iterable<A>): Array<A> = Array.from // 🤨
export const ensure = <A>(self: ReadonlyArray<A> | A): Array<A> => Array.from([self].flat(1)) // if we allow cloning
```

tim-smart: Array.from performs a shallow copy:

image

datner: Is this an anecdote, or an argument for keeping `fromIterable` because it only clones non-`Array`s?

tim-smart: It is there because if you `import { Array } from "effect"`, accessing `Array.from` becomes annoying.

tim-smart:
> Shouldn't it take `Iterable<A>` as input like other APIs in the module?

I'm not so sure about this, as some of our data types implement Iterable (Effect, Option, Either) which would lead to confusing behaviour with this api.

packages/effect/src/Array.ts
datner: So in conclusion we need to have `fromIterable` do a `slice()` and so does `ensure`

datner:
Hmm It seems I need to fix the git tree again

tim-smart (approved):

PR #2830, opened 2024-05-23
add `TupleOf` and `TupleOfAtLeast` types, `isTupleOf` and `isTupleOfAtLeast` type guards

dilame:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- `isTupleOf`: Checks if an array has exactly `N elements.
- `isTupleOfAtLeast`: Checks if an array has at least `N` elements.

```ts
import { isTupleOf, isTupleOfAtLeast } from "effect/Array"

const arr: number[] = [1, 2, 3];

if (isTupleOf(arr, 3)) {
  arr;
  // ^? [number, number, number]
}

if (isTupleOfAtLeast(arr, 2)) {
  arr;
  // ^? [number, number, ...number[]]
}
```

It is often incredibly useful.

For example, i'm going to use this guard for implementation of `zipLatestAll` function, because i will need to check the length of the streams array for zipping

```ts
export const zipLatestAll = <A extends Array<Stream.Stream<unknown, unknown, unknown>>>(
  ...streams: A
): Stream.Stream<?> => {
  if (streams.length === 0) {
    return Stream.empty;
  }
  if (isTupleOf(streams, 1)) {
    return streams[0];
  }
  if (isTupleOfAtLeast(streams, 2)) {
    // can safely proceed to recursively call Stream.zipLatest(streams[0], streams[1])
  }
};
```


gcanti (changes requested):
I suggest the following changes:

- move `TupleOf` and `TupleOfAtLeast` to `Types` module
- add `// @ts-expect-error` to invalid examples to fix `docgen` failures
- add type-level tests for `TupleOf` and `TupleOfAtLeast` to `dtslint/Types.ts`:

  Example

  ```ts
  // TupleOf

  export const tupleOf1: Types.TupleOf<3, number> = [1, 2, 3]
  // @ts-expect-error
  export const tupleOf2: Types.TupleOf<3, number> = [1, 2]
  // @ts-expect-error
  export const tupleOf3: Types.TupleOf<3, number> = [1, 2, 3, 4]
  // @ts-expect-error
  export const tupleOf4: Types.TupleOf<3, number> = [1, "a", 3]

  // TupleOfAtLeast

  export const tupleOfAtLeast1: Types.TupleOfAtLeast<3, number> = [1, 2, 3]
  export const tupleOfAtLeast2: Types.TupleOfAtLeast<3, number> = [1, 2, 3, 4, 5]
  // @ts-expect-error
  export const tupleOfAtLeast3: Types.TupleOfAtLeast<3, number> = [1, 2]
  // @ts-expect-error
  export const tupleOfAtLeast4: Types.TupleOfAtLeast<3, number> = [1, 2, "a"]
  // @ts-expect-error
  export const tupleOfAtLeast5: Types.TupleOfAtLeast<3, number> = [1, 2, 3, "a"]
  ```

- move `isTupleOf` and `isTupleOfAtLeast` to `Predicate` module and re-export them from `Tuple` module for convenience
- add type-level tests for `isTupleOf` and `isTupleOfAtLeast` in `dtslint/Predicate.ts`
- add runtime tests for `isTupleOf` and `isTupleOfAtLeast` in `test/Predicate.ts`

dilame:
Thank you, @gcanti for clean explanation of your suggestions.

I am following it right now and have several notes about it:

1. I can not implement `TypeOf` type with just 2 input generics (as it should be for user) without introducing an internal `_TypeOf` type with 3 generics, because the third generic is internal. Is it ok? If so, where should i put `_TypeOf`?
2. I noticed that all the predicates in the `Predicate.ts` module accepts `(self: unknown)`, however, i can't do the same for the `isTupleOf` and `isTupleOfAtLeast`, because i need to infer the type of `Array<T>` in order to produce `TupleOf<T>`. Otherwise we just loose the info about tuple content type. Is it ok?

gcanti:
1) that's fine having an helper as long as `_TupleOf` is not actually "internal" (meaning that is not tagged with `/** @internal */`). Maybe just move `_TupleOf` up so we are sure the comment clearly applies to `TupleOf`
     ```ts
    type _TupleOf<T, N extends number, R extends Array<unknown>> = R["length"] extends N ? R : _TupleOf<T, N, [T, ...R]>

    /**
     * Represents ...
     * ...
     * @category ...
     * @since ...
     */
    export type TupleOf<N extends number, T> = N extends N ? number extends N ? Array<T> : _TupleOf<T, N, []> : never
    ```
2) in this case it's ok to start from `Array<T>` instead of `unknown`

gcanti:
> Maybe just move _TupleOf up so we are sure the comment clearly applies to TupleOf

should also fix the current CI failure btw

dilame:
Oh, and one more thing, is it correct to do like this?

```ts
export { isTupleOf, isTupleOfAtLeast } from "./Predicate.js"
```

packages/effect/src/Predicate.ts
```diff
@@ -52,6 +53,64 @@ export const mapInput: {
...
+/**
+ * Determine if an `Array` is a tuple with exactly `N` elements, narrowing down the type to `TupleOf`.
+ *
+ * An `Array` is considered to be a `TupleOf` if its length is exactly `N`.
+ *
+ * @param self - The `Array` to check.
+ * @param n - The exact number of elements that the `Array` should have to be considered a `TupleOf`.
+ *
+ * @example
+ * import { isTupleOf } from "effect/Array"
```
gcanti, line 65:
```suggestion
 * import { isTupleOf } from "effect/Predicate"
```

packages/effect/src/Predicate.ts
```diff
@@ -52,6 +53,64 @@ export const mapInput: {
...
+/**
+ * Determine if an `Array` is a tuple with at least `N` elements, narrowing down the type to `TupleOfAtLeast`.
+ *
+ * An `Array` is considered to be a `TupleOfAtLeast` if its length is at least `N`.
+ *
+ * @param self - The `Array` to check.
+ * @param n - The minimum number of elements that the `Array` should have to be considered a `TupleOfAtLeast`.
+ *
+ * @example
+ * import { isTupleOfAtLeast } from "effect/Array"
```
gcanti, line 94:
```suggestion
 * import { isTupleOfAtLeast } from "effect/Predicate"
```

packages/effect/src/Tuple.ts
```diff
@@ -202,3 +202,56 @@ export const appendElement: {
...
+  /**
+   * Determine if an `Array` is a tuple with exactly `N` elements, narrowing down the type to `TupleOf`.
+   *
+   * An `Array` is considered to be a `TupleOf` if its length is exactly `N`.
+   *
+   * @param self - The `Array` to check.
+   * @param n - The exact number of elements that the `Array` should have to be considered a `TupleOf`.
+   *
+   * @example
+   * import { isTupleOf } from "effect/Array"
```
gcanti, line 216:
```suggestion
   * import { isTupleOf } from "effect/Tuple"
```

gcanti:
@dilame there are still a couple of wrong imports in the jsdocs, btw you can run `docgen` locally and fix the errors before pushing:

```ts
cd packages/schema
pnpm docgen
```

dilame:
Thanks for pointing mew out to `docgen` script. I think i resolved everything

gcanti:
@dilame nice, I think we just need to update the changeset that still refers to Array and we're good to go

dilame:
ok, i will do it right now

gcanti (approved):

Issue #2718, opened 2024-05-09
Stream.zipLatestAll


dilame:
### What is the problem this feature would solve?

We have an extremely useful `Stream.concatAll` and `Stream.mergeAll` functions, but we don't have the same concept for zipping latest values.

### What is the feature you are proposing to solve the problem?

Implement 

```ts
type ZipLatestAllValues<T extends Array<Stream<unknown, unknown, unknown>>> = { [K in keyof T]: T[K] extends Stream<infer U> ? U : never }
type ZipLatestAllErrors<T extends Array<Stream<unknown, unknown, unknown>>> = T extends Array<Stream<unknown, infer E, unknown>> ? E : never;
type ZipLatestAllContext<T extends Array<Stream<unknown, unknown, unknown>>> = T extends Array<Stream<unknown, unknown, infer R>> ? R : never;

export declare const zipLatestAll: {
    <A extends Array<Stream<unknown, unknown, unknown>>>(...streams: A): Stream<ZipLatestAllValues<A>,  ZipLatestAllErrors<A>, ZipLatestAllContext<A>>;
};

// Stream.Stream<[number, string, boolean], never, never>
const latest = Stream.zipLatestAll(Stream.make(1,2,3), Stream.make('a','b','c'), Stream.make(true,false))

```

dilame:
Implemented!

PR #2832, opened 2024-05-23
feat(Stream): zipLatestAll

dilame:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR is based on my another PR https://github.com/Effect-TS/effect/pull/2830 so it is better to resolve it first

## Related

- Related Issue #

- Closes https://github.com/Effect-TS/effect/issues/2718


packages/effect/src/Stream.ts
```diff
@@ -4329,6 +4329,44 @@ export const zipLatest: {
...
+ * // Output:
+ * [ 0, 0, 0 ]
+ * [ 1, 0, 0 ]
+ * [ 1, 1, 0 ]
+ * [ 2, 1, 0 ]
+ * [ 3, 1, 0 ]
+ * [ 3, 1, 1 ]
+ * // .....
+ *
+ * @since 2.0.0
```
tim-smart, line 4359:
```suggestion
 * @since 3.3.0
```

packages/effect/src/internal/stream.ts
```diff
@@ -7565,6 +7566,34 @@ export const zipLatest = dual<
...
+  if (Arr.isTupleOfAtLeast(streams, 2)) {
+    const [head, ...tail] = streams
+    return zipLatestWith(
+      head,
+      zipLatestAll(...tail),
+      (first, second) => [first, ...second] as ZipLatestAllValues<T>
+    )
+  }
+  return empty
+}
```
tim-smart, line 7595: I think just using .length is easier to read and maintain.

```suggestion
): Stream.Stream<ZipLatestAllValues<T>, ZipLatestAllErrors<T>, ZipLatestAllContext<T>> => {
  if (streams.length === 0) {
    return empty
  } else if (streams.length === 1) {
    return map(streams[0], (x) => [x] as ZipLatestAllValues<T>))
  }
  const [head, ...tail] = streams
  return zipLatestWith(
    head,
    zipLatestAll(...tail),
    (first, second) => [first, ...second] as ZipLatestAllValues<T>
  )
}
```

packages/effect/src/Stream.ts
```diff
@@ -4329,6 +4329,44 @@ export const zipLatest: {
...
+ * @category zipping
+ */
+export const zipLatestAll: <T extends Array<Stream<unknown>>>(
```
tim-smart, line 4362:
```suggestion
export const zipLatestAll: <T extends ReadonlyArray<Stream<any, any, any>>>(
```

Otherwise you will run into assignability issues.

packages/effect/src/Stream.ts
```diff
@@ -4329,6 +4329,44 @@ export const zipLatest: {
...
+  ...streams: T
+) => Stream<
+  internal.ZipLatestAllValues<T>,
```
tim-smart, line 4365: Type helpers should not be internal, as they are stripped from the final build.

Also, I think these types should just be inlined :)

```suggestion
  T extends ReadonlyArray<Stream<infer _R, infer _E, infer _A>> ? _R : never,
```

Will need to test if this works with empty arguments too.

tim-smart (changes requested):

packages/effect/src/internal/stream.ts
dilame: If you are checking only length of the array you will not get type narrowing, so 

```ts
if(streams.length === 1) {
  streams[0]
  // ^? number | undefined
}
```

My suggestion is 

```ts
if(Preficate.isTupleOf(streams, 1)) {
  streams[0]
  // ^? number
}
```

My PR with `isTupleOf` has already been merged into `next-minor` (https://github.com/Effect-TS/effect/commit/3182c89e57c87c18dbe66271149ca6a23a5b3b1b), so i think it's a perfect solution here:)

packages/effect/src/Stream.ts
dilame: This is the suggestion that i don't understand, because AFAIU
`T extends ReadonlyArray<Stream<infer _R, infer _E, infer _A>> ? _A : never,` is not the same as `ZipLatestAllValues`.

I am also not sure what to do with the type helpers, because if we choose to inline it, we would need to do it in 2 both -`Stream.ts` and `internal/stream.ts`. These types will look pretty heavy inlined. Is it ok? What do you think about it?

packages/effect/src/internal/stream.ts
```diff
@@ -7565,6 +7566,34 @@ export const zipLatest = dual<
...
+  Array<Stream.Stream<unknown, infer E, unknown>> ? E : never
+/** @internal */
+export type ZipLatestAllContext<T extends Array<Stream.Stream<unknown, unknown, unknown>>> = T extends
+  Array<Stream.Stream<unknown, unknown, infer R>> ? R : never
+
+export const zipLatestAll = <T extends Array<Stream.Stream<unknown>>>(
+  ...streams: T
+): Stream.Stream<ZipLatestAllValues<T>, ZipLatestAllErrors<T>, ZipLatestAllContext<T>> => {
+  if (Arr.isTupleOf(streams, 1)) {
+    return pipe(streams[0], map((x) => [x] as ZipLatestAllValues<T>))
```
dilame, line 7584: BTW, @tim-smart , i also use `ZipLatestAllValues` type helper in 2 places inside the function (_here and 7 lines below from here_) for manual type-casting, because i can not think of a way to get rid of it. 
In total, `ZipLatestAllValues` is used in 4 places. Should i  still inline it?

dilame:
Seems like i did something wrong with merging:) Should i open a new PR?

packages/effect/src/internal/stream.ts
tim-smart: Yes still inline it :)

tim-smart:
> Seems like i did something wrong with merging:) Should i open a new PR?

Rebased and added some tests. There is one type level test that needs fixing.

dilame:
Seems like i fixed the problem with extra commits, and also fixed build errors, should be ok now

dilame:
I can see that pages build is failed, but the reason is
`Error: The operation was canceled.`
I'm not sure what to do in this case

packages/effect/src/Stream.ts
```diff
@@ -4350,6 +4350,45 @@ export const zipLatest: {
...
+ *
+ * Effect.runPromise(Stream.runDrain(stream))
+ * // Output:
+ * // [ 0, 0, 0 ]
+ * // [ 1, 0, 0 ]
+ * // [ 1, 1, 0 ]
+ * // [ 2, 1, 0 ]
+ * // [ 3, 1, 0 ]
+ * // [ 3, 1, 1 ]
+ * // .....
```
IMax153, line 4378: @dilame - I believe the reason docs generation is failing is because this example (which is run by our documentation generator to ensure validity` creates a stream that never terminates. Can you please update the example to the following:

```suggestion
 * @example
 * import { Stream, Schedule, Console, Effect } from "effect"
 *
 * const stream = Stream.zipLatestAll(
 *     Stream.fromSchedule(Schedule.spaced('5 millis')),
 *     Stream.fromSchedule(Schedule.spaced('10 millis')),
 *     Stream.fromSchedule(Schedule.spaced('20 millis')),
 * ).pipe(Stream.take(6), Stream.tap(Console.log))
 *
 * Effect.runPromise(Stream.runDrain(stream))
 * // Output:
 * // [ 0, 0, 0 ]
 * // [ 1, 0, 0 ]
 * // [ 1, 1, 0 ]
 * // [ 2, 1, 0 ]
 * // [ 3, 1, 0 ]
 * // [ 3, 1, 1 ]
 * // .....
```

tim-smart (approved):

IMax153 (approved):

Issue #2831, opened 2024-05-23
CLI Wizard Displays Incorrect Output for Optional Choices


gcanti:
### What version of Effect is running?

@effect/cli@0.36.34

### What steps can reproduce the bug?

The CLI wizard incorrectly displays the output for optional choice parameters. Instead of showing a clear and user-friendly way to indicate a lack of selection (`None`), it outputs a JSON-like string which is not intuitive for end-users.

1. Create a TypeScript file named `repro.ts` with the following content:
    ```ts
    import { Command, Options } from "@effect/cli"
    import { NodeContext, NodeRuntime } from "@effect/platform-node"
    import { Console, Effect } from "effect"

    const opt = Options.choice("a", ["b", "c"]).pipe(Options.optional)
    const command = Command.make("hello-world", { opt }, ({ opt }) => Console.log(opt))
    const cli = Command.run(command, {
      name: "Hello World CLI",
      version: "v1.0.0"
    })
    Effect.suspend(() => cli(process.argv)).pipe(
      Effect.provide(NodeContext.layer),
      NodeRuntime.runMain
    )
    ```
2. Run the command in a terminal:
    ```sh
    npx tsx repro.ts --wizard
    ```

### What is the expected behavior?

The CLI wizard should present a clear and user-friendly method to indicate that no option was selected.

### What do you see instead?

The output displayed is `Default ['{"_id":"Option","_tag":"None"}']?`, which appears to be a raw serialization of the internal state rather than a user-friendly message or indication.

### Additional information

Output:

```
% npx tsx repro.ts --wizard
Wizard Mode for CLI Application: Hello World CLI (v1.0.0)

Instructions

  The wizard mode will assist you with constructing commands for Hello World CLI (v1.0.0).

  Please answer all prompts provided by the wizard.

COMMAND: hello-world

Options Wizard - hello-world

? -a b | c
  
    One of the following: b, c
  
  This option is optional - use the default? › 
❯ Default ['{"_id":"Option","_tag":"None"}'] 
  Custom 
```

IMax153:
Closed by #2847.

PR #2847, opened 2024-05-24
display a nicer message in the cli wizard when using Option

ricardo-valero:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue https://github.com/Effect-TS/effect/issues/2831#issue-2313555269
- Closes #

### DIFF
--- a/packages/cli/src/internal/options.ts
+++ b/packages/cli/src/internal/options.ts
@@ -1428,10 +1428,16 @@ const wizardInternal = (self: Instruction, config: CliConfig.CliConfig): Effect.
         getHelpInternal(self.options as Instruction),
         InternalHelpDoc.sequence(defaultHelp)
       )
+      const defaultChoice = Option.isOption(self.fallback)
+        ? Option.match(self.fallback, {
+          onNone: () => ({ title: "None", description: "It has no value as default", value: true }),
+          onSome: (a) => ({ title: `Default [${JSON.stringify(a)}]`, value: true })
+        })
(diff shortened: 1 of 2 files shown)


packages/cli/src/internal/options.ts
```diff
@@ -1428,10 +1428,16 @@ const wizardInternal = (self: Instruction, config: CliConfig.CliConfig): Effect.
         getHelpInternal(self.options as Instruction),
         InternalHelpDoc.sequence(defaultHelp)
       )
+      const defaultChoice = Option.isOption(self.fallback)
+        ? Option.match(self.fallback, {
+          onNone: () => ({ title: "None", description: "It has no value as default", value: true }),
+          onSome: (a) => ({ title: `Default [${JSON.stringify(a)}]`, value: true })
```
IMax153, line 1434: @ricardo-valero - thanks for kick-starting this! 

Since the Wizard asks the user:

> This option is optional - use the default? ›

Perhaps to keep things a bit more semantic we could just make the options: 
- `Yes - use the default (<print default if available>)`
- `No - use a custom value`

What do you think?

IMax153 (changes requested):

packages/cli/src/internal/options.ts
ricardo-valero: Yes that's better

IMax153 (approved):
Thank you @ricardo-valero ! LGTM 👍

PR #2859, opened 2024-05-26
remove temp directory for FileSystem makeTempFileScoped

### LINKED ISSUE #2841: From Discord: Query about `fs.makeTempFileScoped()` behaviour
# Summary
### Summary

- **Wayne1726** observed that `fs.makeTempFileScoped()` creates a temporary directory and a temporary file within it. When the scope is closed, the file is removed, but the directory remains.
- **Patroza** initially thought this behavior was intentional.
- **Wayne1726** expressed concern that this could lead to thousands of empty directories if many files are processed.
- **Patroza** questioned why there would be thousands of directories and mentioned the use of the OS temp directory.
- **Wayne1726** clarified that `makeTempFileScoped` creates a subdirectory under the OS temp directory, leading to the accumulation of empty directories.
- **Patroza** acknowledged the issue and noted that it occurs when a prefix is set.

### Key Takeaways

1. **Behavior of `fs.makeTempFileScoped()`**: The function creates a temporary directory and a file within it. The file is cleaned up after the scope is closed, but the directory is not.
2. **Potential Issue**: This behavior can lead to the accumulation of empty directories in the OS temp directory, especially when processing many files.
3. **Clarification on Prefix**: The issue is particularly noticeable when a prefix is set for the temporary files.
4. **User Preference**: Wayne1726 prefers that both the temporary file and the directory be cleaned up to avoid clutter in the file system.

# Discord thread

https://discord.com/channels/795981131316985866/1243207712032952473

tim-smart:
fixes #2841

### DIFF
--- a/.changeset/old-trains-train.md
+++ b/.changeset/old-trains-train.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform-node-shared": patch
+---
+
+remove temp directory for FileSystem makeTempFileScoped
--- a/packages/platform-node-shared/src/internal/fileSystem.ts
+++ b/packages/platform-node-shared/src/internal/fileSystem.ts
@@ -380,11 +380,11 @@ const makeTempFile = makeTempFileFactory("makeTempFile")
 
 const makeTempFileScoped = (() => {
   const makeFile = makeTempFileFactory("makeTempFileScoped")
-  const removeFile = removeFactory("makeTempFileScoped")
+  const removeDirectory = removeFactory("makeTempFileScoped")
   return (options?: FileSystem.MakeTempFileOptions) =>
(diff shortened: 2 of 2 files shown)


wmaurer:
Thank-you very much!

patroza:
@tim-smart am I mistaken when it seems to me you're deleting too much?
```ts
      const prefix = options?.prefix ?? ""
      const directory = typeof options?.directory === "string"
        ? Path.join(options.directory, ".")
        : OS.tmpdir()

      return nodeMkdtemp(prefix ? Path.join(directory, prefix) : directory + "/")
``` I think that means depending on prefix, and directory options, you may be deleting a lot more than you bargained for

PR #2860, opened 2024-05-27
add @effect/sql-drizzle integration package

tim-smart:


datner:
What is the limitation regarding the non-select queries?

tim-smart:
> What is the limitation regarding the non-select queries?

https://github.com/drizzle-team/drizzle-orm/issues/287

mikearnaldi:
> > What is the limitation regarding the non-select queries?
> 
> [drizzle-team/drizzle-orm#287](https://github.com/drizzle-team/drizzle-orm/issues/287)

The issue is open since ages with no activity, I don't think we can rely on this

packages/sql-drizzle/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+)
+
+/**
+ * @since 1.0.0
+ */
+export const registerDialect = (dialect: unknown, client: Client.Client) => {
+  if (!(Effect.EffectTypeId in TypedQueryBuilder.prototype)) {
+    Object.assign(TypedQueryBuilder.prototype, {
+      ...Effectable.CommitPrototype,
+      commit(this: any) {
```
ecyrbe, line 22: maybe add a definition for toSQL :
```ts
import type { Query } from 'drizzle-orm/sql';

interface DrizzleQueryBuilder {
  toSQL: () => Query;
  dialect: unknown;
}
```

```suggestion
      commit(this: DrizzleQueryBuilder) {
```

mikearnaldi:
> Hello,
> 
> Just a litle feedback on this implementation. It's relying on internals of drizzle that are not exposed (ie: dialect) and may change without a breaking change because it's just internals.
> 
> Maybe one solution would be to use `entityKind` and register by name. Since QueryBuilders just return 'this', we can assume `entityKind` attribute string will be stable.
> 
> what do you think ? did i miss something ?

Even if, given that it doesn't support all kinds of queries I don't see how this can be useful, I think we need the drizzle team to advise if they are interested or not. At the moment we can support kysley without trouble

ecyrbe:
@mikearnaldi i found a way to use drizzle Database directly without using QueryBuilder. This way it supports typesafe select, update, insert and delete internal builders. 
i will provide a PR for discussion once i checked everything works

image

tim-smart:
> @mikearnaldi i found a way to use drizzle Database directly without using QueryBuilder. This way it supports typesafe select, update, insert and delete internal builders. i will provide a PR for discussion once i checked everything works
> 
> image

Inlay hints would drive me crazy lol.

ecyrbe:
PR should be ready for this week-end on my side. SQLite working, hopefully it will also work for PG and MySQL.

datner:
> inlay hints actually make me more productive, no mouse use to know the inferred types :)

*mouse use* ? in my code editor?? can't relate tbh 🤨

ecyrbe:
PR available here #2896

mikearnaldi:
```ts
declare module "drizzle-orm" {
  export interface QueryPromise<T> extends Effect.Effect<T, SqlError> {}
}
```

this is an issue, it will patch `QueryPromise` regardless of it being patched via the layer creation or not, leading to runtime errors

PR #2864, opened 2024-05-27
Add queuing strategy option for Stream.toReadableStream

### LINKED ISSUE #2840: Accept ReadableStream `queuingStrategy` in `Stream.toReadableStream`

When a Stream is converted to a standard ReadableStream with `Stream.toReadableStream`, the stream immediately begins pulling a value even though no attempts to read the stream have been made.
https://effect.website/play#5459e589747e

This is due to the default `queuingStrategy` of the ReadableStream constructor using a`highWaterMark` of 1.
https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream#queuingstrategy

This behavior is not always desirable—in my application I need to set a specific eagerness (similar to `Stream.buffer`), or you may wish for the stream to only pull values when read, matching the behavior of effect's Streams.

I propose an optional `queuingStrategy` argument for `Stream.toReadableStream`, which would get passed onto the underlying `ReadableStream` constructor.

```ts
const stream = Stream.make(1,2,3)

// Default queuingStrategy
const readableStream1 = Stream.toReadableStream(stream)

// Custom queuingStrategy
const queueingStrategy = new CountQueuingStrategy({ highWaterMark: 2 })
const readableStream2 = Stream.toReadableStream(stream, queueingStrategy)
```

This aligns with the syntax for the underlying [ReadableStream() constructor](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream#syntax), which takes an underlying source and an optional queingStrategy

```ts
new ReadableStream()
new ReadableStream(underlyingSource)
new ReadableStream(underlyingSource, queuingStrategy)
```

I thought perhaps `Stream.toReadableStream` could enforce a `highWaterMark` of 0 and let developers control stream eagerness with `Stream.buffer`, however that only allows buffering to start after a read has been attempted. Rather, I think ReadableStreams should be able to start pulling immediately but with a configurable limit.

mattrossman:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a "strategy" option to
- `Stream.toReadableStream`
- `Stream.toReadableStreamEffect`
- `Stream.toReadableStreamRuntime`

This allows configuring the queuing strategy of the underlying ReadableStream.

```ts

// Default queuing strategy
stream.pipe(Stream.toReadableStream)

// Custom queuing strategies
stream.pipe(Stream.toReadableStream({ strategy: { highWaterMark: 2 } }))
stream.pipe(Stream.toReadableStream({ strategy: new CountQueuingStrategy({ highWaterMark: 2 }) }))

```

The name "strategy" was chosen to align with the Streams spec and TypeScript lib types.

References:
- https://streams.spec.whatwg.org/#rs-prototype
- https://github.com/microsoft/TypeScript/blob/af3a61fe4487a92d59f9479aa4249d897b91af14/src/lib/dom.generated.d.ts#L18794-L18799

Note the diff for this PR shows many lines changed in `toReadableStreamRuntime`, these are mostly formatting changes incurred by the extra level of indentation from the expanded signature. The main real changes there are:
- the `ReadableStream` constructor receives a second argument passed down via `options?.strategy`
- the first argument of `dual()` has been changed from an arity of `2` to an `isStream` predicate matching other configurable stream functions 

## Related

- Closes #2840

### DIFF
--- a/.changeset/empty-islands-pump.md
+++ b/.changeset/empty-islands-pump.md
@@ -1,5 +1,5 @@
 ---
-"effect": patch
+"effect": minor
 ---
 
 Add queuing strategy option for Stream.toReadableStream
--- a/.changeset/empty-islands-pump.md
+++ b/.changeset/empty-islands-pump.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add queuing strategy option for Stream.toReadableStream
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -3860,7 +3860,17 @@ export const toQueueOfElements: {
  * @since 2.0.0
  * @category destructors
  */
-export const toReadableStream: <A, E>(self: Stream<A, E>) => ReadableStream<A> = internal.toReadableStream
+export const toReadableStream: {
+  <A>(
+    options?: { readonly strategy?: QueuingStrategy<A> | undefined }
+  ): <A, E>(
+    self: Stream<A, E>
+  ) => ReadableStream<A>
+  <A, E>(
+    self: Stream<A, E>,
+    options?: { readonly strategy?: QueuingStrategy<A> | undefined }
+  ): ReadableStream<A>
+} = internal.toReadableStream
 
 /**
  * Converts the stream to a `Effect<ReadableStream>`.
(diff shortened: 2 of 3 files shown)


.changeset/empty-islands-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
```
tim-smart, line 2:
```suggestion
"effect": minor
```

tim-smart (changes requested):
Looks good, just one small change.

.changeset/empty-islands-pump.md
mattrossman: Fixed. FYI I picked "patch" based on this README instruction, not sure if the pre-release mode part is still accurate in the 3.0 era.

https://github.com/Effect-TS/effect/blob/c12d7ff1935f74cf8d3c3e8e9396dd68dd52a56f/README.md?plain=1#L54

tim-smart (approved):

PR #2865, opened 2024-05-28
add concurrency control to Pool options

tim-smart:


packages/effect/test/Pool.test.ts
```diff
@@ -202,6 +217,70 @@ describe("Pool", () => {
       expect(max).toBe(15)
     }))
 
+  it.scoped("max pool size with concurrency: 3", () =>
+    Effect.gen(function*($) {
```
IMax153, line 221: I see us removing the adapter in some tests but keeping it elsewhere. I suppose it doesn't really matter for this PR, but perhaps we should consider stripping it out of the tests entirely at some point.

IMax153 (commented):
Only minor comments, otherwise this looks great to me - nice work @tim-smart!

Also, I know it's tested through usage but should we add explicit tests for the `timeToLiveStrategy` in the `Pool` options?

packages/effect/src/internal/pool.ts
```diff
@@ -34,467 +26,348 @@ const poolVariance = {
...
+      core.zipRight(fiberRuntime.scopeTag),
+      core.flatMap((scope) =>
+        core.suspend(() => {
+          this.waiters++
+          if (this.isShuttingDown) {
+            return core.interrupt
+          } else if (this.targetSize > this.activeSize) {
+            return core.zipRight(
+              restore(this.resize),
+              core.sync(() => Iterable.unsafeHead(this.available))
```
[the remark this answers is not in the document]
tim-smart: It needs to be lazy here, as the resize will add items.

packages/effect/test/Pool.test.ts
tim-smart: The ts inference issue keeps popping up, so I think we should wait until 5.5 lands.

PR #2895, opened 2024-06-01
update and simplify flake.nix

tim-smart:


IMax153 (approved):

IMax153:
Nice simplification 👍 - Didn't realize the `corepack` Nixpkg runs the install script now.

PR #2896, opened 2024-06-01
Feature: add Drizzle support @effect/sql-drizzle

ecyrbe:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR add support for drizzle in effect.
It uses internal Drizzle DB, and disable drizzle runtime execution (execute and transaction) to leverage on effect sql client

## example
```ts
const SqlLive = Sqlite.client.layer({
  filename: Config.succeed("test.db")
})
const DrizzleLive = SqliteDrizzle.layer.pipe(
  Layer.provide(SqlLive)
)
const DatabaseLive = Layer.mergeAll(SqlLive, DrizzleLive)

// usage

const users = D.sqliteTable("users", {
  id: D.integer("id").primaryKey(),
  name: D.text("name")
})

Effect.gen(function*() {
  const sql = yield* Sql.client.Client
  const db = yield* SqliteDrizzle.SqliteDrizzle
  yield* sql`CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT)`
  const inserted = yield* db.insert(users).values({ name: "Alice" }).returning()
  yield* Console.log(inserted)
  const selected = yield* db.select().from(users)
  yield* Console.log(selected)
  const updated = yield* db.update(users).set({ name: "Bob" }).where(gte(users.id, 1)).returning()
  yield* Console.log(updated)
  const deleted = yield* db.delete(users).where(gte(users.id, 1)).returning()
  yield* Console.log(deleted)
}).pipe(
  Effect.provide(DatabaseLive),
  Effect.runPromise
)
```
## Related

- Related Issue #
- Closes #


packages/sql-drizzle/test/pg.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  id: D.serial("id").primaryKey(),
+  name: D.text("name")
+})
+
+describe("PgDrizzle", () => {
+  let pgContainer: StartedPostgreSqlContainer
+  let makeClient: Effect.Effect<Pg.client.PgClient, never, Scope>
+
+  beforeAll(async () => {
+    pgContainer = await new PostgreSqlContainer("postgres:alpine").start()
```
ecyrbe, line 23: using testcontainer, not sure that's the standard way to test db in effect. tell me if it's ok.

packages/sql-drizzle/src/pg.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @category constructors
+ */
+export const make: Effect.Effect<DrizzlePgDatabase, never, Client.Client> = Effect.gen(function*() {
+  const client = yield* Client.Client
+  // instanciate the db without a client, since we are going to attach the client to the QueryBuilder
+  const db = drizzle({
+    options: {
+      parsers: [],
+      serializers: []
+    }
```
ecyrbe, line 34: This is to make drizzle patching of postgres-js client instance working. drizzle patching is not used in our case, since db instance is just used to convert query builder postgres dialect to sql

packages/sql-drizzle/test/pg.test.ts
mikearnaldi: testcontainer is really great, I usually wrap it in a layer so I can have a TestDb that spins up and down automatically, but this looks fine too

mikearnaldi:
```
packages/sql-drizzle/src/internal/registry.ts
```

Seems like the above file creates a global map Dialect<->Client but the Client is scoped and we might have many different clients in the same project.

Additionally, patching the global drizzle type would make every client effectable, even if registration didn't happen.

I think a better idea would be to yes make every drizzle query effectable but use `Effect<T, SqlError, Client>` so that the client is passed to the query when used

ecyrbe:
> ```
> packages/sql-drizzle/src/internal/registry.ts
> ```
> 
> Seems like the above file creates a global map Dialect<->Client but the Client is scoped and we might have many different clients in the same project.
> 

This is explained in the comment of the code. We store the `instance` of the dialect, and a new instance of the dialect is created for each db... so each client will have a different dialect associated with it, effectively allowing to have many clients in the same project.

> Additionally, patching the global drizzle type would make every client effectable, even if registration didn't happen.
> 
> I think a better idea would be to yes make every drizzle query effectable but use `Effect<T, SqlError, Client>` so that the client is passed to the query when used

Nice idea, it would indeed remove this global registry. i'll try it to see if it works and report back ASAP.

mikearnaldi:
> > ```
> > packages/sql-drizzle/src/internal/registry.ts
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Seems like the above file creates a global map Dialect<->Client but the Client is scoped and we might have many different clients in the same project.
> 
> This is explained in the comment of the code. We store the `instance` of the dialect, and a new instance of the dialect is created for each db... so each client will have a different dialect associated with it, effectively allowing to have many clients in the same project.
> 
> > Additionally, patching the global drizzle type would make every client effectable, even if registration didn't happen.
> > I think a better idea would be to yes make every drizzle query effectable but use `Effect<T, SqlError, Client>` so that the client is passed to the query when used
> 
> Nice idea, it would indeed remove this global registry. i'll try it to see if it works and report back ASAP.

Ok so it isn't the dialect itself but a specific instance created with it's own client, then I think it could work, it might even be better compared to `Effect<T, SqlError, Client>` as it would capture the client on creation but the issue remains on having all db instances effectable regardless of the client being attached.

Also small note, `packages/sql-drizzle/src/internal/drizzle-patch.types.ts` should not be on `/internal` given that it is exposed

ecyrbe:
> Ok so it isn't the dialect itself but a specific instance created with it's own client, then I think it could work, it might even be better compared to `Effect<T, SqlError, Client>` as it would capture the client on creation but the issue remains on having all db instances effectable regardless of the client being attached.

I have a solution for that, patch SQLiteSelectBase, SQLiteInsertBase directly. that's what i did first. but the types are ugly :)
It's less pretty to maintain, but i can do it.

> Also small note, `packages/sql-drizzle/src/internal/drizzle-patch.types.ts` should not be on `/internal` given that it is exposed

Indeed.

packages/sql-drizzle/src/internal/drizzle-patch.types.ts
```diff
@@ -0,0 +1,10 @@
+import type { SqlError } from "@effect/sql/Error"
+import type * as Effect from "effect/Effect"
+/**
+ * QueryPromise class is used for every Drizzle QueryBuilders, it's what allows Builders to be awaitable
+ * So since we need all QueryBuilders to also be effectable, we patch it's interface to extend Effect
+ * We don't however monkey patch the QueryPromise class itself, but the QueryBuilders that extend it to be able to attach the right client to them
```
tim-smart, line 6: Patching QueryPromise might be cleaner. Is it exposed?

ecyrbe: yes, it's exposed

ecyrbe:
So, i updated the PR.
It now uses dependency injection for client instead of registry. I feel it's safer this way.
It allows to no have any dependency on private `dialect` property, making this implementation more stable in time
I added the type patching at Builder level instead of `QueryPromise`. Like i sead it's ugly, and risky, since any parameter change in the builder types (renaming, adding new ones) will break typescript inference for users.

I feal that the best combination is to :
- patch QueryBuilders without registry and use dependency injection
- patch QueryPromise type as it only have one type parameter making it more stable

what do you think ? do i revert back to QueryPromise type Patching ?

packages/sql-drizzle/src/internal/drizzle-patch.types.ts
ecyrbe: But patching QueryPromise with dependency injecting the client will patch all instances (not just the one we declared), it's indeed doable, but @mikearnaldi has concerns about patching globally the types, so patching the propotype globally would be even more intrusive.

mikearnaldi: > But patching QueryPromise with dependency injecting the client will patch all instances (not just for the client we declared), it's indeed doable, but @mikearnaldi has concerns about patching globally the types, so patching the propotype globally would be even more intrusive.

I don't really have concerns if the patching is valid, the issue would be patching instances globally that can't be used in an effect environment, if the patch adds `Effect<X, SqlError, Client>` it is valid because the client will need to be provided so as long as augmentation and patching are done at the same time it should work

ecyrbe: So i tested it, and patching QueryPromise does not work. No idea why, i get this error :
```md
Unknown Error: TypeError: yield* (intermediate value)(intermediate value)(intermediate value)(intermediate value)(intermediate value)(intermediate value) is not iterable
 ❯ test/sqlite.test.ts:31:43
     29|       const inserted = yield* db.insert(users).values({ name: "Alice" }).returning()
     30|       assert.deepStrictEqual(inserted, [{ id: 1, name: "Alice" }])
     31|       const selected = yield* db.select().from(users)
       |                                           ^
```
have you ever seen it ? If no solution, i'll keep the patching at builder level.

tim-smart: I have another idea, will try it out and report back if it works.

tim-smart: Updated #2860

ecyrbe:
closing, since tim re-took the lead on this.

PR #2910, opened 2024-06-03
rethrow assertion errors in @effect/vitest

tim-smart:


mikearnaldi:
this is not ideal as the thrown error won't contain the refined stack traces, this should be done by prettyErrors

mikearnaldi:
This: https://github.com/Effect-TS/effect/blob/main/packages/effect/src/internal/cause.ts#L993 was enough in my tests to have the error shown properly

mikearnaldi:
Screenshot 2024-06-04 at 08 44 34

Something is odd, it looks like the error coming from PrettyError is recognised as UnknownError and the AssertionError becomes part of the message, this is not the ideal behaviour, PrettyError should be basically a clone of the original error with a refined stack. 

Mutating the original error isn't really great, it could lead to undefined behaviour

mikearnaldi:
It appears the issue is the `toJSON` method, removing it leads to correct behaviour

mikearnaldi:
https://github.com/Effect-TS/effect/pull/2912 fixes the issue

PR #2917, opened 2024-06-04
Fix Unify for Stream

mikearnaldi:


tim-smart:
@mikearnaldi ready to merge?

mikearnaldi:
> @mikearnaldi ready to merge?

think so, unless I missed something

dilame:
I can confirm it fixes the Unify issue and doesn't break anything in a large codebase

PR #2920, opened 2024-06-04
fix print to stdout of cli (#1812)

### LINKED ISSUE #1812: Prints to stdout rather than stderr

effect@npm:2.0.0-next.62

Follow the “minigit” example in README and invoke the command.

Messages to user are printed to stderr by default.

### What do you see instead?

Messages are printed to stdout.

### Additional information

Printing to stdout (unless specifically requested, such as with a `--help` flag) disrupts piping and is not conducive to writing well-behaved CLI commands.

Iri-Hor:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Updates the CLI Package: The error message for invalid commands will be printed out to stderr instead of stdout, as indicated by the related issue below.

## Related

- Closes #1812

### DIFF
--- a/.changeset/lazy-keys-wait.md
+++ b/.changeset/lazy-keys-wait.md
@@ -1,0 +1,5 @@
+---
+"@effect/cli": patch
+---
+
+in case of error let cli print to stderr instead of stdout (#1812)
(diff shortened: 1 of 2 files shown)


IMax153 (commented):
While this is definitely a bugfix, it could also be viewed by some as a "breaking" change in that it alters which file descriptor folks expect their logs to be coming from. 

@mikearnaldi @tim-smart - what do you folks think? Should this go out as a minor?

mikearnaldi:
Bug fixes are kind of breaking by definition :) I'd go with a patch

PR #2927, opened 2024-06-05
add `@effect/cluster-*` packages

fubhy:


jessekelly881:
I wonder if it would make sense to start to organize packages by subfolder i.e. /cluster/browser /cluster/pg etc to make things easier to find (in general, not just w/ the cluster packages).

PR #2940, opened 2024-06-06
feat(PubSub): unbounded replay

dilame:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Implementation of replay last N values feature for `PubSub.unbounded(replayBufferSize: N = 0)`.
Can further be used to create shared `Stream` https://github.com/Effect-TS/effect/pull/2943.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/clever-walls-smoke.md
+++ b/.changeset/clever-walls-smoke.md
@@ -1,0 +1,5 @@
+---
+"@effect/sql-d1": minor
+---
+
+Add new cloudflare @effect/sql-d1 package
--- a/.changeset/gorgeous-eels-eat.md
+++ b/.changeset/gorgeous-eels-eat.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+feat(pubsub): unbounded replay N last values
--- a/.changeset/lucky-poets-guess.md
+++ b/.changeset/lucky-poets-guess.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Add `Success` type util for `Config`.
--- a/packages/effect/dtslint/Config.ts
+++ b/packages/effect/dtslint/Config.ts
@@ -1,5 +1,5 @@
 import * as Config from "effect/Config"
-import { pipe } from "effect/Function"
+import { hole, pipe } from "effect/Function"
 
 declare const string: Config.Config<string>
 declare const number: Config.Config<number>
@@ -45,3 +45,18 @@ Config.all(numberRecord)
 
 // $ExpectType Config<{ [x: string]: number; }>
 pipe(numberRecord, Config.all)
+
+// -------------------------------------------------------------------------------------
+// Success
+// -------------------------------------------------------------------------------------
+
+// $ExpectType string
+hole<Config.Config.Success<typeof string>>()
+
+// $ExpectType number
+hole<Config.Config.Success<typeof number>>()
+
+const object = Config.all({ a: string, b: number })
+
+// $ExpectType { a: string; b: number; }
+hole<Config.Config.Success<typeof object>>()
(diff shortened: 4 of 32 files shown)


tim-smart:
I think if we were going to add a replay option, we would need to add it to every strategy.

Also switch to object arguments:

```ts
Pubsub.unbounded<string>({ replaySize: 16 })
```

dilame:
I'm not sure how replay should behave for all the other strategies. 
FMPOV it only makes sense for unbounded, because it's the only unlimited strategy.
For example, `PubSub.bounded<string>({capacity: 2, replay: 3})` looks contradictory to me, same as `PubSub.dropping<string>({capacity: 2, replay: 3})`. But maybe i am wrong. 
I would suggest to focus on `unbounded` strategy in this PR.

I like the idea about object argument. What do you think about `replayBufferSize` name?

tim-smart:
> `PubSub.bounded<string>({capacity: 2, replay: 3})` looks contradictory to me

capacity and replay could be separate buffers, so you could have a higher replay buffer compared to the capacity. Or the capacity could be the ceiling.

Also I think the replay doesn't need to be tracked per strategy, but rather on the PubSubImpl. The replay buffer can be independent of the subscriber-based logic.

dilame:
> Also I think the replay doesn't need to be tracked per strategy, but rather on the PubSubImpl. The replay buffer can be independent of the subscriber-based logic.

I don't see a way to implement it in the `PubSubImpl` within the current architecture because we need to pass a pointer to the node from where to start to the new subscription (https://github.com/dilame/effect/blob/pubsub-unbounded-replay/packages/effect/src/internal/pubsub.ts#L754). The new subscription is being created at the `UnboundedPubSub` level, and `UnboundedPubSub` doesn't have any pointers to the `PubSubImpl`, so even if we track the replay in the `PubSubImpl`, we will not have access to it from the place where we need it.

I think it is possible to reimplement the whole `PubSub` module from scratch with a new architecture that will track replay on a higher abstraction level, but is it really needed?

mikearnaldi:
Not sure replay should be part of pub-sub, but if we decide for it then it should not be part of a specific strategy, I'd at least explore the design where we track replay independently

dilame:
The `replay` feature is pretty commonly used. In RxJS – [ReplaySubject](https://rxjs.dev/api/index/class/ReplaySubject) is one of the most important classes. 
I implemented this feature for myself, because i really need it in my project. I decided to make it part of `PubSub` because `PubSub.unbounded` is the most similar entity to the RxJS `Subject`. I just don't see any other entities in Effect to make it part of. Maybe you have any thought about it, @mikearnaldi ?

TBH i wasn't sure if it should be a part of `unbounded` strategy. Maybe we just should make a new strategy specifically for that – `PubSub.replay(N)` and leave the `PubSub.unbounded` as is?

dilame:
Another reason why i think it should be a `PubSub` is because it will enable us to make a replayable shared stream with the `share()` operator from my another PR https://github.com/Effect-TS/effect/pull/2943, i think it is a really powerful and flexible symbiosis

mikearnaldi:
> Another reason why i think it should be a `PubSub` is because it will enable us to make a replayable shared stream with the `share()` operator from my another PR #2943, i think it is a really powerful and flexible symbiosis



> Another reason why i think it should be a `PubSub` is because it will enable us to make a replayable shared stream with the `share()` operator from my another PR #2943, i think it is a really powerful and flexible symbiosis



> Another reason why i think it should be a `PubSub` is because it will enable us to make a replayable shared stream with the `share()` operator from my another PR #2943, i think it is a really powerful and flexible symbiosis

wouldn't share be memory unsafe?

dilame:
> wouldn't share be memory unsafe?

I don't see any reason for it. Why do you think it could be?
Actually, i just copied the code for`share()` from some of `broadcast` operators, i don't remember which exactly. But conceptually `share` is just more flexible `broadcast`

PR #2954, opened 2024-06-07
Fix runtime error in `Struct.evolve` by enhancing compile-time checks, closes #2953

### LINKED ISSUE #2953: `Struct.evolve` no compile time error for non function transformers

Coming from the archived repository, I was suggesting to make some adjustments to the `Struct.evolve` function in order for it to spot some cases where right now it errors at runtime instead of compile time.

- effect version `3.3.0`
- [old issue](https://github.com/Effect-TS/data/issues/376)

repro:
```ts
const mystruct = { a: 42, b: `hello` }
const bad = Struct.evolve(mystruct, Record.fromEntries([[`a`, `not a function`]])) // not an error ❌
```

Adjusting of the `Struct.evolve` type signature to actively check for the transformation object.

Suggested solution (at least works for me, of course not well battle tested):
```ts
type Transformed<O, T> = unknown & {
    [K in keyof O]: K extends keyof T ? (T[K] extends (...a: any) => any ? ReturnType<T[K]> : O[K]) : O[K]
}
type PartialTransform<O, T> = {
    [K in keyof T]: T[K] extends (a: O[K & keyof O]) => any ? T[K] : (a: O[K & keyof O]) => unknown
}

export const evolve: {
    <O, T>(t: PartialTransform<O, T>): (obj: O) => Transformed<O, T>
    <O, T>(obj: O, t: PartialTransform<O, T>): Transformed<O, T>
} = Struct.evolve


const mystruct = { a: 42, b: `hello` }
const bad = Struct.evolve(mystruct, Record.fromEntries([[`a`, `not a function`]])) // not an error ❌
const good = evolve(mystruct, Record.fromEntries([[`a`, `not a function`]])) // compile error ✅
```

Notice this is allowing for compile time known / typed transformations object with no overlapping with original struct properties.

That is:
```ts
const actual = Struct.evolve(mystruct, { c: () => 42 }) // compile error ✅
const regression = evolve(mystruct, { c: () => 42 }) // not an error ❌
```

This last corner case / regression may be solvable though.

_No response_

jessekelly881:
Closes #2953

### DIFF
--- a/.changeset/long-dodos-poke.md
+++ b/.changeset/long-dodos-poke.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+made Struct.evolve more strict
--- a/.changeset/long-dodos-poke.md
+++ b/.changeset/long-dodos-poke.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-made Struct.evolve more strict
+Fix runtime error in `Struct.evolve` by enhancing compile-time checks, closes #2953
--- a/packages/effect/src/Struct.ts
+++ b/packages/effect/src/Struct.ts
@@ -119,14 +119,14 @@ export const getOrder: <R extends { readonly [x: string]: order.Order<any> }>(
   fields: R
 ) => order.Order<{ [K in keyof R]: [R[K]] extends [order.Order<infer A>] ? A : never }> = order.struct
 
-type PartialTransform<O> = Partial<{ [K in keyof O]: (a: O[K]) => unknown }>
-
-type Transformed<O, T extends PartialTransform<O>> =
+type Transformed<O, T> =
   & unknown
   & {
-    [K in keyof O]: K extends keyof T ? T[K] extends (...a: any) => any ? ReturnType<T[K]> : O[K] : O[K]
+    [K in keyof O]: K extends keyof T ? (T[K] extends (...a: any) => any ? ReturnType<T[K]> : O[K]) : O[K]
   }
-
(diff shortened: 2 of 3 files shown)


.changeset/long-dodos-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+made Struct.evolve more strict
```
gcanti, line 5:
```suggestion
Fix runtime error in `Struct.evolve` by enhancing compile-time checks, closes #2953
```

packages/effect/src/Struct.ts
```diff
@@ -148,13 +148,11 @@ type Transformed<O, T extends PartialTransform<O>> =
...
-  <O, T extends PartialTransform<O>>(obj: O, t: T): Transformed<O, T>
+  <O, T>(t: PartialTransform<O, T>): (obj: O) => Transformed<O, T>
+  <O, T>(obj: O, t: PartialTransform<O, T>): Transformed<O, T>
 } = dual(
   2,
-  <O, T extends PartialTransform<O>>(
+  <O, T extends PartialTransform<O, T>>(
     obj: O,
     t: T
   ): unknown & Transformed<O, T> => {
```
gcanti, line 158:
```suggestion
<O, T>(obj: O, t: PartialTransform<O, T>): Transformed<O, T> => {
```

gcanti (changes requested):

gcanti (approved):

PR #2919, opened 2024-06-04
Add `Either.liftPredicate`

LaureRC:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add `Either.liftPredicate`. In my current project we were missing it so we created a custom helper for it, but I found out in the Discord that other people would like it to be implemented (see thread [here](https://discord.com/channels/795981131316985866/1022125411485896708/1241130231666970744)). So I figured it could be an opportunity to contribute here!

I took inspiration from `Option.fromPredicate` but I went a bit further and made `Either.liftPredicate` dual

## Related

N/A

PR #2956, opened 2024-06-08
Add `Effect.liftPredicate`

LaureRC:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add `Effect.liftPredicate` (similar to recently merged `Either.liftPredicate`: https://github.com/Effect-TS/effect/pull/2919). This is a proposition though, I'm not 100% sure it's really needed, as we might be able to use `Either.liftPredicate` (at least that's the case for the use cases I was looking into in my current project). So I'll keep the PR as a draft for now to see if I find some good use cases where it's needed, otherwise I'll close the PR. Happy to get some feedback about the idea though!

## Related

N/A

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -2160,6 +2160,61 @@ export const uninterruptibleMask: <A, E, R>(
 ) => Effect<A, E, R> = core.uninterruptibleMask
 
 // -------------------------------------------------------------------------------------
+// lifting
+// -------------------------------------------------------------------------------------
+
+/**
+ * Transforms a `Predicate` function into an `Effect` returning the input value if the predicate returns `true`
+ * or failing with specified error if the predicate fails
+ *
+ * @param predicate - A `Predicate` function that takes in a value of type `A` and returns a boolean.
+ *
+ * @example
+ * import { pipe, Effect } from "effect"
+ *
+ * const isPositive = (n: number): boolean => n > 0
+ *
+ * assert.deepStrictEqual(
+ *   pipe(
+ *     1,
+ *     Effect.liftPredicate(isPositive, n => `${n} is not positive`)
+ *   ),
+ *   Effect.succeed(1)
+ * )
+ * assert.deepStrictEqual(
+ *   pipe(
+ *     0,
+ *     Effect.liftPredicate(isPositive, n => `${n} is not positive`)
+ *   ),
+ *   Effect.fail("0 is not positive")
+ * )
+ *
+ * @category lifting
+ * @since 3.4.0
+ */
+export const liftPredicate: {
+  <A, B extends A, E>(refinement: Refinement<NoInfer<A>, B>, orFailWith: (a: NoInfer<A>) => E): (a: A) => Effect<B, E>
+  <A, E>(
+    predicate: Predicate<NoInfer<A>>,
+    orFailWith: (a: NoInfer<A>) => E
+  ): (a: A) => Effect<A, E>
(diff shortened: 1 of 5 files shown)


tim-smart:
I think the Effect.filter* apis are enough:

```ts
Effect.succeed(0).pipe(
  Effect.filterOrFail(...)
)
```

I'll let someone else weigh in before we decide to continue / close this PR.

LaureRC:
Thank you for your reviews!

> I think the Effect.filter* apis are enough:

I understand your point, I thought it would be more convenient to directly have this `liftPredicate`, rather than have to wrap the initial value in an `Effect.succeed` to be able to use the `Effect.filter*` methods

> That's right, this is literally Either.liftPredicate typed differently:

Right, but then does that mean you don't recommend adding it for Effect? We can use `Either.liftPredicate` which is going to be released in the next minor anyway so I'm happy with that 🙂 but apparently some people do not use `Either` that much, so it could be convenient for them to not have to import (and know) about another module than `Effect`. I'm aware this is not a super strong argument though, so if you think that's not interesting enough I won't be mad if we close this subject 😄

gcanti:
Actually, I'm not sure they are *literally* the same,`Effect.succeed`/`Effect.fail` and `Either.right`/`Either.left` return different values.

LaureRC:
> Actually, I'm not sure they are literally the same,Effect.succeed/Effect.fail and Either.right/Either.left return different values.

Indeed, in some cases I identified in my current project, it does not change much to use the Either one. As I said before, I was pushing a little bit for this `Effect.liftPredicate` as I still think it could have value to have this method in both `Either` and `Effect` modules, to avoid having to "switch context" and use the `Either` module if not really needed. 
So I'll reopen the PR as I still think it could be useful, but don't hesitate to push back on it if you disagree 🙂 (as I didn't get an explicit no-go for now I'll insist a little bit more 😄)

packages/effect/src/Effect.ts
```diff
@@ -2159,6 +2159,61 @@ export const uninterruptibleMask: <A, E, R>(
...
+    self: A,
+    refinement: Refinement<A, B>,
+    orFailWith: (a: A) => E
+  ): Effect<B, E>
+  <A, E>(
+    self: A,
+    predicate: Predicate<NoInfer<A>>,
+    orFailWith: (a: NoInfer<A>) => E
+  ): Effect<A, E>
+} = dual(
```
tim-smart, line 2211: If you can move the implementation into `internal/core-effect.ts`, to be consistent with the rest of the apis.

packages/effect/src/Effect.ts
```diff
@@ -2159,6 +2159,61 @@ export const uninterruptibleMask: <A, E, R>(
...
+  3,
+  <A, E>(a: A, predicate: Predicate<A>, orFailWith: (a: A) => E): Effect<A, E> =>
+    predicate(a) ? succeed(a) : fail(orFailWith(a))
```
tim-smart, line 2214: This needs to be wrapped in `Effect.suspend`, in case the predicate throws an error.

tim-smart (changes requested):

tim-smart:
I'll make the changes now to get this into the next minor release.

tim-smart (approved):

LaureRC:
> I'll make the changes now to get this into the next minor release.

Thank you for handling this! I'm glad to see this released 😊

PR #2974, opened 2024-06-10
feat(Stream): switchMap

dilame:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #


tim-smart:
What is the benefit over `flatMap(f, { switch: true })`?

mikearnaldi:
I guess just verbosity, not sure if this is common enough to justify it

dilame:
RxJS also do have a [switchMap](https://rxjs.dev/api/index/function/switchMap) operator, and it's widely used.

I decided to make it as a separate operator not just because of verbosity, but because of clarity and visualization.

This is a real code from my project

```ts
export const BinanceWsUnitLayer = Layer.effect(
  BinanceWsStreamTag,
  Effect.gen(function* () {
    const connection = yield* BinanceWsConnectionContextTag;
    const streams = new Map<
      string,
      Stream.Stream<
        unknown,
        | BinanceWsConnectionError
        | BinanceWsResponseError
        | BinanceWsConnectionCloseError
      >
    >();
    return {
      createStream(
        stream: string,
      ): Stream.Stream<
        unknown,
        | BinanceWsConnectionError
        | BinanceWsResponseError
        | BinanceWsConnectionCloseError
      > {
        return Stream.suspend(() => {
          debug(`Requested stream "${stream}"`);
          let stream$ = streams.get(stream);
          if (stream$) {
            return stream$;
          }
          stream$ = connection.pipe(
            Stream.flatMap(
              (ws) =>
                Effect.request(
                  BinanceWsSubscribeRequest({ stream, ws }),
                  RequestResolver.contextFromEffect(
                    BinanceWsSubscribeRequestResolver,
                  ),
                ),
              { switch: true, concurrency: 1 },
            ),
            Stream.flatten(),
          );
          streams.set(stream, stream$);
          return stream$;
        });
      },
    } satisfies BinanceWsStream;
  }),
);
```

It is not really easy to notice the `options` parameter for `flatMap`. When you fluently read this code, you only see `Stream.flatMap`, but default `flatMap` behaviour is different. In addition, there are no clear documentation about what does `switch` option means. On the site it is written

> if the order of concatenation is not important for us, we can use the `switch` option.

And it doesn't describes this concept. 

Actually, i thinks that the `switch` option changes the behaviour dramatically enough to extract it to the separate operator with it's own documentation and example, which i also provided in this PR.

mikearnaldi:
@tim-smart what should we do? I am kind of in favour of accepting this given the semantic specificity

tim-smart:
> @tim-smart what should we do? I am kind of in favour of accepting this given the semantic specificity

I think we can add it back as `flatMapSwitch` (what it was called before we consolidated the apis).

We would then mark the `switch` option as depreciated and remove it in 4.0?

dilame:
@mikearnaldi done

dilame:
> @dilame can you pls fix the version in `since` to 3.5 and I'll merge it

Hi! Just wanted to remind that i changed the version in `since` that time to `3.5.0` and just now i changed it to `3.6.0` 😃

fubhy:
Closing this as stale. The branch is now conflicted and includes unrelated already-merged sql-kysely commits, with no current owner or active review path. If this API is still wanted, please open a fresh focused PR against the current branch.

PR #2976, opened 2024-06-10
Implement `at` method for tuples

### LINKED ISSUE #2975: Implement `at` method for tuples

The tuple utility module lacks a method to retrieve an element by its index, which limits its functionality.

Implement the at method to retrieve an element at a specified index from a tuple. This method was previously marked as `TODO`.

[Here's](https://github.com/Effect-TS/effect/blob/f085f92dfa204afb41823ffc27d437225137643d/packages/effect/src/Tuple.ts#L201)

msensys:
Closes #2975

### DIFF
--- a/packages/effect/src/Tuple.ts
+++ b/packages/effect/src/Tuple.ts
@@ -209,9 +209,9 @@ export const appendElement: {
  * @since 3.4.0
  */
 export const at: {
-  <A extends ReadonlyArray<unknown>>(index: number): (self: A) => A[number]
-  <A extends ReadonlyArray<unknown>>(self: A, index: number): A[number]
-} = dual(2, <A extends ReadonlyArray<unknown>>(self: A, index: number): A[number] => self[index])
+  <A extends ReadonlyArray<unknown>, N extends number>(index: N): (self: A) => A[N]
+  <A extends ReadonlyArray<unknown>, N extends number>(self: A, index: N): A[N]
+} = dual(2, <A extends ReadonlyArray<unknown>, N extends number>(self: A, index: N): A[N] => self[index])
 
 export {
   /**
--- a/packages/effect/src/Tuple.ts
+++ b/packages/effect/src/Tuple.ts
@@ -201,7 +201,7 @@ export const appendElement: {
  * @param index - The index of the element to retrieve.
  *
  * @example
- * import { at } from "effect/Tuple"
+ * import { Tuple } from "effect"
  *
  * assert.deepStrictEqual(at([1, 'hello', true], 1), 'hello')
  *
--- a/packages/effect/src/Tuple.ts
+++ b/packages/effect/src/Tuple.ts
@@ -203,7 +203,7 @@ export const appendElement: {
  * @example
  * import { Tuple } from "effect"
  *
- * assert.deepStrictEqual(at([1, 'hello', true], 1), 'hello')
+ * assert.deepStrictEqual(Tuple.at([1, 'hello', true], 1), 'hello')
  *
  * @category getters
  * @since 3.4.0
--- a/packages/effect/src/Tuple.ts
+++ b/packages/effect/src/Tuple.ts
@@ -194,14 +194,24 @@ export const appendElement: {
   <A extends ReadonlyArray<unknown>, B>(self: A, that: B): [...A, B]
 } = dual(2, <A extends ReadonlyArray<unknown>, B>(self: A, that: B): [...A, B] => [...self, that])
 
-/*
-
-  TODO:
-
-  - at
-  - swap
-
-*/
+/**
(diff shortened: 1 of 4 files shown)


mikearnaldi:
We could probably do better by using `N extends number` instead of `number` so that `T[N]` is the actual tuple type at `\N`

msensys:
> We could probably do better by using `N extends number` instead of `number` so that `T[N]` is the actual tuple type at `N`

Ready!

mikearnaldi:
```ts
export const at: {
  <A extends ReadonlyArray<unknown>, N extends number>(index: N): (self: A) => A[N]
  <A extends ReadonlyArray<unknown>, N extends number>(self: A, index: N): A[N]
}
```

wondering if `N` should instead `extends keyof A`, if instead we leave it as `number` then it would allow `undefined` access to properties not in the tuple (which may / may not be ok) but in that case a more correct signature would be:

```ts
export const at: {
  <N extends number>(index: N): <A extends ReadonlyArray<unknown>>(self: A) => A[N]
  <A extends ReadonlyArray<unknown>, N extends number>(self: A, index: N): A[N]
}
```

So that it allows `at(n)` to be used as a function even outside of a pipe.

cc @gcanti @tim-smart

gcanti:
I would mirror what we did in `Struct.ts`:

`src/Tuple.ts`

```ts
export const get =
  <I extends number>(index: I) => <T extends ReadonlyArray<any>>(t: T): [] extends T ? T[I] | undefined : T[I] =>
    t[index]

```

`dtslint/Tuple.ts`

```ts
import { hole, pipe } from "effect/Function"

// get

// $ExpectType <T extends readonly any[]>(t: T) => [] extends T ? T[0] | undefined : T[0]
const get0 = get(0)

// $ExpectType undefined
pipe(hole<[]>(), get0)

// $ExpectType undefined
pipe(hole<readonly []>(), get0)

// $ExpectType string
pipe(hole<[string, number]>(), get0)

// $ExpectType string
pipe(hole<readonly [string, number]>(), get0)

// $ExpectType number
pipe(hole<[string, number]>(), get(1))

// $ExpectType number
pipe(hole<readonly [string, number]>(), get(1))

// $ExpectType undefined
pipe(hole<[string, number]>(), get(2))

// $ExpectType undefined
pipe(hole<readonly [string, number]>(), get(2))

// $ExpectType string | number
pipe(hole<[string, number]>(), get(-1))

// $ExpectType string | number
pipe(hole<readonly [string, number]>(), get(-1))

// $ExpectType number | undefined
pipe(hole<Array<number>>(), get(1))

// $ExpectType number | undefined
pipe(hole<Array<number>>(), get(-1))
```

msensys:
thank you @gcanti! what do you think, @mikearnaldi?

msensys:
It would be something as:
```ts
/**
 * Retrieves the element at a specified index from a tuple.
 *
 * @param self - A tuple from which to retrieve the element.
 * @param index - The index of the element to retrieve.
 *
 * @example
 * import { at } from "effect/Tuple"
 *
 * assert.deepStrictEqual(at([1, 'hello', true], 1), 'hello')
 *
 * @category getters
 * @since 3.4.0
 */
export const at: {
  <N extends number>(index: N): <A extends ReadonlyArray<unknown>>(self: A) => A[N]
  <A extends ReadonlyArray<unknown>, N extends number>(self: A, index: N): A[N]
} = dual(2, <A extends ReadonlyArray<unknown>, N extends number>(self: A, index: N): A[N] => self[index])
```

mikearnaldi:
I think I am ok with it, it's in line with schema

msensys:
I will send a PR to add it

msensys:
cc @mikearnaldi  👍🏻

packages/effect/src/Tuple.ts
```diff
@@ -194,14 +194,24 @@ export const appendElement: {
...
+/**
+ * Retrieves the element at a specified index from a tuple.
+ *
+ * @param self - A tuple from which to retrieve the element.
+ * @param index - The index of the element to retrieve.
+ *
+ * @example
+ * import { at } from "effect/Tuple"
+ *
+ * assert.deepStrictEqual(at([1, 'hello', true], 1), 'hello')
```
tim-smart, line 206:
```suggestion
 * import { Tuple } from "effect"
 *
 * assert.deepStrictEqual(Tuple.at([1, 'hello', true], 1), 'hello')
```

We generally recommend using a namespace import.

msensys: @tim-smart, look at here now :D

tim-smart:
I would be good to add the dtslint tests @gcanti quoted above.

msensys:
@tim-smart can you review it?

PR #3017, opened 2024-06-19
Feature: add @effect/sql-kysely package

ecyrbe:
## Type

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add support for Kysely as a DB query builder.

Since kysely is built on pure typescript types, i added no @effect/schema overload, like for @effect/sql-drizzle . 
There are implementations out there doing it, but it's because they don't have effect commit implementation, so they rely on effect/sql schema and resolver to effectify kysely.

### many implementations

- add support for @effect/sql mssql backend (required mssql backend refactoring)
- add support for @effect/sql mysql backend
- add support for @effect/sql postgres backend
- add support for @effect/sql sqlite backend
- add support for all kysely drivers with a special kysely Dialect Tag

### Example usage
```ts
import * as SqliteKysely from "@effect/sql-kysely/Sqlite"
import * as Sqlite from "@effect/sql-sqlite-node"
import { Config, Console, Context, Effect, Exit, Layer } from "effect"
import type { Generated } from "kysely"

export interface User {
  id: Generated<number>
  name: string
}

interface Database {
  users: User
}

class SqliteDB extends Context.Tag("SqliteDB")<SqliteDB, SqliteKysely.EffectKysely<Database>>() {}

const SqliteLive = Sqlite.client.layer({
  filename: Config.succeed(":memory:")
})

const KyselyLive = Layer.effect(SqliteDB, SqliteKysely.make<Database>()).pipe(Layer.provide(SqliteLive))

Effect.gen(function*(_) {
  const db = yield* SqliteDB

  yield* db.schema
    .createTable("users")
    .addColumn("id", "integer", (c) => c.primaryKey().autoIncrement())
    .addColumn("name", "text", (c) => c.notNull())

  const result = yield* db.withTransaction(
    Effect.gen(function*() {
      const inserted = yield* db.insertInto("users").values({ name: "Alice" }).returningAll()
      yield* Console.log(inserted)
      const selected = yield* db.selectFrom("users").selectAll()
      yield* Console.log(selected)
      const updated = yield* db.updateTable("users").set({ name: "Bob" }).returningAll()
      yield* Console.log(updated)
      return yield* Effect.fail(new Error("rollback"))
    })
  ).pipe(Effect.exit)
  if (Exit.isSuccess(result)) {
    return yield* Effect.fail("should not reach here")
  }
  const selected = yield* db.selectFrom("users").selectAll()
  yield* Console.log(selected)
}).pipe(
  Effect.provide(KyselyLive),
  Effect.runPromise
)
```

## Related

Effect has prior external library integration with @effect/sql-drizzle , that's why i propose to also integrate kysely.

It also has a minor change to MSSQL placeholder format. Since Kysely use numbered placeholders and Tedious library supports them, i made a change for mssql to be compatible.

- Related Issue #
- Closes #


packages/sql-kysely/src/internal/kysely.ts
```diff
@@ -0,0 +1,80 @@
...
+import type { EffectKysely } from "../patch.types.js"
+import { effectifyWithExecute, effectifyWithSql, patch } from "./patch.js"
+
+/**
+ * @internal
+ * patch all compilable/executable builders with commit prototypes
+ *
+ * @warning side effect
+ */
+patch(AlterTableColumnAlteringBuilder.prototype)
```
tim-smart, line 34: Hmm this is even more brittle than the drizzle integration. I think we will need to find a more general interface to target if possible.

packages/sql-kysely/test/Kysely.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+describe("Kysely", () => {
+  it.scoped("queries", () =>
+    Effect.gen(function*(_) {
+      const db = yield* KyselyDB
+
+      yield* db.schema
+        .createTable("users")
+        .addColumn("id", "integer", (c) => c.primaryKey().autoIncrement())
+        .addColumn("name", "text", (c) => c.notNull())
```
tim-smart, line 36: Make sure you test column renaming (snake case to camel etc)

packages/sql-kysely/src/internal/kysely.ts
ecyrbe: No Luck here. There is no common impl to rely on.

tim-smart: When I get some time I'll take a look and see if there is anything we can do.

tim-smart: This might have some merit:

```ts
import {
  CompiledQuery,
  DefaultQueryCompiler,
  Kysely,
  SqliteDialect,
} from "kysely"
import { Effect, Effectable } from "effect"

declare module "kysely" {
  export interface CompiledQuery<O = unknown> extends Effect.Effect<O> {}
}

class EffectCompiledQuery<O = unknown>
  extends Effectable.Class<O>
  implements CompiledQuery<O>
{
  constructor(private compiledQuery: CompiledQuery<O>) {
    super()
  }
  commit(): Effect.Effect<O> {
    return Effect.succeed({} as any)
  }
  get sql() {
    return this.compiledQuery.sql
  }
  get query() {
    return this.compiledQuery.query
  }
  get parameters() {
    return this.compiledQuery.parameters
  }
}

Effect.gen(function* () {
  const oldCompileQuery = DefaultQueryCompiler.prototype.compileQuery
  DefaultQueryCompiler.prototype.compileQuery = function (query) {
    console.log("compiling query", query)
    return new EffectCompiledQuery(oldCompileQuery.call(this, query))
  }

  const dialect = new SqliteDialect({} as any)
  const db = new Kysely<any>({
    dialect,
  })
  const result = yield* db.selectFrom("person").where("id", "=", 1).compile()
  console.log(result)
}).pipe(Effect.runPromise)
```

ecyrbe:
Docgen generation seems to fail due to "export * " for types. I'll take a look tonight to check why.

packages/sql-kysely/src/internal/kysely.ts
ecyrbe: My guts tell me that it's not a good enough tradeoff.
it's trading users DX for internal code simplification.
And it don't allow for using native kysely drivers in an Effect integrated way.

ecyrbe:
> Docgen generation seems to fail due to "export * " for types. I'll take a look tonight to check why.

hopefully las commit should fix docgen errors

packages/sql-kysely/test/Kysely.test.ts
ecyrbe: added support for plugin and test with camelCase to snake_case

packages/sql-kysely/src/internal/kysely.ts
mikearnaldi: yeah I'd also try to optimise for usage rather than internal complexity, only wondering if those internals are stable or they are up for change in kysley, if the latter it might be risky to depend on those. I guess we can involve someone from the kysley team to clarify?

ecyrbe: @igalklebanov your feedback would be valuable here.

igalklebanov: Hey 👋

Don't have enough context and understanding of "the Effect way".

The main builders are pretty stable.
`SelectQueryBuilder` will *probably* be exposed again.
Builders do get renamed/split on rare ocassions.
Builders are added sometimes, and it's not always a breaking change 

There is no indication `

ecyrbe: Thanks,

One thing that could simplify things on our side would be to had a Executable class like this :
```ts
class Executable<0> {
  abstract execute() : promise<O>
}
```
And let all builders implement it. This way we would only need to patch this one. And it would be future proof for us. Meaning adding / splitting / removing builders would have no impact on us.
 Would such addition be welcomed in kysely ? Or at least start a discussion on kysely repo about it?

ecyrbe: I got feedback from kysely team, and my proposal is not something they want to add in their code base.
So either we maintain on our side the compatibility with builders evolution or we can close this PR if it's too much of a concern.

igalklebanov: There are more ways to execute a query:

`Kysely`'s `executeQuery` method.
`SelectQueryBuilder`, `UpdateQueryBuilder`, `InsertQueryBuilder` and `DeleteQueryBuilder`'s `stream` method (anything that implements `Streamable<O>`).
`sql` template tag's `execute`.

Regarding more future-proof patching:

Have you tried patching `DefaultQueryExecutor`'s `compileQuery`, `executeQuery` and `stream`? It extends `QueryExecutorBase` a class that'll never be exposed, but these parts are 1-2 years untouched and central to execution flow.

ecyrbe: Unfortunately, Streamable is just an interface, not a class, so prototype is not patchable.
We need a base class to patch and no Builder has a base class.

ecyrbe: I first read it wrong. Sorry for last message (removed).
So DefaultQueryExecutor is used has the executor for all execute calls ?

ecyrbe: Just digged deeper into the code, unfortunately, Patching DefaultQueryExecutor would mean changing behaviour of you `execute()` for all instances to return an effect instead of a Promise even when user don't want to use an Effect.
We really need a way to Patch the Builder itself.

mikearnaldi: I think we can keep current solution and see, if the kysley team has no interest to facilitate integration we are left with two choices:
1) trust they won't change internals and eventually adapt for the changes when they come, we need to make sure using a builder that is not supported will raise an error
2) fork kysley and maintain an effect-first fork

I'd start with 1 and get to 2 only if 1 becomes a pain

mikearnaldi:
@ecyrbe following twitter discussion please add a note in the readme that warns against "future proof" and suggesting to users to use this at their own risk

ecyrbe:
> @ecyrbe following twitter discussion please add a note in the readme that warns against "future proof" and suggesting to users to use this at their own risk

done

mikearnaldi:
@ecyrbe would you mind rebasing pls so we can merge?

TylorS:
I'm probably a bit late to the game here, but FWIW, I have a fairly different approach to this integration which may be of some value in the discussion. https://github.com/TylorS/effect-sql-kysely. The `/test` folder is probably most useful in terms of usage as its not really entirely ready for public consumption (though it is being utilized).

The key differences are:

1.) `@effect/sql`'s `Client` interface is implemented _atop_ of Kysely's connections instead of as "just" a query builder
2.) Kysely's client interface is injected in the `Sql.schema.*` and `Sql.resolver.*` interfaces 
3.) Integrates with @effect/schema for Kysely's `ColumnType` and its derivatives

There's probably plenty of reasons it wouldn't be utilized as-is, and I'm definitely not upset if it's disregarded entirely, but I figured I'd share either way as its fairly different from what's in this PR and doesn't really have any means of breaking unless the public API from Kysely is broken explicitly.

PR #3023, opened 2024-06-20
effect: add Stream.filterMapEffectOption function

### LINKED ISSUE #3018: Stream.filterMapEffect but accepting Effect<Option<A>> instead of Option<Effect<A>>

`Stream.filterMapEffect` is unique to the rest of the `Stream.*Effect` functions: it expects an effect wrapped in another data structure instead of an effect _containing_ a data structure: `(a: A) => Option<Effect<B>>`

or in other words, if you want to `filterMapEffect` you can't filter map based on an effect value, but instead you can choose whether to return an effect or not

Let's say we want to map a stream of numbers into their square roots:

```typescript
declare const sqrt = (number: number) => Effect.Effect<Option<number>>;

const numbers = Stream.fromIterable([-1,2,3])

// What we can do right now
const squareRoots_ = numbers.pipe(
  Stream.mapEffect(number => sqrt(number)),
  Stream.filterMap(opt => opt)
);

// With this filterMapEffectX (name tbd):
const squareRoots = numbers.pipe(
  Stream.filterMapEffectX(number => sqrt(number))
)
```

it seems counter intuitive that this is the signature IMO, and I know that we probably can't break this API as Effect 3 is stable -- so we'll have to think of a different way to call it.

This is obviously can be done in userland as I mentioned previously, which is how I implemented it on my projects:

```typescript
import { Function, Stream, Effect, Option } from 'effect';

export const filterMapEffect = Function.dual<
  <A, A2, E2, R2>(
    fn: (a: A) => Effect.Effect<Option.Option<A2>, E2, R2>,
  ) => <E, R>(
    stream: Stream.Stream<A, E, R>,
  ) => Stream.Stream<A2, E2 | E, R2 | R>,
  <A, E, R, A2, E2, R2>(
    stream: Stream.Stream<A, E, R>,
    fn: (a: A) => Effect.Effect<Option.Option<A2>, E2, R2>,
  ) => Stream.Stream<A2, E2 | E, R2 | R>
>(2, (stream, fn) =>
  Stream.filterMap(Stream.mapEffect(stream, fn), Function.identity),
);
```

The alternative is to simply keep that a userland implementation. But I think this API makes sense in the "standard library".

Schniz:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a `Stream.mapEffectFilter` function that maps items of a stream effectfully,
and then filters them.

## Related

- Closes #3018

### DIFF
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -2094,7 +2094,7 @@ export const mapEffect: {
  * const even = (a: number) => a % 2 === 0 ? Effect.succeedSome(a) : Effect.succeedNone
  *
  * const result = Stream.make(0, 1, 2, 4).pipe(
- *   Stream.mapEffectFilter(even),
+ *   Stream.filterMapEffectOption(even),
  *   Stream.runCollect,
  *   Effect.runPromise,
  * )
@@ -2103,7 +2103,7 @@ export const mapEffect: {
  * @since 3.3.5
  * @category utils
  */
-export const mapEffectFilter: {
+export const filterMapEffectOption: {
   <A, A2, E2, R2>(
     fn: (a: A) => Effect.Effect<Option.Option<A2>, E2, R2>
   ): <E, R>(stream: Stream<A, E, R>) => Stream<A2, E2 | E, R2 | R>
@@ -2111,7 +2111,7 @@ export const mapEffectFilter: {
     stream: Stream<A, E, R>,
     fn: (a: A) => Effect.Effect<Option.Option<A2>, E2, R2>
   ): Stream<A2, E2 | E, R2 | R>
-} = _groupBy.mapEffectFilter
+} = _groupBy.filterMapEffectOption
 
 /**
  * Transforms the errors emitted by this stream using `f`.
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -2088,7 +2088,6 @@ export const mapEffect: {
  * and filters out `None` values
  *
  * @example
- * ```ts
  * import { Effect, Stream, Console } from 'effect';
  *
  * // returns `Some(number)` if the number is even, otherwise `None`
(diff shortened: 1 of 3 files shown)


mikearnaldi:
cc @tim-smart @IMax153 what do we think of the name?

IMax153:
> cc @tim-smart @IMax153 what do we think of the name?

I'll be honest, I'm not a fan of this name. I find it confusing given there's also filterMapEffect. 

That being said I don't have a good alternative at the moment 😬.

Schniz:
There are couple of alternatives I generated with AI: all feel as wrong IMO

https://sdk.vercel.ai/s/7HAZozu43jqI

tim-smart:
Can we overload the predicate return type of `filterMapEffect` to be `Option<Effect> | Effect<Option>` I wonder.

The only case that could cause issues is if you return `Option<Option>`.

mikearnaldi:
> Can we overload the predicate return type of `filterMapEffect` to be `Option<Effect> | Effect<Option>` I wonder.
> 
> The only case that could cause issues is if you return `Option<Option>`.

given Option is an Effect a union such as this would not be possible

tim-smart:
`Stream.filterMapEffectOption`?

mikearnaldi:
> `Stream.filterMapEffectOption`?

let's go for this and re-think in 4.0 if we can have a single function

IMax153:
Yea I second Tim's suggestion - `Stream.filterMapEffectOption` is good for now 👍

Schniz:
<picture data-single-emoji=":done:" title=":done:">:done:</picture> <picture data-single-emoji=":effect-ts:" title=":effect-ts:">:effect-ts:</picture>

mikearnaldi:
apologies @Schniz I merged by mistake, needs to be re-opened given the lack of a changeset

PR #3045, opened 2024-06-22
feature: add cloudflare @effect/sql-d1 package

### LINKED ISSUE #2700: From Discord: @effect/sql with Cloudflare D1
# Summary
The conversation revolves around an attempt to use `@effect/sql` with Cloudflare's D1 (SQLite) database. Here are the key points and takeaways:

1. **Initial Inquiry**: The user, phischer, inquired about the compatibility of `@effect/sql` with Cloudflare D1, wondering if it's possible to use them together given the current clients.

2. **Assumption**: Another user, datner_, suggested that if `@effect/sql` works with `better-sqlite3`, it should theoretically work with Cloudflare D1, as there might not be a significant difference in how they operate at a basic level.

3. **Implementation Attempt**: Phischer managed to get `@effect/sql` working with Cloudflare D1 by modifying the client from `sql-sqlite-node` to work with `D1Database`. However, this solution required multiple `@ts-ignore` annotations to compile, indicating potential type mismatches or unsupported operations. Due to these uncertainties and the makeshift nature of the solution, phischer expressed a lack of trust in the stability and reliability of this approach and decided to use an alternative (drizzle) for the time being.

4. **Code Example**: Phischer shared a code snippet demonstrating how they adapted `@effect/sql` for use with Cloudflare D1. The code involves creating a SQL client that interacts with the D1 database, executing a SQL query, and handling the results.

5. **Challenges**: The main challenges highlighted include:
   - The need to adapt the client to work with a database object (`D1Database`) instead of a filename, which is what the original `@effect/sql-sqlite-node` client expects.
   - Handling the asynchronous nature of D1's query results, which differ from the synchronous "try" pattern used in the existing client.
   - The necessity to bypass TypeScript's type checking with `@ts-ignore` to make the code compile, indicating potential type safety issues.

6. **Conclusion**: While it is technically possible to use `@effect/sql` with Cloudflare D1 by adapting existing clients, the process involves significant modifications and workarounds that may compromise type safety and reliability. As such, users might consider looking for more straightforward or officially supported solutions for their database needs on Cloudflare Workers.

# Discord thread

https://discord.com/channels/795981131316985866/1236321129941958726

ecyrbe:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add a cloudflare D1 sqlite client.  D1 don't support transactions directly (only batches).
Since it was asked multiple times for it to be supported

```ts

import type { D1Database } from "@cloudflare/workers-types"
import { D1Client } from "@effect/sql-d1"
import { Config, Context, Data, Effect, Layer } from "effect"
import { Miniflare } from "miniflare"

export class MiniflareError extends Data.TaggedError("MiniflareError")<{
  cause: unknown
}> {}

export class D1Miniflare extends Context.Tag("test/D1Miniflare")<
  D1Miniflare,
  Miniflare
>() {
  static Live = Layer.scoped(
    this,
    Effect.acquireRelease(
      Effect.try({
        try: () =>
          new Miniflare({
            modules: true,
            d1Databases: {
              DB: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
            },
            script: ""
          }),
        catch: (cause) => new MiniflareError({ cause })
      }),
      (miniflare) => Effect.promise(() => miniflare.dispose())
    )
  )

  static ClientLive = Layer.unwrapEffect(
    Effect.gen(function*() {
      const miniflare = yield* D1Miniflare
      const db: D1Database = yield* Effect.tryPromise(() => miniflare.getD1Database("DB"))
      return D1Client.layer({
        db: Config.succeed(db)
      })
    })
  ).pipe(Layer.provide(this.Live))
}

Effect.gen(function*(_) {
  const sql = yield* D1Client.D1Client
  yield* sql`CREATE TABLE test (id INTEGER PRIMARY KEY, name TEXT)`
  yield* sql`INSERT INTO test (name) VALUES ('hello')`
  let rows = yield* _(sql`SELECT * FROM test`)
  assert.deepStrictEqual(rows, [{ id: 1, name: "hello" }])
  yield* sql`INSERT INTO test (name) VALUES ('world')`
  rows = yield* sql`SELECT * FROM test`
  assert.deepStrictEqual(rows, [
    { id: 1, name: "hello" },
    { id: 2, name: "world" }
  ])
}).pipe(Effect.provide(D1Miniflare.ClientLive)))
```

## Tests
Tests are using miniflare to have a local D1 instance

- Closes #2700

### DIFF
--- a/packages/sql-d1/src/D1Client.ts
+++ b/packages/sql-d1/src/D1Client.ts
@@ -42,15 +42,6 @@ export interface D1Client extends Client.SqlClient {
 }
 
 /**
- * @category models
- * @since 1.0.0
- */
-export interface BackupMetadata {
-  readonly totalPages: number
-  readonly remainingPages: number
-}
-
-/**
  * @category tags
  * @since 1.0.0
  */
@@ -163,10 +154,8 @@ export const make = (
       })
     })
 
-    const semaphore = yield* Effect.makeSemaphore(1)
     const connection = yield* makeConnection
-
-    const acquirer = semaphore.withPermits(1)(Effect.succeed(connection))
+    const acquirer = Effect.succeed(connection)
     const transactionAcquirer = Effect.dieMessage("transactions are not supported in D1")
 
     return Object.assign(
--- a/packages/sql-d1/src/D1Client.ts
+++ b/packages/sql-d1/src/D1Client.ts
@@ -1,0 +1,216 @@
+/**
+ * @since 1.0.0
+ */
+import type { D1Database, D1PreparedStatement } from "@cloudflare/workers-types"
+import * as Client from "@effect/sql/SqlClient"
+import type { Connection } from "@effect/sql/SqlConnection"
+import { SqlError } from "@effect/sql/SqlError"
+import * as Statement from "@effect/sql/Statement"
+import * as Otel from "@opentelemetry/semantic-conventions"
+import * as Cache from "effect/Cache"
+import type { ConfigError } from "effect/ConfigError"
+import * as Context from "effect/Context"
+import * as Duration from "effect/Duration"
+import * as Effect from "effect/Effect"
+import { identity } from "effect/Function"
+import * as Layer from "effect/Layer"
+import * as Scope from "effect/Scope"
+
+/**
+ * @category type ids
+ * @since 1.0.0
+ */
+export const TypeId: unique symbol = Symbol.for("@effect/sql-d1/D1Client")
+
+/**
+ * @category type ids
+ * @since 1.0.0
+ */
+export type TypeId = typeof TypeId
+
+/**
+ * @category models
+ * @since 1.0.0
+ */
+export interface D1Client extends Client.SqlClient {
+  readonly [TypeId]: TypeId
+  readonly config: D1ClientConfig
+
+  /** Not supported in d1 */
+  readonly updateValues: never
+}
+
+/**
+ * @category models
+ * @since 1.0.0
+ */
+export interface BackupMetadata {
+  readonly totalPages: number
+  readonly remainingPages: number
+}
+
+/**
+ * @category tags
+ * @since 1.0.0
+ */
+export const D1Client = Context.GenericTag<D1Client>("@effect/sql-d1/D1Client")
+
+/**
+ * @category models
+ * @since 1.0.0
+ */
+export interface D1ClientConfig {
+  readonly db: D1Database
+  readonly prepareCacheSize?: number | undefined
+  readonly prepareCacheTTL?: Duration.DurationInput | undefined
+  readonly spanAttributes?: Record<string, unknown> | undefined
+
+  readonly transformResultNames?: ((str: string) => string) | undefined
+  readonly transformQueryNames?: ((str: string) => string) | undefined
+}
+
+/**
+ * @category constructor
+ * @since 1.0.0
+ */
+export const make = (
+  options: D1ClientConfig
+): Effect.Effect<D1Client, never, Scope.Scope> =>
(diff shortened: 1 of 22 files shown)


mikearnaldi:
Maybe we need a "batch" api in sql to support batch transactions?

datner:
a batch api would just be the transaction api without injecting markers between statements, though I don't think all dbs support batching in this way. For example mssql doesn't lol

mikearnaldi:
> a batch api would just be the transaction api without injecting markers between statements, though I don't think all dbs support batching in this way. For example mssql doesn't lol

no, a batching api would allow us to specify a number of sql statements in a single shot, every sql database that supports transactions also supports such batching api.

example:

```ts
sql.batchTx(
  sql`SELECT ...`,
  sql`INSERT ...`,
  sql`DELETE...`
)
```

the execution can send a single batch to dbs that support it or simply run a `BEGIN, ..., COMMIT`.

mikearnaldi:
> > a batch api would just be the transaction api without injecting markers between statements, though I don't think all dbs support batching in this way. For example mssql doesn't lol
> 
> no, a batching api would allow us to specify a number of sql statements in a single shot, every sql database that supports transactions also supports such batching api.
> 
> example:
> 
> ```ts
> sql.batchTx(
>   sql`SELECT ...`,
>   sql`INSERT ...`,
>   sql`DELETE...`
> )
> ```
> 
> the execution can send a single batch to dbs that support it or simply run a `BEGIN, ..., COMMIT`.

note that this is also common of products that put http apis on top of databases, while I believe this is a bad practice and that such dbs can't be used in most of the use cases where you'd need a transaction we should probably still support it.

@tim-smart we have to decide though what to do with the `withTransaction` api, how should that behave on a db that doesn't support it? the choices are:
1) make it defect
2) make it a no-op
3) make it a no-op and log a warning

ecyrbe:
@mikearnaldi added first suggestion for now. it defects when using transactions.

datner:
The more user-friendly choice is no-op and warn, but it might come as a surprise to the user

ecyrbe:
My fear is that `no-op + warning` can go undetected to production.  I prefer defect personnaly.

packages/sql-d1/src/D1Client.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  /** Not supported in d1 */
+  readonly updateValues: never
+}
+
+/**
+ * @category models
+ * @since 1.0.0
+ */
+export interface BackupMetadata {
```
tim-smart, line 48: Not used?

packages/sql-d1/src/D1Client.ts
```diff
@@ -0,0 +1,205 @@
...
+        executeRaw(sql, params) {
+          return runRaw(sql, params)
+        },
+        executeStream(_sql, _params) {
+          return Effect.dieMessage("executeStream not implemented")
+        }
+      })
+    })
+
+    const semaphore = yield* Effect.makeSemaphore(1)
```
tim-smart, line 166: You can remove the semaphore, as there is no transactions.

tim-smart (changes requested):

packages/sql-d1/src/D1Client.ts
ecyrbe: Fixed

mikearnaldi:
> My fear is that `no-op + warning` can go undetected to production. I prefer defect personnaly.

Same here

tim-smart (approved):

tim-smart:
🚀

PR #3067, opened 2024-06-24
Cleanup signal "abort" event handler in `Micro.runFork`

KhraksMamtsov:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/eight-points-push.md
+++ b/.changeset/eight-points-push.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Cleanup signal "abort" event handler in `Micro.runFork`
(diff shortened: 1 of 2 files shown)


tim-smart:
I think to do this properly you will need to `removeEventListener` once the Micro has returned a result.

mikearnaldi:
> I think to do this properly you will need to `removeEventListener` once the Micro has returned a result.

Isn't interruption in Micro fire and forget? Why would we need to call the handler multiple times?

tim-smart:
> Isn't interruption in Micro fire and forget? Why would we need to call the handler multiple times?

When you abort a Micro it will fail with a `Failure.Aborted`.

I think we will need both `once: true` and `removeEventListener` (to prevent memory leaks where the same signal is used multiple times).

PR #3068, opened 2024-06-24
more type safe dual signature, closes #2967

### LINKED ISSUE #2967: More precise `dual` signature (prevent type errors/issues)

With the current `dual` signature from the `Function` module you can easily get away with wrong types, since the generics can be inferred in a very wide fashion. This can lead to subtle type errors that are not discovered at compile time, but rather at runtime.

Given this `f` definition:
```ts
type f_data_first = (a: number, b: number) => number
type f_data_last = (b: number, a: number) => number
type f = f_data_last & f_data_first
```

The following problems could arise with no type error / wrong type inferred:
```ts
const f_no_type_old = effect_dual(2, (a: number, b: number) => a + b) // (args: ...any[]) => any ❌
const f_wrong_type_old: f = effect_dual(2, (a: string, b: string) => a + b) // no error! ❌
```

Change the `dual` function type signature to be more precise over what you actually used as the implementation for the dual function.

```ts
export const dual: {
    <Impl extends (...args: readonly any[]) => any, Signature extends Impl = Impl>(
        arity: Parameters<Impl>["length"],
        body: Impl
    ): Signature
    <Impl extends (...args: readonly any[]) => any, Signature extends Impl = Impl>(
        // eslint-disable-next-line @typescript-eslint/unified-signatures
        isDataFirst: (args: IArguments) => boolean,
        body: Impl
    ): Signature
} = effect_dual



const define_f_new: f = dual(2, (a: number, b: number) => a + b)
const defined_f_old: f = effect_dual(2, (a: number, b: number) => a + b)

const define_f_new_explicit_types = dual<f_data_first, f>(2, (a: number, b: number) => a + b)
const defined_f_old_explicit_types = effect_dual<f_data_last, f_data_first>(2, (a: number, b: number) => a + b)

const f_no_type_new = dual(2, (a: number, b: number) => a + b) // (a: number, b: number) => number ✅
const f_no_type_old = effect_dual(2, (a: number, b: number) => a + b) // (args: ...any[]) => any ❌

const f_wrong_type_new: f = dual(2, (a: string, b: string) => a + b) // compile time error ✅
const f_wrong_type_old: f = effect_dual(2, (a: string, b: string) => a + b) // no error! ❌
```

_No response_

kristiannotari:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Ad discussed in #2967, the `dual` type signature could be improved to avoid inferring an `(...args: any[]) => any` function type, which is assignable to  every function signature you define for your dual api, and instead be precise over the implementation signature used. Of course the rest of the type signature would still be unsafe, but that's unavoidable.

This should help prevent errors when the type signature of the implementation used for the `dual` api is different to the one used as the final signature of such dual api.

NOTICE:

There are a couple of tests failing, regarding the schema package, which should not be related to this change. In fact, if you run the tests on the main this pr is based on, those tests keep failing.

This PR should be type level only. I added dtslint tests for the dual signature type checks I did in the first place that brought this change to exist. This should help to assert those cases are preserved in the future.

## Related

- Closes #2967

### DIFF
--- a/packages/schema/src/Schema.ts
+++ b/packages/schema/src/Schema.ts
@@ -2246,12 +2246,15 @@ export const optional: {
     self: Schema.All extends S ? "you can't apply optional implicitly, use optional() instead" : S,
     options?: Options
   ): [undefined] extends [Options] ? optional<S> : optionalWithOptions<S, Options>
-} = dual((args) => isSchema(args[0]), (from, options) => {
+} = (dual((args) => isSchema(args[0]), (
+  from,
+  options
+) => {
   // Note: `Schema.All extends S ? "you can't...` is used to prevent the case where `optional` is implicitly applied.
   // For example: `S.String.pipe(S.optional)` would result in `S.String` being inferred as `Schema.All`,
   // which is not the intended behavior. This is mostly an aesthetic consideration, so if it causes issues, we can remove it.
   return new PropertySignatureWithFromImpl(optionalPropertySignatureAST(from, options), from)
-})
+})) as unknown as typeof optional
 
 /**
  * @since 0.67.0
@@ -3073,9 +3076,12 @@ export interface extend<Self extends Schema.Any, That extends Schema.Any> extend
 export const extend: {
   <That extends Schema.Any>(that: That): <Self extends Schema.Any>(self: Self) => extend<Self, That>
   <Self extends Schema.Any, That extends Schema.Any>(self: Self, that: That): extend<Self, That>
-} = dual(
+} = dual<
+  typeof extend,
+  <Self extends Schema.Any, That extends Schema.Any>(self: Self, that: That) => extend<Self, That>
+>(
   2,
-  <Self extends Schema.Any, That extends Schema.Any>(self: Self, that: That) => make(extendAST(self.ast, that.ast, []))
+  (self, that) => make(extendAST(self.ast, that.ast, []))
 )
 
 /**
--- a/.changeset/nice-laws-admire.md
+++ b/.changeset/nice-laws-admire.md
@@ -1,0 +1,8 @@
+---
+"@effect/typeclass": minor
+"@effect/platform": minor
+"effect": minor
+"@effect/schema": minor
+---
+
+more type safe `Function.dual` signature
--- a/packages/effect/dtslint/Function.ts
+++ b/packages/effect/dtslint/Function.ts
@@ -1,0 +1,24 @@
+import * as Function from "effect/Function"
+
+// -------------------------------------------------------------------------------------
+// dual
+// -------------------------------------------------------------------------------------
+
+type f_data_first = (a: number, b: number) => number
+type f_data_last = (b: number) => (a: number) => number
+
+const f: f_data_first & f_data_last = Function.dual(2, (a: number, b: number) => a + b)
+// $ExpectType f_data_first & f_data_last
+f
+
+const f_explicit_types = Function.dual<f_data_last, f_data_first>(2, (a: number, b: number) => a + b)
+// $ExpectType f_data_first & f_data_last
+f_explicit_types
+
+const f_no_type = Function.dual(2, (a: number, b: number) => a + b)
+// $ExpectType (a: number, b: number) => number
+f_no_type
+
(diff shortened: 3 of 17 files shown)


packages/schema/src/Schema.ts
```diff
@@ -2244,12 +2244,15 @@ export const optional: {
...
+} = (dual((args) => isSchema(args[0]), (
+  from,
+  options
+) => {
   // Note: `Schema.All extends S ? "you can't...` is used to prevent the case where `optional` is implicitly applied.
   // For example: `S.String.pipe(S.optional)` would result in `S.String` being inferred as `Schema.All`,
   // which is not the intended behavior. This is mostly an aesthetic consideration, so if it causes issues, we can remove it.
   return new PropertySignatureWithFromImpl(optionalPropertySignatureAST(from, options), from)
-})
+})) as unknown as typeof optional
```
kristiannotari, line 2255: Not sure why this is needed, but I had to put it here to make it work. Remember the `dual` application before was inferred as `(...args: any[]) => any` so this stuff is expected on complex cases, still, should be doable without casting.

mikearnaldi:
Imho seems to be over complexifying usage, also it would be a breaking change that potentially breaks user code, not sure about it

kristiannotari:
> Imho seems to be over complexifying usage, also it would be a breaking change that potentially breaks user code, not sure about it

Usage should be the same in the "common case" scenario. There are a couple of instances where it can't infer automatically the whole type you'd like your dual signature to be, and require you to type the `Other` and `Impl` parts of it as generic types. This, btw, remove the need to type the implementation internally though, so it's more like moving and typing more directly than just inferring.

It's true some user code could be breaking due to this change, at least at the type level, since this PR is all about types, even if it's just about reordering types so to put generics in place.

tim-smart:
This would be a major change, the changeset will need adjusting.

`dual` was also made quite "relaxed" on purpose, so I'm on fence about this change.

kristiannotari:
> This would be a major change, the changeset will need adjusting.
> 
> `dual` was also made quite "relaxed" on purpose, so I'm on fence about this change.

Just to have a clear view on this so that everyone is aligned, the "new" signature would be as relaxed as the old one. You can always tell ts your dual signature is something else other than the actual implementation, which is always the case, no matter how strict you write the `dual` signature. This only enforces the implementation you write don't end up being treated as an "any" function early on without you noticing.

fubhy:
Thanks for the PR. This still looks potentially actionable, but it currently has merge conflicts. Could you please rebase/update the branch against the target branch so maintainers can review it?

kristiannotari:
Thanks for the ping. I'll try to find some time to update this asap.

kristiannotari:
@fubhy just to be clear, the updated PR would be exactly a 1:1 porting of the old PR logic with the updated main branch. The idea and implementation around the `dual` signature would not change. This requires a change in every (or at least many) points of the codebase that uses it. Based on how the signature it's approached, we may end up or not with the issues found when discussing about it in the related (now closed) issue #2967. There could be a change as suggested in the order of the generics so to avoid a lot of changes in the repo, but it's not a guaranteed perfect migration, even for users's codebases, which may need to adjust things manually here and there, not to speak about the internal typings exposing Mike noticed.

I have no problem in submitting an update to this pr so that it reflects the original scope, but I guess the same issues could arise. So if there are any unsolvable point from the discussions happened around this PR that would stop this PR from going on, I'd rather know them now so not to put more effort on this only to see it closed after a couple of hours for the cited points above.

Not to sound rude, just I want to know if it's actually reasonable to keep this PR and its usecase or not, to avoid wasting both mine and your time on this if I'm the only one requesting it and there's almost 0% chance for it to being actually accepted.

fubhy:
Yes. Thanks for the push back. My prior comment came from AI assisted issue triaging and likely misjudged this. I am not certain that this would land in the previously suggested form. Plus, it would indeed be quite disruptive considering how widely `dual` is used.

tim-smart:
v3 is under feature freeze now, so for the time being will close this :)

PR #3041, opened 2024-06-21
docs: remove adapter examples

meesvandongen:
## Type

- [x] Documentation Update

## Description

Remove adapter usage from examples

## Related

- Related Issue https://github.com/Effect-TS/effect/pull/2602

PR #3072, opened 2024-06-24
Remove adapter internal

meesvandongen:
## Type

- [x] Refactor

## Description

Remove usage of adapter from internal code.

## Related

- Related Issue https://github.com/Effect-TS/effect/pull/3041


meesvandongen:
Note that there are still 1861 functions in spec files with `function*($)`.

tim-smart:
Thanks! If you could add a `patch` changeset for this, as it touches implementation code.

tim-smart:
If you rebase against the latest `main` branch we can get this merged :)

PR #3080, opened 2024-06-24
add Stream.share api

dilame:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Stream.share({ connector: PubSub.unbounded() })` returns a new Stream that multicasts (shares) the original `Stream` by forking `runIntoPubSub` process in `forkScoped` mode. As long as there is at least one consumer, this `Stream` will be run and emitting data. When all consumers have exited, it will kill forked daemon.

I understand that this PR is kinda incomplete, i just need to know if you are interested in this operator at all, so i could finish.
Please, let me know.

## Related

I have not filled an issue on GitHub, but i posted a question in Discord, i will repeat it here

I have a WebSocket connection to a remote server. To subscribe to a topic, I send a `subscribe@someTopic` message, and to unsubscribe, I send an `unsubscribe@someTopic` message. Each topic is represented as a broadcasted Stream.

The challenge is that if two different parts of my code subscribe to the same topic, they should use the same WS connection, but should be able to work independently at the same time. This is difficult to achieve with the classic `Scope` finalization model because if one of the two consumers closes the scope, it will send the `unsubscribe@someTopic` signal to the server, causing the second consumer to stop receiving events without notice.

Using a global scope is not a viable solution because I need to ensure the WebSocket connection remains clean and unsubscribe from the topic when there are no active consumers.

The share operator addresses this issue by allowing multiple independent consumers to subscribe to the same topic without interfering with each other. This ensures that each consumer can independently manage its subscription and receive events without being affected by other consumers.

And, actually, there are much more scenarios when such `share` is invaluable.


tim-smart:
I think a better primitive is a kind of `ScopedRef` that does ref counting. You can then wrap a `PubSub` or `Queue` with it and call `Stream.fromQueue` etc.

It could also be used for other applications.

dilame:
As i said, this PR is WIP. It should have options like 
```ts
interface Options {
  readonly connector: Effect.Effect<
      PubSub.PubSub<Take.Take<A, E>> | Queue.Queue<Take.Take<A, E>>
    >;
  readonly resetOnRefCountZero?: boolean;
  readonly resetOnError?: boolean;
  readonly resetOnComplete?: boolean;
}
```

Let's say we have 
```ts
Stream.share({
  connector: PubSub.replay(1),
  resetOnRefCountZero: false;
})
```

That means if there was 2 consumers, then they both unsubscribed (0 consumers at now), then 1 consumer subscribed – it will receive the last value from stream. As i understand, it is impossible to achieve with `ScopedRef` idea. 
Also, i don't quite understand how to implement such a `ScopedRef` primitive

dilame:
@tim-smart could you maybe provide an example of code of how to use hypothetical `ScopedRef` to solve the described problem?

tim-smart:
> @tim-smart could you maybe provide an example of code of how to use hypothetical `ScopedRef` to solve the described problem?

https://github.com/Effect-TS/effect/pull/3179

You could do:

```ts
import { Effect, PubSub, RcMap, Stream } from "effect"

Effect.gen(function*() {
  const map = yield* RcMap.make((topic: string) =>
    Effect.acquireRelease(
      PubSub.unbounded<string>(),
      () => Effect.log(`unsubscribing from ${topic}`)
    )
  )

  const subscribe = (topic: string): Stream.Stream<string> =>
    RcMap.get(map, topic).pipe(
      Effect.map((_) => Stream.fromPubSub(_)),
      Stream.unwrapScoped
    )
})
```

dilame:
@tim-smart wow, i didn't noticed your message here, saw it just now after pushed refined version of `share` to ask you for a help.

Anyway, as i understand, it it not supposed to be a replacement for `Stream.share` operator, but it could be used under the hood. Not sure if it will make it better or just more complex with no benefit, though.

I'm sorry distrubing you here, but i spent the whole day trying to understand why my second test case doesn't work.
It works if i replace this
```ts
fiber ??= yield* Effect.forkScoped(runIntoPubSubScoped(upstream, connector))
```
with this
```ts
fiber ??= yield* Effect.forkDaemon(runIntoPubSub(upstream, connector))
```

Which makes me think that something is wrong with my scope management, but everything looks correct FMPOV. But effect doesn't think so:)

I really need someone's help. Could you take a look at it, please?

tim-smart (changes requested):

packages/effect/src/internal/stream.ts
```diff
@@ -5678,6 +5678,71 @@ export const scoped = <A, E, R>(
...
+    self: Stream.Stream<A, E, R>
+  ) => Stream.Stream<A, E, R | Scope.Scope>,
+  <A, E, R>(
+    self: Stream.Stream<A, E, R>,
+    config: Stream.ShareConfig<A, E>
+  ) => Stream.Stream<A, E, R | Scope.Scope>
+>(
+  2,
+  <A, E, R>(self: Stream.Stream<A, E, R>, config: Stream.ShareConfig<A, E>) => {
+    let consumersCount = 0
```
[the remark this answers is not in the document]
dilame: If fork the upstream inside the parent scope – there is no option to do ref counting. It will be just subscribed until the scope is closed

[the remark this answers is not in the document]
dilame: As i understand, the current implementation is resource safe already. There is no resource leaks or logical mistakes.
The issue is somewhere in the forking logic. Looks like a bug in effect, i opened an issue with reproduction links in sandbox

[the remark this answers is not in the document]
tim-smart: The current implementation is creating state eagerly in the closure. It needs to be allocated lazily / effectfully and bound to a Scope.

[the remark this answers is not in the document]
dilame: I found a bug here

https://github.com/Effect-TS/effect/blob/main/packages/effect/src/internal/channel/channelExecutor.ts#L1098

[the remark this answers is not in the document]
dilame: > The current implementation is creating state in the closure. It needs to be allocated lazily / effectfully and bound to a Scope.

Is it a hard requirement? I am pretty sure the state should be in the closure in this case because of a nature of this operator. And it's already bounded in the scope.
Of course i can wrap it inside `Effect.suspend`, but it will be an extra action for the user to unpack it, and i don't see benefit of it. Is there any?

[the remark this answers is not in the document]
dilame: I always highly value your reviews, but in this case i would ask you to consider this implementation more thoroughly. 
Of course i will wrap it in `Effect.suspend ` if you reject, because i need this operator:) 

I'm sure that now it is implemented properly.

We don't usually need a shared stream to be shared multiple times. Because it is already shared.

It is a very rare scenario (i can't even think of a use-case) when we need to do like this

```ts
const shared = someStream.pipe(share({...}))
const shared1 = yield* shared;
const shared2 = yield* shared;
```

And even if we need to, we can do
```ts
const shared1 = someStream.pipe(share({...}))
const shared2 = someStream.pipe(share({...}))

```

But most of the times we just want to share it once and use it everywhere. In that case an extra `yield*` will be just an extra boilerplate

dilame:
I fixed the test to bypass the founded bug

packages/effect/src/internal/stream.ts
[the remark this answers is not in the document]
tim-smart: It would be:

```ts
const shared = yield* someStream.pipe(share(...))

shared.pipe(...)
shared.pipe(...)
```

You run the Effect to allocate the state and resources etc

packages/effect/src/internal/stream.ts
```diff
@@ -5678,6 +5678,79 @@ export const scoped = <A, E, R>(
...
+    }
+  ) => Stream.Stream<A, E, R>
+>(
+  2,
+  <A, E, R>(
+    self: Stream.Stream<A, E, R>,
+    config: {
+      readonly connector?: Effect.Effect<PubSub.PubSub<Take.Take<A, E>>>
+    }
+  ): Stream.Stream<A, E, R> => {
```
tim-smart, line 5729: Again, this should be `Effect<Stream<A, E>, never, R | Scope>`.

The context should be captured at the point of creation, so the R is eliminated from the consumption point, and any fibers forked inside the Scope.

packages/effect/src/internal/stream.ts
```diff
@@ -5678,6 +5678,79 @@ export const scoped = <A, E, R>(
...
+    let consumersCount = 0
+    let connector: PubSub.PubSub<Take.Take<A, E>> | null = null
+    let fiber: Fiber.RuntimeFiber<void> | null = null
+    return Effect.gen(function*() {
+      consumersCount++
+      connector ??= yield* (config.connector ?? PubSub.unbounded<Take.Take<A, E>>())
```
tim-smart, line 5735: It should probably default to `PubSub.bounded(16)`, so back-pressure is applied.

dilame:
> Again, this should be `Effect<Stream<A, E>, never, R | Scope>`.
> 
> The context should be captured at the point of creation, so the R is eliminated from the consumption point, and any fibers forked inside the Scope.

I feel there is a misunderstanding about this operator. Forking upstream consumption inside the resulting effect's scope is impossible due to its nature. This operator is designed to subscribe to the upstream only after the first consumer subscribes. Even if I fork upstream consumption during effect initialization, I must kill this fork when the ref count hits zero and restart it when it hits one. This is only feasible within the stream execution pipeline.

As we discussed earlier, streams operate within a `Scope` they create themselves if using non-scoped `Stream.run*`.

I hope this clarifies why I fork inside the stream pipeline—it's simply not feasible otherwise. Now, let’s consider using `forkScoped` instead of `forkDaemon`:

If you fork upstream consumption with `forkScoped` inside the stream pipeline, this fork will terminate upon downstream completion regardless of the ref count. This is unacceptable since other shared downstreams might still be consuming it. This contradicts the logic of `shareRefCount`.

You can run all shared downstreams with `Stream.runScoped`, but this requires informing users because it’s not obvious. Additionally, it restricts users as they might attach business logic to `Stream.ensuring` on shared downstreams, needing it to execute upon stream termination. With `Stream.runScoped`, this only happens upon parent scope termination. Thus, users face a limitation: either use `shareRefCount` or have `Stream.ensuring` guaranteed to execute upon stream termination, but not both simultaneously. This is unacceptable.

Even if we run shared downstreams with `Stream.runScoped`, the subscriber count hits zero only when the scope closes, not upon actual stream consumption termination. The count decreases only with `Stream.ensuring`, which, in the case of `runScoped`, executes upon parent scope termination, defeating the purpose of `shareRefCount`.

Using `forkDaemon` is the only correct and 100% resource-safe solution since it kills within `Stream.ensuring` when `refCount === 0`. The `Effect` runtime ensures that `Stream.ensuring` executes upon scope closure or program termination—eventually, it will execute, giving users full control.

Based on this, I believe the implementation is correct.

Regarding wrapping this in `Effect`, I think it's pointless. Only the counter initializes; no effects execute during initialization. Furthermore, we can't "lift" the context `R` from `Stream` to `Effect`, as it results in a compilation error.

For example, this declaration:

```ts
shareRefCount(): Effect.Effect<Stream.Stream<A, E>, never, R>
```

will result in the following error:

```ts
packages/effect/src/internal/stream.ts:5730:5 - error TS2375: Type 'Effect<Stream<A, E, R>, never, never>' is not assignable to type 'Effect<Stream<A, E, never>, never, Scope | R>' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Type 'Stream<A, E, R>' is not assignable to type 'Stream<A, E, never>' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
    Type 'R' is not assignable to type 'never'.

5730     return Effect.gen(function*(){
         ~~~~~~

  packages/effect/src/internal/stream.ts:5724:10
    5724   <A, E, R>(
                  ~
    This type parameter might need an `extends never` constraint.
```

This is correct because the resulting `Stream` actually requires the context `R` as it runs `runIntoPubSub(self, connector)` within its pipeline, not upon `yield* shareRefCount()`.

Thus, the correct declaration should be:

```ts
declare function shareRefCount(...): Effect.Effect<Stream.Stream<A, E, R>>
```

which, in my opinion, offers no practical benefit.

The only reason to use this form:

```ts
declare function shareRefCount(...): Effect.Effect<Stream.Stream<A, E, R>>
```

is to create two different instances with separate refCount counters. But this can also be done with:

```ts
declare function shareRefCount(...): Stream.Stream<A, E, R>
```

In the first case, you'd write:

```ts
declare const upstream: Stream.Stream<number>
const sharedStreamConstructor = upstream.pipe(Stream.shareRefCount());

const sharedStreamInstance1 = yield* sharedStreamConstructor;
const sharedStreamInstance2 = yield* sharedStreamConstructor;
```

In the second case, you'd write:

```ts
declare const upstream: Stream.Stream<number>
const sharedStreamInstance1 = upstream.pipe(Stream.shareRefCount());
const sharedStreamInstance2 = upstream.pipe(Stream.shareRefCount());
```

The constructor step seems unnecessary in this chain.

However, if this is necessary, I will do it as you want. I would appreciate an explanation of why you think it's better—I'm genuinely curious. If you don't want to explain, just let me know, and I will still implement it as requested.

dilame:
I also implemented a test that dead-locks in case of using `Effect.forkScoped` under the hood of `shareRefCount`

tim-smart:
> I also implemented a test that dead-locks in case of using `Effect.forkScoped` under the hood of `shareRefCount`

You should be using `forkIn` in this situation, to fork in the parent scope: ```ts
const forkingStream = <A, E, R>(self: Stream.Stream<A, E, R>) =>
  Effect.gen(function*() {
    const scope = yield* Effect.scope
    const context = yield* Effect.context<R>()
    let fiber: Fiber.RuntimeFiber<void, E> | undefined = undefined
    return Effect.gen(function*() {
      fiber = yield* Stream.runDrain(self).pipe(
        Effect.locally(FiberRef.currentContext, context),
        Effect.forkIn(scope)
      )
    })
  })
```

**Comment by dilame** [comment]
Do you mean like this (only about forking mechanism)?

**Comment by tim-smart** [comment]
> Do you mean like this (only about forking mechanism)?

Yes

File: packages/effect/src/internal/stream.ts [hunk 4]
```diff
@@ -5709,42 +5710,49 @@ export const shareRefCount = dual<
...
+          Effect.forkIn(scope)
+        )
+        return flattenTake(fromPubSub(connector))
+      }).pipe(
+        unwrap,
+        ensuring(
+          Effect.suspend(() => {
+            refCount--
+            const cleanup = fiber
+            fiber = null
```
tim-smart, line 5748: You are clearing the fiber even if the ref count is not 0.

datner:
@dilame Thank you for your detailed explanation! I'll try to address what I can you lower the cognitive load for @tim-smart  and I'm sure he would correct me if I say something wrong :)

> I feel there is a misunderstanding about this operator. Forking upstream consumption inside the resulting effect's scope is impossible due to its nature.

> As we discussed earlier, streams operate within a Scope they create themselves if using non-scoped Stream.run*.

> Regarding wrapping this in Effect, I think it's pointless. Only the counter initializes; no effects execute during initialization.

The forking is not the relevant action that is being addressed (it's just the affected party, in this case). You should stagger the construction of the stream so that the environment could be accessed as soon as possible. I don't mean the `Scope` of the `Stream`. I mean the scope of the `PubSub`. By _not_ staggering the construction you create a brand new `PubSub` for every time a stream is evaluated that lives only for the duration of said stream. Practically speaking, you can remove it entirely without changing the behavior of the util at all.

> Using forkDaemon is the only correct and 100% resource-safe solution

It's really not. I am pretty sure that for `n` listeners you actually zombify `n-1` of them.

packages/effect/src/internal/stream.ts
```diff
@@ -5709,42 +5710,49 @@ export const shareRefCount = dual<
...
-        })
+      let refCount = 0
+      let connector: PubSub.PubSub<Take.Take<A, E>> | null = null
+      let fiber: Fiber.RuntimeFiber<void> | null = null
+      const scope = yield* Effect.scope
+      const context = yield* Effect.context<R>()
+      return Effect.gen(function*() {
+        refCount++
+        connector ??= yield* (config.connector ?? PubSub.bounded<Take.Take<A, E>>(16))
+        fiber ??= yield* runIntoPubSub(self, connector).pipe(
```
tim-smart, line 5737: There is a potential race condition here, as the fiber could yield to the scheduler and another have another one sneak in.

Hard to reproduce and unlikely to happen, but possible.

dilame: Is there a way to prevent it?

packages/effect/src/internal/stream.ts
datner: it doesn't matter what is the ref count, theres a new pubsub and new forked fiber for every evaluation of listening

packages/effect/src/internal/stream.ts
tim-smart: You can use a semaphore or a SynchronizedRef to ensure only one access at a time.

packages/effect/src/internal/stream.ts
dilame: I use [Nullish coalescing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment), so i don't think it's the case here

datner: You're right, only a new forked fiber on every eval, not new `PubSub`, I glazed over the `??=`

You should still instantiate the `PubSub` earlier imo. Theres no reason that a pubsub should be instantiated as a hidden side-effect of listening to a stream

packages/effect/src/internal/stream.ts
dilame: Is there a way to just declare an Effect as "unswitchable" or something like this? So the scheduler can not switch to another context until this effect is executed?

dilame: Do you mean like this?

packages/effect/src/internal/stream.ts
```diff
@@ -5675,6 +5676,90 @@ export const scoped = <A, E, R>(
...
+      const scope = yield* Effect.scope
+      const context = yield* Effect.context<R>()
+      const sem = yield* Effect.makeSemaphore(1)
+
+      return Effect.gen(function*() {
+        refCount++
+        connector ??= yield* (config.connector ?? PubSub.bounded<Take.Take<A, E>>(16))
+        fiber ??= yield* runIntoPubSub(self, connector).pipe(
+          Effect.locally(FiberRef.currentContext, context),
+          Effect.forkIn(scope)
```
datner, line 5741:
```ts
declare const stream: Stream<number, never, Foo>
Effect.gen(function* () {
  const shared = yield* sharedRefCount(stream)
  yield* doSomethingWith(shared).pipe(
    Effect.provide(Foo1Layer),
    Effect.forkScoped,
  )
  yield* doSomethingWith(shared).pipe(
    Effect.provide(Foo2Layer),
    Effect.forkScoped
  )
})

```
@tim-smart if I understand correctly, both instances would use the `Foo1` version. No?

tim-smart: They would use whatever was supplied to the `sharedRefCount` effect.

packages/effect/src/internal/stream.ts
tim-smart: Yes, a semaphore will prevent this.

packages/effect/src/internal/stream.ts
datner: despite this effect being ran as part of the stream evaluation?  How?

packages/effect/src/Stream.ts
```diff
@@ -483,28 +484,62 @@ export const branchAfter: {
 export const broadcast: {
   <N extends number>(
     n: N,
-    maximumLag: number
+    bounded: number | { readonly capacity: number; readonly replay?: number | undefined }
```
tim-smart, line 487: Why did you rename this?

packages/effect/src/internal/stream.ts
```diff
@@ -707,41 +708,79 @@ export const broadcast = dual<
...
+    }: {
+      readonly capacity: number
+      readonly replay?: number | undefined
+      readonly idleTimeToLive?: Duration.DurationInput | undefined
+    }
+  ): Effect.Effect<Stream.Stream<A, E, Scope.Scope>, never, R | Scope.Scope> =>
+    RcRef.make({
+      acquire: broadcastDynamic(self, { capacity, replay }),
+      idleTimeToLive
+    }).pipe(Effect.map((rcRef) => unwrap(RcRef.get(rcRef))))
```
tim-smart, line 762: This should be `unwrapScoped`

tim-smart (changes requested):
I think we should stick with the `maximumLag` terminology.

packages/effect/src/Stream.ts
tim-smart: I have addressed additional pubsub options in another PR.

dilame: Ok, got it, i should focus on shareRefCount in this PR. I don't work today, but will finish when get back to work

tim-smart (approved):

PR #3083, opened 2024-06-25
Micro: update `fromOption` to change the failure type to `NoSuchElementException`

gcanti:


packages/effect/src/Micro.ts
```diff
@@ -981,17 +981,17 @@ export const sync = <A>(evaluate: LazyArg<A>): Micro<A> =>
   })
 
 /**
- * Converts an `Option` into a `Micro` effect, that will fail with a
- * `Option.None` if the option is `None`. Otherwise, it will succeed with the
+ * Converts an `Option` into a `Micro` effect, that will fail with
+ * `void` if the option is `None`. Otherwise, it will succeed with the
```
tim-smart, line 985:
```suggestion
 * `NoSuchElementException` if the option is `None`. Otherwise, it will succeed with the
```

tim-smart (approved):

PR #3085, opened 2024-06-25
Micro: add `zipWith`

gcanti:
Note: I think `Micro.zip` is redundant since we have `Micro.all`. We could remove it.

### DIFF
--- a/packages/effect/src/Micro.ts
+++ b/packages/effect/src/Micro.ts
@@ -1659,9 +1659,7 @@ export const zip: {
   self: Micro<A, E, R>,
   that: Micro<A2, E2, R2>,
   options?: { readonly concurrent?: boolean | undefined }
-): Micro<[A, A2], E | E2, R | R2> => {
-  return zipWith(self, that, (a, a2) => [a, a2], options)
-})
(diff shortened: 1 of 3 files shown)


packages/effect/src/Micro.ts
```diff
@@ -1653,23 +1653,48 @@ export const zip: {
...
+    options?: { readonly concurrent?: boolean | undefined }
+  ): Micro<B, E2 | E, R2 | R>
+} = dual((args) => isMicro(args[1]), <A, E, R, A2, E2, R2, B>(
+  self: Micro<A, E, R>,
+  that: Micro<A2, E2, R2>,
+  f: (a: A, b: A2) => B,
+  options?: { readonly concurrent?: boolean | undefined }
+): Micro<B, E2 | E, R2 | R> =>
+  map(
+    all([self, that], { concurrency: options?.concurrent ? 2 : 1 }),
```
tim-smart, line 1694: You would only use `all` for the concurrent case. It adds a bit of overhead with all the concurrency book-keeping.

tim-smart (approved):

PR #3107, opened 2024-06-27
Adds a `--log-level` built-in option to all commands

### LINKED ISSUE #2978: Cli: --log-level flag
It would be nice if @effect/cli generated a --log-level flag that set Logger.withMinimumLogLevel w/in the context of the cli commands. It's not too tedious to write by hand but seems general enough to be available for all cli applications.

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a `--log-level` built-in option to all commands which can be used to control the **minimum** `LogLevel` of the command's associated handler.

cc @jessekelly881 given you were the one who requested the feature.

The question really is whether I should set the **minumum** log-level with the option, or just set the log-level. I thought setting the **minumum** log-level made more sense from a UX perspective, but would like to hear others' thoughts.

## Related

- Related Issue #2978
- Closes #2978


jessekelly881:
+1 for minimum. If I set --log-level=debug I would still expect errors to be shown. Hmm. Maybe another name besides --log-level would make that clearer..

IMax153:
> Vite has a --logLevel flag (w/ "silent" as an option as an option. 

I think the equivalent here would be --log-level=none. I want to try to avoid introducing extraneous terminology that deviates from the LogLevel in effect. 

If minimum sounds good though, I'll probably proceed with the PR as is.

PR #3121, opened 2024-06-28
Support for the tacit usage of external handlers for `Match.tag` and `Match.tagStartsWith` functions

KhraksMamtsov:
## Type

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
+ used namespace for Types module to avoid collisions with the native `NoInfer` intrinsic

## Related

- Related Issue #
- Closes  https://github.com/Effect-TS/effect/issues/3120

### DIFF
--- a/.changeset/lovely-monkeys-compete.md
+++ b/.changeset/lovely-monkeys-compete.md
@@ -1,0 +1,17 @@
+---
+"effect": patch
+---
+
+Support for the tacit usage of external handlers for `Match.tag` and `Match.tagStartsWith` functions
+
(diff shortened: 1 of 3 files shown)


datner:
Does using the native `NoInfer` works here? Or only the `NoInfer` hack?

KhraksMamtsov:
> Does using the native `NoInfer` works here? Or only the `NoInfer` hack?

Only hack. Native intrinsic infers handler argument as `never`

PR #3131, opened 2024-06-30
feat(Stream): raceAll implementation

dilame:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I implemented raceAll using high-level API, not sure if it's a correct way. I just don't understand the low-level API yet:)

### DIFF
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -2696,7 +2696,7 @@ export const provideSomeLayer: {
 
 /**
  * Returns a stream that mirrors the first upstream to emit an item.
- * As soon as one of the upstream emits a first value, all the others are being interrupted.
+ * As soon as one of the upstream emits a first value, all the others are interrupted.
  * The resulting stream will forward all items from the "winning" source stream.
  * Any upstream failures will cause the returned stream to fail.
  *
@@ -2704,9 +2704,9 @@ export const provideSomeLayer: {
  * import { Stream, Schedule, Console, Effect } from "effect"
  *
  * const stream = Stream.raceAll(
- *     Stream.fromSchedule(Schedule.spaced('1 millis')),
- *     Stream.fromSchedule(Schedule.spaced('2 millis')),
- *     Stream.fromSchedule(Schedule.spaced('4 millis')),
+ *   Stream.fromSchedule(Schedule.spaced('1 millis')),
+ *   Stream.fromSchedule(Schedule.spaced('2 millis')),
+ *   Stream.fromSchedule(Schedule.spaced('4 millis')),
  * ).pipe(Stream.take(6), Stream.tap(Console.log))
  *
  * Effect.runPromise(Stream.runDrain(stream))
@@ -2718,7 +2718,7 @@ export const provideSomeLayer: {
  * // 4
  * // 5
  * @since 3.5.0
- * @category utils
+ * @category racing
  */
 export const raceAll: <S extends ReadonlyArray<Stream<any, any, any>>>(
   ...streams: S
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -2716,12 +2716,12 @@ export const provideSomeLayer: {
  * @since 3.5.0
  * @category utils
  */
-export const raceAll: <T extends ReadonlyArray<Stream<any, any, any>>>(
-  ...streams: T
+export const raceAll: <S extends ReadonlyArray<Stream<any, any, any>>>(
+  ...streams: S
 ) => Stream<
-  [T[number]] extends [never] ? never : T[number] extends Stream<infer A, any, any> ? A : never,
-  [T[number]] extends [never] ? never : T[number] extends Stream<any, infer E, any> ? E : never,
-  [T[number]] extends [never] ? never : T[number] extends Stream<any, any, infer R> ? R : never
+  Stream.Success<S[number]>,
+  Stream.Error<S[number]>,
+  Stream.Context<S[number]>
 > = internal.raceAll
 
 /**
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -2695,7 +2695,11 @@ export const provideSomeLayer: {
 } = internal.provideSomeLayer
 
 /**
- * Returns a stream that mirrors the first source stream to emit an item.
+ * Returns a stream that mirrors the first upstream to emit an item.
+ * As soon as one of the upstream emits a first value, all the others are being interrupted.
+ * The resulting stream will forward all items from the "winning" source stream.
+ * Any upstream failures will cause the returned stream to fail.
+ *
  * @example
  * import { Stream, Schedule, Console, Effect } from "effect"
  *
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -2695,6 +2695,36 @@ export const provideSomeLayer: {
 } = internal.provideSomeLayer
 
 /**
+ * Returns a stream that mirrors the first source stream to emit an item.
+ * @example
+ * import { Stream, Schedule, Console, Effect } from "effect"
+ *
+ * const stream = Stream.raceAll(
+ *     Stream.fromSchedule(Schedule.spaced('1 millis')),
+ *     Stream.fromSchedule(Schedule.spaced('2 millis')),
+ *     Stream.fromSchedule(Schedule.spaced('4 millis')),
+ * ).pipe(Stream.take(6), Stream.tap(Console.log))
+ *
+ * Effect.runPromise(Stream.runDrain(stream))
+ * // Output each millisecond from the first stream, the rest streams are interrupted
+ * // 0
+ * // 1
+ * // 2
+ * // 3
+ * // 4
+ * // 5
+ * @since 3.5.0
+ * @category utils
+ */
+export const raceAll: <T extends ReadonlyArray<Stream<any, any, any>>>(
+  ...streams: T
+) => Stream<
+  [T[number]] extends [never] ? never : T[number] extends Stream<infer A, any, any> ? A : never,
+  [T[number]] extends [never] ? never : T[number] extends Stream<any, infer E, any> ? E : never,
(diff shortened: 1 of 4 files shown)


dilame:
Now it should be a correct way

packages/effect/src/internal/stream.ts
```diff
@@ -4686,6 +4686,39 @@ export const range = (min: number, max: number, chunkSize = DefaultChunkSize): S
     return new StreamImpl(go(min, max, chunkSize))
   })
 
+export const raceAll = <T extends ReadonlyArray<Stream.Stream<any, any, any>>>(
+  ...streams: T
+): Stream.Stream<
+  [T[number]] extends [never] ? never : T[number] extends Stream.Stream<infer A, any, any> ? A : never,
```
tim-smart, line 4692: You can use `Stream.Stream.Success` etc here

packages/effect/src/internal/stream.ts
```diff
@@ -4686,6 +4686,39 @@ export const range = (min: number, max: number, chunkSize = DefaultChunkSize): S
...
+  [T[number]] extends [never] ? never : T[number] extends Stream.Stream<any, infer E, any> ? E : never,
+  [T[number]] extends [never] ? never : T[number] extends Stream.Stream<any, any, infer R> ? R : never
+> => {
+  let finished = false
```
tim-smart, line 4696: Any state needs to be constructed in the context of the running stream (as they are lazy). So inside the Effect.gen in this case.

packages/effect/src/internal/stream.ts
```diff
@@ -4686,6 +4686,39 @@ export const range = (min: number, max: number, chunkSize = DefaultChunkSize): S
...
+              if (finished) {
+                return Effect.void
+              }
+              finished = true
+              return Effect.all(
+                halts.map((def, i) => i === index ? Effect.void : Deferred.succeed(def, void 0))
+              )
+            }),
+            interruptWhenDeferred(halts[index]!),
+            haltWhenDeferred(halts[index]!)
```
tim-smart, line 4713: Hmm you should only need the interrupt api here.

packages/effect/src/internal/stream.ts
```diff
@@ -4686,6 +4686,39 @@ export const range = (min: number, max: number, chunkSize = DefaultChunkSize): S
...
+  ...streams: T
+): Stream.Stream<
+  [T[number]] extends [never] ? never : T[number] extends Stream.Stream<infer A, any, any> ? A : never,
+  [T[number]] extends [never] ? never : T[number] extends Stream.Stream<any, infer E, any> ? E : never,
+  [T[number]] extends [never] ? never : T[number] extends Stream.Stream<any, any, infer R> ? R : never
+> => {
+  let finished = false
+  return unwrap(
+    Effect.gen(function*() {
+      const halts = yield* Effect.all(streams.map(() => Deferred.make<void>()))
```
tim-smart, line 4699: You should be able to do this with only one Deferred, and catching the signal for the elected stream.

packages/effect/src/internal/stream.ts
dilame: The funny thing is that it was me who created `Stream.Stream.Success` about a week ago, but i forgot that it was merged:) Definetely it should be used here

packages/effect/src/internal/stream.ts
dilame: this is the most important comment, thank you

dilame:
@tim-smart I think i'm done, could you review, please?

packages/effect/src/internal/stream.ts
```diff
@@ -4686,6 +4686,44 @@ export const range = (min: number, max: number, chunkSize = DefaultChunkSize): S
...
+            takeWhile(() => {
+              if (winner === index) {
+                return true;
+              } else if (winner === null) {
+                winner = index;
+                Deferred.unsafeDone(halt, Exit.void);
+                return true;
+              }
+              return false;
+            }),
```
tim-smart, line 4711: This feels a bit cleaner

```suggestion
            takeWhile(() => {
              if (winner === null) {
                winner = index;
                Deferred.unsafeDone(halt, Exit.void);
                return true;
              }
              return winner === index;
            }),
```

packages/effect/src/Stream.ts
```diff
@@ -2694,6 +2694,36 @@ export const provideSomeLayer: {
...
+ * import { Stream, Schedule, Console, Effect } from "effect"
+ *
+ * const stream = Stream.raceAll(
+ *     Stream.fromSchedule(Schedule.spaced('1 millis')),
+ *     Stream.fromSchedule(Schedule.spaced('2 millis')),
+ *     Stream.fromSchedule(Schedule.spaced('4 millis')),
+ * ).pipe(Stream.take(6), Stream.tap(Console.log))
+ *
+ * Effect.runPromise(Stream.runDrain(stream))
+ * // Output each millisecond from the first stream, the rest streams are interrupted
```
tim-smart, line 2709:
```suggestion
 * Returns a stream that mirrors the first source stream to emit an item.
 * Any upstream failures will cause the returned stream to fail.
 *
 * @example
 * import { Stream, Schedule, Console, Effect } from "effect"
 *
 * const stream = Stream.raceAll(
 *   Stream.fromSchedule(Schedule.spaced('1 millis')),
 *   Stream.fromSchedule(Schedule.spaced('2 millis')),
 *   Stream.fromSchedule(Schedule.spaced('4 millis')),
 * ).pipe(Stream.take(6), Stream.tap(Console.log))
 *
 * Effect.runPromise(Stream.runDrain(stream))
 * // Output each millisecond from the first stream, the rest of the streams are interrupted.
```

tim-smart (changes requested):
Couple more small things, then we can merge :)

packages/effect/src/internal/stream.ts
dilame: Yes, i thought about it, but while it really feels cleaner, i decided to go that way because of perfomance.

The predicate function will be executed only once for all the loosers, but can be executed an infinite amount of times for the winner. 

It means that if i apply your suggestion, each item will be checked to `null` first, and then checked to `winner === index`, so it is 2 checks per each emission. 

I optimised it for the winner stream by placing the check for `winner === index` first, so the first check is successful for the winner branch, which is the most sensitive branch than the loosers.

packages/effect/src/Stream.ts
dilame: Yes, i will make the documentation more descriptive now

packages/effect/src/internal/stream.ts
tim-smart: The performance impact here will be zero, so I generally lead towards readability. But I'm OK with either approach.

dilame:
@tim-smart a strange thing is happening.

https://effect.website/play#a7005d4f9f10

The program is finished, but the process is hanging. I didn't get the reason yet, will investigate, but it looks like a deep bug somewhere out of this PR scope

dilame:
I improved the docs, refactored the `takeWhile` predicate function body to be more readable.
It stucks on the docs generation now, and the reason is this behaviour that looks like a bug

tim-smart:
> @tim-smart a strange thing is happening.
> 
> https://effect.website/play#a7005d4f9f10
> 
> The program is finished, but the process is hanging. I didn't get the reason yet, will investigate, but it looks like a deep bug somewhere out of this PR scope

That seems to be exiting fine. Not sure what the docs issue is then.

dilame:
No, it's not exiting. I am checking it locally on my machine.

dilame:
I'm sorry, i provided a wrong code. This one is not exiting

https://effect.website/play#2bd210998527

dilame:
It does exiting as expected if run it with `NodeRuntime.runMain`, but doesn't with 'Effect.runPromise`

dilame:
`NodeRuntime` waits until program is finished and forcefully kills the process. `Effect.runPromise` just runs the program.

I assume the issue is somewhere here https://github.com/Effect-TS/effect/blob/main/packages/effect/src/internal/core.ts#L1532-L1537

dilame:
Yes, the issue is here.

I modified the `node_modules/effect/dist/esm/internal/core.js` like this

```
export const never = /*#__PURE__*/async(() => {
  const interval = setInterval(() => void 0, 2 ** 31 - 1);
  console.log('set interval');
  return sync(() => {
    console.log('clear interval');
    clearInterval(interval)
  });
});
```

and the output is 

```
tsx ./src/bugs/effect.interrupt.bug.ts
set interval
0
1
2
3
4
5
```

There are no `clear interval` log

dilame:
I modified the code like this, and it works now. Looks like a dirty workaround, not sure if it's a good idea to do it. What do you think about it?

```ts
export const raceAll = <S extends ReadonlyArray<Stream.Stream<any, any, any>>>(
  ...streams: S
): Stream.Stream<
  Stream.Stream.Success<S[number]>,
  Stream.Stream.Error<S[number]>,
  Stream.Stream.Context<S[number]>
> =>
  Effect.all([Deferred.make<void>(), Deferred.make<void>()]).pipe(
    Effect.map(([halt, forever]) => {
      let winner: number | null = null;
      return Stream.mergeAll(
        streams.map((stream, index) =>
          stream.pipe(
            Stream.takeWhile(() => {
              if (winner === null) {
                winner = index;
                Deferred.unsafeDone(halt, Exit.void);
                return true;
              }
              return winner === index;
            }),
            Stream.interruptWhen(
              Deferred.await(halt).pipe(
                Effect.flatMap(() =>
                  winner === index ? Deferred.await(forever) : Effect.void,
                ),
              ),
            ),
          ),
        ),
        { concurrency: streams.length },
      );
    }),
    Stream.unwrap,
  );
```

dilame:
I think we need @mikearnaldi here, because the bug looks serious

dilame:
Here is an even easier code to reproduce

```ts
import { Effect, Schedule, Stream } from 'effect';

const program = Stream.fromSchedule(Schedule.spaced('1 second')).pipe(
  Stream.interruptWhen(Effect.never),
  Stream.take(1),
  Stream.runCollect,
);

Effect.runPromise(program);
```

tim-smart (approved):

PR #3135, opened 2024-07-01
add PubSub `replay` option

tim-smart:


packages/effect/src/internal/pubsub.ts
```diff
@@ -1141,6 +1176,105 @@ class PubSubImpl<in out A> implements PubSub.PubSub<A> {
...
+  readonly [queue.DequeueTypeId] = queue.dequeueVariance
+
+  constructor(
+    readonly backing: PubSubImpl<A>,
+    readonly replayBufferSize: number
+  ) {}
+
+  replayBuffer = Chunk.empty<A>()
+  offerReplay(value: A): void {
+    this.replayBuffer = Chunk.append(this.replayBuffer, value)
```
dilame, line 1191: Maybe it could be a good idea to derive `offerReplay` from `offerAllReplay` 
```ts
this.offerAllReplay(Chunk.of(value))
```

dilame:
While the `withReplay` implementation is more flexible, I would prefer having a separate `PubSub.replay` for several reasons.

First, `PubSub.replay` uses a Linked List, which makes it faster in terms of performance.

Second, there are issues with consistency when combining `withReplay` with `PubSub.sliding`. `PubSub.sliding` is designed to discard the oldest unprocessed value, and I expect this value to be the one in the `replayBuffer`. For example:

```ts
const pubsub = yield* PubSub.sliding<number>(3).pipe(
    Effect.map(PubSub.withReplay(3))
)
```

In this scenario, if we emit `1, 2, 3`, they are stored in the `replayBuffer`. When a new subscription `S1` starts processing from `1`, and we emit `4, 5, 6, 7`, S1 will process `1, 2, 3, 5, 6, 7`. 
However, based on the sliding behavior, the logical expectation would be for `S1` to process `1, 3, 4, 5, 6, 7`

Third, the `replayBufferSize` is not considered when determining if the `dropping` or `bounded` pubsub is full. This undermines the point of these types. If we use `dropping` or `bounded` or `sliding`, it's usually to save resources like CPU or memory. Adding replay behaviour that doesn't take `replayBuffer` into capacity accounting is kinda "_poisons_" these pubsub types.

In summary, while `PubSub.withReplay` is functional, its use with sliding, dropping, and bounded variants reveals design flaws. The only clear and effective use case is with `unbounded` pubsub. Therefore, i would prefer a separate `PubSub.replay` as it provides a simpler and faster solution, ensuring better performance and clearer behavior consistency.

dilame:
I think it only makes sense to track replays for `sliding`/`bounded`/`dropping` separately in each of these implementations.

So, for `PubSub.sliding({capacity: 5, replay: 10})` we are sliding the real "oldest value".
For `PubSub.bounded({capacity: 5, replay: 10})` we are rejecting the offer if the current `replayBuffer` length is more than capacity.
And for `PubSub.dropping({capacity: 5, replay: 10})` we are dropping the offer if the current `replayBuffer` length is more than capacity.

tim-smart:
> its use with sliding, dropping, and bounded variants

Bounded and dropping is an easy fix - only add items to the replay buffer if they are successfully added to the pub sub.

Sliding is the only tricky one, but I want to see if we can support it as I have a use case for it.

packages/effect/src/internal/pubsub.ts
```diff
@@ -137,6 +138,12 @@ export const publishAll = dual<
 export const subscribe = <A>(self: PubSub.PubSub<A>): Effect.Effect<Queue.Dequeue<A>, never, Scope.Scope> =>
   self.subscribe
 
+/** @internal */
+export const withReplay = dual<
+  (n: number) => <A>(self: PubSub.PubSub<A>) => PubSub.PubSub<A>,
```
dilame, line 143: What do you think about 
```ts
  (n: number) => <A>(self: Effect.Effect<PubSub.PubSub<A>>) => Effect.Effect<PubSub.PubSub<A>>,
```

So we can write

```ts
const replayPubSub = yield* withReplay(3)(PubSub.unbounded())
```

Because 
```ts
const pubsub = yield* PubSub.unbounded<number>().pipe(
    Effect.map(PubSub.withReplay(3))
  )
```
looks verbose to me

dilame:
I think i have a good idea.

We can expose the most efficient API for each use-case. I offered a good (as i think) implementation for unbounded pubsub replay, which is better than independent tracking in terms of perfomance.

We also have a good implementation of tracking replays for `bounded` and `dropping` pubsubs in this PR.

We can also write a separated implementation of replays for `sliding` variant.

And then we can expose such APIs

```ts
PubSub.unboundedReplay();
PubSub.boundedReplay();
PubSub.droppingReplay();
PubSub.slidingReplay();
```

Which would use different replay tracking mechanisms under the hood.

packages/effect/src/internal/pubsub.ts
tim-smart: Went with a constructor option.

tim-smart:
> We can expose the most efficient API for each use-case.

From a maintainability perspective, I would prefer not to have a replay version of each AtomicPubSub implementation.
Also, things like `capacity` would be different depending on the strategy, so keeping the replay buffer separate keeps it cleaner.

dilame:
>  I would prefer not to have a replay version of each AtomicPubSub implementation.

Yes, of course, i didn't mean that. I just mean that we could use my implementation of unbounded replay, and your implementation for all the rest AtomicPubSub implementations, just because IMHO my implementaion is more efficient for unbounded type, and we already have it:)

packages/effect/src/PubSub.ts
```diff
@@ -46,7 +46,10 @@ export interface PubSub<in out A> extends Queue.Enqueue<A>, Pipeable {
  * @since 2.0.0
  * @category constructors
  */
-export const bounded: <A>(requestedCapacity: number) => Effect.Effect<PubSub<A>> = internal.bounded
+export const bounded: <A>(
+  requestedCapacity: number,
+  options?: { readonly replayCapacity?: number }
+) => Effect.Effect<PubSub<A>> = internal.bounded
```
dilame, line 52: Maybe it makes sense to implement like that?
```ts
export const bounded: <A>(
  capacity: number | {capacity: number; replay?: number},
) => Effect.Effect<PubSub<A>> = internal.bounded
```

Because the usage looks non-consistent

```ts
PubSub.bounded(5, {replayCapacity: 10})
```

in comparison with

```ts
PubSub.bounded({capacity: 5, replay: 10})
```

tim-smart:
> my implementaion is more efficient for unbounded type

I ran some benchmarks and they are pretty much the same, so I think we should just go with the less complex option.

packages/effect/src/PubSub.ts
tim-smart: Yeah that might be better.

dilame:
> I ran some benchmarks and they are pretty much the same

Yep, i saw that you implemented `ReplayBuffer` based on linked list, and it should be the same from the performance point of view, but less effective from RAM point of view.

Let's say we have `PubSub.unbounded({replay: Infinity})` and one slow subscriber. We emit 10000 elements, but we have 20000 in RAM because we keep the same elements in 2 places

tim-smart:
> We emit 10000 elements, but we have 20000 in RAM because we keep the same elements in 2 places.

It will just be overhead in terms of the replay buffer nodes, as the actual elements will be the same reference.

dilame:
> It will just be overhead in terms of the replay buffer nodes, as the actual elements will be the same reference.

Yes, it will work for objects, but not for primitives, especially if it is a long strings

tim-smart:
> Yes, it will work for objects, but not for primitives, especially if it is a long strings

If you have a large string, and rebind it to another variable, it will not clone the string:

```ts
const largeString = "....."

const obj = {
  value: largeString // still the same string in memory
}
```

dilame:
As far as i understand the JS – each time you pass a primitive to somewhere – you get the new copy of it.

https://javascript.info/object-copy

tim-smart:
> As far as i understand the JS – each time you pass a primitive to somewhere – you get the new copy of it in memory.
> 
> https://javascript.info/object-copy

No, only mutating the string will cause a new one to be allocated.

dilame:
I checked the memory snapshot and i was wrong – it is the same reference
Screenshot 2024-07-05 at 07 25 26

packages/effect/src/internal/pubsub.ts
```diff
@@ -368,13 +406,15 @@ class BoundedPubSubPow2<in out A> implements AtomicPubSub<A> {
   subscriberCount = 0
   subscribersIndex = 0
 
-  readonly capacity: number
+  constructor(readonly capacity: number, readonly replayBuffer: ReplayBuffer<A> | undefined) {
+    this.array = Array.from({ length: capacity })
+    this.mask = capacity - 1
+    this.subscribers = Array.from({ length: capacity })
+    this.capacity = capacity
```
tim-smart, line 413: Remove

packages/effect/src/internal/pubsub.ts
```diff
@@ -1562,3 +1657,102 @@ const unsafeStrategyCompleteSubscribers = <A>(
...
+
+interface ReplayWindow<A> {
+  take(): A | undefined
+  takeN(n: number): Chunk.Chunk<A>
+  takeAll(): Chunk.Chunk<A>
+  readonly remaining: number
+}
+
+class ReplayWindowImpl<A> implements ReplayWindow<A> {
+  readonly size: number
```
tim-smart, line 1705: Unused

dilame:
WOW WOW WOW HOT IN HERE!!! 🔥

PR #3146, opened 2024-07-03
add Logger.prettyLogger & Logger.pretty

tim-smart:


tim-smart:
> It would be really nice if this was more customizable

I think we should lean towards a consistent experience over customizability here. In saying that, I think date formatting is an OK option.

PR #3147, opened 2024-07-03
return Exit values from `Fiber.awaitAll`

patroza:


packages/effect/src/Fiber.ts
```diff
@@ -275,7 +275,8 @@ export {
  * @since 2.0.0
  * @category destructors
  */
-export const awaitAll: (fibers: Iterable<Fiber<any, any>>) => Effect.Effect<void> = fiberRuntime.fiberAwaitAll
+export const awaitAll: <A, E>(fibers: Iterable<Fiber<A, E>>) => Effect.Effect<Array<Exit.Exit<A, E>>> =
```
patroza, line 278: Hmm, this is a major departure from the original implementation.
I wonder why this was implemented as `await(fiberAll(fibres))`, instead of `forEach(fibers, Fiber.await)`

mikearnaldi: any failing tests?

patroza: Just the one I updated which was expecting the void result.

mikearnaldi: then I think it is fine to change the behaviour, the old one is odd

tim-smart: Technically this is a breaking change. Did we still want to keep this as a minor changeset?

patroza: Technically it's also a bug fix. hmm

tim-smart: @mikearnaldi I'm ok with this change as minor. What do you think?

tim-smart:
The types need to be improved to handle Iterable's correctly.

PR #3157, opened 2024-07-04
add PrimaryKey.valueOrHash for generating a primary key from `unknown`

tim-smart:


mikearnaldi:
Hash is not supposed to be unique, it can't be used as a key. Also it is not generally stable

tim-smart:
Yeah I was starting to think that too.

PR #3175, opened 2024-07-04
ensure fibers are interrupted in Stream.mergeWith

tim-smart:


dilame:
@tim-smart thank you for this fix. Just curious if [this line](https://github.com/Effect-TS/effect/blob/next-minor/packages/effect/src/internal/channel.ts#L1714) should also be replaced with `forkIn(pull, scope)`?

tim-smart:
> @tim-smart thank you for this fix. Just curious if [this line](https://github.com/Effect-TS/effect/blob/next-minor/packages/effect/src/internal/channel.ts#L1714) should also be replaced with `forkIn(pull, scope)`?

No, it only was occurring when one of the sides was empty.

PR #3179, opened 2024-07-05
add RcMap & RcRef modules

tim-smart:


packages/effect/src/RcMap.ts
```diff
@@ -0,0 +1,188 @@
...
+declare namespace State {
+  interface Open<K, A> {
+    readonly _tag: "Open"
+    readonly map: MutableHashMap.MutableHashMap<K, Entry<A>>
+  }
+  interface Closed {
+    readonly _tag: "Closed"
+  }
+  interface Entry<A> {
+    readonly value: A
```
tim-smart, line 54: Use Deferred to allow concurrent acquisition.

packages/effect/src/RcMap.ts
```diff
@@ -0,0 +1,206 @@
...
+ *   // Get "foo" from the map twice, which will only acquire it once.
+ *   // It will then be released once the scope closes.
+ *   yield* RcMap.get(map, "foo").pipe(
+ *     Effect.andThen(RcMap.get(map, "foo")),
+ *     Effect.scoped
+ *   )
+ * })
+ */
+export const make = <K, A, E, R>(
+  lookup: (key: K) => Effect.Effect<A, E, R>
```
tim-smart, line 95: Support a `idleTimeToLive` option to prevent excessive churn.

tim-smart: Adds quite a bit of complexity, so will leave it out.

dilame:
Would also be nice to have some sort of _debouncing_ to release the resource.

For example, i would like to wait for 10 seconds after ref count became zero, and if no new consumers – release

```ts
const ref = yield* RcRef.make({
  resource: Effect.acquireRelease(Effect.succeed("foo"), () =>
      Effect.log("release foo"),
    ),
  release: Effect.sleep('10 seconds')
});
```

P.S. What does `RcRef` name stands for? `RefCounterRef`?

tim-smart:
> Would also be nice to have some sort of _debouncing_ to release the resource.

Yeah I'm thinking about it already, not sure on the best implementation yet.

RcRef stands for "reference counted ref".

dilame:
Another a little bit crazy but real-world idea. 
Implement some sort of maximum resource capacity. 
It is an often happens that websocket API providers limits topic amount per one connection (binance WS api docs https://binance-docs.github.io/apidocs/spot/en/#limits )

Screenshot 2024-07-08 at 06 10 27

So something like 

```ts
const connection = yield* RcRef.make({
  resource: acquireAndReleaseWsConnection(),
  release: Effect.sleep('10 seconds'),
  capacity: 1024
});

```

Would be a killer feature.

Once the `refCount` exceeds the `capacity` – the new resource is being created

tim-smart:
> I think it will make `RcRef` fully functional-complete

I think you mean `RcMap`? RcRef will only ever have one resource acquired at a time.

I can take a look.

dilame:
> I think you mean `RcMap`?

Oh, yes, sorry

packages/effect/src/internal/rcMap.ts
```diff
@@ -0,0 +1,194 @@
...
+        self.semaphore.withPermits(1),
+        coreEffect.bindTo("entry"),
+        coreEffect.bind("scope", () => fiberRuntime.scopeTag),
+        core.tap(({ entry, scope }) =>
+          scope.addFinalizer(() =>
+            core.suspend(() => {
+              entry.refCount--
+              if (entry.refCount > 0) {
+                return core.void
+              } else if (Duration.isZero(self.idleTimeToLive)) {
```
dilame, line 163: Actually, when i use this, i only use zero-duration to release resource at the next event loop tick.
I think it's better to check if `idleTimeToLive` is undefined, because `undefined` and '0 millis' is not the same

dilame:
Do you think it is better to throw an exception? I am thinking about creating a new resource

```
const map = yield* RcMap.make({
        lookup: (key: string) => Clock.currentTimeMillis,
        capacity: 2
      })
const first = yield* RcMap.get(map, 1); // 0
const second = yield* RcMap.get(map, 2); // 0
yield* Effect.sleep(1000);
const second = yield* RcMap.get(map, 3); // 1000
```

dilame:
Forget about it, i seem to be sleepy already:)

dilame:
I'm thinking about a way to create a new WS connection after the first one is out of capacity.

tim-smart:
> Do you think it is better to throw an exception

I moved it to the error channel, so it is easier to retry / handle.

packages/effect/src/RcMap.ts
```diff
@@ -0,0 +1,97 @@
...
+    readonly capacity?: number | undefined
+  }
+) => Effect.Effect<RcMap<K, A, E>, never, R | Scope.Scope> = internal.make
+
+/**
+ * @since 3.5.0
+ * @category combinators
+ */
+export const get: {
+  <K>(key: K): <A, E>(self: RcMap<K, A, E>) => Effect.Effect<A, E | Cause.ExceededCapacityException, Scope.Scope>
```
dilame, line 95: What if put `capacity` into `RcMap` generics when `make()` it, and then conditionally check if there is a capacity – then `Cause.ExceededCapacityException` could be emitted, otherwise – no?

packages/effect/src/RcMap.ts
```diff
@@ -0,0 +1,97 @@
...
+     * When the reference count reaches zero, the resource will be released
+     * after this duration.
+     */
+    readonly idleTimeToLive?: Duration.DurationInput | undefined
+    /**
+     * The maximum number of resources that can be held in the map.
+     */
+    readonly capacity?: number | undefined
+  }
+) => Effect.Effect<RcMap<K, A, E>, never, R | Scope.Scope> = internal.make
```
dilame, line 88:
```ts
export interface RcMap<in K, out A, out E, out C extends number | undefined> extends Pipeable {
  readonly [TypeId]: RcMap.Variance<K, A, E>
}

export const make: <K, A, E, R, C extends number | undefined>(
  options: {
    readonly lookup: (key: K) => Effect.Effect<A, E, R>
    /**
     * When the reference count reaches zero, the resource will be released
     * after this duration.
     */
    readonly idleTimeToLive?: Duration.DurationInput | undefined
    /**
     * The maximum number of resources that can be held in the map.
     */
    readonly capacity?: C
  }
) => Effect.Effect<RcMap<K, A, E, C>, never, R | Scope.Scope> = internal.make


export const get: {
  <K>(key: K): <A, E, C>(self: RcMap<K, A, E, C>) => 
  Effect.Effect<A, E | (C extends number ? Cause.ExceededCapacityException : never), Scope.Scope>
```

dilame:
What do you think about `RcMapBalancer`? Here is the example of usage:

```ts
const rcMapBalancer = yield* RcMapBalancer.make(
  upsource: makeWebSocketConnection,
  map: (connection) => RcMap.make({
    lookup: (key: string) =>
      Effect.acquireRelease(connection.subscribe(key), () =>
        connection.unsubscribe(key),
      ),
    capacity: 2
  }),
)
```

dilame:
It provides it's own `RcMapBalancer.get` which tries to aquire the resource from the `RcMap` with minimum resources, but if it encounters `Cause.ExceededCapacityException` – it creates a new `upsource` and a new `RcMap` for it. Once the `RcMap` is `0` – it kills the `upsource`

dilame:
Maybe it even should be like this

```ts
const rcMapBalancer = yield* RcMapBalancer.make(
  upsource: Effect.acquireRelease(createConnection(), (connection) =>
        connection.close(),
      ),
  map: (connection) => RcMap.make({
    lookup: (key: string) =>
      Effect.acquireRelease(connection.subscribe(key), () =>
        connection.unsubscribe(key),
      ),
    capacity: 2
  }),
  keepAliveAtLeast: 2
)
```

which keeps at least 2 upsources alive no matter it is empty.

tim-smart:
> What do you think about `RcMapBalancer`? Here is the example of usage:

You can use a pool for this:

```ts
import type { Scope, Stream } from "effect"
import { Effect, Pool, RcMap } from "effect"

interface Connection {
  readonly messages: Stream.Stream<string>
  readonly send: (message: string) => Effect.Effect<void>
}

declare const acquireConnection: Effect.Effect<Connection, never, Scope.Scope>

Effect.gen(function*() {
  const pool = yield* Pool.makeWithTTL({
    acquire: acquireConnection,
    concurrency: 500, // max number of concurrent subscriptions per connection
    min: 0,
    max: 50,
    timeToLive: "1 minute"
  })

  const map = yield* RcMap.make({
    lookup: (topic: string) =>
      Pool.get(pool).pipe(
        Effect.tap((_) => _.send(`subscribe ${topic}`)),
        Effect.acquireRelease((_) => _.send(`unsubscribe ${topic}`))
      )
  })
})
```

dilame:
As i understand Pool – it is not the same.
I mean the WS API provider allows to subscribe 1024 topics within single connection. It's kinda maximum bandwidth one one resource. If we reach that limit – we want to create a new resource for the 1025th topic. So we have first resource with 1024 topics subscribed, and  second resource with 1 topic subscribed. 
Pool is able to count resources, but not able to estimate the amount of usage of particular resource. For the pool the resource is either busy or free.

.changeset/spicy-hats-travel.md
```diff
@@ -0,0 +1,29 @@
...
+An `RcMap` can contain multiple reference counted resources that can be indexed
+by a key. The resources are lazily acquired on the first call to `get` and
+released when the last reference is released.
+
+```ts
+import { Effect, RcMap } from "effect";
+
+Effect.gen(function* () {
+  const map = yield* RcMap.make({
+    lookup: (key: string) =>
```
IMax153, line 16: It may be worth noting both here and in the relevant JSDocs that they `key` of `lookup` will be compared by `Hash`.

packages/effect/test/RcMap.test.ts
```diff
@@ -0,0 +1,114 @@
+import { Cause, Effect, Exit, RcMap, Scope, TestClock } from "effect"
```
IMax153, line 1: Probably worth adding at least one small test with a complex key.

IMax153 (approved):

dilame:
Yes, the pool looks even more suitable, thank you!

PR #3188, opened 2024-07-06
ensure "cause" is rendered in Data.Error output

tim-smart:


packages/effect/src/internal/core.ts
```diff
@@ -2174,23 +2174,19 @@ export const causeSquashWith = dual<
 // -----------------------------------------------------------------------------
 
 /** @internal */
-export const YieldableError: new(message?: string) => Cause.YieldableError = (function() {
+export const YieldableError: new(message?: string, options?: ErrorOptions) => Cause.YieldableError = (function() {
```
johtso, line 2177: What effect is this change supposed to have? Is it supposed to change the default arguments to a `Data.TaggedError` for example?

PR #3194, opened 2024-07-07
include Error.cause stack in log output

tim-smart:


johtso:
Very exciting! Thanks for picking this up

mikearnaldi:
Shall we merge this @tim-smart ?

tim-smart:
> Shall we merge this @tim-smart ?

I'll test bun and a browser tomorrow, then I think we can.

PR #3203, opened 2024-07-09
export Random.make taking hashable values as seed

giacomoran:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Proposed changes:

- Export `Random.make`
- Allow any hashable value as a seed for `Random.make`, instead of 32-bit integers

These changes enable developers to create instances of the `Random` service with a specified seed.

Use cases:

- For visual tests where the UI data is randomly generated, we use the string "test" as a seed.
- To compute data that should vary daily but remain consistent within the same day, we use a date as a seed.

### DIFF
--- a/.changeset/rich-insects-jog.md
+++ b/.changeset/rich-insects-jog.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+export Random.make taking hashable values as seed
--- a/.changeset/rich-insects-jog.md
+++ b/.changeset/rich-insects-jog.md
@@ -1,5 +1,5 @@
 ---
-"effect": patch
+"effect": minor
 ---
 
 export Random.make taking hashable values as seed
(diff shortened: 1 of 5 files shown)


.changeset/rich-insects-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
```
tim-smart, line 2:
```suggestion
"effect": minor
```

packages/effect/src/Random.ts
```diff
@@ -118,3 +118,12 @@ export const randomWith: <A, E, R>(f: (random: Random) => Effect.Effect<A, E, R>
  * @category context
  */
 export const Random: Context.Tag<Random, Random> = internal.randomTag
+
+/**
+ * Constructs the `Random` service, seeding the pseudo-random number generator
+ * with an hash of the specified seed.
+ *
+ * @since 3.4.9
```
tim-smart, line 126:
```suggestion
 * @since 3.5.0
```

tim-smart (changes requested):

tim-smart (approved):

PR #3213, opened 2024-07-10
switch to object options for Stream.async apis

tim-smart:
Up for discussion :)

### DIFF
--- a/.changeset/chilled-ducks-sniff.md
+++ b/.changeset/chilled-ducks-sniff.md
@@ -1,0 +1,24 @@
+---
+"effect": major
+---
+
+Use object options for Stream.async apis
+
+Instead of:
+
+```ts
+Stream.async((emit) => {
+  //...
+}, 16);
+```
+
+You can now write:
+
+```ts
+Stream.async(
+  (emit) => {
+    //...
+  },
+  { bufferSize: 16 },
+);
+```
--- a/.changeset/cool-birds-exercise.md
+++ b/.changeset/cool-birds-exercise.md
@@ -1,0 +1,5 @@
+---
+"effect": major
+---
+
+use "dropping" strategy by default for Stream.async apis
--- a/.changeset/cool-birds-exercise.md
+++ b/.changeset/cool-birds-exercise.md
@@ -1,5 +1,0 @@
----
-"effect": major
----
-
-use "dropping" strategy by default for Stream.async apis
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -312,7 +312,7 @@ export const as: {
 
 const _async: <A, E = never, R = never>(
   register: (emit: Emit.Emit<R, E, A, void>) => Effect.Effect<void, never, R> | void,
-  bufferSize?: number | "unbounded" | {
+  options?: { readonly bufferSize: "unbounded" } | {
     readonly bufferSize?: number | undefined
     readonly strategy?: "dropping" | "sliding" | "suspend" | undefined
   } | undefined
@@ -367,7 +367,7 @@ export {
  */
 export const asyncEffect: <A, E = never, R = never>(
   register: (emit: Emit.Emit<R, E, A, void>) => Effect.Effect<unknown, E, R>,
-  bufferSize?: number | "unbounded" | {
+  options?: { readonly bufferSize: "unbounded" } | {
     readonly bufferSize?: number | undefined
     readonly strategy?: "dropping" | "sliding" | "suspend" | undefined
   } | undefined
@@ -424,7 +424,7 @@ export const asyncPush: <A, E = never, R = never>(
  */
 export const asyncScoped: <A, E = never, R = never>(
   register: (emit: Emit.Emit<R, E, A, void>) => Effect.Effect<unknown, E, R | Scope.Scope>,
-  bufferSize?: number | "unbounded" | {
+  options?: { readonly bufferSize: "unbounded" } | {
     readonly bufferSize?: number | undefined
     readonly strategy?: "dropping" | "sliding" | "suspend" | undefined
   } | undefined
--- a/packages/effect/src/internal/stream.ts
+++ b/packages/effect/src/internal/stream.ts
@@ -474,12 +474,12 @@ const queueFromBufferOptions = <A, E>(
     return Queue.unbounded()
   }
   switch (options?.strategy) {
+    case "dropping":
+      return Queue.dropping(options.bufferSize ?? 16)
     case "sliding":
       return Queue.sliding(options.bufferSize ?? 16)
-    case "suspend":
-      return Queue.bounded(options.bufferSize ?? 16)
     default:
-      return Queue.dropping(options?.bufferSize ?? 16)
+      return Queue.bounded(options?.bufferSize ?? 16)
   }
 }
 
--- a/packages/effect/src/internal/stream.ts
+++ b/packages/effect/src/internal/stream.ts
@@ -463,23 +463,23 @@ export const as = dual<
 >(2, <A, E, R, B>(self: Stream.Stream<A, E, R>, value: B): Stream.Stream<B, E, R> => map(self, () => value))
 
 const queueFromBufferOptions = <A, E>(
-  bufferSize?: number | "unbounded" | {
+  options?: {
+    readonly bufferSize: "unbounded"
+  } | {
     readonly bufferSize?: number | undefined
     readonly strategy?: "dropping" | "sliding" | "suspend" | undefined
   } | undefined
 ): Effect.Effect<Queue.Queue<Take.Take<A, E>>> => {
-  if (bufferSize === "unbounded") {
+  if (options?.bufferSize === "unbounded") {
     return Queue.unbounded()
-  } else if (typeof bufferSize === "number" || bufferSize === undefined) {
-    return Queue.bounded(bufferSize ?? 16)
   }
-  switch (bufferSize.strategy) {
-    case "dropping":
-      return Queue.dropping(bufferSize.bufferSize ?? 16)
+  switch (options?.strategy) {
     case "sliding":
-      return Queue.sliding(bufferSize.bufferSize ?? 16)
+      return Queue.sliding(options.bufferSize ?? 16)
+    case "suspend":
+      return Queue.bounded(options.bufferSize ?? 16)
     default:
-      return Queue.bounded(bufferSize.bufferSize ?? 16)
+      return Queue.dropping(options?.bufferSize ?? 16)
   }
 }
 
@@ -488,13 +488,15 @@ export const _async = <A, E = never, R = never>(
   register: (
     emit: Emit.Emit<R, E, A, void>
   ) => Effect.Effect<void, never, R> | void,
-  bufferSize?: number | "unbounded" | {
+  options?: {
+    readonly bufferSize: "unbounded"
+  } | {
     readonly bufferSize?: number | undefined
     readonly strategy?: "dropping" | "sliding" | "suspend" | undefined
   } | undefined
 ): Stream.Stream<A, E, R> =>
   Effect.acquireRelease(
-    queueFromBufferOptions<A, E>(bufferSize),
+    queueFromBufferOptions<A, E>(options),
     (queue) => Queue.shutdown(queue)
   ).pipe(
     Effect.flatMap((output) =>
@@ -543,14 +545,16 @@ export const _async = <A, E = never, R = never>(
 /** @internal */
 export const asyncEffect = <A, E = never, R = never>(
   register: (emit: Emit.Emit<R, E, A, void>) => Effect.Effect<unknown, E, R>,
-  bufferSize?: number | "unbounded" | {
+  options?: {
+    readonly bufferSize: "unbounded"
+  } | {
     readonly bufferSize?: number | undefined
     readonly strategy?: "dropping" | "sliding" | "suspend" | undefined
   } | undefined
 ): Stream.Stream<A, E, R> =>
   pipe(
     Effect.acquireRelease(
-      queueFromBufferOptions<A, E>(bufferSize),
+      queueFromBufferOptions<A, E>(options),
       (queue) => Queue.shutdown(queue)
     ),
     Effect.flatMap((output) =>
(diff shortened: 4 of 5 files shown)


dilame:
As usual, i would vote for the composable API. 

```ts
Stream.async(
  (emit) => {
    //...
  },
  { buffer: Queue.unbounded() },
);
```

I believe that the main point of having a single-responsibility primitives is having the ability to compose them. 

In addition, this way we don't hide the internal details from the user – if i provided a `Queue` by my own – i understand that this Queue is using under the hood, which grants me extra knowledge about how my code works.

Plus if we will have a new options for `Queue` in the future – we will not need to re-implement all the methods that just passes props from one place to another. I can't think of any new options for Queue now, but just in case.

Don't see any reason to hide such implementation details from the user.

IMax153:
> Don't see any reason to hide such implementation details from the user.

We do not provide such an API anywhere else within Effect, so I disagree that this _should not_ be hidden from the user. 

Given the underlying complexity in stream, I think giving users the flexibility to choose the buffer **strategy** (i.e. `unbounded`, `dropping`, etc.) is a better UX than having to pass the buffer to use. Furthermore, for this case specifically the `Queue` constructors are effectful, and while that's not an issue since the effect can be resolved internally, it obfuscates things further.

From my perspective, I like the refined API proposed by @tim-smart. 

This also closes #3200.

mikearnaldi:
> Up for discussion :)

Dropping elements by default looks like a very bad choice. Bounded by default is kind of the only thing that make sense

IMax153:
> Dropping elements by default looks like a very bad choice. Bounded by default is kind of the only thing that make sense

I'm fine with leaving it bounded by default I guess but we should probably add a note to the JSDoc that explains the semantics of each option.

dilame:
> I'm fine with leaving it bounded by default I guess but we should probably add a note to the JSDoc that explains the semantics of each option.

Or just let user to pass `Queue` by himself, so no need to describe things, everything is described in the documentation 
https://effect.website/docs/guides/concurrency/queue

IMax153:
> Or just let user to pass `Queue` by himself

Again, this is not a pattern we use anywhere else in the library and I'm not sure it's a desirable one.

mikearnaldi:
Also note that passing `Queue.unbounded()` is not "passing a queue" but rather passing a queue constructor. Per se as an additional option it isn't necessarily bad

dilame:
Right now we are discussing the necessity of `share` operator, and seems like there will be the first place where this pattern will be used

https://discord.com/channels/795981131316985866/1258936229219532851

tim-smart:
> Or just let user to pass `Queue` by himself, so no need to describe things, everything is described in the documentation https://effect.website/docs/guides/concurrency/queue

The issue with passing a constructor is that it doesn't guarantee exclusive use by the stream, and it surfaces implementation detail. While unlikely, we may want to swap the usage of Queue with another buffer implementation in the future.

dilame:
> The issue with passing a constructor is that it doesn't guarantee exclusive use by the stream

Yes, indeed, a constructor does not guarantee exclusive use, but that isn't necessarily a bad thing. I wouldn't call this an "issue" – it's more about providing flexibility to the user. Actually it's not an easy thing to share a Queue between two streams – if user already did it – he definetely did it intentionally and for some reason.

> and it surfaces implementation detail

This is actually one of the main points – I don't just see this as not being bad, I assert that exposing this implementation detail is beneficial. As I mentioned earlier, one of the advantages of simple primitives with limited responsibilities is the ability to compose them.

Let's imagine a newcomer joins a project written in Effect, having never worked with Effect before. If they see:

```ts
Stream.async(
  (emit) => {
    //...
  },
  { buffer: Queue.bounded(16) },
);
```
It's immediately clear that queues are being used for buffering in this stream. They can look up the documentation on queues and get a complete picture of what's happening under the hood. They will understand that Stream.async calls Queue.offer, and so on. The code stops being magical or a black box – they see the composition of familiar primitives.

On the other hand, if the newcomer sees:

```ts
Stream.async(
  (emit) => {
    //...
  },
  { bufferSize: 16 },
);
```

It's unclear what's actually happening. What is this buffer? How does it work? They would still need to delve into the source code to find out that a queue is being used underneath.

So, in the end, every developer will end up "surfacing the implementation detail" for themselves by looking into the Effect source code, instead of understanding it through type declarations and their own source code.

dilame:
This actually highlights a broader issue known as "**composition vs configuration**." In React, for example, the principle of "**Composition over Configuration**" is widely embraced. 

> Again, this is not a pattern we use anywhere else in the library and I'm not sure it's a desirable one.

I would argue that Effect also adheres to this principle, as it is inherently applicable to functional programming – building complex programs from the composition of small functions.

Additionally, on the main page of Effect's website (https://effect.website/), there are sections emphasizing this philosophy:


> **Composable & Reusable**
> Everything in the Effect ecosystem is designed to work together. Building applications with Effect feels like playing with Lego.



> **Powerful building blocks**
> Every part of the Effect ecosystem is designed to be composable. The Effect primitives can be combined in many different ways to tackle the most complex problems.


Given this, the approach of passing a queue constructor instead of a configuration seems to be the most appropriate solution in this case. It aligns well with the overall approach of composition over configuration.

tim-smart:
I would argue that this is why apis such as `Stream.fromQueue` & `Stream.fromPubSub` exist.

I'm not completely against the idea of passing constructors, but maybe as an additional option. It should also support the error and requirements types, and fully offload the lifetime responsibility to the user by unwrapping a Scope (`Effect<Queue, E, R | Scope>`)

mikearnaldi:
@tim-smart given that dropping elements by default isn't something we should do, can we close this?

mikearnaldi:
> I would argue that this is why apis such as `Stream.fromQueue` & `Stream.fromPubSub` exist.
> 
> I'm not completely against the idea of passing constructors, but maybe as an additional option. It should also support the error and requirements type, and fully offload the lifetime responsibility to the user by unwrapping a Scope (`Effect<Queue, E, R | Scope>`).

unwrapping a Scope would not delegate the responsibility of the Scope to the user

tim-smart:
> @tim-smart given that dropping elements by default isn't something we should do, can we close this?

I'll rework it to just make the change to object arguments.

Regarding keeping `bounded`, would switching to `unbounded` be better? You are either accumulating unbounded suspended fibers, or unbounded elements either way, and suspended fibers would likely be more expensive in terms of memory.

dilame:
`unbounded` sounds like the only natural choice for `Stream.async*` family, which produces a push-based stream

mikearnaldi:
> > @tim-smart given that dropping elements by default isn't something we should do, can we close this?
> 
> I'll rework it to just make the change to object arguments.
> 
> Regarding keeping `bounded`, would switching to `unbounded` be better? You are either accumulating unbounded suspended fibers, or unbounded elements either way, and suspended fibers would likely be more expensive in terms of memory.

I think we broke something in the async constructor typing but the advised usage is:

```ts
Stream.async<number>(async (emit) => {
  while (Math.random() <= 0.99) {
    await emit.single(100)
  }
})
```

which would back-pressure the producer thanks to the bounded buffer, the same will leak with unbounded. I think we need an option to potentially use unbounded but unbounded by default is not ideal

mikearnaldi:
> `unbounded` sounds like the only natural choice for `Stream.async*` family, which produces a push-based stream

no, for the reason in: https://github.com/Effect-TS/effect/pull/3213#issuecomment-2230579231

dilame:
> I think we broke something in the async constructor typing but the advised usage is

It sounds good when the implementation is just emitting sequential `Math.random()`, but it looks more of a "Hello World". 

In reality, the `Stream.async` is used when you have some `EventEmitter` (like this, for example https://www.npmjs.com/package/ws) and you want to turn it into a stream.

```ts
Stream.async((emit) => {
  const ws = new WebSocket();
  ws.on('message', m=>emit(m))
})
```

dilame:
> We don't have any push-based Stream, async is not producing a push-based stream it is producing a plain and normal stream

Yes, I understand that Effect streams are pull-based, but it doesn't change the fact that there are data sources that are inherently push-based. 

I see `Stream.async` as a boundary between the external push-based world and Effect. I think placing a bounded buffer here by default doesn't make sense.

Because all other cases (such as sequencing random, etc) are covered by other operators and primitives in Effect for various scenarios, you don't need to do it "manually".

Do you have any real-life examples where a boundary for async actually worked and made sense?

mikearnaldi:
> > I think we broke something in the async constructor typing but the advised usage is
> 
> It sounds good when the implementation is just emitting sequential `Math.random()`, but it looks more of a "Hello World".
> 
> In reality, the `Stream.async` is used when you have some `EventEmitter` (like this, for example https://www.npmjs.com/package/ws) and you want to turn it into a stream.
> 
> ```ts
> Stream.async((emit) => {
>   const ws = new WebSocket();
>   ws.on('message', m=>emit(m))
> })
> ```

That's really not an hello world, you're confusing one usage with every usage, event emitters are problematic in JS no matter what and in fact node streams support backpressure by calling read explicitly instead of using emitters. Not everything is listening to an event emitter, if you're listening to an event emitter the best way is to use unbounded but that's not the general case

mikearnaldi:
> > We don't have any push-based Stream, async is not producing a push-based stream it is producing a plain and normal stream
> 
> Yes, I understand that Effect streams are pull-based, but it doesn't change the fact that there are data sources that are inherently push-based.
> 
> I see `Stream.async` as a boundary between the external push-based world and Effect. I think placing a bounded buffer here by default doesn't make sense.
> 
> Because all other cases (such as sequencing random, etc) are covered by other operators and primitives in Effect for various scenarios, you don't need to do it "manually".
> 
> Do you have any real-life examples where a boundary for async actually worked and made sense?

"but it doesn't change the fact that there are data sources that are inherently push-based" that's untrue, there are bad implementations that don't expose a pull-based approach but at the lower level (incl kernel) nothing is push based

"I see Stream.async as a boundary between the external push-based world and Effect." that's your view, it's not what the async combinator is for, the async combinator gives you a way to emit and emit returns a Promise

"Do you have any real-life examples where a boundary for async actually worked and made sense?" calling APIs that may return cursors, integrating to async sources, reading from streams where the read is returning a Promise

mikearnaldi:
if anything it might actually be better to expose a `fromEventEmitter` which is purpose built for handling emitters, with the quirks of emitters

tim-smart:
> if you're listening to an event emitter the best way is to use unbounded

https://github.com/Effect-TS/effect/pull/3266

dilame:
> that's untrue, there are bad implementations that don't expose a pull-based approach but at the lower level (incl kernel) nothing is push based

Regarding push-based data sources, network sockets are a fundamental example. At the lowest level, a network socket receives data packets from the network interface card (NIC). When a packet arrives, it triggers an interrupt in the NIC, which notifies the operating system that new data is available. The OS then places this data into a buffer and signals the application that data is ready to be read. This signaling mechanism is inherently push-based because the data arrival drives the processing flow. The application doesn't continuously poll for data; instead, it reacts to incoming data as it arrives. This model is efficient and necessary for handling asynchronous events in network communication, where the timing of data arrival is unpredictable.

Keyboard events are another clear example of a push-based system. When a user presses a key, the operating system generates an event and sends it to the application immediately. The application doesn't need to poll for keypresses; it simply responds to the events as they occur.

Both push-based and pull-based models are essential in different contexts. While pull-based systems are useful for scenarios where the application needs to control the rate of data processing, push-based systems are ideal for event-driven architectures where immediate reaction to incoming data is required.

dilame:
> "Do you have any real-life examples where a boundary for async actually worked and made sense?" calling APIs that may return cursors, integrating to async sources, reading from streams where the read is returning a Promise

If you need to call APIs that may return cursors, we can use `Stream.unfold` or `Stream.paginate` in userland, correct? Why would `Stream.async` be preferable in this case?

When you mention integrating with async sources, could you clarify what you mean? For example, would an `EventEmitter` be considered an async source?

In the context of reading from streams where the read operation returns a Promise, how would a bounded buffer be beneficial?

mikearnaldi:
> > that's untrue, there are bad implementations that don't expose a pull-based approach but at the lower level (incl kernel) nothing is push based
> 
> Regarding push-based data sources, network sockets are a fundamental example. At the lowest level, a network socket receives data packets from the network interface card (NIC). When a packet arrives, it triggers an interrupt in the NIC, which notifies the operating system that new data is available. The OS then places this data into a buffer and signals the application that data is ready to be read. This signaling mechanism is inherently push-based because the data arrival drives the processing flow. The application doesn't continuously poll for data; instead, it reacts to incoming data as it arrives. This model is efficient and necessary for handling asynchronous events in network communication, where the timing of data arrival is unpredictable.
> 
> Keyboard events are another clear example of a push-based system. When a user presses a key, the operating system generates an event and sends it to the application immediately. The application doesn't need to poll for keypresses; it simply responds to the events as they occur.
> 
> Both push-based and pull-based models are essential in different contexts. While pull-based systems are useful for scenarios where the application needs to control the rate of data processing, push-based systems are ideal for event-driven architectures where immediate reaction to incoming data is required.

push-based streams are not modelled by Effect Stream and working around it leads to issues, if you want such system you can use an unbounded buffer, the use case is supported making it the default isn't ideal and we will not do it. Note that even if you use an unbounded buffer, the Effect Stream is still pull based, if there is no consumer pulling nothing happens so while your assumption may be correct (arguable if it is, see for example why Kafka is pull-based and heavily used in event-driven systems) having unbounded/bounded/whatever won't get you there.

> If you need to call APIs that may return cursors, we can use Stream.unfold or Stream.paginate in userland, correct? Why would Stream.async be preferable in this case?

depends on the API, if you have a promise based API to interact with and you need custom control flow then paginate & co might end up being restrictive, an alternative would be to hook into the lower level API for channels and use an Effect generator.

> When you mention integrating with async sources, could you clarify what you mean? For example, would an EventEmitter be considered an async source?

no an EventEmitter is not something I would consider an async source, a node stream with an async read would be one

> In the context of reading from streams where the read operation returns a Promise, how would a bounded buffer be beneficial?

because it allows you to wait before reading the next element given that the push will be suspended until there is a taker

tim-smart:
> At the lowest level, a network socket receives data packets from the network interface card (NIC). When a packet arrives, it triggers an interrupt in the NIC, which notifies the operating system that new data is available. The OS then places this data into a buffer and signals the application that data is ready to be read.

You still have to explicitly read the data to access it though, it isn't pushed to the program. You are just notified when there is something ready.

And even technically keyboard events are buffered and processed on-demand, but you don't get access to that buffer at the dom level.

Push based streams still need to decide what to do if you can't handle the events fast enough (i.e. using a rxjs .concatMap into a slow resource will cause memory issues).

I think we can introduce a `Stream.asyncPush` variant that handles integrating with apis without backpressure more seamlessly.

tim-smart:
> I think we can introduce a `Stream.asyncPush` variant that handles integrating with apis without backpressure more seamlessly.

Exploring .asyncPush here: https://github.com/Effect-TS/effect/pull/3277/files

PR #3225, opened 2024-07-11
Render a more helpful error message in `Effect.timeout`

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/famous-windows-poke.md
+++ b/.changeset/famous-windows-poke.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Render a more helpful error message when timing out an effect
--- a/.changeset/kind-peaches-tap.md
+++ b/.changeset/kind-peaches-tap.md
@@ -1,5 +1,0 @@
----
-"effect": patch
----
-
-fix Logger.pretty on bun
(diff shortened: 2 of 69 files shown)


IMax153:
I suppose we could make this a patch release since I made the `duration` optional?

packages/effect/src/Cause.ts
```diff
@@ -910,7 +911,10 @@ export const isRuntimeException: (u: unknown) => u is RuntimeException = core.is
  * @since 2.0.0
  * @category errors
  */
-export const TimeoutException: new(message?: string | undefined) => TimeoutException = core.TimeoutException
+export const TimeoutException: new(
+  duration?: Duration.Duration,
```
gcanti, line 915: this is a breaking change, maybe

```ts
new(message?: string, duration?: Duration.Duration)
```

or

```ts
new(
  options?: string | {
    readonly message?: string
    readonly duration?: Duration.DurationInput
  }
)
```

IMax153: Good call @gcanti - I'll opt for the second strategy for now unless someone else has strong opposition.

IMax153:
Going to close and re-open a new PR given this is no longer a minor version bump.

PR #3228, opened 2024-07-11
Render a more helpful error message in Effect.timeout 

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/test/Effect/timeout.test.ts
+++ b/packages/effect/test/Effect/timeout.test.ts
@@ -10,6 +10,21 @@ import * as TestClock from "effect/TestClock"
 import { assert, describe } from "vitest"
 
 describe("Effect", () => {
+  it.effect("timeout produces a useful error message", () =>
+    Effect.gen(function*() {
+      const duration = Duration.millis(1500)
+      const fiber = yield* Effect.never.pipe(
+        Effect.timeout(duration),
+        Effect.flip,
+        Effect.fork
+      )
+      yield* TestClock.adjust(Duration.millis(2000))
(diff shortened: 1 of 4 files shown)


gcanti (approved):

packages/effect/test/Effect/timeout.test.ts
```diff
@@ -10,6 +10,21 @@ import * as TestClock from "effect/TestClock"
...
+        Effect.flip,
+        Effect.fork
+      )
+      yield* TestClock.adjust(Duration.millis(2000))
+      const result = yield* Fiber.join(fiber)
+      assert.include(
+        result.toString(),
+        "TimeoutException: Operation timed out before the specified duration of '1s 500ms' elapsed"
+      )
+    }))
```
riwsky, line 27: I'm confused: isn't this operation timing out _because_ the specified duration elapsed? This message sounds like how one would describe e.g. a `Effect.timeout(Duration.millis(1000))` blowing up "within" an `Effect.timeout(Duration.millis(1500))`, if one wanted to include some error wrapping to make it clear "hey, it's not this outer timeout which caused the error".

riwsky: that would make more sense to me, yes (assuming I'm properly understanding the underlying semantics)

IMax153: https://github.com/Effect-TS/effect/pull/4591

PR #3190, opened 2024-07-06
fix(Stream|Channel): exclude `Scope` from the resulting effect env of all the non-scoped `.run*` methods

dilame:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Fixed the implementation of non-scoped `run*` functions across the `Channel` and `Stream`. 
This fix ensures that `Scope` is now properly excluded from the resulting effect environment. 
The affected functions include `run`, `runCollect`, `runCount`, `runDrain` and other non-scoped `run*` in both `Stream` and `Channel` modules. 
This fix brings the type declaration in line with the runtime implementation.

## Related

- Related Issue #
- Closes #
- https://github.com/Effect-TS/effect/issues/3187

PR #3238, opened 2024-07-12
Align Stream.run public function signatures

leonitousconforti:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I remembered reading something in the Discord about how Scope was removed from the resulting effect of all the non-scoped .run* methods, and yet Stream.run didn't seem to be behaving like this. I reviewed [this commit](https://github.com/Effect-TS/effect/commit/ac71f378f2413e5aa91c95f649ffe898d6a26114) from pr #3190, and it looks to me like one of the function signatures of `Stream.run` was not updated (I couldn't find any others that were missed). The internal signatures are correct as well.

## Related

#3190

### DIFF
--- a/.changeset/twenty-toes-brake.md
+++ b/.changeset/twenty-toes-brake.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Align Stream.run public function signatures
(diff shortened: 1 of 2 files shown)


tim-smart:
👍🏼

PR #3247, opened 2024-07-13
if performance.timeOrigin is 0, use performance.now() directly in Clock

tim-smart:


johtso:
After some investigation it seems like this is an inconsistency between miniflare and the live cloudflare runtime. Really `Date.now()` and `performance.now()` should _both_ return 0.. but due to a quirk in miniflare `Date.now()` returns the current time even pre-io, but `performance.now()` returns `0`.

https://github.com/cloudflare/workers-sdk/issues/6261

PR #3267, opened 2024-07-16
Support `Refinement` in `Predicate.tuple` and `Predicate.struct`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Support `Refinement` in `Preficate.tuple` and `Predicate.struct`
Add namespace for type-helpers with inference

## Related

- Related Issue #
- Closes #


.changeset/purple-onions-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+support `Refinement` in `Preficate.tuple` and `Predicate.struct`
```
fubhy, line 5:
```suggestion
support `Refinement` in `Predicate.tuple` and `Predicate.struct`
```

fubhy (changes requested):

.changeset/purple-onions-drive.md
KhraksMamtsov: Done, also I've marked this change as minor

packages/effect/src/Predicate.ts
```diff
@@ -29,6 +29,19 @@ export interface Refinement<in A, out B extends A> {
   (a: A): a is B
 }
 
+export declare namespace Predicate {
+  /**
+   * @since 2.0.0
```
gcanti, line 34: should be `3.6.0` I guess

packages/effect/src/Predicate.ts
```diff
@@ -29,6 +29,19 @@ export interface Refinement<in A, out B extends A> {
   (a: A): a is B
 }
 
+export declare namespace Predicate {
```
gcanti, line 32: this namespace is not actually used

gcanti (changes requested):
I would suggest splitting the namespaces and then using them in `tuple` and `struct`

```ts
/**
 * @since 3.6.0
 * @category type-level
 */
export declare namespace Predicate {
  /**
   * @since 3.6.0
   * @category type-level
   */
  export type In<T extends Predicate<any>> = [T] extends [Predicate<infer _A>] ? _A : never
}

/**
 * @since 3.6.0
 * @category type-level
 */
export declare namespace Refinement {
  /**
   * @since 3.6.0
   * @category type-level
   */
  export type In<T extends Refinement<any, any>> = [T] extends [Refinement<infer _A, infer _>] ? _A : never
  /**
   * @since 3.6.0
   * @category type-level
   */
  export type Out<T extends Refinement<any, any>> = [T] extends [Refinement<infer _, infer _B>] ? _B : never
}

/**
 * Similar to `Promise.all` but operates on `Predicate`s.
 *
 * ```
 * [Refinement<A, B>, Refinement<C, D>, ...] -> Refinement<[A, C, ...], [B, D, ...]>
 * [Predicate<A>, Predicate<B>, ...] -> Predicate<[A, B, ...]>
 * ```
 *
 * @since 2.0.0
 */
export const tuple: {
  <T extends ReadonlyArray<Refinement<any, any>>>(
    ...elements: T
  ): Refinement<
    Readonly<{ [I in keyof T]: Refinement.In<T[I]> }>,
    Readonly<{ [I in keyof T]: Refinement.Out<T[I]> }>
  >
  <T extends ReadonlyArray<Predicate<any>>>(
    ...elements: T
  ): Predicate<Readonly<{ [I in keyof T]: Predicate.In<T[I]> }>>
} = (...elements: ReadonlyArray<any>) => all(elements) as any

/**
 * ```
 * { ab: Refinement<A, B>; cd: Refinement<C, D>, ... } -> Refinement<{ ab: A; cd: C; ... }, { ab: B; cd: D; ... }>
 * { a: Predicate<A, B>; b: Predicate<B>, ... } -> Predicate<{ a: A; b: B; ... }>
 * ```
 *
 * @since 2.0.0
 */
export const struct: {
  <R extends Record<string, Refinement<any, any>>>(
    fields: R
  ): Refinement<
    { readonly [K in keyof R]: Refinement.In<R[K]> },
    { readonly [K in keyof R]: Refinement.Out<R[K]> }
  >
  <R extends Record<string, Predicate<any>>>(
    fields: R
  ): Predicate<{ readonly [K in keyof R]: Predicate.In<R[K]> }>
} = (<R extends Record<string, Predicate<any>>>(fields: R) => {
  const keys = Object.keys(fields)
  return (a: Record<string, unknown>) => {
    for (const key of keys) {
      if (!fields[key](a[key])) {
        return false
      }
    }
    return true
  }
}) as any
```

gcanti:
The other CI failure, however, does not pertain to this PR:

```
Error: packages/sql-kysely/src/internal/patch.ts(70,38): error TS2353: Object literal may only specify known properties, and 'error' does not exist in type '{ cause: unknown; message?: string; }'.
```

it seems that the `next-minor` branch is currently broken

packages/effect/src/Predicate.ts
KhraksMamtsov: I have applied your changes

KhraksMamtsov:
What should I do to get rid of these errors (`sql-kysely`)?

tim-smart:
I think we could support partial refinements too:

```ts
Predicate.struct({

KhraksMamtsov:
open new PR - https://github.com/Effect-TS/effect/pull/3366

PR #3269, opened 2024-07-16
Use `pkg.pr.new` for snapshots

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Swaps our snapshot strategy from publishing to the NPM registry to using [pkg.pr.new](https://github.com/stackblitz-labs/pkg.pr.new) from the folks at Stackblitz.

This has numerous benefits, not the least of which is avoiding pollution of the NPM registry with extraneous snapshot packages.

Unfortunately, it's not possible to test for the `effect` repo unless it's merged into main, but I did test it on the `dadinator` repository and it [seemed to work quite well](https://github.com/IMax153/dadinator/pull/6#issuecomment-2231253315). The only annoyance is that we still need the custom steps in the action for comments because `pkg.pr.new` doesn't really handle workflows triggered via `issue_comment` well (i.e. it doesn't write comments to PRs triggered via that modality).

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -61,18 +61,8 @@ jobs:
         if: ${{ steps.branch.outputs.branch == 'changeset-release/main' }}
         run: git checkout origin/main
 
-      - name: Install pnpm
-        uses: pnpm/action-setup@v3
-
-      - name: Install node
-        uses: actions/setup-node@v4
-        with:
-          cache: pnpm
-          node-version: 20.12.2
-
       - name: Install dependencies
-        shell: bash
-        run: pnpm install
+        uses: ./.github/actions/setup
 
       - name: Build package
         run: pnpm build
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -70,7 +70,7 @@ jobs:
       - name: Create snapshot
         id: snapshot
         run: |
-          snapshots=$(pnpm pkg-pr-new publish --compact --pnpm --comment=off ./packages/*)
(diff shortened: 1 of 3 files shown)


.github/workflows/snapshot.yml
```diff
@@ -63,41 +64,43 @@ jobs:
...
+      - name: Create snapshot
         id: snapshot
         run: |
-          # Publish and extract published tags from stdout.
-          output=$(pnpm changeset publish --tag ${{ steps.command.outputs.snapshot }} --no-git-tag)
-          output=$(echo "$output" | awk '/packages published successfully:/{flag=1; next} flag')
-          output=$(echo "$output" | grep -o '[^ ]*@[^ ]*' | awk '{print "\"" $0 "\""}' | paste -sd ',')
-          echo "tags=[$output]" >> $GITHUB_OUTPUT
+          snapshots=$(pnpm pkg-pr-new publish --compact --pnpm --comment=off ./packages/*)
+          echo "${snapshots}"
```
fubhy, line 74: For debugging?

package.json
```diff
@@ -65,6 +65,7 @@
     "fast-check": "^3.20.0",
     "glob": "^11.0.0",
     "madge": "^7.0.0",
+    "pkg-pr-new": "^0.0.17",
```
fubhy, line 68: Instead of installing this as a dependency, let'\s use `npx` with a version lock (`npx pkg-pr-new@0.0.17`) as this is purely a CI thing.

fubhy (changes requested):

PR #3276, opened 2024-07-16
Next minor

ecyrbe:
Thanks to ray, we detect that i forgot to whitelist "compile" method from kysely effectified instance.

This fixes the issue and add test for it.


tim-smart:
Not sure how this got merged, it is currently failing CI due to some linting issues.

PR #3277, opened 2024-07-16
add Stream.asyncPush api

tim-smart:


mikearnaldi:
Same as in the other PR, dropping elements by default is not a good choice

mikearnaldi:
Options could be `{ strategy: "unbounded" }` by default and potentially accept `{ strategy: "dropping" / "sliding", size: 100 }`

The nice aspect about this is that assuming we don't support backpressure we can guarantee sync push, which would be faster than the plain async way

tim-smart:
> Same as in the other PR, dropping elements by default is not a good choice

Have changed it to "unbounded" by default.

PR #3282, opened 2024-07-17
Implement Struct.keys as a typed alternative to Object.keys

mikearnaldi:
Allows for use cases such as:

```ts
import { Struct } from "effect"

export const Version = S.String.pipe(
  S.pattern(/^\d+\.\d+\.\d+$/) // Ensure "<major>.<minor>.<patch>" and no leading or trailing
)

export const Email = S.String.pipe(
  S.pattern(
    /^(?!\.)(?!.*\.\.)([A-Z0-9_+-.]*)[A-Z0-9_+-]@([A-Z0-9][A-Z0-9-]*\.)+[A-Z]{2,}$/i
  )
)
export const Entity = S.Struct({
  id: S.UUID,
  version: Version
})
export class Creator extends S.Class<Creator>("creator")({
  ...Entity.fields,
  createdAt: S.Date,
  communicationChannels: S.Array(S.String),
  email: Email,
  platforms: S.Array(S.String)
}) {}

////// Usecase
export const UseCaseCreateCreator = S.Struct(
  Struct.omit(Creator.fields, ...Struct.keys(Entity.fields))
)
```

### DIFF
--- a/.changeset/lovely-buckets-sip.md
+++ b/.changeset/lovely-buckets-sip.md
@@ -1,0 +1,19 @@
+---
+"effect": minor
+---
+
+Implement Struct.keys as a typed alternative to Object.keys
+
+```ts
+import { Struct } from "effect"
+
+const symbol: unique symbol = Symbol()
+
+const value = {
+  a: 1,
+  b: 2,
+  [symbol]: 3
+}
+
+const keys: Array<"a" | "b"> = Struct.keys(value)
+```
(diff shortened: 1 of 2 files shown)


tim-smart:
We also have `Record.keys` which is similar, but incorrectly removes symbols from the type.

tim-smart (approved):
I guess it doesn't hurt to have it in both modules.

PR #3289, opened 2024-07-17
Changed `Stream.groupByKey` JSDoc category from `utils` to `grouping`

dilame:


gcanti:
What about `grouped` and `groupedWithin`?

dilame:
Do you think it is also `grouping`? I was based on assumption that `grouping` is the category that works with `GroupBy` module, and `grouped*` family doesn't do it

gcanti:
not sure, groupAdjacentBy is also tagged as a grouping function

dilame:
FMPOV it is more of a `buffering` based on time than grouping, but changed it also

PR #3292, opened 2024-07-17
feat: add onlyEffect option to Effect.tap

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Effect.tap` is permissive as it allows passing pretty much anything. While it's practical, it's also a bit less safe than forcing to pass an Effect or a function returning an Effect.
This P.R. add the option `onlyEffect` to be set to `true` to force the use of an Effect for more safety

## Related

- Related Issue #
- Closes #


vinassefranche:
While I only added 4 commits, I suppose that the minor in the changeset has changed the target branch of the P.R. adding other commits to it.
Should I change the target branch? Re-create a P.R. with commits starting from the branch `next-minor`? Something else?

tim-smart:
I have rebased on next-minor for you.

Have you encountered some situations where this change was needed? Could you share some examples?
Just trying to understand the issue better.

vinassefranche:
Thanks @tim-smart 
There was still a commit that was not part of my P.R. I've removed it so we're good now.

The issue was faced a few times was in the process of migrating the code from fp-ts. There were some readertaskeither that were returned by the function in the `tap` that were then never executed.

Apart from this specific case, I can easily see this happening, specially during refactorings:
```ts
const program = pipe(
  Effect.void,
  Effect.tap(() => fireSideEffect) // oups. I forgot to call the function. Nothing happens
  Effect.flatMap(...)
)

const fireSideEffect = () => Effect.void
```

vinassefranche:
Some other commits have been added to my P.R. I don't know why...

tim-smart:
> Some other commits have been added to my P.R. I don't know why...

Don't worry too much. The minor release branch is constantly rebased against main which can make PRs go out of sync.

vinassefranche:
@tim-smart or @mikearnaldi any thoughts on this? Should I close the P.R. or there's a chance that it's accepted?

mikearnaldi:
> @tim-smart or @mikearnaldi any thoughts on this? Should I close the P.R. or there's a chance that it's accepted?

didn't have the time to check just yet, I am positive it will be merged

mikearnaldi:
@vinassefranche there is a test failure in `test/Effect/sequencing.test.ts:42:36`

fubhy:
Are we settled on the `onlyEffect` option? I'd personally be in favour of a `tapEffect` variant instead. Purely for esthetic reasons. We don't do anything like this elsewhere, do we?

mikearnaldi:
> Are we settled on the `onlyEffect` option? I'd personally be in favour of a `tapEffect` variant instead. Purely for esthetic reasons. We don't do anything like this elsewhere, do we?

given that we are progressively widening the signatures, like `andThen`, adding a variant would lead explosion of functions. I have expressed being 100% against adding variants in the discussion that lead to this PR

vinassefranche:
> @vinassefranche there is a test failure in `test/Effect/sequencing.test.ts:42:36`

Thanks @mikearnaldi for the fix! I completely missed that...
Shouldn't I modify also the types of the internal tap? The `onlyEffect` word in the implementation feels random in this context.

tim-smart:
> Shouldn't I modify also the types of the internal tap

Yes that would be good thanks. Often the public types are "toggled" from the internal types using the language service plugin.

vinassefranche:
> > Shouldn't I modify also the types of the internal tap
> 
> Yes that would be good thanks. Often the public types are "toggled" from the internal types using the language service plugin.

@tim-smart I did the change. I hope I did it right!

PR #3296, opened 2024-07-18
add DateTime module

tim-smart:


packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,333 @@
...
+export const matchUtc: {
+  <A>(f: (_: DateTime.Utc) => A): (self: DateTime) => A
+  <A>(self: DateTime, f: (_: DateTime.Utc) => A): A
+} = dual(2, <A>(self: DateTime, f: (_: DateTime.Utc) => A) => f(self._tag === "WithZone" ? self.utc : self))
+
+/**
+ * @since 3.6.0
+ * @category time zones
+ */
+export const withZone: {
```
tim-smart, line 301:
```suggestion
export const withCurrentZone: {
```

packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,333 @@
...
+  (zone: TimeZone): <A, E, R>(effect: Effect.Effect<A, E, R>) => Effect.Effect<A, E, R>
```
tim-smart, line 302:
```suggestion
  (zone: TimeZone): <A, E, R>(effect: Effect.Effect<A, E, R>) => Effect.Effect<A, E, Exclude<R, CurrentTimeZone>>
```

packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,333 @@
...
+/**
+ * @since 3.6.0
+ * @category constructors
+ */
+export const unsafeFromDate = (date: Date): DateTime => {
+  if (isNaN(date.getTime())) {
+    throw new IllegalArgumentException("Invalid date")
+  }
+  const self = Object.create(ProtoUtc)
+  self.date = date
```
tim-smart, line 160: Clone date

johtso:
Exciting! Would you consider a type for representing a Date with no time component?

tim-smart:
> Exciting! Would you consider a type for representing a Date with no time component?

Maybe as a separate module in a new PR.

packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,1590 @@
...
+ * If the `Date` is invalid, an `IllegalArgumentException` will be thrown.
+ *
+ * @since 3.6.0
+ * @category constructors
+ */
+export const unsafeFromDate = (date: Date): DateTime.Utc => {
+  const epochMillis = date.getTime()
+  if (Number.isNaN(epochMillis)) {
+    throw new IllegalArgumentException("Invalid date")
+  }
```
fubhy, line 413: What about infinity & negative infinity?

packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,1590 @@
...
+  (clock) => unsafeFromEpochMillis(clock.unsafeCurrentTimeMillis())
+)
+
+/**
+ * Get the current time using `Date.now`.
+ *
+ * @since 3.6.0
+ * @category constructors
+ */
+export const unsafeNow: LazyArg<DateTime.Utc> = () => unsafeFromEpochMillis(Date.now())
```
fubhy, line 504: Why is this unsafe? Shouldn't `Date.now` always yield a numerical output that should be convertable safely?

packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,1590 @@
+/**
```
fubhy, line 1: The docblocks in this module could use some `@example` annotations.

packages/schema/src/Schema.ts
```diff
@@ -5846,6 +5847,79 @@ export class DateFromNumber extends transform(
...
+  Number$,
+  DateTimeUtcFromSelf,
+  {
+    strict: true,
+    decode: (n, _, ast) => {
+      try {
+        return ParseResult.succeed(dateTime.unsafeFromEpochMillis(n))
+      } catch (_) {
+        return ParseResult.fail(new ParseResult.Type(ast, n))
+      }
```
fubhy, line 5887:
```suggestion
      try {
        return ParseResult.succeed(dateTime.unsafeFromEpochMillis(n))
      } catch {
        return ParseResult.fail(new ParseResult.Type(ast, n))
      }
```

packages/schema/src/Schema.ts
```diff
@@ -5846,6 +5847,79 @@ export class DateFromNumber extends transform(
...
+  String$,
+  DateTimeUtcFromSelf,
+  {
+    strict: true,
+    decode: (s, _, ast) => {
+      try {
+        return ParseResult.succeed(dateTime.unsafeFromString(s))
+      } catch (_) {
+        return ParseResult.fail(new ParseResult.Type(ast, s))
+      }
```
fubhy, line 5913:
```suggestion
      try {
        return ParseResult.succeed(dateTime.unsafeFromString(s))
      } catch {
        return ParseResult.fail(new ParseResult.Type(ast, s))
      }
```

fubhy (changes requested):
Beautiful api design. Nice.

fubhy:
One nitpick: I don't like the `WithZone` naming. The Temporal API uses `ZonedDateTime`. I think that has a better ring to it. Wdyt?

packages/effect/src/DateTime.ts
tim-smart: Because `Date.now()` is a side effect that should be an `Effect`. And not using `Clock` means testing is harder.

packages/effect/src/DateTime.ts
tim-smart:
```ts
new Date(Infinity).getTime() // NaN
```

tim-smart:
> One nitpick: I don't like the `WithZone` naming. The Temporal API uses `ZonedDateTime`. I think that has a better ring to it. Wdyt?

Trying out `DateTime.Zoned`, will see if it sticks :)

> I'm wondering if we should align even more with the Temporal API suggestion and also use `PlainDateTime` for those without a timezone. But I'm not sure if I like that namechoice. But on the other hand it might make sense to preemptively align with that API.

I'll stick with `Utc`, as technically this isn't a plain date time as it wraps a epoch timestamp.

fubhy:
> Trying out `DateTime.Zoned`, will see if it sticks :)

Sounds much better imho. Especially in the `onUtc` vs. `onZoned` matcher. :-)

packages/effect/src/DateTime.ts
jessekelly881: Similarly, why use unsafeCurrentTimeMillis with `now` vs something like this?

```ts
export const now: Effect.Effect<Utc> = Effect.map(
  Effect.flatMap(Effect.clock, (clock) => clock.currentTimeMillis),
  (ms) => makeUtc(ms)
)
```

tim-smart: We are already in an effect context here so it is fine.

fubhy (approved):
Hot.

packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,1839 @@
...
+ * Create a Layer from the given IANA time zone identifier.
+ *
+ * @since 3.6.0
+ * @category current time zone
+ */
+export const layerCurrentZoneNamed = (zoneId: string): Layer.Layer<CurrentTimeZone, IllegalArgumentException> =>
+  Layer.effect(CurrentTimeZone, zoneMakeNamedEffect(zoneId))
+
+/**
+ * Create a Layer from the given IANA time zone identifier.
```
tim-smart, line 1273:
```suggestion
 * Create a Layer from the systems local time zone.
```

packages/schema/src/Schema.ts
```diff
@@ -5846,6 +5847,248 @@ export class DateFromNumber extends transform(
...
+  (u) => dateTime.isDateTime(u) && dateTime.isUtc(u),
+  {
+    identifier: "DateTimeUtcFromSelf",
+    description: "a DateTime.Utc instance",
+    pretty: (): pretty_.Pretty<dateTime.Utc> => (dateTime) => dateTime.toString(),
+    arbitrary: (): LazyArbitrary<dateTime.Utc> => (fc) => fc.date().map((date) => dateTime.unsafeFromDate(date)),
+    equivalence: () => dateTime.Equivalence
+  }
+) {
+  static override annotations: (annotations: Annotations.Schema<dateTime.Utc>) => typeof DateTimeUtcFromSelf = super
```
gcanti, line 5866: I removed all `static override annotations` in `main` because they interfere with tree shaking

tim-smart: Did you use an alternative?

gcanti: No because they were also incorrect (it's not true that it returns a `typeof DateTimeUtcFromSelf`)

packages/schema/src/Schema.ts
```diff
@@ -5746,6 +5747,209 @@ export class DateFromNumber extends transform(
...
+ */
+export class DateTimeUtc extends transformOrFail(
+  String$,
+  DateTimeUtcFromSelf,
+  {
+    strict: true,
+    decode: decodeDateTime,
+    encode: (dt) => ParseResult.succeed(dateTime.formatIso(dt))
+  }
+).annotations({ identifier: "DateTime.Utc" }) {}
```
gcanti, line 5803: As an `identifier`, it's better, if possible, to use a value that is a valid TypeScript identifier (in this case "DateTimeUtc" rather than "DateTime.Utc"). This is because, for example, it might be used by a compiler to produce a TypeScript type or to generate code.

riordanpawley:
Love your work!

I don't think we should wait on this issue but just want to note for people who may be looking for an answer in the future that this may not work perfectly on react-native/hermes https://github.com/facebook/hermes/issues/1172

tim-smart:
> Love your work!
> 
> I don't think we should wait on this issue but just want to note for people who may be looking for an answer in the future that this may not work perfectly on react-native/hermes [facebook/hermes#1172](https://github.com/facebook/hermes/issues/1172)

I'll have a look and see what the support is like. This is only using numeric formatting from the Intl APIs to work with time zones.

tim-smart:
> Love your work!
> 
> I don't think we should wait on this issue but just want to note for people who may be looking for an answer in the future that this may not work perfectly on react-native/hermes https://github.com/facebook/hermes/issues/1172

Just realised I don't have a physical ios device to test with. Could you see the output of the following?

```ts
console.log(new Intl.DateFormat("en-US", {
  day: "numeric",
  month: "numeric",
  year: "numeric",
  hour: "numeric",
  minute: "numeric",
  second: "numeric",
  timeZoneName: "longOffset",
  fractionalSecondDigits: 3,
  hourCycle: "h23",
  timeZone: "UTC"
}).formatToParts(new Date())
```

riordanpawley:
> > Love your work!
> > I don't think we should wait on this issue but just want to note for people who may be looking for an answer in the future that this may not work perfectly on react-native/hermes [facebook/hermes#1172](https://github.com/facebook/hermes/issues/1172)
> 
> Just realised I don't have a physical ios device to test with. Could you see the output of the following?
> 
> ```ts
> console.log(new Intl.DateFormat("en-US", {
>   day: "numeric",
>   month: "numeric",
>   year: "numeric",
>   hour: "numeric",
>   minute: "numeric",
>   second: "numeric",
>   timeZoneName: "longOffset",
>   fractionalSecondDigits: 3,
>   hourCycle: "h23",
>   timeZone: "UTC"
> }).formatToParts(new Date()))
> ```

just spun up an expo app from create-expo-app. 
ran it via expo go on a physical iphone 12

    "react-native": "0.74.3",
    "expo": "~51.0.22",
DateFormat was missing, I got:
"Cannot read property 'prototype' of undefined"

Assuming you meant DateTimeFormat:
```json
[{"type": "month", "value": "7"}, {"type": "literal", "value": "/"}, {"type": "day", "value": "25"}, {"type": "literal", "value": "/"}, {"type": "year", "value": "2024"}, {"type": "literal", "value": ","}, {"type": "literal", "value": " "}, {"type": "hour", "value": "01"}, {"type": "literal", "value": ":"}, {"type": "minute", "value": "19"}, {"type": "literal", "value": ":"}, {"type": "second", "value": "41"}, {"type": "literal", "value": "."}, {"type": "fractionalSecond", "value": "724"}, {"type": "literal", "value": " "}, {"type": "timeZoneName", "value": "GMT"}]
```

tim-smart:
> Assuming you meant DateTimeFormat:

Yes thanks. The output is slightly different which I have added a fix for.

packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,1991 @@
...
+  }
+  return result
+}
+
+/**
+ * Modify a `DateTime` by applying a function to the underlying `Date`.
+ *
+ * The `Date` will first have the time zone applied if possible, and then be
+ * converted back to a `DateTime` within the same time zone.
+ *
```
jdharrisnz, line 1454: Does `mutate` and `mutateUtc` need a comment here explaining the meaning/intention of the function name? i.e., that it is safe and expected to use `Date`'s mutative methods because it's been pre-cloned before being passed to the function

packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,1991 @@
...
+ * @category math
+ * @example
+ * import { DateTime } from "effect"
+ *
+ * // add 5 minutes
+ * DateTime.unsafeMake(0).pipe(
+ *   DateTime.add(5, "minutes")
+ * )
+ */
+export const add: {
```
jdharrisnz, line 1633: For this and `subtract`, would it make sense for it to instead accept a `Partial<DateTime.Parts>` object? That way, users could add various units at once, and you'd be saved from dealing with so many strings. This also aligns with the `Temporal` API.

packages/effect/src/DateTime.ts
```diff
@@ -0,0 +1,1991 @@
...
+        return date
+      }
+      case "months":
+      case "month": {
+        const day = date.getUTCDate()
+        date.setUTCMonth(date.getUTCMonth() + amount + 1, 0)
+        if (day < date.getUTCDate()) {
+          date.setUTCDate(day)
+        }
+        return date
```
jdharrisnz, line 1670: This aligns with `Temporal`'s default `{ overflow: "constrain" }` option, but it's different to how standard `Date`s work. Just want to point it out to make sure it's intentional and that there's also no intention to add another option for it.

jdharrisnz (commented):
Thanks for doing this work, Tim, it's looking great! Sorry I didn't add my stuff, time hasn't been on my side lately. You've done a way better job anyway.

Did you think about adding a `toNearest` function?

packages/effect/src/DateTime.ts
tim-smart: Yes it is intentional, I think it is temporals default for good reason.

If you want the more exact behaviour, you can use .addDuration.

tim-smart:
> Did you think about adding a `toNearest` function?

What does it do?

jdharrisnz:
> > Did you think about adding a `toNearest` function?
> 
> 
> 
> What does it do?

Sorry, ambiguous name! I mean round to the nearest whole 'part'.

jdharrisnz:
Also, there are a few pairs of things here that involve `Date.now()`. Half are Effects, the other half are plain functions.

The plain functions are very welcome, but should they be prefixed with `unsafe` like this?

Elsewhere in Effect this is used to mean "this could throw", "you could end up with `undefined`", "the result could be invalid", "you better know what you're doing", etc. It doesn't seem like a similar case here.

I'm not sure I have better naming to suggest though. Maybe add "Sync" to the end? Or add "Effect" to the end of the Effects and use normal names for the plain functions?

tim-smart:
> Sorry, ambiguous name! I mean round to the nearest whole given 'part'

Maybe we can just add more units to `startOf`? I.e. `startOf("hour")` would be the same as rounding off to the nearest hour.

tim-smart:
> Elsewhere in Effect this is used to mean "this could throw"

No, elsewhere in Effect it means "this can perform side effects / is not pure without using the Effect type"

thewilkybarkid:
> One nitpick: I don't like the `WithZone` naming. The Temporal API uses `ZonedDateTime`. I think that has a better ring to it. Wdyt?
> 
> I'm wondering if we should align even more with the Temporal API suggestion and also use `PlainDateTime` for those without a timezone. But I'm not sure if I like that namechoice. But on the other hand it might make sense to preemptively align with that API.

Is there meant to be a relationship with Temporal? Can't see much about it in the comments.

Was just wondering as I've been using it quite a bit because of the other types like `PlainYearMonth` and `PlainDate` (e.g. https://github.com/PREreview/coar-notify/blob/d3baf2ba7c4e78027a958be57099c79eca899bd6/src/Temporal.ts).

tim-smart:
> Is there meant to be a relationship with Temporal? Can't see much about it in the comments.

At this stage no, but I think we can add more "Plain" data types over time and align more with Temporal.

PR #3302, opened 2024-07-18
feat(Stream): add Stream.onStart api

dilame:
Not sure about name, maybe it should be `tapSubscribe`?

## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Implement `Stream.tapStart` that adds an effect to be executed at the start of the stream.

```ts
import { Console, Effect, Stream } from "effect";

const stream = Stream.make(1, 2, 3).pipe(
  Stream.tapStart(Console.log("Stream started")),
  Stream.map((n) => n * 2),
  Stream.tap((n) => Console.log(`after mapping: ${n}`))
) 

Effect. runPromise(Stream. runCollect(stream)).then(console. log) 
// Stream started 
// after mapping: 2 
// after mapping: 4 
// after mapping: 6 
// { _id: 'Chunk', values: [ 2, 4, 6 ] }
```

### DIFF
--- a/packages/effect/src/internal/stream.ts
+++ b/packages/effect/src/internal/stream.ts
@@ -6208,7 +6208,7 @@ export const tapStart: {
   <A, E, R, _, E2, R2>(
     self: Stream.Stream<A, E, R>,
     effect: Effect.Effect<_, E2, R2>
-  ): Stream.Stream<A, E | E2, R | R2> => concat(drain(fromEffect(effect)), self)
+  ): Stream.Stream<A, E | E2, R | R2> => unwrap(Effect.as(effect, self))
 )
 
 /** @internal */
--- a/packages/effect/src/internal/stream.ts
+++ b/packages/effect/src/internal/stream.ts
@@ -6195,6 +6195,23 @@ export const tap: {
 )
 
 /** @internal */
+export const tapStart: {
+  <_, E2, R2>(
+    effect: Effect.Effect<_, E2, R2>
+  ): <A, E, R>(self: Stream.Stream<A, E, R>) => Stream.Stream<A, E2 | E, R2 | R>
+  <A, E, R, _, E2, R2>(
+    self: Stream.Stream<A, E, R>,
+    effect: Effect.Effect<_, E2, R2>
+  ): Stream.Stream<A, E | E2, R | R2>
+} = dual(
+  2,
+  <A, E, R, _, E2, R2>(
+    self: Stream.Stream<A, E, R>,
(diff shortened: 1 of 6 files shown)


packages/effect/src/internal/stream.ts
```diff
@@ -6194,6 +6194,23 @@ export const tap: {
   ): Stream.Stream<A, E | E2, R | R2> => mapEffectSequential(self, (a) => Effect.as(f(a), a))
 )
 
+/** @internal */
+export const tapStart: {
```
tim-smart, line 6198: I think `onStart` is a better name, aligns with `onInterrupt` etc.

packages/effect/src/internal/stream.ts
```diff
@@ -6194,6 +6194,23 @@ export const tap: {
...
+  <A, E, R, _, E2, R2>(
+    self: Stream.Stream<A, E, R>,
+    effect: Effect.Effect<_, E2, R2>
+  ): Stream.Stream<A, E | E2, R | R2>
+} = dual(
+  2,
+  <A, E, R, _, E2, R2>(
+    self: Stream.Stream<A, E, R>,
+    effect: Effect.Effect<_, E2, R2>
+  ): Stream.Stream<A, E | E2, R | R2> => concat(drain(fromEffect(effect)), self)
```
tim-smart, line 6211: This is a bit cleaner:

```suggestion
  ): Stream.Stream<A, E | E2, R | R2> => unwrap(Effect.as(effect, self))
```

tim-smart (changes requested):

packages/effect/src/internal/stream.ts
dilame: Or maybe `onSubscribe`? 😃

dilame: But yeah, `on` prefix sounds better, because `tap` implies some input value

tim-smart: We don't use the `subscribe` terminology in effect, so I prefer "start".

dilame:
Done. What do you think about `Stream.onEnd` with `concat(self, drain(fromEffect(effect)))`?

tim-smart:
> Done. What do you think about `Stream.onEnd` with `concat(self, drain(fromEffect(effect)))`?

I'm OK with it. Feel free to open a seperate PR.

tim-smart (approved):

PR #3309, opened 2024-07-20
feat(@effect/sql): SqlQuery module

dilame:
Add `SqlQuery` module with `findAll`, `void`, `findOne`, and `single` **dual** APIs for running SQL queries with schema validation.

### SqlQuery.findAll

Runs an SQL query and validates the results against a provided schema, returning all results as an array.

```ts
import { Schema } from "@effect/schema"
import { SqlClient, SqlQuery } from "@effect/sql"

const userSchema = Schema.struct({
  id: Schema.number,
  name: Schema.string,
})

const findAllUsers = Effect.gen(function* () {
  const sql = yield* SqlClient.SqlClient

  return yield* SqlQuery.findAll(userSchema)(
    sql`SELECT id, name FROM users`
  )
})
```

### SqlQuery.void

Runs an SQL query and discards the result.

```ts
import { SqlClient, SqlQuery } from "@effect/sql"

const updateUser = Effect.gen(function* () {
  const sql = yield* SqlClient.SqlClient

  return yield* SqlQuery.void(
    sql`UPDATE users SET name = 'Alice' WHERE id = 1`
  )
})

```

### SqlQuery.findOne

Runs an SQL query and validates the results against a provided schema, returning the first result as an `Option`.

```ts
import { Schema } from "@effect/schema"
import { SqlClient, SqlQuery } from "@effect/sql"
import { Effect, Option } from "effect"

const userSchema = Schema.struct({
  id: Schema.number,
  name: Schema.string,
})

const findUser = Effect.gen(function* () {
  const sql = yield* SqlClient.SqlClient

  const result = yield* SqlQuery.findOne(userSchema)(
    sql`SELECT id, name FROM users WHERE id = 1`
  )
  if(Option.isSome(result)) {
    console.log(result.value)
  } else {
    console.log('User not found')
  }
})
```

### SqlQuery.single

Runs an SQL query and validates the results against a provided schema, returning the first result. 
If no result is found, it throws a `NoSuchElementException`.

```ts
import { Schema } from "@effect/schema"
import { SqlClient, SqlQuery } from "@effect/sql"
import { Effect, Console } from "effect"
import * as Cause from "effect/Cause"

const userSchema = Schema.struct({
  id: Schema.number,
  name: Schema.string,
})

const findSingleUser = Effect.gen(function* () {
  const sql = yield* SqlClient.SqlClient

  return yield* SqlQuery.single(userSchema)(
    sql`SELECT id, name FROM users WHERE id = 1`
  ).pipe(
    Effect.catchTag('NoSuchElementException', () => Console.log('User not found'))
  )
})
```


dilame:
Not sure if we really need `single` because it's too simple, just `Effect.flatten`

tim-smart:
There is also these schema APIs https://effect-ts.github.io/effect/sql/SqlSchema.ts.html#findall

dilame:
> There is also these schema APIs https://effect-ts.github.io/effect/sql/SqlSchema.ts.html#findall

Of course i've seen it:) And implemented this PR because `SqlQuery` doesn't fit my needs. I don't need to parse input, because it's already parsed. I would say that it's a very rare case when you pass input **directly** to SQL, because usually SQL is not the input edge of the application.

Also, i use "_ports and adapters_" or "_hexagonal_" (call as you wish) architecture, where my persistence layer is abstracted, so my application should not know about the underlying driver. It could be SQL DB, redis, or filesystem. `SqlSchema` module emits `SqlError`, my application is not allowed to know about it. So in order to map error type i anyway do 

```ts
export const FetchBotLaunchConfigPostgres = Layer.effect(
  PersistenceFetchBotLaunchConfigImpl,
  Effect.gen(function* () {
    const sql = yield* SqlClient.SqlClient

    return {
      fetchLaunchConfig: (instanceId: string) => {
        return SqlSchema.single({
          Request: Schema.String,
          Result: PersistenceFetchBotLaunchConfigRes,
          execute: (instanceId) =>
            sql`SELECT *
                FROM fxngine.instance_launch_config
                WHERE instance_id = ${instanceId}::bigint`
        })(instanceId).pipe(
          Effect.catchTags({
            SqlError: (cause) =>
              new PersistenceError({ source: 'fetchBotLaunchConfig', cause }),
            NoSuchElementException: () =>
              new PersistenceFetchBotLaunchConfigNotFoundError({ instanceId })
          })
        )
      }
    }
  })
)

```

Which makes no sense

tim-smart:
Going to close this, I don't think it adds enough value to be an addition to /sql.

dilame:
I understand that the schema is also about encoding, not just validation. However, in my case, encoding is unnecessary for SQL queries. I deal with dozens of SQL queries, and i am forced to use `S.Any` as a placeholder. The example of encoding an ID is an edge case that applies only in specific, rare situations. I respectfully ask for reconsideration of the PR, as it provides a solution that is suitable for the majority of users who don’t need this kind of encoding for SQL.

Let's view at this PR from another perspective – how would you suggest to make an SQL that does not requires input encoding?

PR #3314, opened 2024-07-21
add `Random.choice`

sukovanej:


packages/effect/src/internal/random.ts
```diff
@@ -79,6 +81,12 @@ const shuffleWith = <A>(
   )
 }
 
+export const choice = <A>(elements: array.NonEmptyReadonlyArray<A>): Effect.Effect<A> =>
```
tim-smart, line 84: Will need to move this to the "defaultServices" module.

dilame: Should the elements array really be `NonEmpty`?

tim-smart: Yes, otherwise you will need to return an Option or fail with a NoSuchElementException.

sukovanej: Fixed

datner (commented):
Cool feature, I like

packages/effect/src/internal/random.ts
mikearnaldi: wondering if we shouldn't make this conditional, like `X extends ReadonlyArray` and if not empty then return `Effect<A>`, otherwise `Effect<A, NoSuchElement...>`, forcing `NonEmpty` proved tedious to use in other APIs

packages/effect/test/Random.test.ts
```diff
@@ -25,4 +25,10 @@ describe("Random", () => {
       assert.strictEqual(n2, n3)
       assert.notStrictEqual(n0, n2)
     }))
+
+  it.live("choice", () =>
+    Effect.gen(function*() {
+      expect(yield* Random.choice([1])).toEqual(1)
+      expect(yield* Random.choice([1, 2, 3])).oneOf([1, 2, 3])
+    }))
```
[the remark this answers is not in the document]
mikearnaldi: 1. `it.effect` provides a `TestEnv` by default, `it.live` opts-out
2. fast-check is probably a good way but even a repeated test would be fine

packages/effect/src/internal/random.ts
sukovanej: Updated

KhraksMamtsov: What about widening `Array` to `Iterable`?

sukovanej: The length must be known upfront so I don't it's possible to make `choice` work for any iterable.

KhraksMamtsov: You can materialize an `Iterable` to an `Array` before accepting an element. 
I think this request will appear in the future anyway.

datner: Suggest an edit then lol
also I don't think this api is going to be very popular, like not even top 100, it's quite niche in the end of the day

packages/effect/src/internal/defaultServices.ts
```diff
@@ -133,6 +133,17 @@ export const nextIntBetween = (min: number, max: number): Effect.Effect<number>
...
+/** @internal */
+export const choice = <Self extends ReadonlyArray<unknown>>(
+  elements: Self
+) =>
+  core.map(
+    elements.length === 0
+      ? core.fail(new core.NoSuchElementException("Cannot select a random element from an empty array"))
+      : randomWith((random) => random.nextIntBetween(0, elements.length)),
+    (i) => elements[i]
+  ) as any
```
KhraksMamtsov, line 145:
```suggestion
export const choice = <Self extends Iterable<unknown>>(
  elements: Self
) => {
const _elements = Array.fromIterable(elements)
  return core.map(
    _elements.length === 0
      ? core.fail(new core.NoSuchElementException("Cannot select a random element from an empty iterable"))
      : randomWith((random) => random.nextIntBetween(0, _elements.length)),
    (i) => _elements[i]
  ) as any
}
```

packages/effect/src/Random.ts
```diff
@@ -103,6 +105,26 @@ export const nextIntBetween: (min: number, max: number) => Effect.Effect<number>
...
+ *   console.log(randomItem)
+ * })
+ *
+ * @since 2.0.0
+ * @category constructors
+ */
+export const choice: <Self extends ReadonlyArray<unknown>>(
+  elements: Self
+) => Self extends Array.NonEmptyReadonlyArray<infer A> ? Effect.Effect<A>
+  : Self extends ReadonlyArray<infer A> ? Effect.Effect<A, Cause.NoSuchElementException>
```
KhraksMamtsov, line 125:
```suggestion
  : Self extends Iterable<infer A> ? Effect.Effect<A, Cause.NoSuchElementException>
```

packages/effect/src/internal/defaultServices.ts
KhraksMamtsov: What do you think?

packages/effect/src/internal/random.ts
sukovanej: My problem with allowing any iterable is that we'd allow infinite iterators as well and in such a case it would throw an exception (by trying it locally, the `Array.from` gives up after ~125M iterations and throws a `RangeError` exception). If we're okay with this, I'll add the support, document this edge-case behaviour in the doc-string and convert the exception to `Effect.die`. WDYT?

mikearnaldi: > My problem with allowing any iterable is that we'd allow infinite iterators as well and in such a case it would throw an exception (by trying it locally, the `Array.from` gives up after ~125M iterations and throws a `RangeError` exception). If we're okay with this, I'll add the support, document this edge-case behaviour in the doc-string and convert the exception to `Effect.die`. WDYT?

kind of the same for every single api that allows for an iterable and we have plenty, so I'd not consider this an issue. You can have `X extends Iterable<any>` and then use a conditional to branch the cases

sukovanej: ok

packages/effect/src/Random.ts
```diff
@@ -119,10 +119,10 @@ export const shuffle: <A>(elements: Iterable<A>) => Effect.Effect<Chunk.Chunk<A>
  * @since 2.0.0
  * @category constructors
  */
-export const choice: <Self extends ReadonlyArray<unknown>>(
+export const choice: <Self extends Iterable<unknown>>(
   elements: Self
 ) => Self extends Array.NonEmptyReadonlyArray<infer A> ? Effect.Effect<A>
-  : Self extends ReadonlyArray<infer A> ? Effect.Effect<A, Cause.NoSuchElementException>
+  : Self extends Iterable<infer A> ? Effect.Effect<A, Cause.NoSuchElementException>
```
tim-smart, line 125: You can use the `NonEmptyIterable` interface to check for `NonEmptyChunk` etc.

packages/effect/src/Random.ts
```diff
@@ -103,6 +106,27 @@ export const nextIntBetween: (min: number, max: number) => Effect.Effect<number>
...
+ *
+ * @example
+ * import { Effect, Random } from "effect"
+ *
+ * Effect.gen(function* () {
+ *   const randomItem = yield* Random.choice([1, 2, 3])
+ *   console.log(randomItem)
+ * })
+ *
+ * @since 2.0.0
```
tim-smart, line 120:
```suggestion
 * @since 3.6.0
```

packages/effect/src/Random.ts
```diff
@@ -103,6 +106,27 @@ export const nextIntBetween: (min: number, max: number) => Effect.Effect<number>
...
+ * @category constructors
+ */
+export const choice: <Self extends Iterable<unknown>>(
+  elements: Self
+) => Self extends NonEmptyIterable.NonEmptyIterable<infer A> ? Effect.Effect<A>
+  : Self extends Array.NonEmptyReadonlyArray<infer A> ? Effect.Effect<A, never>
```
tim-smart, line 126:
```suggestion
  : Self extends Array.NonEmptyReadonlyArray<infer A> ? Effect.Effect<A>
```

tim-smart (commented):
8y94vp

PR #3329, opened 2024-07-24
Support record argument in `Prompt.all`

### LINKED ISSUE #3327: From Discord: Confusion with Prompt.all: Handling Record/Struct Arguments in TypeScript
# Summary
### Summary:

**User danielo515** is encountering an issue with the `Prompt.all` function in TypeScript. According to the documentation, `Prompt.all` should support multiple arguments, a single argument tuple/array, or a record/struct, allowing the user to get back an object of responses. However, when danielo515 tries to execute it with a record/struct, TypeScript throws an error: 

```
Object literal may only specify known properties, and 'run' does not exist in type 'Iterable<Prompt<any>>'.ts(2353)
```

**User gcanti** responds by stating that the documentation is incorrect.

### Key Takeaways:

1. **Documentation Issue**: The `Prompt.all` function documentation may not accurately describe its capabilities or usage.
2. **TypeScript Error**: The error indicates that the provided object structure is not compatible with the expected type `Iterable<Prompt<any>>`.
3. **Need for Clarification**: There is a need to clarify how to correctly use `Prompt.all` with record/struct arguments in TypeScript.

# Discord thread

https://discord.com/channels/795981131316985866/1265632416568049724

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #3327 
- Closes #3327

### DIFF
--- a/packages/cli/src/Prompt.ts
+++ b/packages/cli/src/Prompt.ts
@@ -463,8 +463,21 @@ export declare namespace All {
   /**
    * @since 1.0.0
    */
-  export type Return<Arg extends Iterable<PromptAny>> = [Arg] extends [ReadonlyArray<PromptAny>] ? ReturnTuple<Arg>
+  export type ReturnObject<T> = [T] extends [{ [K: string]: PromptAny }] ? Prompt<
+      {
+        -readonly [K in keyof T]: [T[K]] extends [Prompt.Variance<infer _A>] ? _A : never
+      }
+    >
+    : never
+
+  /**
+   * @since 1.0.0
+   */
+  export type Return<
+    Arg extends Iterable<PromptAny> | Record<string, PromptAny>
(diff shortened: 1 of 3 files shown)


packages/cli/src/Prompt.ts
```diff
@@ -478,9 +491,8 @@ export declare namespace All {
  * @since 1.0.0
  * @category collecting & elements
  */
-export const all: <
-  const Arg extends Iterable<Prompt<any>>
->(arg: Arg) => All.Return<Arg> = InternalPrompt.all
```
gcanti, line 483: RE "Supports **multiple arguments**" in jsdocs, AFAIK it only supports a single argument

IMax153: That's true - only an array or a record. I can update the JSDoc and add the example from the changelog as a separate PR.

PR #3340, opened 2024-07-25
adjusted various function signatures to return `Array`

fubhy:
~~`Effect.forEach` should probably also return `Array<A>` instead.~~


mikearnaldi:
```ts

const res = Effect.forEach([0, 1, 2], (n) => Effect.succeed(n))
// Effect.Effect<number[], never, never>
```

Seems like it does

fubhy:
> ```ts
> import { Effect } from "effect"
> 
> const res = Effect.forEach([0, 1, 2], (n) => Effect.succeed(n))
> // Effect.Effect<number[], never, never>
> ```
> 
> Seems like it does

Ah yeah. I was confused by the `Effect<RA.ReadonlyArray.With<S, B>, E, R>` and check to see what that actually resolves to.

tim-smart (approved):

PR #3364, opened 2024-07-27
add `applicationName` option to `PgClient`

sukovanej:


packages/sql-pg/src/PgClient.ts
```diff
@@ -129,7 +130,10 @@ export const make = (
       fetch_types: options.fetchTypes ?? true,
       prepare: options.prepare ?? true,
       types: options.types,
-      debug: options.debug
+      debug: options.debug,
+      connection: {
+        application_name: options.applicationName ?? "postgres.js"
```
tim-smart, line 135:
```suggestion
        application_name: options.applicationName ?? "@effect/sql-pg"
```

May as well make it more accurate.

PR #3365, opened 2024-07-27
add `socket` option to `PgClient`

sukovanej:


packages/sql-pg/src/PgClient.ts
```diff
@@ -64,6 +63,27 @@ export interface PgClientConfig {
...
+   * const layer = Effect.gen(function*() {
+   *   const connector = new Connector();
+   *   const clientOpts = yield* Effect.promise(() => connector.getOptions({
+   *     instanceConnectionName: "project:region:instance",
+   *     authType: AuthTypes.IAM,
+   *   }));
+   *   return PgClient.layer({ socket: clientOpts.stream, user: "iam-user" });
+   * }).pipe(Layer.unwrapEffect)
+   */
+  readonly socket?: (() => unknown) | undefined
```
tim-smart, line 85: Can unknown be replaced with a node.js Duplex?

sukovanej: Honestly, I wasn't sure which one of `TLSSocket`, `Socket`, `Duplex`,... to constrain it with and hoped for your input. I'll fix it. Thanks.

packages/sql-pg/src/PgClient.ts
```diff
@@ -80,11 +101,15 @@ export interface PgClientConfig {
   readonly prepare?: boolean | undefined
   readonly types?: Record<string, postgres.PostgresType> | undefined
 
-  readonly debug?: postgres.Options<{}>["debug"] | undefined
+  readonly debug?: PostgresOptions["debug"] | undefined
 }
 
 type PartialWithUndefined<T> = { [K in keyof T]?: T[K] | undefined }
 
+interface PostgresOptions extends postgres.Options<{}> {
```
tim-smart, line 109: This should be exported if we are using it on a public interface. Or you can revert the "debug" change.

PR #3366, opened 2024-07-27
Support `Refinement` in `Preficate.tuple` and `Predicate.struct`

KhraksMamtsov:
support combination predicates and refinements

## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Predicate.ts
+++ b/packages/effect/src/Predicate.ts
@@ -30,6 +30,45 @@ export interface Refinement<in A, out B extends A> {
 }
 
 /**
+ * @since 3.6.0
+ * @category type-level
+ */
+export declare namespace Predicate {
+  /**
+   * @since 3.6.0
+   * @category type-level
+   */
+  export type In<T extends Any> = [T] extends [Predicate<infer _A>] ? _A : never
+  /**
+   * @since 3.6.0
+   * @category type-level
+   */
+  export type Any = Predicate<any>
+}
+
+/**
+ * @since 3.6.0
+ * @category type-level
+ */
+export declare namespace Refinement {
+  /**
+   * @since 3.6.0
+   * @category type-level
+   */
+  export type In<T extends Any> = [T] extends [Refinement<infer _A, infer _>] ? _A : never
+  /**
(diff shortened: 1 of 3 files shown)


packages/effect/src/Predicate.ts
```diff
@@ -686,31 +725,51 @@ export const productMany = <A>(
...
  *
  * @since 2.0.0
  */
-export const tuple = <T extends ReadonlyArray<Predicate<any>>>(
-  ...elements: T
-): Predicate<Readonly<{ [I in keyof T]: [T[I]] extends [Predicate<infer A>] ? A : never }>> => all(elements) as any
+export const tuple: {
+  <T extends ReadonlyArray<Predicate.Any>>(
+    ...elements: T
+  ): true extends { [K in keyof T]: T[K] extends Refinement.Any ? true : false }[number] ? Refinement<
```
tim-smart, line 738: Would this be better? Then you can remove the intermediate record:

```ts
[Extract<A[keyof A], Predicate.Refinement<any, any>>] extends [never]
```

packages/effect/src/Predicate.ts
```diff
@@ -686,31 +725,51 @@ export const productMany = <A>(
...
+      Readonly<{ [I in keyof T]: T[I] extends Refinement.Any ? Refinement.In<T[I]> : Predicate.In<T[I]> }>,
```
tim-smart, line 739:
```suggestion
      { readonly [I in keyof T]: T[I] extends Refinement.Any ? Refinement.In<T[I]> : Predicate.In<T[I]> },
```

You can do it in one pass.

packages/effect/src/Predicate.ts
KhraksMamtsov: Yes, It works and more elegant.

packages/effect/src/Predicate.ts
KhraksMamtsov: done

PR #3398, opened 2024-08-01
add `listen` / `notify` to `PgClient`

sukovanej:


packages/sql-pg/src/PgClient.ts
```diff
@@ -249,7 +251,26 @@ export const make = (
           database: client.options.database
         },
         json: (_: unknown) => PgJson(_),
-        array: (_: ReadonlyArray<Primitive>) => PgArray(_)
+        array: (_: ReadonlyArray<Primitive>) => PgArray(_),
+        listen: (channel: string) =>
+          Stream.asyncScoped<string, SqlError>((emit) =>
```
tim-smart, line 256: The new `Stream.asyncPush` api is a better fit here.

tim-smart (changes requested):

packages/sql-pg/src/PgClient.ts
sukovanej: The `asyncPush` doesn't allow the register effect to fail in the type signature. I can catch on the promise and emit failure but that doesn't feel ergonomic. Can we adjust the type signature of `asyncPush` to allow the register to fail or is the `never` there intentional?

tim-smart: No it wasn't intentional. Feel free to patch it in this PR and add a separate changeset.

tim-smart (approved):

PR #3404, opened 2024-08-03
Fix `Cache<_, Value, _>` type parameter variance (covariant -> invariant)

KhraksMamtsov:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Cache.ts
+++ b/packages/effect/src/Cache.ts
@@ -23,13 +23,13 @@ export const CacheTypeId: unique symbol = internal.CacheTypeId
 export type CacheTypeId = typeof CacheTypeId
 
 /**
- * @since 2.0.0
+ * @since 3.6.2
  * @category symbols
  */
 export const ConsumerCacheTypeId: unique symbol = internal.ConsumerCacheTypeId
 
 /**
- * @since 2.0.0
+ * @since 3.6.2
  * @category symbols
  */
 export type ConsumerCacheTypeId = typeof ConsumerCacheTypeId
@@ -180,7 +180,7 @@ export declare namespace Cache {
     }
   }
   /**
-   * @since 2.0.0
+   * @since 3.6.2
    * @category models
    */
   export interface ConsumerVariance<in out Key, out Value, out Error> {
--- a/packages/effect/src/Cache.ts
+++ b/packages/effect/src/Cache.ts
@@ -43,7 +43,7 @@ export type CacheTypeId = typeof CacheTypeId
  * @since 2.0.0
  * @category models
  */
-export interface Cache<in out Key, out Value, out Error = never> extends ConsumerCache<Key, Value, Error> {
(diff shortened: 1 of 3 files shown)


mikearnaldi:
just changing `out` to `in out` doesn't actually affect variance when compared structurally, we should use the `Variance` trick we use in `Effect` and on the other types to force a specific variance

KhraksMamtsov:
Done.
I used the `Cache.Variance` type for the regular `Cache`, and for `ConsumerCache` - `Cache.ConsumerVariance`.
And now `CacheImpl` contains both `Variance` symbols.
I was looking at `Queue` which implements `Enqueue` + `Dequeue`

packages/effect/src/Cache.ts
```diff
@@ -22,6 +22,18 @@ export const CacheTypeId: unique symbol = internal.CacheTypeId
  */
 export type CacheTypeId = typeof CacheTypeId
 
+/**
+ * @since 2.0.0
```
tim-smart, line 26:
```suggestion
 * @since 3.6.2
```

tim-smart (changes requested):

packages/effect/src/Cache.ts
KhraksMamtsov: Done

packages/effect/src/Cache.ts
```diff
@@ -22,6 +22,18 @@ export const CacheTypeId: unique symbol = internal.CacheTypeId
  */
 export type CacheTypeId = typeof CacheTypeId
 
+/**
+ * @since 3.6.2
```
tim-smart, line 26:
```suggestion
 * @since 3.6.4
```

packages/effect/src/Cache.ts
```diff
@@ -22,6 +22,18 @@ export const CacheTypeId: unique symbol = internal.CacheTypeId
...
+ * @category symbols
+ */
+export const ConsumerCacheTypeId: unique symbol = internal.ConsumerCacheTypeId
+
+/**
+ * @since 3.6.2
```
tim-smart, line 32:
```suggestion
 * @since 3.6.4
```

packages/effect/src/Cache.ts
```diff
@@ -156,8 +172,19 @@ export declare namespace Cache {
...
-  export interface Variance<in out Key, out Value, out Error> {
+  export interface Variance<in out Key, in out Value, out Error> {
     readonly [CacheTypeId]: {
+      readonly _Key: Types.Invariant<Key>
+      readonly _Error: Types.Covariant<Error>
+      readonly _Value: Types.Invariant<Value>
+    }
+  }
+  /**
+   * @since 3.6.2
```
tim-smart, line 183:
```suggestion
   * @since 3.6.4
```

tim-smart (changes requested):

tim-smart (approved):

PR #3409, opened 2024-08-04
add `{Bun,Node}HttpServer.layerTest` for testing http servers

sukovanej:
POC of http testing modules, it is introducing three new modules

- `NodeHttpTest` and `BunHttpTest` - exposing layers enabling live testing of http apps, I have similar implementation in the `effect-http` package, and one implementation was already used in the test suite for /platform-node so this is only about exposing this functionality for convenient testing in the user-land
- `HttpTest` exposing layers for "in-memory" testing of http apps, it mimics the client <> server comm using a queue, there's an implementation of request / response classes enabling conversions `client request ->  server request` and `server response -> client response` and finally layers with the implementation of `HttpClient` and `HttpServer` that exchange requests / responses through the queue and map the requests / responses.

UPDATE: the in-memory implementation was removed and this PR now introduces only the live testing layers.


sukovanej:
@tim-smart could you please take a look and give me some feedback? I'd like to know what you think about this in general so I can decide whether to move forward with it.

packages/platform-bun/src/BunHttpTest.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * @since 0.41.3
```
tim-smart, line 2: If you keep all the since tags at 1.0.0, as we are not yet api stable.

packages/platform-node/src/NodeHttpTest.ts
```diff
@@ -0,0 +1,24 @@
...
+import type * as HttpServerError from "@effect/platform/HttpServerError"
+import type * as Layer from "effect/Layer"
+import * as internal from "./internal/nodeHttpTest.js"
+import type * as NodeContext from "./NodeContext.js"
+
+/**
+ * @since 0.56.3
+ * @category layers
+ */
+export const layer: Layer.Layer<
```
tim-smart, line 17: I think we can export this from the `NodeHttpServer` module as `layerTest`.

packages/platform/src/internal/httpInMemoryMessage.ts
```diff
@@ -0,0 +1,248 @@
...
+
+    if (body._tag === "FormData") {
+      return Effect.succeed(body.formData)
+    }
+
+    return Effect.dieMessage("not implemented")
+  }
+  toJSON(): unknown {
+    return HttpIncomingMessage.inspect(this, {
+      _id: "@effect/platform/InMemoryServerRequestImpl",
```
tim-smart, line 243:
```suggestion
      _id: "@effect/platform/HttpClientResponse",
```

packages/platform/src/HttpTest.ts
```diff
@@ -0,0 +1,89 @@
...
+import type * as HttpServer from "./HttpServer.js"
+import type * as HttpServerRequest from "./HttpServerRequest.js"
+import type * as HttpServerResponse from "./HttpServerResponse.js"
+import * as internal from "./internal/httpTest.js"
+
+/**
+ * @since 0.56.3
+ * @category models
+ */
+export interface HttpInMemoryServerQueue {
```
tim-smart, line 22:
```suggestion
export interface InMemoryHttpServerQueue {
```

packages/platform/src/HttpTest.ts
```diff
@@ -0,0 +1,89 @@
...
+  Exclude<R, HttpServerRequest.HttpServerRequest> | HttpClient.HttpClient.Default | HttpServer.HttpServer | Scope.Scope
+> = internal.serve
+
+/**
+ * Layer producing an `HttpClient` with prepended url of the running http server.
+ *
+ * @since 0.56.3
+ * @category layers
+ */
+export const clientLayer: Layer.Layer<
```
tim-smart, line 57: Generally for layer apis, the naming starts with "layer" and then has a suffix.

If we move this to the HttpServer module, then this could be `layerTestClient`.

```suggestion
export const layerTestClient: Layer.Layer<
```

tim-smart (commented):
I think the test live layers are a good addition - I think we can just add them to the HttpServer modules as `layerTest`.

Not sure about the in-memory addition. Maybe add those changes in a separate PR and we can discuss there.

packages/platform-bun/src/BunHttpTest.ts
sukovanej: I'll change it. I noticed Giulio uses precise `@since` versions in the /schema and I think it's beneficial for the consumers of the lib.

sukovanej:
Thanks for the feedback. I removed the in-memory impl from this PR and adjusted.

packages/platform-bun/src/BunHttpServer.ts
```diff
@@ -36,6 +38,35 @@ export const layer: (
...
+ *   Effect.gen(function*(_) {
+ *     yield* HttpServer.serveEffect(HttpRouter.empty)
+ *     const response = yield* HttpClientRequest.get("/non-existing")
+ *     expect(response.status).toEqual(404)
+ *   }).pipe(Effect.provide(BunHttpServer.layerTest), Effect.scoped, Effect.runPromise))
+ *
+ * @since 1.0.0
+ * @category layers
+ */
+export const layerTest: Layer.Layer<
```
tim-smart, line 61: I think this should be moved to the server module as `layerTestClient`, as it is concerned with server testing.

PR #3424, opened 2024-08-07
Improve custom matchers

mikearnaldi:


IMax153:
@mikearnaldi @tim-smart - Might be worth circling back to this discussion - the same issue with `vitest`'s handling of `Error` has bitten another user here: https://discord.com/channels/795981131316985866/795983589644304396/1285321401213124618

We should discuss Mike's proposed solution here so we can try to resolve it for users at least on our end.

fubhy:
This is still marked draft and currently has conflicts. Could you confirm whether this is still planned and update it if it is ready for review?

tim-smart:

PR #3435, opened 2024-08-08
ensure fiber is properly cleared in FiberHandle.unsafeSet

Andarist:
This was caught by this TS experiment I'm running: https://github.com/microsoft/TypeScript/pull/59559 . I believe this to be the correct change :P


tim-smart (approved):
Thanks :)

PR #3441, opened 2024-08-11
Avoid automatic propagation of finalizer concurrency, closes #3440

### LINKED ISSUE #3440: Dynamic Scope access causes a memory leak

3.x

The following code ends up creating a memory leak:

```ts
const program = Effect.scoped(Effect.gen(function*() {
  while (true) {
    yield* Effect.all([Effect.void, Effect.void], { concurrency: "unbounded" })
    yield* Effect.sleep("5 millis")
  }
}))
```

The reason is that `all` uses `forEach` and `forEach` dynamically accesses `Scope` in order to execute finalizers according to the strategy specified in `concurrency`. It all sounds like a nice feature but to do that we need to create a child scope that cannot close until the parent closes, and doing this repeatedly ends up adding an unbounded amount of finalizers into the parent `Scope` which may be created further up (and in theory may never close).

This bug was discovered by the JUCR folks and was also found in ZIO, the choice of action for ZIO has been to remove the feature altogether, I think we can do better and allow the feature with a specific flag since this only materialise when the `forEach` itself is repeated & when you have a parent `Scope` that never closes.

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

mikearnaldi:
closes #3440

### DIFF
--- a/.changeset/witty-kings-lie.md
+++ b/.changeset/witty-kings-lie.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Avoid automatic propagation of finalizer concurrency, closes #3440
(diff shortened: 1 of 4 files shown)


mikearnaldi:
Cc @tim-smart

tim-smart (approved):

PR #3456, opened 2024-08-14
Introduced `RandomScheduler`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

 Same as `MixedScheduler`, but performs tasks with the same priority randomly
 
 It is useful in scenarios where you want to introduce non-deterministic behavior in task execution while still
 ensuring that higher-priority tasks are generally executed before lower-priority ones.
 Uses a Pseudo-Random Number Generator (PRNG) for selecting the next task to execute within a tasks with the same priority.
 `RandomScheduler` can be seeded for reproducibility.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/cyan-coats-peel.md
+++ b/.changeset/cyan-coats-peel.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Introduced `RandomScheduler`
--- a/.changeset/cyan-coats-peel.md
+++ b/.changeset/cyan-coats-peel.md
@@ -1,5 +1,5 @@
 ---
-"effect": patch
+"effect": minor
 ---
 
 Introduced `RandomScheduler`
--- a/packages/effect/src/Scheduler.ts
+++ b/packages/effect/src/Scheduler.ts
@@ -7,7 +7,9 @@ import type { RuntimeFiber } from "./Fiber.js"
 import type { FiberRef } from "./FiberRef.js"
 import { dual } from "./Function.js"
 import { globalValue } from "./GlobalValue.js"
+import * as Hash from "./Hash.js"
 import * as core from "./internal/core.js"
+import * as Utils from "./Utils.js"
 
 /**
  * @since 2.0.0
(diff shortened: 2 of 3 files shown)


.changeset/cyan-coats-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
```
tim-smart, line 2:
```suggestion
"effect": minor
```

packages/effect/src/Scheduler.ts
```diff
@@ -238,6 +240,102 @@ export class ControlledScheduler implements Scheduler {
...
+/**
+ * Same as `MixedScheduler`, but performs tasks with the same priority randomly
+ *
+ * It is useful in scenarios where you want to introduce non-deterministic behavior in task execution while still
+ * ensuring that higher-priority tasks are generally executed before lower-priority ones.
+ * Uses a Pseudo-Random Number Generator (PRNG) for selecting the next task to execute within a tasks with the same priority.
+ * `RandomScheduler` can be seeded for reproducibility.
+ *
+ * @see MixedScheduler
+ * @since 3.6.4
```
tim-smart, line 252:
```suggestion
 * @since 3.7.0
```

tim-smart (changes requested):

.changeset/cyan-coats-peel.md
KhraksMamtsov: done

packages/effect/src/Scheduler.ts
KhraksMamtsov: done

mikearnaldi:
Order of execution within the same priority is semantically important, introducing non determinism at that level may lead to very hard to debug issues and is unwarranted. I think if you want to introduce non-determinism you can do so by yielding to a random priority instead of processing with a random order

KhraksMamtsov:
```
[1,[a,b,c]]
[2,[d,e,f]]

[c-b-a, e-d-f] // respect priority but not order inside priority
[d-e-f, a-b-c] // respect order inside priority but no priority itself
```
in the first case, the semantics of priorities are preserved.
and in the second case, the priorities can simply be called a group of tasks.
Are you sure about this?

mikearnaldi:
```
I believe that usage RandomScheduler will highlight this kind of bugs.
And we shouldn't rely on execution order within one priority and use correct concurrent primitives correctly)
It is assumed that it will be used in testing
```

No, order of execution has to be guaranteed otherwise semantics are broken. That's a required invariant for Effect to work, we have to rely on order of execution.

KhraksMamtsov:
Ok, It's clear, thank you

KhraksMamtsov:
> No, order of execution has to be guaranteed otherwise semantics are broken. That's a required invariant for Effect to work, we have to rely on order of execution. Both priority and order within priority.

I think it should be reflected somewhere in the documentation.

KhraksMamtsov:
Also I am glad that another bug was discovered and fixed during the development of this PR)

mikearnaldi:
Which bug? Re docs we can add it to the scheduler docs, it's fairly low level to make schedulers manually and there can be a degree of freedom e.g. one may decide to execute random within a specific priority that is not used elsewhere. Will try to find a good explainer

KhraksMamtsov:
@mikearnaldi 
https://discord.com/channels/795981131316985866/795983589644304396/1273026396779974807
https://github.com/Effect-TS/effect/pull/3452

mikearnaldi:
Ah yeah indeed

PR #3457, opened 2024-08-14
Add support for obtaining raw SQL query results from the underlying SQL client

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR adds support for executing raw SQL queries with the underlying SQL client.

This is primarily useful when the SQL client returns special results for certain query types.

For example, because MySQL does not support the `RETURNING` clause, the `mysql2` client will return a [`ResultSetHeader`](https://sidorares.github.io/node-mysql2/docs/documentation/typescript-examples#resultsetheader) for `INSERT`, `UPDATE`, `DELETE`, and `TRUNCATE` operations.

To gain access to the raw results of a query, you can use the `.raw` property on the `Statement`:

```ts
import * as SqlClient from "@effect/sql/SqlClient"
import * as MysqlClient from "@effect/sql/MysqlClient"

const DatabaseLive = MysqlClient.layer({
  database: Config.succeed("database"),
  username: Config.succeed("root"),
  password: Config.succeed(Redacted.make("password")),
})

  const sql = yield* SqlClient.SqlClient

  const result = yield* sql`INSERT INTO usernames VALUES ("Bob")`.raw

  console.log(result)
  /**
   * ResultSetHeader {
   *   fieldCount: 0,
   *   affectedRows: 1,
   *   insertId: 0,
   *   info: '',
   *   serverStatus: 2,
   *   warningStatus: 0,
   *   changedRows: 0
   * }
   */
})

program.pipe(
  Effect.provide(DatabaseLive),
  Effect.runPromise
)
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/sql-mysql2/src/MysqlClient.ts
+++ b/packages/sql-mysql2/src/MysqlClient.ts
@@ -89,7 +89,7 @@ export const make = (
     class ConnectionImpl implements Connection {
       constructor(private readonly conn: Mysql.PoolConnection | Mysql.Pool) {}
 
-      private runUnprepared(
+      private runRaw(
         sql: string,
         values?: ReadonlyArray<any>,
         rowsAsArray = false,
@@ -117,7 +117,7 @@ export const make = (
         rowsAsArray = false,
         method: "execute" | "query" = "execute"
       ) {
-        return this.runUnprepared(sql, values, rowsAsArray, method).pipe(
+        return this.runRaw(sql, values, rowsAsArray, method).pipe(
           Effect.map((results) => {
             if (transform && !rowsAsArray && options.transformResultNames) {
               return transformRows(results as ReadonlyArray<any>)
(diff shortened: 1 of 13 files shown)


packages/sql-mysql2/src/MysqlClient.ts
```diff
@@ -89,40 +89,57 @@ export const make = (
     class ConnectionImpl implements Connection {
       constructor(private readonly conn: Mysql.PoolConnection | Mysql.Pool) {}
 
-      private run(
+      private runUnprepared(
```
tim-smart, line 92:
```suggestion
      private runRaw(
```

packages/sql/src/internal/statement.ts
```diff
@@ -130,6 +130,13 @@ export class StatementPrimitive<A> extends Effectable.Class<ReadonlyArray<A>, Er
     )
   }
 
+  get raw(): Effect.Effect<unknown, Error.SqlError> {
+    return this.withConnection(
+      "executeRaw",
+      (connection, sql, params) => connection.executeRaw(sql, params)
+    )
+  }
```
tim-smart, line 138:
```suggestion
  get raw(): Effect.Effect<unknown, Error.SqlError> {
    return this.withConnection(
      "executeRaw",
      (connection, sql, params) => connection.executeRaw(sql, params),
      true
    )
  }
```

Remove the transformations too.

tim-smart (changes requested):

PR #3458, opened 2024-08-14
ensure makeZonedFromString supports 00:00 offset

thomasvargiu:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently `effect/DateTime::makeZonedFromString()` doesn't parse dates with `00:00` offset.
Fixed in order to allow zero offset.


tim-smart (approved):
Thanks :)

PR #3464, opened 2024-08-16
Lock down workflow permissions

fubhy:
Default all workflows to "no permissions" and enable speicifc permissions on the job level.


IMax153 (approved):
LGTM 👍

PR #3466, opened 2024-08-16
fix: respect the default for a text prompt

### LINKED ISSUE #3465: Passing a default value to `Prompt.text` from @effect/cli has no effect.

"@effect/cli@0.41.0","@effect/platform@0.62.0","@effect/platform-node@0.57.0","effect@3.6.4"

https://stackblitz.com/edit/stackblitz-starters-m1qrud?file=index.ts

The default value passed in as the `TextOptions` argument on a `Prompt.text` call will be respected.

### What do you see instead?

There is no default value set.

### Additional information

_No response_

cdierkens:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR respects the default value passed in when creating a 

### Example
```ts
const prompt = Prompt.text({
  message: "This is a prompt with a default value",
  default: "default-value", // <-- This value does not populate before this PR
});
```

> [!WARNING]  
> I attempted to model the `Wizard` tests, but could not get the `Enter` key to effect the program exit.

## Related

- Closes #3465

### DIFF
--- a/packages/cli/src/internal/prompt/text.ts
+++ b/packages/cli/src/internal/prompt/text.ts
@@ -5,7 +5,6 @@ import * as Optimize from "@effect/printer/Optimize"
 import * as Arr from "effect/Array"
 import * as Effect from "effect/Effect"
 import * as Option from "effect/Option"
-import * as Predicate from "effect/Predicate"
 import * as Redacted from "effect/Redacted"
 import type * as Prompt from "../../Prompt.js"
 import * as InternalPrompt from "../prompt.js"
@@ -233,10 +232,11 @@ function handleProcess(options: Options) {
       }
       case "enter":
       case "return": {
-        return Effect.match(options.validate(state.value), {
+        const value = state.value.length > 0 ? state.value : options.default
+        return Effect.match(options.validate(value), {
           onFailure: (error) =>
             Action.NextFrame({
-              state: { ...state, error: Option.some(error) }
+              state: { ...state, value, error: Option.some(error) }
             }),
           onSuccess: (value) => Action.Submit({ value })
         })
@@ -266,17 +266,7 @@ function basePrompt(
     ...options
   }
 
-  const state: State = {
-    ...initialState,
-    ...(Predicate.isString(options.default) ?
-      {
-        value: options.default,
-        cursor: options.default.length
-      } :
-      undefined)
-  }
-
-  return InternalPrompt.custom(state, {
+  return InternalPrompt.custom(initialState, {
     render: handleRender(opts),
     process: handleProcess(opts),
     clear: handleClear(opts)
--- a/packages/cli/src/internal/prompt/text.ts
+++ b/packages/cli/src/internal/prompt/text.ts
@@ -5,6 +5,7 @@ import * as Optimize from "@effect/printer/Optimize"
 import * as Arr from "effect/Array"
 import * as Effect from "effect/Effect"
 import * as Option from "effect/Option"
+import * as Predicate from "effect/Predicate"
 import * as Redacted from "effect/Redacted"
 import type * as Prompt from "../../Prompt.js"
 import * as InternalPrompt from "../prompt.js"
@@ -264,7 +265,18 @@ function basePrompt(
     validate: Effect.succeed,
     ...options
   }
-  return InternalPrompt.custom(initialState, {
+
+  const state: State = {
+    ...initialState,
+    ...(Predicate.isString(options.default) ?
+      {
+        value: options.default,
+        cursor: options.default.length
+      } :
+      undefined)
+  }
+
+  return InternalPrompt.custom(state, {
     render: handleRender(opts),
     process: handleProcess(opts),
     clear: handleClear(opts)
--- a/packages/cli/test/Prompt.test.ts
+++ b/packages/cli/test/Prompt.test.ts
@@ -1,21 +1,15 @@
 import * as Prompt from "@effect/cli/Prompt"
 import * as MockTerminal from "@effect/cli/test/services/MockTerminal"
-import type { Terminal } from "@effect/platform"
-import { Effect } from "effect"
+import type { Terminal } from "@effect/platform/Terminal"
+import * as Effect from "effect/Effect"
 import * as Fiber from "effect/Fiber"
-import * as Layer from "effect/Layer"
 import { describe, expect, it } from "vitest"
 
 
-const MainLive = Effect.gen(function*(_) {
-  return Layer.mergeAll(
-    MockTerminal.layer,
-  )
-}).pipe(Layer.unwrapEffect)
 
 const runEffect = <E, A>(
-  self: Effect.Effect<A, E, Terminal.Terminal>
-): Promise<A> => Effect.provide(self, MainLive).pipe(Effect.runPromise)
+  self: Effect.Effect<A, E, Terminal>
+): Promise<A> => Effect.provide(self, MockTerminal.layer).pipe(Effect.runPromise)
 
 describe("Prompt", () => {
   describe("text", () => {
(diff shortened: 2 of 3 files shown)


packages/cli/src/internal/prompt/text.ts
```diff
@@ -264,7 +265,18 @@ function basePrompt(
...
+
+  const state: State = {
+    ...initialState,
+    ...(Predicate.isString(options.default) ?
+      {
+        value: options.default,
+        cursor: options.default.length
+      } :
+      undefined)
+  }
```
IMax153, line 277: @cdierkens - I don't think we actually need to modify the initial state here.

I think the behavior we are looking for is to have the `default` value be used in place of the `state.value` if `state.value` is empty when the user actually submits the prompt.

Since we have access to the `default` value in the `opts` that are passed to the prompt processing functions, I think we can just make use of that instead.

I think the only change we need to make is in `handleProcess`, to use the `default` if `state.value` is empty:

```diff
      case "enter":
      case "return": {
+        const value = state.value.length > 0 ? state.value : options.default
-        return Effect.match(options.validate(state.value), {
+        return Effect.match(options.validate(value), {
          onFailure: (error) =>
            Action.NextFrame({
              state: { ...state, error: Option.some(error) }
            }),
          onSuccess: (value) => Action.Submit({ value })
        })
      }
```

packages/cli/test/Prompt.test.ts
```diff
@@ -0,0 +1,84 @@
...
+          (value) =>
+            Console.log(
+              `Prompt value: "${value}"`
+            )
+        )
+
+        const cli = Command.run(command, {
+          name: "Default Value App",
+          version: "0.0.1"
+        })
```
IMax153, line 45: Because a `Prompt` is a subtype of `Effect`, you should be able to avoid spinning up a whole CLI application here.

You can just directly `Effect.fork(prompt)`.

packages/cli/test/Prompt.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        const fiber = yield* _(Effect.fork(cli([])))
+        yield* _(MockTerminal.inputKey("Enter"))
```
IMax153, line 48: I'm not 100% sure but I believe the key name should be `enter` (not `Enter`).

packages/cli/test/Prompt.test.ts
```diff
@@ -0,0 +1,84 @@
...
+}).pipe(Layer.unwrapEffect)
+
+const runEffect = <E, A>(
+  self: Effect.Effect<A, E, CliApp.CliApp.Environment>
+): Promise<A> => Effect.provide(self, MainLive).pipe(Effect.runPromise)
+
+describe("Prompt", () => {
+  describe("text", () => {
+    it("should return an empty string when `default` on `Prompt.TextOptions` is not provided", () =>
+      Effect.gen(function*(_) {
```
IMax153, line 28:
```suggestion
      Effect.gen(function*() {
```

Given that the adapter in `Effect.gen` is no longer required, we are trying to avoid introducing new code into the codebase that makes use of it.

packages/cli/test/Prompt.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  )
+}).pipe(Layer.unwrapEffect)
+
+const runEffect = <E, A>(
+  self: Effect.Effect<A, E, CliApp.CliApp.Environment>
+): Promise<A> => Effect.provide(self, MainLive).pipe(Effect.runPromise)
+
+describe("Prompt", () => {
+  describe("text", () => {
+    it("should return an empty string when `default` on `Prompt.TextOptions` is not provided", () =>
```
IMax153, line 27: I think the behavior we're testing here is:

```
should use the prompt value when no default is provided
```

IMax153 (changes requested):

packages/cli/src/internal/prompt/text.ts
cdierkens: I've updated with your suggestion.  It's a nice bite size and testable increment of value. 

I was hoping to follow up with another PR to render the default value during the prompt. Here's an example from `create-remix`

https://github.com/user-attachments/assets/5c224f4a-5026-4acb-a3bd-8971b15bfa64

What's happening here is that the initial state is empty, so the default is rendered in a grayed out manner. You can tab to accept the default value. Once you delete the state value again, you are re-presented with the grayed out default.

I can follow up with an update to `renderInput` where the default value is rendered in gray when the state's value is empty. Let me know your thoughts!

packages/cli/test/Prompt.test.ts
cdierkens: Perfect. Thanks for the tip. I was able to greatly simplify the tests.

packages/cli/test/Prompt.test.ts
cdierkens: That was it, lowercase `enter`. 🙏

packages/cli/test/Prompt.test.ts
cdierkens: I matched to phrasing for both tests to be more behavior driven.

packages/cli/src/internal/prompt/text.ts
IMax153: That sounds good to me - I can help with the finer points of that PR as well if you like 👍

packages/cli/test/Prompt.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import * as Fiber from "effect/Fiber"
+import * as Layer from "effect/Layer"
+import { describe, expect, it } from "vitest"
+
+
+const MainLive = Effect.gen(function*(_) {
+  return Layer.mergeAll(
+    MockTerminal.layer,
+  )
+}).pipe(Layer.unwrapEffect)
```
IMax153, line 14: Given that `MockTerminal.layer` is already a `Layer` and we don't have any other layers to merge, there is no real reason to use `Effect.gen` here or `Layer.mergeAll`.

```suggestion
const MainLive = MockTerminal.layer
```

packages/cli/test/Prompt.test.ts
```diff
@@ -0,0 +1,49 @@
+import * as Prompt from "@effect/cli/Prompt"
+import * as MockTerminal from "@effect/cli/test/services/MockTerminal"
+import type { Terminal } from "@effect/platform"
+import { Effect } from "effect"
```
IMax153, line 4: To keep things consistent:

```suggestion
import * as Effect from "effect/Effect"
```

packages/cli/test/Prompt.test.ts
```diff
@@ -0,0 +1,49 @@
+import * as Prompt from "@effect/cli/Prompt"
+import * as MockTerminal from "@effect/cli/test/services/MockTerminal"
+import type { Terminal } from "@effect/platform"
```
IMax153, line 3:
```suggestion
import type { Terminal } from "@effect/platform/Terminal"
```

To keep imports consistent.

IMax153 (changes requested):
@cdierkens - just a few tiny additional comments. Otherwise everything looks good. Once you make these additional changes, I'll trigger CI and as long as everything passes we can get this merged 👍

IMax153 (approved):

IMax153:
@cdierkens - looks like there are some linting issues. If you run `pnpm lint-fix` it should take care of those issues.

cdierkens:
> @cdierkens - looks like there are some linting issues. If you run `pnpm lint-fix` it should take care of those issues.

@IMax153 Done and done.

IMax153 (approved):

PR #3474, opened 2024-08-18
Add support for incrementing / decrementing a gauge

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

For context, take a look at the detailed discussion that was had within ZIO regarding this change https://github.com/zio/zio/pull/7284.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Metric.ts
+++ b/packages/effect/src/Metric.ts
@@ -45,7 +45,7 @@ export type MetricTypeId = typeof MetricTypeId
  *   - Histograms
  *   - Summaries
  *
- * @since 3.6.5
+ * @since 2.0.0
  * @category models
  */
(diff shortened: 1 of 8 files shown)


packages/effect/src/Metric.ts
```diff
@@ -45,7 +45,7 @@ export type MetricTypeId = typeof MetricTypeId
  *   - Histograms
  *   - Summaries
  *
- * @since 2.0.0
+ * @since 3.6.5
```
tim-smart, line 48: The since tags would stay unchanged, as they reflect when something was added rather than modified.

IMax153: @tim-smart - reverted the since tags that did not reflect additions.

tim-smart (approved):

PR #3486, opened 2024-08-19
Introduce `bindAll`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue  #
- Closes https://github.com/Effect-TS/effect/issues/3484

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -595,22 +595,17 @@ export declare namespace All {
   export type ExtractMode<A> = [A] extends [{ mode: infer M }] ? M : "default"
 
   /**
-   * @since 3.7.0
-   */
-  export type Options = {
-    readonly concurrency?: Concurrency | undefined
-    readonly batching?: boolean | "inherit" | undefined
-    readonly discard?: boolean | undefined
-    readonly mode?: "default" | "validate" | "either" | undefined
-    readonly concurrentFinalizers?: boolean | undefined
-  }
-
-  /**
    * @since 2.0.0
    */
   export type Return<
     Arg extends Iterable<EffectAny> | Record<string, EffectAny>,
-    O extends Options
+    O extends {
+      readonly concurrency?: Concurrency | undefined
+      readonly batching?: boolean | "inherit" | undefined
+      readonly discard?: boolean | undefined
+      readonly mode?: "default" | "validate" | "either" | undefined
+      readonly concurrentFinalizers?: boolean | undefined
+    }
   > = [Arg] extends [ReadonlyArray<EffectAny>] ? ReturnTuple<Arg, IsDiscard<O>, ExtractMode<O>>
     : [Arg] extends [Iterable<EffectAny>] ? ReturnIterable<Arg, IsDiscard<O>, ExtractMode<O>>
     : [Arg] extends [Record<string, EffectAny>] ? ReturnObject<Arg, IsDiscard<O>, ExtractMode<O>>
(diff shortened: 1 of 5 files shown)


gcanti:
is `bindAll` truly necessary when nesting can achieve pretty much the same result without introducing new APIs?

```ts
Effect.Do.pipe(
  Effect.bind("x", () => Effect.succeed(2)),
  Effect.bind("y", ({ x }) =>
    Effect.all({
      a: Effect.succeed(x),
      b: Effect.succeed("ops")
    }, { mode: "either" }))
)
```

vs

```ts
Effect.Do.pipe(
  Effect.bind("x", () => Effect.succeed(2)),
  Effect.bindAll(({ x }) => ({
    a: Effect.succeed(x),
    b: Effect.succeed("ops")
  }), { mode: "either" })
)
```

tim-smart:
> is `bindAll` truly necessary when nesting can achieve pretty much the same result without introducing new APIs?

With `Effect.all` you will need to follow that up with merging `y` into the result. So the equivalent would be:

```ts
Effect.Do.pipe(
  Effect.bind("x", () => Effect.succeed(2)),
  Effect.flatMap((value) =>
    Effect.all(
      {
        a: Effect.succeed(value.x),
        b: Effect.succeed("ops"),
      },
      { concurrency: "unbounded" },
    ).pipe(Effect.map((result) => ({ ...value, ...result }))),
  ),
)
```

KhraksMamtsov:
On the one hand, I agree with Guilio. On the other hand, I saw that Michael approved of the idea. I had time to contribute to the effect and practice with the types)

packages/effect/src/Effect.ts
```diff
@@ -594,18 +594,23 @@ export declare namespace All {
...
+    readonly mode?: "default" | "validate" | "either" | undefined
+    readonly concurrentFinalizers?: boolean | undefined
+  }
+
   /**
    * @since 2.0.0
    */
   export type Return<
     Arg extends Iterable<EffectAny> | Record<string, EffectAny>,
-    O extends {
```
tim-smart, line 602: I think this should be reverted, and then remove the `discard` option.

tim-smart (changes requested):

packages/effect/src/Effect.ts
mikearnaldi: we also need the "concurrentFinalizers" option

KhraksMamtsov:
@tim-smart, please confirm that everything is fine

packages/effect/src/Effect.ts
```diff
@@ -3540,6 +3540,91 @@ export const bind: {
...
+ *   Effect.Do,
+ *   Effect.bind("x", () => Effect.succeed(2)),
+ *   Effect.bindAll(({ x }) => ({
+ *     a: Effect.succeed(x),
+ *     b: Effect.fail('ops'),
+ *   }), { concurrency: 2, mode: 'either' })
+ * )
+ * assert.deepStrictEqual(Effect.runSync(result), { x: 2, a: Either.right(2), b: Either.left('ops') })
+ *
+ * @see All.Options
```
tim-smart, line 3561: Can remove this

packages/effect/src/Effect.ts
```diff
@@ -3540,6 +3540,91 @@ export const bind: {
...
+  >(
+    f: (a: A) => [Extract<keyof X, keyof A>] extends [never] ? X : `Duplicate keys`,
+    options?: undefined | O
+  ): <E1, R1>(
+    self: Effect<A, E1, R1>
+  ) => Effect<
+    {
+      [K in keyof X | keyof A]: K extends keyof A ? A[K] :
+        K extends keyof Effect.Success<
+          All.ReturnObject<X, All.IsDiscard<O>, All.ExtractMode<O>>
```
tim-smart, line 3584: You can remove the discard checks here.

tim-smart (changes requested):

tim-smart (approved):

PR #3491, opened 2024-08-20
preserve `Array.replace` `Array.replaceOption` non emptiness

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Array.ts
+++ b/packages/effect/src/Array.ts
@@ -1055,8 +1055,17 @@ export const insertAt: {
  * @since 2.0.0
  */
 export const replace: {
-  <B>(i: number, b: B): <A>(self: Iterable<A>) => Array<A | B>
-  <A, B>(self: Iterable<A>, i: number, b: B): Array<A | B>
+  <B>(
+    i: number,
+    b: B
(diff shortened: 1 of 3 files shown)


tim-smart (commented):
Thanks. Could you add some dtslint tests?

packages/effect/src/Array.ts
```diff
@@ -1055,8 +1055,17 @@ export const insertAt: {
  * @since 2.0.0
  */
 export const replace: {
-  <B>(i: number, b: B): <A>(self: Iterable<A>) => Array<A | B>
-  <A, B>(self: Iterable<A>, i: number, b: B): Array<A | B>
+  <B>(
+    i: number,
+    b: B
+  ): <A, S extends Iterable<A> = Iterable<A>>(
```
gcanti, line 1061: what's the purpose of these defaults? (`= Iterable<A>`)

KhraksMamtsov: keep backward compatibility in case of use with passing type arguments
in the previous version, there was only one required parameter

PR #3494, opened 2024-08-20
add binary support to KeyValueStore

tim-smart:


nounder:
Oh wow. Thanks @tim-smart

PR #3495, opened 2024-08-22
add HttpApi modules to /platform

tim-smart:


packages/platform-node/examples/api.ts
```diff
@@ -0,0 +1,126 @@
...
+      })),
+      ApiEndpoint.setSuccess(User),
+      ApiEndpoint.addError(Schema.String.pipe(
+        ApiSchema.asEmpty({ status: 413, decode: () => "boom" })
+      ))
+    )
+  ),
+  ApiGroup.add(
+    ApiEndpoint.post("create", "/").pipe(
+      ApiEndpoint.setPayload(ApiSchema.Multipart(Schema.Struct({
```
danielo515, line 53: Are all this supposed to set validation schemas? If that is the case (and I'm being a bit nitpicking here) reflecting that in the name will be easier to understand and more consistent with the rest of the methods of the platform package (eg setCookiesSchema or setHeadersSchema)

tim-smart: All Api* modules are schema only, and the `ApiBuilder` module is for the implementation side.

The Api naming might change to make it more obvious that everything is a schema.

kattsushi:
I'm not sure if it makes sense (im consider my self very new on effect and fp paradigms), but I would like the @effect/RPC API to eventually be as similar to this HttpApi as possible. While RPC is currently very composable, I understand that it is not as declarative as this (HttpApi) in terms of separating the blueprint or declaration (with the help of schemas) from the implementation.

Great work, congratulations! 🚀

PR #3508, opened 2024-08-28
Renders the default for all `Prompt` types that accepts `TextOptions`.

cdierkens:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR renders all `Prompt` types that accepts `TextOptions`.

1. Intially, the text is rendered in a muted fashion.
2. When the user hits the tab key the value is accepted and rendered in cyan.
3. Finally, the user can edit the entered text or hit enter to submit the text.

https://github.com/user-attachments/assets/ca381dcb-40dc-46ce-b147-fdfda15debec

### DIFF
--- a/packages/cli/src/internal/prompt/text.ts
+++ b/packages/cli/src/internal/prompt/text.ts
@@ -66,17 +66,19 @@ function renderInput(nextState: State, options: Options, submitted: boolean) {
   const text = getValue(nextState, options)
 
   const annotation = Option.match(nextState.error, {
-    onNone: () => Match.value(submitted)
-    .pipe(
-      Match.when(true, () => Ansi.green),
-      Match.orElse(() => Match.value(nextState.value).pipe(
-        Match.when('', () => Ansi.blackBright),
-        Match.orElse(() => Ansi.combine(Ansi.underlined, Ansi.cyanBright))
-      ))
-    ),
+    onNone: () =>
+      Match.value(submitted)
+        .pipe(
+          Match.when(true, () => Ansi.green),
+          Match.orElse(() =>
+            Match.value(nextState.value).pipe(
+              Match.when("", () => Ansi.blackBright),
+              Match.orElse(() => Ansi.combine(Ansi.underlined, Ansi.cyanBright))
+            )
+          )
+        ),
     onSome: () => Ansi.red
   })
-
 
   switch (options.type) {
     case "hidden": {
@@ -259,7 +261,7 @@ function handleProcess(options: Options) {
       }
       case "enter":
       case "return": {
-        const value = getValue(state, options)   
+        const value = getValue(state, options)
         return Effect.match(options.validate(value), {
           onFailure: (error) =>
             Action.NextFrame({
@@ -269,7 +271,7 @@ function handleProcess(options: Options) {
         })
       }
       case "tab": {
-        return processTab(state, options);
+        return processTab(state, options)
       }
       default: {
         const value = Option.getOrElse(input.input, () => "")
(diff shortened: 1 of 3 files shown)


packages/cli/src/internal/prompt/text.ts
```diff
@@ -58,19 +63,32 @@ function renderClearScreen(state: State, options: Options) {
 }
 
 function renderInput(nextState: State, options: Options, submitted: boolean) {
+  const text = getValue(nextState, options)
+
   const annotation = Option.match(nextState.error, {
-    onNone: () => submitted ? Ansi.white : Ansi.combine(Ansi.underlined, Ansi.cyanBright),
+    onNone: () =>
+      Match.value(submitted)
```
IMax153, line 70: I would remove the usage of `Match` here - it is just adding unnecessary overhead and we try to keep the internals of the libs as fast as possible.

packages/cli/src/internal/prompt/text.ts
```diff
@@ -58,19 +63,32 @@ function renderClearScreen(state: State, options: Options) {
...
+        .pipe(
+          Match.when(true, () => Ansi.green),
```
IMax153, line 72: I don't think we should change the rendering of a submitted value. This should stay white as it was before.

packages/cli/test/Prompt.test.ts
```diff
@@ -37,5 +55,295 @@ describe("Prompt", () => {
...
       }).pipe(runEffect))
+
+    it("should render the default value when the default is provided", () =>
+      Effect.gen(function*() {
+        const prompt = Prompt.text({
+          message: "Test Prompt",
+          default: "default-value"
+        })
+
+        const cli = Command.prompt("test-command", prompt, () => Effect.void).pipe(Command.run({
```
IMax153, line 66: Similar to the prior PR, I let's remove the CLI in favor of running the `Prompt` directly.

IMax153 (changes requested):

packages/cli/src/internal/prompt/text.ts
cdierkens: Sounds good!

packages/cli/src/internal/prompt/text.ts
cdierkens: Whoops, did not mean to change the color. Good catch.

packages/cli/test/Prompt.test.ts
cdierkens: How would I get the terminal lines out without creating a cli?

IMax153: Since MockTerminal uses Console you should still be able to use the same APIs

cdierkens: I'll give it a try. There was a type error when I previously tried it, with just adding in the MockConsole but I just did what worked and moved on.

packages/cli/test/Prompt.test.ts
```diff
@@ -1,13 +1,30 @@
...
+import * as Doc from "@effect/printer-ansi/AnsiDoc"
+import * as Console from "effect/Console"
 import * as Effect from "effect/Effect"
 import * as Fiber from "effect/Fiber"
+import * as Layer from "effect/Layer"
+import * as Redacted from "effect/Redacted"
 import { describe, expect, it } from "vitest"
 
+const MainLive = Effect.gen(function*(_) {
+  const console = yield* _(MockConsole.make)
```
IMax153, line 16:
```suggestion
const MainLive = Effect.gen(function*() {
  const console = yield* MockConsole.make
```

IMax153 (changes requested):
Sorry - I missed this on the first pass. After this small change this PR looks good to go. I'll trigger CI to make sure now.

cdierkens:
> Sorry - I missed this on the first pass. After this small change this PR looks good to go. I'll trigger CI to make sure now.

No worries.

IMax153 (approved):

Issue #3500, opened 2024-08-25
DX improvement: Support `yield*` in even more places: Fibers, Deferreds, ...


schickling:
### What is the problem this feature would solve?

Using the Effect generator syntax is really nice as some data types are "yieldable" however even more data types could be yieldable.

### What is the feature you are proposing to solve the problem?

DX improvement: Support `yield*` in even more places: Fibers, Deferreds, ...

mikearnaldi:
clarification: make Fiber, Ref, Deferred, etc subtypes of Effect

KhraksMamtsov:
@mikearnaldi , What subtype should `Fiber` be?
|||semantic|
|--|--|--|
| `Fiber<A, E>` | `Effect<A, E>` | `Fiber.join`
| `Fiber<A, E>` | `Effect<Exit<A, E>>` | `Fiber.await`

```typescript
// Fiber.join semantic
const value = yield* joinFiber
// Fiber.await
const exit = yield* awaitFiber 
const value = yield* exit
```
the first case can be expressed by the second one , but not vice versa

KhraksMamtsov:
Are there any other entities that can be affected? Resource, Pool, PubSub? or is there no need for it yet?

fubhy:
Closing this as implemented. Current Effect types support `yield*` for the requested core cases including Fiber, Deferred, Ref, RcRef, and SubscriptionRef/SynchronizedRef.

PR #3511, opened 2024-08-28
Made `Ref`, `SynchronizedRef`,`RcRef, and `SubscriptionRef` a subtype of `Effect`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue https://github.com/Effect-TS/effect/issues/3500
- Closes #

### DIFF
--- a/packages/effect/src/internal/subscriptionRef.ts
+++ b/packages/effect/src/internal/subscriptionRef.ts
@@ -27,8 +27,8 @@ const subscriptionRefVariance = {
 
 /** @internal */
 class SubscriptionRefImpl<in out A> extends Effectable.Class<A> implements SubscriptionRef.SubscriptionRef<A> {
-  readonly [Readable.TypeId]: Readable.TypeId
-  readonly [Subscribable.TypeId]: Subscribable.TypeId
+  readonly [Readable.TypeId]: Readable.TypeId = Readable.TypeId
+  readonly [Subscribable.TypeId]: Subscribable.TypeId = Subscribable.TypeId
   readonly [Ref.RefTypeId] = _ref.refVariance
   readonly [Synchronized.SynchronizedRefTypeId] = _circular.synchronizedVariance
   readonly [SubscriptionRefTypeId] = subscriptionRefVariance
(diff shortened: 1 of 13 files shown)


tim-smart:
I wonder if making `Readable` a sub-type of Effect is more appropriate.

KhraksMamtsov:
If I make `Readable` a subtype of `Effect` this error arise:
image
Readable looks more like mixin or trait - and only finalized types should be Effect's subtypes

Also where is problem with `Unify` - now it is designed to work with covariant types
But since `*Ref` has an invariant type parameter, this does not work
```ts
type Test = Unify<Ref<1> | Ref<"a">> // Ref<1> | Ref<"a"> | Ref<1 | "a">
```
because `Unify` uses ` ... | Exclude<X, Z> | ...` and `Exclude<Ref<in out 1> | Ref<in out 'a'>, Ref<in out 1 | 'a'>` takes no effect

KhraksMamtsov:
I figured out with Unify module - all invariant types must remain as they are
add dtslint tests

KhraksMamtsov:
@tim-smart Could you please tell me if this is a move in the right direction? or is the correct implementation much more complicated than this approach? Thank you

tim-smart:
I'm still not sure if I like having `Ref`'s a subtype of Effect, it feels a bit strange. If others like it I'm happy to be veto'ed though.

mikearnaldi:
> I'm still not sure if I like having `Ref`'s a subtype of Effect, it feels a bit strange. If others like it I'm happy to be veto'ed though.

kind of goes in the same direction of having Option be an Effect

mikearnaldi:
I suppose we can accept for now and if we don't like it revise in 4.0

KhraksMamtsov:
> kind of goes in the same direction of having Option be an Effect

yes, after turning Option and Either into an Effect, the other subtypes don't look so wild, but even more ergonomic. It seems to be the next logical step.

packages/effect/src/internal/subscriptionRef.ts
```diff
@@ -37,17 +37,18 @@ class SubscriptionRefImpl<in out A> implements SubscriptionRef.SubscriptionRef<A
     readonly pubsub: PubSub.PubSub<A>,
     readonly semaphore: Effect.Semaphore
   ) {
+    super()
     this[Readable.TypeId] = Readable.TypeId
     this[Subscribable.TypeId] = Subscribable.TypeId
-    this.get = Ref.get(this.ref)
```
tim-smart, line 42: I think internally we should still use the module APIs.

tim-smart (changes requested):

packages/effect/src/internal/subscriptionRef.ts
KhraksMamtsov: Done

tim-smart (approved):

PR #3514, opened 2024-08-29
Add a Config constructor : `Config.stringNonEmpty `

### LINKED ISSUE #3515: Add a Config.nonEmptyString constructor

Most of the time, empty string is not a valid configuration, it often comes from a .env file not filled in by the user (file that comes from .env.example or something)

```env
# Redirect the emails to this user
EMAIL_REDIRECT=

DB_URL=
```

Some other packages (like https://env.t3.gg/docs/core) have an option to treat empty string as undefined.

I believe this case is common enough to be a part of the Effect repo.  

Add a new constructor:

```ts
const dbUrl = yield* Config. stringNonEmpty("DB_URL");
```

Making my own function like the one @tim-smart did:

```ts
import { Config, ConfigError, Either, Redacted } from "effect"

export const nonEmptyString = (name: string) =>
  Config.string(name).pipe(
    Config.mapOrFail(_ => {
      const trimmed = _.trim()
      return trimmed !== ""
        ? Either.right(trimmed)
        : Either.left(ConfigError.MissingData([], "must not be empty"))
    }),
  )

```

titouancreach:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added a new constructor to config.

## Related

- Related Issue #3515
- Closes #3515

### DIFF
--- a/packages/effect/src/Config.ts
+++ b/packages/effect/src/Config.ts
@@ -370,7 +370,7 @@ export const string: (name?: string) => Config<string> = internal.string
 /**
  * Constructs a config for a non-empty string value.
  *
- * @since ??
+ * @since 3.7.0
  * @category constructors
  */
 export const nonEmptyString: (name?: string) => Config<string> = internal.nonEmptyString
--- a/packages/effect/src/Config.ts
+++ b/packages/effect/src/Config.ts
@@ -373,7 +373,7 @@ export const string: (name?: string) => Config<string> = internal.string
  * @since 3.7.0
  * @category constructors
  */
-export const nonEmptyString: (name?: string) => Config<string> = internal.nonEmptyString
+export const stringNonEmpty: (name?: string) => Config<string> = internal.stringNonEmpty
 
 /**
  * Constructs a config which contains the specified value.
--- a/packages/effect/src/Config.ts
+++ b/packages/effect/src/Config.ts
@@ -368,6 +368,14 @@ export const hashSet: <A>(config: Config<A>, name?: string) => Config<HashSet.Ha
 export const string: (name?: string) => Config<string> = internal.string
 
 /**
+ * Constructs a config for a non-empty string value.
+ *
+ * @since ??
+ * @category constructors
+ */
+export const nonEmptyString: (name?: string) => Config<string> = internal.nonEmptyString
+
+/**
  * Constructs a config which contains the specified value.
  *
  * @since 2.0.0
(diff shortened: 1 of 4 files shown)


packages/effect/src/Config.ts
```diff
@@ -367,6 +367,14 @@ export const hashSet: <A>(config: Config<A>, name?: string) => Config<HashSet.Ha
  */
 export const string: (name?: string) => Config<string> = internal.string
 
+/**
+ * Constructs a config for a non-empty string value.
+ *
+ * @since 3.7.0
+ * @category constructors
+ */
+export const nonEmptyString: (name?: string) => Config<string> = internal.nonEmptyString
```
tim-smart, line 376: I think it should be called `stringNonEmpty`

packages/effect/src/internal/config.ts
```diff
@@ -447,6 +447,16 @@ export const string = (name?: string): Config.Config<string> => {
   return name === undefined ? config : nested(config, name)
 }
 
+/** @internal */
+export const nonEmptyString = (name?: string): Config.Config<string> => {
+  const config = primitive(
+    "a non-empty text property",
+    Either.liftPredicate((text) => text.length > 0, () => configError.InvalidData([], "Expected a non-empty string"))
```
tim-smart, line 454: I wonder if a missing data error is better, then you can use a fallback value.

tim-smart (changes requested):

titouancreach:
Thanks for the review ! I updated the PR accordingly

tim-smart (approved):

packages/effect/src/Config.ts
gcanti: note that there's already a `Match.nonEmptyString`

tim-smart: Maybe we can consolidate at some point.

I prefer `stringNonEmpty` as the developer can start typing "string" and see all the options.

gcanti: > developer can start typing "string" and see all the options

yeah that's nice, however until we decide to make the change across the board, I think it would be better to maintain consistency

there are also `NonEmptyReadonlyArray`, `Schema.nonEmptyString`, `Schema.NonEmptyArray`, etc...

tim-smart: Yeah sure, can put it on the 4.0 list.

PR #3528, opened 2024-09-01
chore: migrate eslint to flat config

pnodet:


tim-smart (approved):
Look good to me. I'll let @fubhy our tooling guru weigh in before merging.

fubhy:
Thanks @pnodet, this is great.

If I remember correctly, this was the only missing piece before we can upgrade eslint to v9. I've added a commit to try exactly that. Let's see if it works and then it should be ready to go!

PR #3533, opened 2024-09-02
Add number round

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adds `Number.round` method to which you can pass a precision.
`* assert.deepStrictEqual(round(5,1234, 2), 5.12)`

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Number.ts
+++ b/packages/effect/src/Number.ts
@@ -502,8 +502,8 @@ export const parse = (s: string): Option<number> => {
  * @example
  * import { round } from "effect/Number"
  *
- * assert.deepStrictEqual(round(5,1234, 2), 5.12)
- * assert.deepStrictEqual(nextPow2(17), 32)
+ * assert.deepStrictEqual(round(1.1234, 2), 1.12)
+ * assert.deepStrictEqual(round(1.567, 2), 1.57)
  *
  * @category math
  * @since 3.8.0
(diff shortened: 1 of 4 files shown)


vinassefranche:
I'm trying to remove lodash from my project in favor of native node features and effect features and this one was missing.
As the implementation was already in the repository, I thought it could be a good idea to export it.

packages/effect/src/Number.ts
```diff
@@ -492,3 +492,26 @@ export const parse = (s: string): Option<number> => {
...
+ * Returns the number rounded with the given precision.
+ *
+ * @param self - The number to round
+ * @param precision - The precision
+ *
+ * @example
+ * import { round } from "effect/Number"
+ *
+ * assert.deepStrictEqual(round(5,1234, 2), 5.12)
+ * assert.deepStrictEqual(nextPow2(17), 32)
```
tim-smart, line 506: This can be removed.

tim-smart (changes requested):

packages/effect/src/Number.ts
vinassefranche: Sorry for that. I started to write the doc and stopped to write the tests first and forgot to come back to it.
There was also a mistake on the first line, by the way

tim-smart (approved):

PR #3540, opened 2024-09-03
add Logger.withLeveledConsole

Schniz:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In browsers and different platforms, `console.error` renders differently than `console.info`. This helps to distinguish between different levels of logging. `Logger.withLeveledConsole` takes any logger and calls the respective `Console` method based on the log level. For instance, `Effect.logError` will call `Console.error` and `Effect.logInfo` will call `Console.info`.

To use it, you can replace the default logger with a `Logger.withLeveledConsole` logger:

```ts
import { Logger, Effect } from "effect"

const loggerLayer = Logger.withLeveledConsole(Logger.stringLogger)

Effect.gen(function* () {
  yield* Effect.logError("an error")
  yield* Effect.logInfo("an info")
}).pipe(Effect.provide(loggerLayer))
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/internal/fiberRuntime.ts
+++ b/packages/effect/src/internal/fiberRuntime.ts
@@ -1401,6 +1401,23 @@ export const loggerWithConsoleLog = <M, O>(self: Logger<M, O>): Logger<M, void> 
   })
 
 /** @internal */
+export const loggerWithLeveledLog = <M, O>(self: Logger<M, O>): Logger<M, void> =>
+  internalLogger.makeLogger((opts) => {
+    const services = FiberRefs.getOrDefault(opts.context, defaultServices.currentServices)
+    const unsafeLogger = Context.get(services, consoleTag).unsafe
+    const levels: Partial<Record<LogLevel.LogLevel["_tag"], (v: unknown) => void>> = {
+      All: unsafeLogger.log,
+      Info: unsafeLogger.info,
+      Debug: unsafeLogger.debug,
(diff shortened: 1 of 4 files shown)


datner:
Honestly I think this should be the default even, great idea

packages/effect/src/internal/fiberRuntime.ts
```diff
@@ -1396,6 +1396,23 @@ export const loggerWithConsoleLog = <M, O>(self: Logger<M, O>): Logger<M, void>
     Context.get(services, consoleTag).unsafe.log(self.log(opts))
   })
 
+/** @internal */
+export const loggerWithLeveledLog = <M, O>(self: Logger<M, O>): Logger<M, void> =>
+  internalLogger.makeLogger((opts) => {
+    const services = FiberRefs.getOrDefault(opts.context, defaultServices.currentServices)
+    const unsafeLogger = Context.get(services, consoleTag).unsafe
+    const levels: Partial<Record<LogLevel.LogLevel["_tag"], (v: unknown) => void>> = {
```
tim-smart, line 1404: A `switch` would be better here, instead of creating an intermediary object for every log.

tim-smart (changes requested):

tim-smart (approved):

Issue #3257, opened 2024-07-15
Suggestion: improve DX of type errors from inside `pipe` and `flow`


OliverJAsh:
### What is the problem this feature would solve?

When writing code inside `pipe` and `flow` we frequently run into scenarios where a definition is missing e.g. because it hasn't been imported yet. This can result in very confusing error messages due to TypeScript inferring `unknown` for the type parameters.

Here's a small example comparing the behaviour with and without `pipe`:

```ts
// Deliberately not defined
// declare const add: (x: number) => (n: number) => number;

declare const identity: <T>(x: T) => T;

declare function pipe<A, B, C>(a: A, ab: (a: A) => B, bc: (b: B) => C): C

// ## Arrow function with block body

// ### without pipe

// ✅ Error is clear
// ✅ Just one error
(): number => identity(add(1)(1));

// ### with pipe

// ❌ Two errors instead of one
(): number => {
  return pipe(1, add(1), identity);
};

// ## Arrow function with expression body

// ❌ Two errors instead of one
// ❌ The most important error is masked ("Cannot find name 'add'") by another
// error ("Type 'unknown' is not assignable to type 'number'.")
(): number => pipe(1, add(1), identity);
```

This is especially problematic inside large pipelines. In my experience it's not uncommon for TypeScript to report multiple errors and highlight hundreds of lines of code just because of one missing import/definition.

### What is the feature you are proposing to solve the problem?

This isn't a problem when we're not using `pipe` or `flow`:

```ts
// Deliberately not defined
// declare const add: (x: number) => (n: number) => number;

declare const identity: <T>(x: T) => T;

// ✅ Error is clear
// ✅ Just one error
(): number => identity(add(1)(1));
```

In this case, TypeScript infers the return type of `add` as `any`, so there are no further errors.

This is the DX we want for `pipe` and `flow`: there's just one error and it's clear what and where the error is.

How can we mimic this behaviour with `pipe`/`flow`? We can use default type parameters:

```diff
-declare function pipe<A, B, C>(a: A, ab: (a: A) => B, bc: (b: B) => C): C
+declare function pipe<A, B = never, C = never>(a: A, ab: (a: A) => B, bc: (b: B) => C): C
```

Complete example:

```ts
// Deliberately not defined
// declare const add: (x: number) => (n: number) => number;

declare const identity: <T>(x: T) => T;

declare function pipe<A, B = never, C = never>(a: A, ab: (a: A) => B, bc: (b: B) => C): C

[...]

PR #3549, opened 2024-09-05
Improve DX of type errors from inside `pipe` and `flow`

OliverJAsh:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

See https://github.com/Effect-TS/effect/issues/3257.

## Related

- Closes https://github.com/Effect-TS/effect/issues/3257

### DIFF
--- a/packages/effect/test/Function.test.ts
+++ b/packages/effect/test/Function.test.ts
@@ -89,6 +89,15 @@ describe("Function", () => {
   })
 
   it("pipe()", () => {
+    // We should only have one error for the missing definition.
+  (): number =>
+    pipe(
+      1,
+      // @ts-expect-error
+      add(1),
+      Function.identity,
+    );
+
     const pipe = Function.pipe // this alias is required in order to exclude the `@effect/babel-plugin` compiler and get 100% coverage
     deepStrictEqual(pipe(2), 2)
     deepStrictEqual(pipe(2, f), 3)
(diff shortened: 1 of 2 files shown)


packages/effect/test/Function.test.ts
```diff
@@ -89,6 +89,15 @@ describe("Function", () => {
   })
 
   it("pipe()", () => {
+    // We should only have one error for the missing definition.
```
gcanti, line 92: I'd move these tests to a `Function.ts` file under the dtslint folder

gcanti (changes requested):

gcanti:
@OliverJAsh I think we can apply the same trick to the `Pipeable` interface

```ts
import { identity, Option } from "effect"

// We should only have one error for the missing definition.
const _x = (): number =>
  Option.some(1).pipe(
    // @ts-expect-error
    add(1),
    identity
  )
```

datner:
Thank you for the PR!
Do we know what are the limits of this? Like, was there any casualty or is this a straight-up DX win? I'd assume the latter, but we should make sure 😄 

After adding that to the `Pipeable` interface and checking if this fails in any case (in a way that was worse than before) I think we can merge it in 💪🏻

patroza:
@OliverJAsh I made the change for pipeable interface here: https://github.com/Effect-TS/effect/pull/3731/commits/8e0907c59c8e4dc805d474c28a403f6e65407758

PR #3559, opened 2024-09-06
Add `RcMap.keys` and `MutableHashMap.keys` lookup functions

fubhy:


packages/effect/src/internal/rcMap.ts
```diff
@@ -211,3 +211,9 @@ export const get: {
     )
   }
 )
+
+/** @internal */
+export const keys = <K, A, E>(self: RcMap.RcMap<K, A, E>): Effect<Array<K>> => {
+  const impl = self as RcMapImpl<K, A, E>
+  return impl.state._tag === "Closed" ? core.interrupt : core.succeed(MutableHashMap.keys(impl.state.map))
```
tim-smart, line 218:
```suggestion
  return impl.state._tag === "Closed" ? core.interrupt : core.sync(() => MutableHashMap.keys(impl.state.map))
```

tim-smart (changes requested):

packages/effect/src/internal/rcMap.ts
tim-smart: Actually the whole thing should be wrapped with .suspend

packages/effect/src/internal/rcMap.ts
```diff
@@ -211,3 +211,9 @@ export const get: {
     )
   }
 )
+
+/** @internal */
+export const keys = <K, A, E>(self: RcMap.RcMap<K, A, E>): Effect<Array<K>> => {
+  const impl = self as RcMapImpl<K, A, E>
+  return core.suspend(() => impl.state._tag === "Closed" ? core.interrupt : core.succeed(MutableHashMap.keys(impl.state.map))
```
tim-smart, line 218:
```suggestion
  return core.suspend(() => impl.state._tag === "Closed" ? core.interrupt : core.succeed(MutableHashMap.keys(impl.state.map)))
```

tim-smart (approved):

PR #3572, opened 2024-09-08
`Deferred` is subtype of `Effect`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The `Deferred<A>` is now a subtype of `Effect<A>`. This change simplifies handling of deferred values, removing the need for explicit call `Deferred.await`.

```typescript
import { Effect, Deferred } from "effect"

Effect.gen(function* () {
  const deferred = yield* Deferred.make<string>()

  const before = yield* Deferred.await(deferred)
  const after = yield* deferred
})
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/gorgeous-toes-help.md
+++ b/.changeset/gorgeous-toes-help.md
@@ -1,0 +1,16 @@
+---
+"effect": minor
+---
+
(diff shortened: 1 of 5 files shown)


tim-smart (approved):

mikearnaldi:
@KhraksMamtsov better to rebase instead of adding merge commits

PR #3578, opened 2024-09-09
Preserve Span on Interruption

mikearnaldi:
The following code:

```ts
import { Effect, Fiber } from "effect"

  const fiber = yield* Effect.fork(
    Effect.uninterruptibleMask((restore) =>
      restore(
        Effect.sleep("1 second").pipe(
          Effect.withSpan("sleep")
        )
      ).pipe(
        Effect.tapErrorCause((cause) => Effect.logError("Got cause:", cause))
      )
    )
  )

  yield* Effect.sleep("250 millis")
  yield* Fiber.interrupt(fiber)
  yield* Fiber.await(fiber)
  yield* Effect.logInfo("Done")
})

Effect.runFork(program)
```

now prints:

```ts
michaelarnaldi@Michaels-MacBook-Pro effect % pnpm tsx scratchpad/interrupt-span.ts 
timestamp=2024-09-09T15:43:01.173Z level=ERROR fiber=#1 message="Got cause:" cause="InterruptedException: Fiber Interrupted by: #0
    at sleep (/Users/michaelarnaldi/repositories/effect/scratchpad/interrupt-span.ts:8:18)"
timestamp=2024-09-09T15:43:01.175Z level=INFO fiber=#0 message=Done
michaelarnaldi@Michaels-MacBook-Pro effect % 
```

I am not too sure this is a wanted & helpful feature, up for discussion

### DIFF
--- a/.changeset/eighty-lobsters-refuse.md
+++ b/.changeset/eighty-lobsters-refuse.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Made `Ref`, `SynchronizedRed` and `SubscriptionRef` a subtype of `Effect`
--- a/.changeset/gorgeous-toes-help.md
+++ b/.changeset/gorgeous-toes-help.md
@@ -1,0 +1,16 @@
+---
+"effect": minor
+---
+
+The `Deferred<A>` is now a subtype of `Effect<A>`. This change simplifies handling of deferred values, removing the need for explicit call `Deffer.await`.
+
+```typescript
+import { Effect, Deferred } from "effect"
+
+Effect.gen(function* () {
+  const deferred = yield* Deferred.make<string>()
+
+  const before = yield* Deferred.await(deferred)
+  const after = yield* deferred
+})
+```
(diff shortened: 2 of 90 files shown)


tim-smart:
I think it is worthwhile, debugging interrupts is painful at the moment.

mikearnaldi:
> I think it is worthwhile, debugging interrupts is painful at the moment.

it is but not sure how this helps debugging interrupts, it tells you where the program stopped and not who stopped it

tim-smart:
> it is but not sure how this helps debugging interrupts, it tells you where the program stopped and not who stopped it

I feel like we could capture a span where `.interrupt` is called.

tim-smart:
It seems that is what you did in the latest commit? We could have the origin span and the receiver span, and then maybe set the error `cause` to the origin if present.

mikearnaldi:
> It seems that is what you did in the latest commit? We could have the origin span and the receiver span, and then maybe set the error `cause` to the origin if present.

No what I did is capture where Effect.interruptible is called not where Fiber.interrupt is

fubhy:
This is still draft and now conflicts with current code. Is this behavior still intended? If so, please rebase and confirm the desired span semantics for interruptions; otherwise we can close this out.

tim-smart:

PR #3579, opened 2024-09-09
Update to TS 5.6

mikearnaldi:
Had to update `@types/node` and add `node` as type entry in the bun package to make `Buffer` a valid `UInt8Array` due to the Iterator methods that have been added in 5.6, also had to force some `!` on iterators, else all seems to work.

Notable improvements in LSP performance on files such as "core.ts"

### DIFF
--- a/packages/platform-bun/tsconfig.src.json
+++ b/packages/platform-bun/tsconfig.src.json
@@ -6,6 +6,6 @@
     "tsBuildInfoFile": ".tsbuildinfo/src.tsbuildinfo",
     "rootDir": "src",
     "outDir": "build/src",
-    "types": ["bun-types", "node"]
+    "types": ["bun-types"]
   }
 }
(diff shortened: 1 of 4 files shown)


mikearnaldi:
That's odd it actually doesn't work but apparently building twice makes the error go away

tim-smart (approved):
Everything seems to look ok 👍🏼

packages/platform-bun/tsconfig.src.json
```diff
@@ -6,6 +6,6 @@
     "tsBuildInfoFile": ".tsbuildinfo/src.tsbuildinfo",
     "rootDir": "src",
     "outDir": "build/src",
-    "types": ["bun-types"]
+    "types": ["bun-types", "node"]
```
fubhy, line 9: We should remove this again when possible.

fubhy (approved):

packages/platform-bun/tsconfig.src.json
mikearnaldi: Done, it was not needed anyway

PR #3591, opened 2024-09-11
`Dequeue<A>` and `Queue<A>` is subtype of `Effect<A>`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
`Dequeue<A>` and `Queue<A>` is subtype of `Effect<A>`. This means that now it can be used as an `Effect`, and when called, it will automatically extract and return an item from the queue, without having to explicitly use the `Queue.take` function.

```ts
Effect.gen(function* () {
  const queue = yield* Queue.unbounded<number>()
  yield* Queue.offer(queue, 1)
  yield* Queue.offer(queue, 2)
  const oldWay = yield* Queue.take(queue)
  const newWay = yield* queue
})
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/proud-cats-shake.md
+++ b/.changeset/proud-cats-shake.md
@@ -1,0 +1,15 @@
+---
+"effect": minor
+---
+
+`Dequeue<A>` and `Queue<A>` is subtype of `Effect<A>`. This means that now it can be used as an `Effect`, and when called, it will automatically extract and return an item from the queue, without having to explicitly use the `Queue.take` function.
+
+```ts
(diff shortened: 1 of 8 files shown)


tim-smart:
The tests are failing because there are some other implementations of Queue that will need updating.

KhraksMamtsov:
is there any doubt about this?

tim-smart (approved):

PR #3594, opened 2024-09-12
add Cause.Annotated

tim-smart:


packages/effect/src/internal/cause.ts
```diff
@@ -127,6 +130,48 @@ export const sequential = <E, E2>(left: Cause.Cause<E>, right: Cause.Cause<E2>):
...
 
+/** @internal */
+export const annotated = dual<
+  (context: Context.Context<never>) => <E>(self: Cause.Cause<E>) => Cause.Cause<E>,
+  <E>(self: Cause.Cause<E>, context: Context.Context<never>) => Cause.Cause<E>
+>(2, (self, context) => {
+  const o = Object.create(proto)
+  o._tag = OpCodes.OP_ANNOTATED
+  if (self._tag === OpCodes.OP_ANNOTATED) {
+    o.context = Context.merge(self.context, context)
```
tim-smart, line 141:
```suggestion
    o.context = Context.merge(context, self.context)
```

tim-smart:
@mikearnaldi This was an idea for a more flexible annotation system for Cause. Maybe something for 4.0?

mikearnaldi:
> @mikearnaldi This was an idea for a more flexible annotation system for Cause. Maybe something for 4.0?

We used to have annotations but we removed them in favor of carrying the span in a proxy due to re-fail losing the span

tim-smart:
> We used to have annotations but we removed them in favor of carrying the span in a proxy due to re-fail losing the span

This goes the route of a re-fail rehydrating the annotations.

mikearnaldi:
Will need to check the details but as far as I understand it's not really possible, spans are linked to a specific error, re-failing with a different error should not hydrate spans

tim-smart:
> Will need to check the details but as far as I understand it's not really possible, spans are linked to a specific error, re-failing with a different error should not hydrate spans

When you annotate a Die or Fail, it will also add the annotations to the error object using a Proxy (like we do now). If during a re-fail it finds existing annotations, then it will re-add them to the cause.

mikearnaldi:
> > Will need to check the details but as far as I understand it's not really possible, spans are linked to a specific error, re-failing with a different error should not hydrate spans
> 
> When you annotate a Die or Fail, it will also add the annotations to the error object using a Proxy (like we do now). If during a re-fail it finds existing annotations, then it will re-add them to the cause.

but then why annotating the cause? what's the advantage, are we planning to add other kinds of annotations and if so those would also be lost on re-fail unless proxied?

mikearnaldi:
> > Will need to check the details but as far as I understand it's not really possible, spans are linked to a specific error, re-failing with a different error should not hydrate spans
> 
> When you annotate a Die or Fail, it will also add the annotations to the error object using a Proxy (like we do now). If during a re-fail it finds existing annotations, then it will re-add them to the cause.

if it re-checks then we might even be able to not annotate / proxy, we just check if the failure is `===` to the previous and we preserve annotations

fubhy:
Is this still intended for a future release? If so, please resolve conflicts or clarify the target; otherwise we can close it as superseded/parked.

PR #3595, opened 2024-09-12
add filter refinement overloads to http client

fubhy:


tim-smart:
If you could use next-minor as the base, as the client has been refactor recently.

tim-smart (approved):

PR #3603, opened 2024-09-13
add `Duration.parts` api

fubhy:
This is useful when working with the new DateTime module (e.g. when extracting the difference between two DateTime objects as a duration).

### DIFF
--- a/.changeset/weak-moles-repair.md
+++ b/.changeset/weak-moles-repair.md
@@ -1,8 +1,8 @@
 ---
-"effect": patch
+"effect": minor
 ---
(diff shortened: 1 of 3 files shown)


.changeset/weak-moles-repair.md
```diff
@@ -0,0 +1,11 @@
+---
+"effect": patch
```
tim-smart, line 2:
```suggestion
"effect": minor
```

And update since tags.

tim-smart (changes requested):

PR #3605, opened 2024-09-13
add additional duration conversion apis

fubhy:
With the recent addition of the `DateTime` module, it's tempting to add some additional duration conversion utilities.

### DIFF
--- a/.changeset/rich-timers-fix.md
+++ b/.changeset/rich-timers-fix.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Added additional `Duration` conversion apis
--- a/.changeset/rich-timers-fix.md
+++ b/.changeset/rich-timers-fix.md
@@ -1,5 +1,11 @@
 ---
-"effect": patch
+"effect": minor
 ---
 
-Added additional `Duration` conversion apis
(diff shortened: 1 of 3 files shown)


packages/effect/test/Duration.test.ts
```diff
@@ -388,7 +388,7 @@ describe("Duration", () => {
 
   it("toSeconds", () => {
     expect(Duration.millis(1).pipe(Duration.toSeconds)).toBe(0.001)
-    expect(Duration.nanos(1n).pipe(Duration.toSeconds)).toBe(9.999999999999999e-10)
+    expect(Duration.nanos(1n).pipe(Duration.toSeconds)).toBe(1e-9)
```
fubhy, line 391: This imprecision was unnecessarily caused by the double division we performed before.

I consider this a bug fix.

.changeset/rich-timers-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
```
tim-smart, line 2:
```suggestion
"effect": minor
```

And adjust the since tags :)

tim-smart (changes requested):

PR #3606, opened 2024-09-13
Add support for date filtering (after and before) to Schema.Date.

achaconm2001:
Add support for date filtering (after and before) to `Schema.Date`.

New APIs include:

- `Schema.after`
- `Schema.afterOrEqualTo`
- `Schema.before`
- `Schema.beforeOrEqualTo`

Important note:
The "decoding" tests not working because I din't know how to fix them. They are commented out.

### DIFF
--- a/.changeset/grumpy-seas-chew.md
+++ b/.changeset/grumpy-seas-chew.md
@@ -1,0 +1,12 @@
+---
+"@effect/schema": minor
+---
+
+Add support for date filtering (after and before) to `Schema.Date`.
+
+New APIs include:
+
+- `Schema.after`
+- `Schema.afterOrEqualTo`
+- `Schema.before`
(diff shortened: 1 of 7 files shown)


gcanti (changes requested):
- changeset: you can replace minor with patch
- naming convention: `lessThan<entity>` so `lessThanDate`, `lessThanOrEqualToDate`, etc...
- no need for a `jsonSchema` annotation
- no need to add the type ids to the internal `filter.ts` file

i.e. in practice you can take inspiration from the same filters defined for `Duration`

https://github.com/Effect-TS/effect/blob/7fdf9d9aa1e2c1c125cbf87991e6efbf4abb7b07/packages/schema/src/Schema.ts#L5519-L5637

PR #3607, opened 2024-09-13
Return raw SQL query results in sql-sqlite-node

thewilkybarkid:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds support for #3457 to `sql-sqlite-node`, so you can find out if a row was inserted etc.

## Related

- Related Issue #3457
- Closes #

### DIFF
--- a/packages/sql-sqlite-node/src/SqliteClient.ts
+++ b/packages/sql-sqlite-node/src/SqliteClient.ts
@@ -118,12 +118,16 @@ export const make = (
 
       const runStatement = (
         statement: Sqlite.Statement,
-        params: ReadonlyArray<Statement.Primitive> = []
+        params: ReadonlyArray<Statement.Primitive> = [],
+        raw = false
       ) =>
         Effect.try({
           try: () => {
             if (statement.reader) {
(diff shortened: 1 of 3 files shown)


packages/sql-sqlite-node/src/SqliteClient.ts
```diff
@@ -116,26 +116,44 @@ export const make = (
...
+            return statement.run(...params)
           },
           catch: (cause) => new SqlError({ cause, message: "Failed to execute statement" })
         })
 
+      const runStatement = (
+        statement: Sqlite.Statement,
+        params: ReadonlyArray<Statement.Primitive> = []
+      ) =>
+        Effect.gen(function*() {
```
tim-smart, line 137: I would just use `Effect.map` here, or add a `raw` option to `runStatement`.

tim-smart (changes requested):

tim-smart (approved):

tim-smart:
Thanks! Just did a bit of maintenance while I was here :)

PR #3626, opened 2024-09-17
`Resource<A, E>` and `ScopedRed<A>` is subtype of `Effect`.

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Resource<A, E>` is subtype of `Effect<A, E>`.
`ScopedRed<A>` is subtype of `Effect<A>`.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/internal/resource.ts
+++ b/packages/effect/src/internal/resource.ts
@@ -4,6 +4,7 @@ import type * as Resource from "../Resource.js"
 import type * as Schedule from "../Schedule.js"
 import type * as Scope from "../Scope.js"
 import * as core from "./core.js"
+import * as effectable from "./effectable.js"
 import * as fiberRuntime from "./fiberRuntime.js"
 import * as _schedule from "./schedule.js"
 import * as scopedRef from "./scopedRef.js"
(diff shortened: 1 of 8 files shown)


packages/effect/src/internal/resource.ts
```diff
@@ -46,11 +47,18 @@ export const manual = <A, E, R>(
   core.flatMap(core.context<R>(), (env) =>
     pipe(
       scopedRef.fromAcquire(core.exit(acquire)),
-      core.map((ref) => ({
-        [ResourceTypeId]: resourceVariance,
-        scopedRef: ref,
-        acquire: core.provideContext(acquire, env)
-      }))
+      core.map((ref) => {
+        const resource = {
```
tim-smart, line 51: I think we can move the prototype out of here and use `Object.create`.

packages/effect/src/internal/scopedRef.ts
```diff
@@ -41,11 +41,12 @@ export const fromAcquire = <A, E, R>(
         core.flatMap((value) =>
           circular.makeSynchronized([newScope, value] as const).pipe(
             core.flatMap((ref) => {
-              const scopedRef: ScopedRef.ScopedRef<A> = {
-                [ScopedRefTypeId]: scopedRefVariance,
-                pipe() {
-                  return pipeArguments(this, arguments)
+              const scopedRef = {
```
tim-smart, line 44: Same as resource, move the prototype out so it is shared by all instances.

packages/effect/src/internal/resource.ts
KhraksMamtsov: sure

packages/effect/src/internal/scopedRef.ts
KhraksMamtsov: done

tim-smart (approved):

PR #3649, opened 2024-09-20
Add libSQL package

### LINKED ISSUE #3368: Effect SQL Turso Adapter

There currently is no Turso/LibSQL adapter in the Effect ecosystem, although there are adapters for other providers such as d1.

Add a Turso/LibSQL Adapter called `sql-turso` for connections with remote databases on Turso, and perhaps local LibSQL databases on my machine.

This would be similar to how Drizzle has a Turso driver in their `drizzle.config.ts` as noted here:

https://orm.drizzle.team/learn/tutorials/drizzle-with-turso#setup-drizzle-config-file

And would be added to the collection of existing adapters:

- sql-d1
- sql-mssql
- sql-mysql2
- sql-pg
- sql-sqlite-bun
- sql-sqlite-node
- sql-sqlite-react-native
- sql-sqlite-wasm

Not currently using Effect SQL Adapters, instead just wrapping queries with `Effect.promise` and `Effect.tryPromise`.

thewilkybarkid:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This change (hopefully!) adds a libSQL implementation for use with [Turso](https://turso.tech/). It should be a minimal implementation (e.g. no migrations), using the same test cases as `sql-sqlite-node`. It's not been tested beyond this, however.

## Related

- Related Issue #
- Closes #3368


packages/sql-libsql/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "codegen": "build-utils prepare-v2",
+    "build": "pnpm build-esm && pnpm build-annotate && pnpm build-cjs && build-utils pack-v2",
+    "build-esm": "tsc -b tsconfig.build.json",
+    "build-cjs": "babel build/esm --plugins @babel/transform-export-namespace-from --plugins @babel/transform-modules-commonjs --out-dir build/cjs --source-maps",
+    "build-annotate": "babel build/esm --plugins annotate-pure-calls --out-dir build/esm --source-maps",
+    "check": "tsc -b tsconfig.json",
+    "test": "vitest",
+    "coverage": "vitest --coverage"
+  },
+  "devDependencies": {
```
tim-smart, line 43: You can remove /platform here, as you don't have a Migrator module.

For migrations, do you use some tooling from turso?

tim-smart: Hmm after some research I think we need to include a Migrator module too.

thewilkybarkid: Yeah, I was just trying to add the minimum to start with.

PR #3658, opened 2024-09-23
Add `Tuple.map`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Tuple.map` transforms each element of tuple using the given function, treating tuple homomorphically

```ts
import { pipe, Tuple } from "effect"

const result = pipe(
  //  ^? [string, string, string]
  ["a", 1, false] as const,
  T.map((el) => {
    //   ^? "a" | 1 | false
    return el.toString().toUppercase()
  })
)
assert.deepStrictEqual(result, ["A", "1", "FALSE"])
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/strong-pans-flash.md
+++ b/.changeset/strong-pans-flash.md
@@ -1,0 +1,19 @@
+---
+"effect": minor
+---
+
+`Tuple.map` transforms each element of tuple using the given function, treating tuple homomorphically
+
+```ts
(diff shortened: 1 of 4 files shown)


KhraksMamtsov:
@mikearnaldi what you think about it?

PR #3661, opened 2024-09-23
`Micro.EnvRef` and `Micro.Handle` is subtype of `Micro`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
`Micro.EnvRef` and `Micro.Handle` is subtype of `Micro`
`MicroableClass` has been exported as an analogue of `Effectable.Class`

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Micro.ts
+++ b/packages/effect/src/Micro.ts
@@ -180,7 +180,10 @@ const MicroBase: MicroClass = (function() {
  * @category constructors
  */
 export abstract class Class<out A, out E = never, out R = never> extends MicroBase<A, E, R> {
-  abstract [runSymbol](env: Env<any>, onExit: (exit: MicroExit<A, E>) => void): void
+  abstract asMicro(): Micro<A, E, R>
+  [runSymbol](env: Env<any>, onExit: (exit: MicroExit<A, E>) => void): void {
+    this.asMicro()[runSymbol](env, onExit)
+  }
 }
 
 // ----------------------------------------------------------------------------
@@ -881,7 +884,7 @@ const EnvRefProto: ThisType<EnvRef<any>> = {
   ...MicroProto,
   [EnvRefTypeId]: EnvRefTypeId,
   [runSymbol](env: Env<any>, onExit: (exit: MicroExit<any, any>) => void) {
-    onExit(Either.right(envGet(env, this)))
+    getEnvRef(this)[runSymbol](env, onExit)
   }
 }
 
@@ -3742,23 +3745,6 @@ export const isHandle = (u: unknown): u is Handle<unknown, unknown> =>
   typeof u === "object" && u !== null && HandleTypeId in u
 
 class HandleImpl<A, E> extends Class<A, E> implements Handle<A, E> {
-  [runSymbol](_env: Env<any>, onExit: (exit: MicroExit<A, E>) => void) {
-    if (this._controller.signal.aborted) {
-      return onExit(exitInterrupt)
-    }
-
-    const observer = (exit: MicroExit<A, E>) => {
-      this.removeObserver(observer)
-      onExit(exit)
-    }
-
-    this._controller.signal.addEventListener("abort", () => {
-      this.removeObserver(observer)
-      onExit(exitInterrupt)
-    }, { once: true })
-
-    this.addObserver(observer)
-  }
   readonly [HandleTypeId]: HandleTypeId
 
   readonly observers: Set<(exit: MicroExit<A, E>) => void> = new Set()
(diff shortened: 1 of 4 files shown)


packages/effect/src/Micro.ts
```diff
@@ -144,6 +144,45 @@ export interface MicroIterator<T extends Micro<any, any, any>> {
...
+  Base.prototype = MicroProto
+  return Base as any
+})()
+
+/**
+ * @since 3.9.0
+ * @category constructors
+ */
+export abstract class MicroableClass<out A, out E = never, out R = never> extends MicroBase<A, E, R> {
+  abstract readonly [runSymbol]: (env: Env<any>, onExit: (exit: MicroExit<A, E>) => void) => void
```
tim-smart, line 183:
```suggestion
  abstract [runSymbol](env: Env<any>, onExit: (exit: MicroExit<A, E>) => void): void
```

packages/effect/src/Micro.ts
```diff
@@ -144,6 +144,45 @@ export interface MicroIterator<T extends Micro<any, any, any>> {
...
+  function Base() {}
+  Base.prototype = MicroProto
+  return Base as any
+})()
+
+/**
+ * @since 3.9.0
+ * @category constructors
+ */
+export abstract class MicroableClass<out A, out E = never, out R = never> extends MicroBase<A, E, R> {
```
tim-smart, line 182: I think `Class` is fine. `MicroableClass` is a bit of a mouthful,  and the usage could be:

```ts
import { Micro } from "effect"

class Thing extends Micro.Class<A, E> {
  // ...
}
```

packages/effect/src/Micro.ts
```diff
@@ -3671,7 +3741,14 @@ export interface Handle<A, E = never> {
 export const isHandle = (u: unknown): u is Handle<unknown, unknown> =>
   typeof u === "object" && u !== null && HandleTypeId in u
 
-class HandleImpl<A, E> implements Handle<A, E> {
+class HandleImpl<A, E> extends MicroableClass<A, E> implements Handle<A, E> {
+  readonly [runSymbol] = (_env: Env<any>, onExit: (exit: MicroExit<A, E>) => void) => {
```
tim-smart, line 3745:
```suggestion
  [runSymbol](_env: Env<any>, onExit: (exit: MicroExit<A, E>) => void) {
```

packages/effect/src/Micro.ts
```diff
@@ -144,6 +144,45 @@ export interface MicroIterator<T extends Micro<any, any, any>> {
   next(...args: ReadonlyArray<any>): IteratorResult<YieldWrap<T>, Micro.Success<T>>
 }
 
+/**
+ * @since 3.9.0
+ * @category models
+ */
+export interface MicroPrimitive {
```
tim-smart, line 151: I think we can call this `MicroClass`.

packages/effect/src/Micro.ts
```diff
@@ -3671,7 +3741,14 @@ export interface Handle<A, E = never> {
 export const isHandle = (u: unknown): u is Handle<unknown, unknown> =>
   typeof u === "object" && u !== null && HandleTypeId in u
 
-class HandleImpl<A, E> implements Handle<A, E> {
+class HandleImpl<A, E> extends MicroableClass<A, E> implements Handle<A, E> {
+  readonly [runSymbol] = (_env: Env<any>, onExit: (exit: MicroExit<A, E>) => void) => {
+    const observer = (exit: MicroExit<A, E>) => {
```
tim-smart, line 3746: This should handle interruption too.

packages/effect/src/Micro.ts
KhraksMamtsov: done

packages/effect/src/Micro.ts
KhraksMamtsov: done

packages/effect/src/Micro.ts
KhraksMamtsov: done

packages/effect/src/Micro.ts
KhraksMamtsov: Done, but please check it again)

tim-smart: I added an abstract `asMicro` method to the Class, as implementing the `runSymbol` function has a bunch of footguns that you have to be aware of.

KhraksMamtsov:
looks like flaky

PR #3682, opened 2024-09-25
move ManagedRuntime.TypeId to fix circular imports

leonitousconforti:
## Type

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Right now, there is two typeId definitions for `effect/ManagedRuntime`, one in layer and one in managedRuntime. If you change one of them then everything still type-checks, but it would not be correct. Would this be an appropriate solution? Didn't make a public MemoMap file because I didn't think it warranted changing the public Layer file 

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/silly-glasses-deliver.md
+++ b/.changeset/silly-glasses-deliver.md
@@ -1,5 +1,5 @@
 ---
-"effect": minor
+"effect": patch
 ---
 
-Create managedRuntime/circular.ts
+move ManagedRuntime.TypeId to fix circular imports
--- a/.changeset/silly-glasses-deliver.md
+++ b/.changeset/silly-glasses-deliver.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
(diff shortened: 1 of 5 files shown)


packages/effect/src/Layer.ts
```diff
@@ -31,6 +31,7 @@ import * as defaultServices from "./internal/defaultServices.js"
 import * as fiberRuntime from "./internal/fiberRuntime.js"
 import * as internal from "./internal/layer.js"
```
leonitousconforti, line 32: Didn't update this from

```ts
import * as internal from "./internal/layer.js"
```
to
```ts
import * as internalLayer from "./internal/layer.js"
```

just because it made the diff so much larger, but I could certainly do that for consistency if thats desired

tim-smart:
I think in this case we could just move the ManagedRuntime.TypeId to a `internal/managedRuntime/circular.ts` file.

PR #3687, opened 2024-09-25
Calling HashMap constructors with empty entries returns "untouched" empty HashMap

AlexGeb:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Make `HashMap.make` and `HashMap.fromIterable` return an "untouched" empty `HashMap` when called with empty entries.

Before this change, the below assertion failed because even when the given `Iterable` is empty, we would create an `HashMap` with one edit.

```typescript
deepStrictEqual(HM.fromIterable([]), HM.empty())
```
Screenshot 2024-09-25 at 20 54 31

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/internal/hashMap.ts
+++ b/packages/effect/src/internal/hashMap.ts
@@ -191,11 +191,12 @@ export const make = <Entries extends ReadonlyArray<readonly [any, any]>>(
 
 /** @internal */
 export const fromIterable = <K, V>(entries: Iterable<readonly [K, V]>): HM.HashMap<K, V> => {
-  const map = beginMutation(empty<K, V>())
+  const emptyMap = empty<K, V>()
+  const map = beginMutation(emptyMap)
   for (const entry of entries) {
     set(map, entry[0], entry[1])
   }
-  return endMutation(map)
+  return isEmpty(map) ? emptyMap : endMutation(map)
 }
 
 /** @internal */
--- a/.changeset/pretty-trees-tickle.md
+++ b/.changeset/pretty-trees-tickle.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Make `HashMap.make` and `HashMap.fromIterable` return an "untouched" empty `HashMap` when called with an empty iterable
--- a/packages/effect/test/HashMap.test.ts
+++ b/packages/effect/test/HashMap.test.ts
@@ -59,6 +59,37 @@ describe("HashMap", () => {
 }`)
   })
 
+  it("empty", () => {
+    assert.isTrue(HM.isEmpty(HM.empty()))
+  })
+
+  describe("fromIterable", () => {
+    it("returns an immutable HashMap", () => {
+      const map = HM.fromIterable([[0, "a"]])
+      expect(HM.has(HM.remove(map, 0), 0)).toBe(false)
+      expect(HM.has(map, 0)).toBe(true)
+    })
+    it("returns an untouched empty HashMap", () => {
+      const emptyIterable = new Map().entries()
+      deepStrictEqual(HM.fromIterable([]), HM.empty())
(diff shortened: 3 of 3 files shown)


packages/effect/src/internal/hashMap.ts
```diff
@@ -191,11 +191,12 @@ export const make = <Entries extends ReadonlyArray<readonly [any, any]>>(
...
 export const fromIterable = <K, V>(entries: Iterable<readonly [K, V]>): HM.HashMap<K, V> => {
-  const map = beginMutation(empty<K, V>())
+  const emptyMap = empty<K, V>()
+  const map = beginMutation(emptyMap)
   for (const entry of entries) {
     set(map, entry[0], entry[1])
   }
-  return endMutation(map)
+  return isEmpty(map) ? emptyMap : endMutation(map)
 }
```
AlexGeb, line 200: We could also do:
```ts
/** @internal */
export const fromIterable = <K, V>(entries: Iterable<readonly [K, V]>): HM.HashMap<K, V> => {
  const emptyMap = empty<K, V>()
  const map = beginMutation(emptyMap)
  for (const entry of entries) {
    set(map, entry[0], entry[1])
  }
  endMutation(map)
  ;(map as HashMapImpl<K, V>)._edit = 0
  return map
}
```

mikearnaldi:
It's not a bug, deepStrictEqual doesn't generally work to compare HashMap values. No point special casing it for empty maps

AlexGeb:
> It's not a bug, deepStrictEqual doesn't generally work to compare HashMap values. No point special casing it for empty maps

Ok, we might not need deep strict equality for comparing HashMap (it's the default in our codebase, every tests we write are using deep strict equality, we never challenged this before, it's maybe the time to do it).

BUT there is still the fact that the `_edit` property is equal to 1 whereas we just called a constructor. I would expect that calling an HashMap constructor would give me an "untouched" HashMap..

mikearnaldi:
Those are to be considered internals, a consumer of the module should not rely on internals. BTW what function are you using? In the vitest custom matchers we add rules to compare effect modules so you should have a positive result in comparing two empty maps regardless of the internals

AlexGeb:
> Those are to be considered internals, a consumer of the module should not rely on internals. BTW what function are you using? In the vitest custom matchers we add rules to compare effect modules so you should have a positive result in comparing two empty maps regardless of the internals

Ah we use the `expect().toStrictEqual` but with expect being automatically injected by vitest in test files (we do not import anything related to tests, its globally scoped for test files).
We only import `it` from '@effect/vitest' when we need test effects.

I think that what would be great for us would be to replace vitest default methods by the one from `@effect/vitest` (this way we can keep same devX)

mikearnaldi:
> > Those are to be considered internals, a consumer of the module should not rely on internals. BTW what function are you using? In the vitest custom matchers we add rules to compare effect modules so you should have a positive result in comparing two empty maps regardless of the internals
> 
> Ah we use the `expect().toStrictEqual` but with expect being automatically injected by vitest in test files (we do not import anything related to tests, its globally scoped for test files). We only import `it` from '@effect/vitest' when we need test effects.
> 
> I think that what would be great for us would be to replace vitest default methods by the one from `@effect/vitest` (this way we can keep same devX)

it's fine to have vitest auto-inject the methods `expect` plugins are a default feature, you need to register them though, like we do in: https://github.com/Effect-TS/effect/blob/main/setupTests.ts

AlexGeb:
> it's fine to have vitest auto-inject the methods expect plugins are a default feature, you need to register them though, like we do in: https://github.com/Effect-TS/effect/blob/main/setupTests.ts

Awesome, I didn't knew about it! I'm closing this then, thank you for your answers

PR #3689, opened 2024-09-26
Add vitest layer api

tim-smart:


jessekelly881:
What about

```ts
layer(Foo.Live)("test name", (it) => {})

const describe = layer(Foo.Live);
describe("test/a", (it) => {})
```

tim-smart:
> Using Vi.describe internally to group tests

Added support for an optional name.

mikearnaldi:
Looks like in beforeAll / afterAll we are using promises that will throw FiberFailure that will be problematic in rendering of errors, we should use the same method we use for it.effect

PR #3690, opened 2024-09-26
Implement Effect.Service and allow multiple layers to be provided in Effect.provide

mikearnaldi:
Namely the following is now possible:

```ts
class Prefix extends Effect.Service<Prefix>()("Prefix", {
  sync: () => ({
    prefix: "PRE"
  })
}) {}

class Postfix extends Effect.Service<Postfix>()("Postfix", {
  sync: () => ({
    postfix: "POST"
  })
}) {}

const messages: Array<string> = []

class Logger extends Effect.Service<Logger>()("Logger", {
  accessors: true,
  effect: Effect.gen(function* () {
    const { prefix } = yield* Prefix
    const { postfix } = yield* Postfix
    return {
      info: (message: string) =>
        Effect.sync(() => {
          messages.push(`[${prefix}][${message}][${postfix}]`)
        })
    }
  }),
  dependencies: [Prefix.Default, Postfix.Default]
}) {}

describe("Effect", () => {
  it.effect("Service correctly wires dependencies", () =>
    Effect.gen(function* () {
      const { _tag } = yield* Logger
      expect(_tag).toEqual("Logger")
      yield* Logger.info("Ok")
      expect(messages).toEqual(["[PRE][Ok][POST]"])
      const { prefix } = yield* Prefix
      expect(prefix).toEqual("PRE")
      const { postfix } = yield* Postfix
      expect(postfix).toEqual("POST")
    }).pipe(Effect.provide([Logger.Default, Prefix.Default, Postfix.Default]))
  )
})
```

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -57,7 +57,7 @@ import * as Scheduler from "./Scheduler.js"
 import type * as Scope from "./Scope.js"
 import type * as Supervisor from "./Supervisor.js"
 import type * as Tracer from "./Tracer.js"
-import type { Concurrency, Contravariant, Covariant, NoExtraKeys, NoInfer, NotFunction } from "./Types.js"
+import type { Concurrency, Contravariant, Covariant, NoExcessProperties, NoInfer, NotFunction } from "./Types.js"
 import type * as Unify from "./Unify.js"
 import type { YieldWrap } from "./Utils.js"
 
@@ -6336,21 +6336,21 @@ export const Service: <Self>() => {
   <
     const Key extends string,
     const Make extends
-      | NoExtraKeys<{
+      | NoExcessProperties<{
         readonly scoped: Effect<Service.AllowedType<Key, Make["accessors"]>, any, any>
         readonly dependencies?: ReadonlyArray<Layer.Layer.Any>
         readonly accessors?: boolean
       }, Make>
-      | NoExtraKeys<{
+      | NoExcessProperties<{
         readonly effect: Effect<Service.AllowedType<Key, Make["accessors"]>, any, any>
         readonly dependencies?: ReadonlyArray<Layer.Layer.Any>
         readonly accessors?: boolean
       }, Make>
-      | NoExtraKeys<{
+      | NoExcessProperties<{
         readonly sync: LazyArg<Service.AllowedType<Key, Make["accessors"]>>
         readonly accessors?: boolean
       }, Make>
-      | NoExtraKeys<{
+      | NoExcessProperties<{
         readonly succeed: Service.AllowedType<Key, Make["accessors"]>
         readonly accessors?: boolean
       }, Make>
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -6514,16 +6514,16 @@ export const Service: {
   <Self>(): {
     <
       const Id extends string,
-      Maker extends (Service.Maker<Service.ProhibitedType> & { proxy?: true }) | (Service.Maker<{}> & { proxy: false })
+      Maker extends (Service.Maker<Service.ProhibitedType> & { proxy: true }) | (Service.Maker<{}> & { proxy?: false })
     >(
       id: Id,
       maker: Maker
-    ): Service.ReturnWithMaker<Self, Id, Maker, Maker extends { proxy: false } ? false : true>
+    ): Service.ReturnWithMaker<Self, Id, Maker, Maker extends { proxy: true } ? true : false>
   }
 } = function() {
   return function() {
     const [id, maker] = arguments
-    const proxy = "proxy" in maker ? maker["proxy"] : true
+    const proxy = "proxy" in maker ? maker["proxy"] : false
     const limit = Error.stackTraceLimit
     Error.stackTraceLimit = 2
     const creationError = new Error()
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -2,7 +2,6 @@
  * @since 2.0.0
  */
 import type * as RA from "./Array.js"
-import type { BrandTypeId } from "./Brand.js"
 import type * as Cause from "./Cause.js"
 import type * as Chunk from "./Chunk.js"
 import type * as Clock from "./Clock.js"
@@ -6334,6 +6333,7 @@ export const Tag: <const Id extends string>(id: Id) => <
 /**
  * @since 3.9.0
  * @category context
+ * @experimental might be up for breaking changes
  */
 export const Service: <Self>() => {
   <
@@ -6344,28 +6344,28 @@ export const Service: <Self>() => {
         readonly dependencies?: ReadonlyArray<Layer.Layer.Any>
         readonly accessors?: boolean
         /** @deprecated */
-        readonly [BrandTypeId]: never
+        readonly ಠ_ಠ: never
       }
       | {
         readonly effect: Effect<Service.AllowedType<Key, Make>, any, any>
         readonly dependencies?: ReadonlyArray<Layer.Layer.Any>
         readonly accessors?: boolean
         /** @deprecated */
-        readonly [BrandTypeId]: never
+        readonly ಠ_ಠ: never
       }
       | {
         readonly sync: LazyArg<Service.AllowedType<Key, Make>>
         readonly dependencies?: ReadonlyArray<Layer.Layer.Any>
         readonly accessors?: boolean
         /** @deprecated */
-        readonly [BrandTypeId]: never
+        readonly ಠ_ಠ: never
       }
       | {
         readonly succeed: Service.AllowedType<Key, Make>
         readonly dependencies?: ReadonlyArray<Layer.Layer.Any>
         readonly accessors?: boolean
         /** @deprecated */
-        readonly [BrandTypeId]: never
+        readonly ಠ_ಠ: never
       }
   >(
     key: Key,
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -6522,10 +6522,12 @@ export declare namespace Service {
   export interface ProhibitedType {
     Service?: `property "Service" is forbidden`
     Identifier?: `property "Identifier" is forbidden`
-    Layer?: `property "Layer" is forbidden`
+    Default?: `property "Default" is forbidden`
+    DefaultWithoutDependencies?: `property "DefaultWithoutDependencies" is forbidden`
     _op_layer?: `property "_op_layer" is forbidden`
     _op?: `property "_op" is forbidden`
     of?: `property "of" is forbidden`
+    make?: `property "make" is forbidden`
     context?: `property "context" is forbidden`
     key?: `property "key" is forbidden`
     stack?: `property "stack" is forbidden`
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -6235,12 +6235,13 @@ export declare namespace Tag {
     Service?: `property "Service" is forbidden`
     Identifier?: `property "Identifier" is forbidden`
     _op?: `property "_op" is forbidden`
-    _tag?: `property "_tag" is forbidden`
     of?: `property "of" is forbidden`
     context?: `property "context" is forbidden`
     key?: `property "key" is forbidden`
     stack?: `property "stack" is forbidden`
     name?: `property "name" is forbidden`
+    pipe?: `property "pipe" is forbidden`
+    use?: `property "use" is forbidden`
   }
 
   /**
@@ -6347,6 +6348,7 @@ export declare namespace Service {
     stack?: `property "stack" is forbidden`
     name?: `property "name" is forbidden`
     pipe?: `property "pipe" is forbidden`
+    use?: `property "use" is forbidden`
   }
 
   /**
@@ -6404,19 +6406,18 @@ export declare namespace Service {
    * @since 2.0.0
    * @category models
    */
-  export type Return<Self, Id extends string, Type, Proxy extends boolean> =
+  export type Return<Self, Id extends string, Type, Proxy extends boolean, T = Type & { readonly _tag: Id }> =
     & TagClass<Self, Id, Type>
-    & (Proxy extends true ? (Type extends Record<PropertyKey, any> ? {
+    & (Proxy extends true ? (T extends Record<PropertyKey, any> ? {
           [
-            k in keyof Type as Type[k] extends ((...args: [...infer Args]) => infer Ret)
-              ? ((...args: Readonly<Args>) => Ret) extends Type[k] ? k : never
+            k in keyof T as T[k] extends ((...args: [...infer Args]) => infer Ret)
+              ? ((...args: Readonly<Args>) => Ret) extends T[k] ? k : never
               : k
-          ]: Type[k] extends (...args: [...infer Args]) => Effect<infer A, infer E, infer R>
+          ]: T[k] extends (...args: [...infer Args]) => Effect<infer A, infer E, infer R>
             ? (...args: Readonly<Args>) => Effect<A, E, Self | R>
-            : Type[k] extends (...args: [...infer Args]) => infer A ?
-              (...args: Readonly<Args>) => Effect<A, never, Self>
-            : Type[k] extends Effect<infer A, infer E, infer R> ? Effect<A, E, Self | R>
-            : Effect<Type[k], never, Self>
+            : T[k] extends (...args: [...infer Args]) => infer A ? (...args: Readonly<Args>) => Effect<A, never, Self>
+            : T[k] extends Effect<infer A, infer E, infer R> ? Effect<A, E, Self | R>
+            : Effect<T[k], never, Self>
         } :
         {}) :
       {})
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -6353,38 +6353,38 @@ export declare namespace Service {
    * @since 2.0.0
    * @category models
    */
-  export type AllowedType = Record<PropertyKey, any> & ProhibitedType
+  export type AllowedType<P> = Record<PropertyKey, any> & P
 
   /**
    * @since 2.0.0
    * @category models
    */
-  export type Maker = {
-    effect: Effect<AllowedType, any, any>
+  export type Maker<P> = {
+    effect: Effect<AllowedType<P>, any, any>
     dependencies?: never
   } | {
-    effect: Effect<AllowedType, any, any>
+    effect: Effect<AllowedType<P>, any, any>
     dependencies: []
   } | {
-    effect: Effect<AllowedType, any, any>
+    effect: Effect<AllowedType<P>, any, any>
     dependencies: [Layer.Layer<any, any, any>, ...Array<Layer.Layer<any, any, any>>]
   } | {
-    scoped: Effect<AllowedType, any, any>
+    scoped: Effect<AllowedType<P>, any, any>
     dependencies?: never
   } | {
-    scoped: Effect<AllowedType, any, any>
+    scoped: Effect<AllowedType<P>, any, any>
     dependencies: []
   } | {
-    scoped: Effect<AllowedType, any, any>
+    scoped: Effect<AllowedType<P>, any, any>
     dependencies: [Layer.Layer<any, any, any>, ...Array<Layer.Layer<any, any, any>>]
   } | {
-    sync: () => AllowedType
+    sync: () => AllowedType<P>
     dependencies?: never
   } | {
-    sync: () => AllowedType
+    sync: () => AllowedType<P>
     dependencies: [Layer.Layer<any, any, any>, ...Array<Layer.Layer<any, any, any>>]
   } | {
-    sync: () => AllowedType
+    sync: () => AllowedType<P>
     dependencies: []
   }
 
@@ -6404,37 +6404,40 @@ export declare namespace Service {
    * @since 2.0.0
    * @category models
    */
-  export type Return<Self, Id extends string, Type extends AllowedType> =
+  export type Return<Self, Id extends string, Type, Proxy extends boolean> =
     & TagClass<Self, Id, Type>
-    & (Type extends Record<PropertyKey, any> ? {
-        [
-          k in keyof Type as Type[k] extends ((...args: [...infer Args]) => infer Ret)
-            ? ((...args: Readonly<Args>) => Ret) extends Type[k] ? k : never
-            : k
-        ]: Type[k] extends (...args: [...infer Args]) => Effect<infer A, infer E, infer R>
-          ? (...args: Readonly<Args>) => Effect<A, E, Self | R>
-          : Type[k] extends (...args: [...infer Args]) => infer A ? (...args: Readonly<Args>) => Effect<A, never, Self>
-          : Type[k] extends Effect<infer A, infer E, infer R> ? Effect<A, E, Self | R>
-          : Effect<Type[k], never, Self>
+    & (Proxy extends true ? (Type extends Record<PropertyKey, any> ? {
+          [
+            k in keyof Type as Type[k] extends ((...args: [...infer Args]) => infer Ret)
+              ? ((...args: Readonly<Args>) => Ret) extends Type[k] ? k : never
+              : k
+          ]: Type[k] extends (...args: [...infer Args]) => Effect<infer A, infer E, infer R>
+            ? (...args: Readonly<Args>) => Effect<A, E, Self | R>
+            : Type[k] extends (...args: [...infer Args]) => infer A ?
+              (...args: Readonly<Args>) => Effect<A, never, Self>
+            : Type[k] extends Effect<infer A, infer E, infer R> ? Effect<A, E, Self | R>
+            : Effect<Type[k], never, Self>
+        } :
+        {}) :
+      {})
+    & (Proxy extends true ? {
+        use: <X>(
+          body: (_: Type) => X
+        ) => X extends Effect<infer A, infer E, infer R> ? Effect<A, E, R | Self> : Effect<X, never, Self>
       } :
       {})
-    & {
-      use: <X>(
-        body: (_: Type) => X
-      ) => X extends Effect<infer A, infer E, infer R> ? Effect<A, E, R | Self> : Effect<X, never, Self>
-    }
 
   /**
    * @since 2.0.0
    * @category models
    */
-  export type ReturnWithMaker<Self, Id extends string, Maker extends Service.Maker> =
+  export type ReturnWithMaker<Self, Id extends string, Maker, Proxy extends boolean> =
     & {}
     & Maker extends {
-    scoped: Effect<infer Type extends AllowedType, infer E, infer R>
+    scoped: Effect<infer Type, infer E, infer R>
     dependencies: infer Layers extends [Layer.Layer<any, any, any>, ...Array<Layer.Layer<any, any, any>>]
   } ?
-      & Return<Self, Id, Type>
+      & Return<Self, Id, Type, Proxy>
       & Layer.Layer<
         Self,
         E | { [k in keyof Layers]: Layer.Layer.Error<Layers[k]> }[number],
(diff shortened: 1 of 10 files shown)


tim-smart:
`Effect.Tag.effect/scoped/sync` could be a good option too.

patroza:
Yes this rocks!
getting closer to mine https://github.com/effect-ts-app/libs/blob/main/packages/core/src/Context.ts#L247
scoped would be great, mine exposes toLayer and toLayerScoped.
I would still expose a `to` variant here too so you can easily build e.g Test or Fake layers too?

mikearnaldi:
>> I would still expose a to variant here too so you can easily build e.g Test or Fake layers too?

```ts
class DateTag extends Effect.Tag("DateTag", {
  sync: () => new Date(1970, 1, 1)
})<DateTag>() {
  static Test = Layer.sync(this, () => new Date(1970, 1, 2))
}
```

not sure I understand

mikearnaldi:
Updated:

Namely the following is now possible:

```ts

class Logger extends Effect.Tag("Logger", {
  sync: () => ({
    info: (message: string) => Effect.logInfo(message)
  })
})<Logger>() {}
```

Together with:

```ts

class Logger extends Effect.Tag("Logger", {
  effect: Effect.sync(() => ({
    info: (message: string) => Effect.logInfo(message)
  }))
})<Logger>() {}
```

And:

```ts

class Logger extends Effect.Tag("Logger", {
  scoped: Effect.gen(function* () {
    yield* Effect.addFinalizer(() => Effect.logInfo("Done"))
    return {
      info: (message: string) => Effect.logInfo(message)
    }
  })
})<Logger>() {}
```

All the above will lead to automatic layer creation named: `Logger` so you can directly do `Effect.provide(Logger)`.

The created layer is also exported with the name `Logger.Layer` to expose non packed dependencies for testing:

```ts
import { Effect, Layer } from "effect"

class Dummy extends Effect.Tag("Dummy", {
  sync: () => ({})
})<Dummy>() {}

class Logger extends Effect.Tag("Logger", {
  effect: Effect.gen(function* () {
    yield* Dummy
    return {
      info: (message: string) => Effect.logInfo(message)
    }
  }),
  dependencies: [Dummy]
})<Logger>() {}
```

You'll have that `Logger` is `Layer<Logger, never, never>` while `Logger.Layer` is `Layer<Logger, never, Dummy>`.

mikearnaldi:
If we make `Effect.provide` variadic (with auto-merging) we are basically hiding layers almost fully, `Effect.provide(Logger, Db, ..)`

patroza:
> You'll have that `Logger` is `Layer<Logger, never, never>` while `Logger.Layer` is `Layer<Logger, never, Dummy>`.

Is kind of neat yes. maybe call it `liveDependencies`, not sure.

mikearnaldi:
> Do you mean typeof Logger (which is also the Tag).

yes

> In my implementation, Logger is the opaque service interface.

that won't be implemented here as you'll fill up your service with irrelevant fields and lose assignability to plain objects

mikearnaldi:
> > that won't be implemented here as you'll fill up your service with irrelevant fields and lose assignability to plain objects
> 
> well, I was also the one fully against that when we were discussing the Context, and making it a **requirement** for service interfaces to be polluted with said fields, but now that it's optional, an implementation detail of your tag/service, I am happy with it for almost all my services. I expose: `static of = (service: Omit<Id, keyof Context.TagClassShape<any, any>>)) => Id` aka `static of = (service: ServiceShape) => Id`, which is just implemented as identity, as they are virtual fields anyway.

but the pollution isn't only on assign, it's also on usage, you'll have stuff like `_op`, etc in your service when you consume it. Anyway your code your decision :)

patroza:
> but the pollution isn't only on assign, it's also on usage, you'll have stuff like `_op`, etc in your service when you consume it. 

No we're conflating`Logger` with `typeof Logger`
typeof Logger, has the op/tag/Layer, Id is clean right, only has TagClassShape,
but yes you're right, Type and Id show up:
Screenshot 2024-09-26 at 16 15 06
but those could be similarly hidden under a symbol like `[TagTypeId]`.. but yea I get your point :)


> Anyway your code your decision :)

of course

patroza:
Oh it used to be that way when I implemented mine
IMG_5493

mikearnaldi:
> > but the pollution isn't only on assign, it's also on usage, you'll have stuff like `_op`, etc in your service when you consume it.
> 
> No we're conflating`Logger` with `typeof Logger` typeof Logger, has the op/tag/Layer, Id is clean right, only has TagClassShape, but yes you're right, Type and Id show up: Screenshot 2024-09-26 at 16 15 06 but those could be similarly hidden under a symbol like `[TagTypeId]`.. but yea I get your point :)
> 
> > Anyway your code your decision :)
> 
> of course

why the abstract?

patroza:
> why the abstract?

leftover I think, I guess because using the constructor in userland wasn't really useful for anything.

mikearnaldi:
> > why the abstract?
> 
> leftover I think, I guess because using the constructor in userland wasn't really useful for anything, so `abstract` kinda tells you not to call `new` :D but otoh it begs "inherit me", so not sure we had a net win here.

also how did you manage to override `of` from `Tag`?

mikearnaldi:
I think I am kind of happy with: https://github.com/Effect-TS/effect/pull/3690/commits/e17d08533950f990811264012bea0f18dc897c1a

cc @patroza @tim-smart @schickling @IMax153

IMax153:
I'm a fan of this - I especially like the idea of a variadic `Effect.provide` with auto-merging.

mikearnaldi:
I guess https://github.com/Effect-TS/effect/pull/3690/commits/b95558e3d6183fe3310982707c7c17b73da1c6ff is fine though, looks like:

```ts
describe("Effect", () => {
  it.effect("Service correctly wires dependencies", () =>
    Effect.gen(function*() {
      const { _tag } = yield* Logger
      expect(_tag).toEqual("Logger")
      yield* Logger.info("Ok")
      expect(messages).toEqual(["[PRE][Ok][POST]"])
      const { prefix } = yield* Prefix
      expect(prefix).toEqual("PRE")
      const { postfix } = yield* Postfix
      expect(postfix).toEqual("POST")
    }).pipe(
      Effect.provide([Logger, Prefix, Postfix])
    ))
})
```

mikearnaldi:
The only remaining footgun is calling provide multiple times, wondering if we should work on having a shared global memoMap to avoid layer re-instantiation

tim-smart:
Given that the prohibited list is growing, a non-proxy version would be nice.

tim-smart:
Also with everything becoming a sub type of effect, the accessors can get annoying when the data types are getting flattened.

mikearnaldi:
> Given that the prohibited list is growing, a non-proxy version would be nice.

added a `proxy: false` option

IMax153:
In the spirit of avoiding footguns for users, should we default the `proxy` option to `false`?

mikearnaldi:
> In the spirit of avoiding footguns for users, should we default the `proxy` option to `false`?

probably a good idea

mikearnaldi:
Screenshot 2024-09-26 at 22 57 54

there seems to be a bug, the tag is not typed properly...

mikearnaldi:
So the core issue was that extending `of` makes the tag no longer a tag, and it doesn't work in places like `Micro.service(tag)`.

In 4.0 we can consider extracting out `.of` from the base tag definition but for now that would be a breaking change, so I added a second `.make` method to create the class

mikearnaldi:
> So the core issue was that extending `of` makes the tag no longer a tag, and it doesn't work in places like `Micro.service(tag)`.
> 
> In 4.0 we can consider extracting out `.of` from the base tag definition but for now that would be a breaking change, so I added a second `.make` method to create the class

@patroza this is an issue in your code too most likely

tim-smart:
Found a couple type issues, and some solutions: https://gist.github.com/tim-smart/646e9d5edfb7e95cc5c8b484966a4e5f

The DX of making the tag itself a Layer has some issues, mainly around not being able to inspect the type to easier see what errors or requirements are on the Layer.

packages/effect/src/Effect.ts
```diff
@@ -6305,3 +6325,309 @@ export const Tag: <const Id extends string>(id: Id) => <
...
+
+    if ("scoped" in maker) {
+      // @ts-expect-error
+      TagClass["Layer"] = layer.scoped(TagClass, maker["scoped"]).pipe(
+        // @ts-expect-error
+        layer.flatMap((c) => layer.succeedContext(Context.add(c, TagClass, of(Context.unsafeGet(c, TagClass)))))
+      )
+      let live = TagClass["Layer"]
+      if ("dependencies" in maker) {
+        live = live.pipe(layer.provide(layer.mergeAll(...maker["dependencies"])))
```
patroza, line 6575: Could we make Layer.provide variadric aswell?

mikearnaldi: Yes

patroza:
Here's an interesting case imo:
```ts
export class Export extends Effect.Service<Export>()("Export", { effect: makeEmpasa }) {
  static LiveMako = Layer.effect(this, makeMako.pipe(Effect.map(this.make))).pipe(
    Layer.provide(
      Layer.mergeAll(Mako.DeliveryNoteRepo.Live, shared)
    )
  )
  static LiveEmpasa = this.pipe(
    Layer.provide(
      Layer.mergeAll(
        MultiPick.DeliveryNoteRepo.Live,
        Manufacturing.DeliveryNoteRepo.Live,
        PickItemRepo.Live,
        shared
      )
    )
  )
  static Live = BaseConfig.pipe(
    Effect.map((_) => _.environment === "empasa" ? this.LiveEmpasa : this.LiveMako),
    Layer.unwrapEffect
  )
}
```

mikearnaldi:
> Found a couple type issues, and some solutions: https://gist.github.com/tim-smart/646e9d5edfb7e95cc5c8b484966a4e5f
> 
> Edit: moved to a PR here #3694
> 
> The DX of making the tag itself a Layer has some issues, mainly around not being able to inspect the type to easily see what errors or requirements are on the Layer.

It regresses on two things:
1) the feedback has been that the name `Live` is confusing, that's why I made the tag a layer
2) it doesn't use the class as opaque type

patroza:
> @patroza this is an issue in your code too most likely
Yea probably :D 

Love that you took on the class as opaque type now! doing it with _tag is brilliant.
it's funny, how we come full circle. it blends all the ideas from the past years, but with the right mix.

mikearnaldi:
> > Found a couple type issues, and some solutions: https://gist.github.com/tim-smart/646e9d5edfb7e95cc5c8b484966a4e5f
> > Edit: moved to a PR here #3694
> > The DX of making the tag itself a Layer has some issues, mainly around not being able to inspect the type to easily see what errors or requirements are on the Layer.
> 
> It regresses on two things:
> 
> 1. the feedback has been that the name `Live` is confusing, that's why I made the tag a layer
> 2. it doesn't use the class as opaque type

Re 1 we could instead call it `Default`
Re 2 this is kind of a nice to have especially when referencing the service type, we could get rid of the make though allowing constructor like `new Logger({...}`

patroza:
> Re 2 this is kind of a nice to have especially when referencing the service type, we could get rid of the make though allowing constructor like `new Logger({...}`

Yea, it's awesome, been enjoying it for a long time.

> get rid of make

well, everywhere else where we supported new we also supported make, because it can be used in pipe, and is also just a common standard in Effect.

What will you return at the end of the constructor? I basically did an `Object.assign(svc, this)` instead, but I wasn't happy with it, as you kinda loose the original object.

mikearnaldi:
> > Re 2 this is kind of a nice to have especially when referencing the service type, we could get rid of the make though allowing constructor like `new Logger({...}`
> 
> Yea, it's awesome, been enjoying it for a long time.
> 
> > get rid of make
> 
> well, everywhere else where we supported `new` we also supported `make`, because it can be passed like in a pipe, and is also just a common standard in Effect.
> 
> What will you return from the constructor? I basically did an `Object.assign(svc, this)` instead, but I wasn't happy with it, as you kinda loose the original object.

Considering that the default instance won't be able to actually be an instance of the class I just return the service ignoring `this`

patroza:
> The DX of making the tag itself a Layer has some issues, mainly around not being able to inspect the type to easily see what errors or requirements are on the Layer.

I had the same experience, however you can inspect `effect:` or `scoped:`

mikearnaldi:
> > The DX of making the tag itself a Layer has some issues, mainly around not being able to inspect the type to easily see what errors or requirements are on the Layer.
> 
> I had the same experience, however you can inspect `effect:` or `scoped:`
> 
> > Considering that the default instance won't be able to actually be an instance of the class
> 
> why not? at least it was in my case. though i'm not suggesting we should support it

The default layer is created from the function Service and the function Service does not have access to the final constructor, `class A extends f()` here `f` has no way of accessing `A.constructor`

patroza:
> The default layer is created from the function Service and the function Service does not have access to the final constructor, `class A extends f()` here `f` has no way of accessing `A.constructor`

Guess it's a limitation of making the tag also the layer, otherwise it could if it were lazy: `new this(...)`

mikearnaldi:
> > The default layer is created from the function Service and the function Service does not have access to the final constructor, `class A extends f()` here `f` has no way of accessing `A.constructor`
> 
> Guess it's a limitation of making the tag also the layer, otherwise it could if it were lazy: `new this(...)`, in my case one had to call: `static Live = this.toLayer()` or `this.toLayer(effect)`

The same applies to `Logger.Default`, yes in your case you call a function even `Logger.Default()` could do it but it's a footgun as it would lose instance equality and lead to duplicated layers

patroza:
> The same applies to `Logger.Default`, yes in your case you call a function even `Logger.Default()` could do it but it's a footgun as it would lose instance equality and lead to duplicated layers

could be a cached getter, but I guess you get trouble then if for whatever reason you decided to extend the class

mikearnaldi:
> > The same applies to `Logger.Default`, yes in your case you call a function even `Logger.Default()` could do it but it's a footgun as it would lose instance equality and lead to duplicated layers
> 
> could be a cached getter, but I guess you get trouble then if for whatever reason you decided to extend the class.
> 
> I do kinda like `Logger` as Id, Shape, Tag, Layer all wrapped into one though.

Oh god extending the class, that's not something I even thought lol. Same here for liking packing all in one. I don't think the structure object being not an instance of the actual class is of any importance imho, instanceof wouldn't be supported anyway (and we can make it work based on _tag)

patroza:
> Oh god extending the class, that's not something I even thought lol. Same here for liking packing all in one. I don't think the structure object being not an instance of the actual class is of any importance imho, instanceof wouldn't be supported anyway (and we can make it work based on _tag)

It's kinda cool as extensions to the make:
`readonly Extension = this.someEffectFromMake.pipe(....`
but im not sure it's too valuable, or not confusing.

patroza:
@mikearnaldi `make` function is undefined when `proxy: false`

patroza:
`Logger.Test` is broken: 
```ts
  it.effect("Service testcorrectly wires dependencies", () =>
    Effect.gen(function*() {
      const { _tag } = yield* Logger
      expect(_tag).toEqual("Logger")
      yield* Logger.info("Ok")
      expect(messages).toEqual(["[PRE][Ok][POST]"])
      const { prefix } = yield* Prefix
      expect(prefix).toEqual("PRE")
      const { postfix } = yield* Postfix
      expect(postfix).toEqual("POST")
    }).pipe(
      Effect.provide([Logger.Test, Prefix, Postfix])
    ))```

AssertionError: expected undefined to deeply equal 'Logger'

- Expected: 
"Logger"

+ Received: 
undefined

 

> ❯ test/Effect/service.test.ts:69:20
>      67|     Effect.gen(function*() {
>      68|       const { _tag } = yield* Logger
>      69|       expect(_tag).toEqual("Logger")
>        |                    ^
>      70|       yield* Logger.info("Ok")
>      71|       expect(messages).toEqual(["[PRE][Ok][POST]"])

**Comment by mikearnaldi** [comment]
> Found a couple type issues, and some solutions: https://gist.github.com/tim-smart/646e9d5edfb7e95cc5c8b484966a4e5f
> 
> Edit: moved to a PR here #3694
> 
> The DX of making the tag itself a Layer has some issues, mainly around not being able to inspect the type to easily see what errors or requirements are on the Layer.

```ts
// @ts-expect-error
export class WithoutProxy extends Effect.Service<WithoutProxy>()("WithoutProxy", {
  scoped: Effect.succeed("foo"),
  proxy: false
}) {}
```

this is expected as `Effect.Service` produces an opaque tagged object as the final service, non objects are not allowed

patroza:
In terms of the opaque Layer not being inspectable, perhaps we should only make the tag also a Layer when it has no more R?

mikearnaldi:
> In terms of the opaque Layer not being inspectable, perhaps we should only make the class also a Layer when it has no more R? I think R is more important than E, as they usually bubble up to the root.

feels like an arbitrary choice, in the last commit I added a `Logger.Default` layer too, I think it's fine to have the instance as layer as long as we also have something to inspect, in this case when you want to inspect `Logger` you can quickly add `.Default` and see

mikearnaldi:
> > In terms of the opaque Layer not being inspectable, perhaps we should only make the class also a Layer when it has no more R? I think R is more important than E, as they usually bubble up to the root.
> 
> feels like an arbitrary choice, in the last commit I added a `Logger.Default` layer too, I think it's fine to have the instance as layer as long as we also have something to inspect, in this case when you want to inspect `Logger` you can quickly add `.Default` and see

I am growing slightly unsure about the tag being a layer, in spirit of recent discussions on subtyping and in spirit of avoiding having two ways of doing the same thing I am starting to think `Logger.Default` would be enough.

@schickling would `.Default` be a better name for `.Live`, we can reference it as "the default implementation of a service" which would be kind of self-explainatory

patroza:
> I am growing slightly unsure about the tag being a layer, in spirit of recent discussions on subtyping and in spirit of avoiding having two ways of doing the same thing I am starting to think `Logger.Default` would be enough.

I have similar reservations, however one cannot deny the awesomeness and cleanliness of just `Logger`.
In any case I think Tag in typeof `Logger`, and Service/Type as `Logger` is already a huge win.
Having `dependencies` in there too, you have another nice chunk of the pattern captured.

I think I could live with `.Default` (dependencies provided) and `.Layer` (not provided), with good docs this should be clear enough.

mikearnaldi:
Ok, removing the tag as layer thing I was able to have Default & Layer return actual instances, now the only issue is that methods defined on the instance like `hello` in the following example:

```ts
class Prefix extends Effect.Service<Prefix>()("Prefix", {
  sync: () => ({ prefix: "PRE" })
}) {
  hello = 100
}
```

can't be included in the proxy methods due to circularity, at the type level TS complains, not sure if there are workarounds.

Funnily enough something like `Prefix.$.hello` could work (no complains), it's kind of horrendous though. I think it's a good compromise, having instance methods isn't really the goal anyway

patroza:
Had my first issue, getting invalid Date:
```ts
import { subMinutes } from "date-fns"
import { Effect } from "effect-app"

export class TimeLive {
  #activity: Date | undefined
  #closable: Date | undefined
  #startOfDay: Date | undefined
  #now: Date | undefined
  constructor(now?: Date) {
    this.#now = now
  }

  get now() {
    return this.#now ||= new Date()
  }

  get activity() {
    return this.#activity ||= subMinutes(this.now, 5)
  }
  get closable() {
    return this.#closable ||= subMinutes(this.now, 30)
  }
  get startOfDay() {
    return this.#startOfDay ||= new Date(new Date(this.now).setHours(0, 0, 0, 0))
  }
}
export abstract class Time
  extends Effect.Service<Time>()("Time", { effect: Effect.sync(() => new TimeLive()), accessors: true })
{
}
```

interestingly enough it worked fine with my `TagMakeId` implementation.

patroza:
ah the difference is, I don't return an instance of the class :)

patroza:
this fixes it, but I don't know what edge cases we just bought ourselves
```ts
      constructor(args: any) {
        // @ts-expect-error
        this._tag = id
        return new Proxy(this, {
          get(_target: any, prop: any, receiver) {
            if (prop in _target) {
              if (typeof _target[prop] === "function") {
                return (...a) => _target[prop](...a)
              }
              return _target[prop]
            }
            if (typeof args[prop] === "function") {
              return (...a) => args[prop](...a)
            }

            return args[prop]
          }
        })
      }
      ```

**Comment by mikearnaldi** [comment]
It should be OK

**Comment by patroza** [comment]
At some point it makes you wonder, if we just get back to basics :D
```ts
export class SomeService extends Effect.ServiceClass<SomeService>()({ dependencies: { someOtherService: SomeOtherService.Default } }) {

  doStuff = this.someOtherService.something("abc")
                            // ^ protected inherited
}
```

packages/effect/src/Effect.ts
```diff
@@ -6277,3 +6336,240 @@ export const Tag: <const Id extends string>(id: Id) => <
...
+  export type Options =
+    | (Service.Maker<Service.ProhibitedType> & { accessors: true })
+    | (Service.Maker<{}> & { accessors?: false })
+
+  /**
+   * @since 2.0.0
+   * @category models
+   */
+  export interface TagClass<Self, Id, Type> extends Context.Tag<Self, Self> {
+    make(_: Type): Self
```
patroza, line 6403: `make(this: void, _: Type): Self`

mikearnaldi: why?

patroza: So I can Effect.andThen(x, this.make) without an ESLint error about binding.

tim-smart:
> Had my first issue

I don't think we can support both inheriting and preserving the prototype of a return value?

On my play branch I got instance getters working: https://github.com/Effect-TS/effect/pull/3694/files#diff-b88ab56a1877c3d6e6d9ffa2e297dbb4f204e612148cc29df5e4b4dc58a5c1c9R50

mikearnaldi:
> > Had my first issue
> 
> I don't think we can support both inheriting the service prototype and preserving the prototype of a return value?
> 
> On my play branch I got instance getters working: https://github.com/Effect-TS/effect/pull/3694/files#diff-b88ab56a1877c3d6e6d9ffa2e297dbb4f204e612148cc29df5e4b4dc58a5c1c9R50

Instance getters are already working in this branch too as the returned type is a class instance, the limitation was that they can't be added to the proxy due to type circularity

mikearnaldi:
> this fixes it, but I don't know what edge cases we just bought ourselves
> 
> ```ts
>       constructor(args: any) {
>         const cache = new Map()
> 
>         // @ts-expect-error
>         this._tag = id
>         return new Proxy(this, {
>           get(_target: any, prop: any, receiver) {
>             if (cache.has(prop)) return cache.get(prop)
>             if (prop in _target) {
>               if (typeof _target[prop] === "function") {
>                 cache.set(prop, (...a) => _target[prop](...a))
>                 return cache.get(prop)
>               }
>               return _target[prop]
>             }
>             if (typeof args[prop] === "function") {
>               cache.set(prop, (...a) => args[prop](...a))
>               return cache.get(prop)
>             }
> 
>             return args[prop]
>           }
>         })
>       }
>       ```
> ```

this doesn't fix it, with this the test gives: `TypeError: Cannot read from private field`

patroza:
> I took a stab at handling Pat's madness: https://github.com/Effect-TS/effect/pull/3694/files#diff-b88ab56a1877c3d6e6d9ffa2e297dbb4f204e612148cc29df5e4b4dc58a5c1c9R88
> 
> Relevant code: https://github.com/Effect-TS/effect/blob/service/packages/effect/src/Effect.ts#L6370

I feel honoured

patroza:
> this doesn't fix it, with this the test gives: `TypeError: Cannot read from private field`

The original version, not the workaround.
The workaround case is fixed by Tim though it seems.

tim-smart:
I'll try it out in the discord bot project and see if everything feels good.

patroza:
Ill bring it to prod tomorrow and shake the tree a bit.

patroza:
- `TagClass.make` is an actual (disguised as `= function()`) method now that requires this access, so we need to flag it as method again, or we have unchecked binding issues.
- `scoped` doesn't work; (change to `effect` works)  will look for a repro

```ts
TypeError: Cannot use 'in' operator to search for '_op' in undefined
    at runLoop (/Users/patrickroza/pj/macs/scanner/node_modules/.pnpm/effect@3.8.4_patch_hash=zrt3ymoqeerndxhv46ijtc74fa/node_modules/effect/src/internal/fiberRuntime.ts:1331:20)
    at evaluateEffect (/Users/patrickroza/pj/macs/scanner/node_modules/.pnpm/effect@3.8.4_patch_hash=zrt3ymoqeerndxhv46ijtc74fa/node_modules/effect/src/internal/fiberRuntime.ts:900:27)
    at evaluateMessageWhileSuspended (/Users/patrickroza/pj/macs/scanner/node_modules/.pnpm/effect@3.8.4_patch_hash=zrt3ymoqeerndxhv46ijtc74fa/node_modules/effect/src/internal/fiberRuntime.ts:868:14)
    at drainQueueOnCurrentThread (/Users/patrickroza/pj/macs/scanner/node_modules/.pnpm/effect@3.8.4_patch_hash=zrt3ymoqeerndxhv46ijtc74fa/node_modules/effect/src/internal/fiberRuntime.ts:630:18)
    at run (/Users/patrickroza/pj/macs/scanner/node_modules/.pnpm/effect@3.8.4_patch_hash=zrt3ymoqeerndxhv46ijtc74fa/node_modules/effect/src/internal/fiberRuntime.ts:1384:10)
    at starveInternal (/Users/patrickroza/pj/macs/scanner/node_modules/.pnpm/effect@3.8.4_patch_hash=zrt3ymoqeerndxhv46ijtc74fa/node_modules/effect/src/Scheduler.ts:89:17)
    at <anonymous> (/Users/patrickroza/pj/macs/scanner/node_modules/.pnpm/effect@3.8.4_patch_hash=zrt3ymoqeerndxhv46ijtc74fa/node_modules/effect/src/Scheduler.ts:106:47)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

patroza:
failing test added, and flagged methods

mikearnaldi:
Screenshot 2024-10-03 at 15 05 38

I see the message...

patroza:
> I see the message...

ok, im blind

https://github.com/user-attachments/assets/776d3632-046b-44f5-8202-10e8c9d753a5

IMax153 (approved):
LGTM 🚢

patroza:
The one thought that's haunting me since this whole thing started, is .. should we maybe demand `R` to be never, to even expose `.Default`? aka one must provide all dependencies, or none.

It removes some flexibility, but at least it would promote the localisation of "finalizing" layers.
It could even still expose Default, but as a literal `"The Layer has not eliminated all of it's dependencies (${REST})"`

mikearnaldi:
> The one thought that's haunting me since this whole thing started, is .. should we maybe demand `R` to be never, to even expose `.Default`? aka one must provide all `dependencies`, or none.
> 
> It removes some flexibility, but at least it would promote the localisation of "finalizing" layers. It could even still expose Default, but as a literal `"The Layer has not eliminated all of it's dependencies (${REST})"`, though maybe even better to just error on incomplete `dependencies` arg

No there are valid cases for some dependencies not to be packed and also the error will be too hidden to notice

patroza:
> No there are valid cases for some dependencies not to be packed and also the error will be too hidden to notice.
Yes this would also be my go-to veto reason.
but, at the same time, we should make the 80% very well supported, and the 20%, more advanced cases, more hidden.
Perhaps not unlike the `strict: true` being default for schema transforms with the `strict: false` being an opt out for those "i know what im doing" cases/folk.

mikearnaldi:
> > No there are valid cases for some dependencies not to be packed and also the error will be too hidden to notice.
> 
> Yes this would also be my go-to veto reason. but, at the same time, we should make the 80% very well supported, and the 20%, more advanced cases, more hidden. Perhaps not unlike the `strict: true` being default for schema transforms with the `strict: false` being an opt out for those "i know what im doing" cases/folk.

Again, no.

tim-smart (approved):

patroza:
Still trouble even with using the BrandId:
Screenshot 2024-10-05 at 16 14 53

In this case I typed `proxy:  true` instead of `accessors: true`

PR #3694, opened 2024-09-27
play with Effect.Service

tim-smart:


mikearnaldi:
Screenshot 2024-09-28 at 11 11 35

The reason for the type level dance in the original implementation is because otherwise auto-complete doesn't work

mikearnaldi:
I think I prefer https://github.com/Effect-TS/effect/pull/3690 as this has the issue of not creating an opaque type for the service (and also doesn't make the tag a layer)

tim-smart:
> type-level still has some issues

Maybe a chainable api would be better. 

```ts
class Users extends Effect.Service.scoped<Users>()(
  Effect.gen(function*() {
    // ...
  }),
  { dependencies: [] }
) {}
```

Hmm, not sure I like it.

mikearnaldi:
> > type-level still has some issues
> 
> Maybe a chainable api would be better.
> 
> ```ts
> class Users extends Effect.Service.scoped<Users>()(
>   Effect.gen(function*() {
>     // ...
>   }),
>   { dependencies: [] }
> ) {}
> ```
> 
> Hmm, not sure I like it.

not sure, wondering if we can find a way to not have the excessive keys show in auto-complete, similar issue is:

Screenshot 2024-10-01 at 10 05 19

mikearnaldi:
also I don't quite like `.layer` and `.layerWithoutDependencies`, we should probably standardise on `.Default | .DefaultWithoutDependencies`

PR #3699, opened 2024-09-28
added Stream.mergeWithTag

jessekelly881:
Adds Stream.mergeWithTag

```ts
// Stream.Stream<{ _tag: "a"; value: number; } | { _tag: "b"; value: string; }>
Stream.mergeWithTag({
  a: Stream.make(0),
  b: Stream.make("")
}, { concurrency: "unbounded"  })
```

### DIFF
--- a/.changeset/silver-tools-collect.md
+++ b/.changeset/silver-tools-collect.md
@@ -2,7 +2,7 @@
 "effect": patch
 ---
 
-added Stream.mergeStruct
+added Stream.mergeWithTag
 
 Combines a struct of streams into a single stream of tagged values where the tag is the key of the struct.
 
@@ -10,7 +10,7 @@ Combines a struct of streams into a single stream of tagged values where the tag
 import { Stream } from "effect"
 
 // Stream.Stream<{ _tag: "a"; value: number; } | { _tag: "b"; value: string; }>
-const stream = Stream.mergeStruct(
+const stream = Stream.mergeWithTag(
   {
     a: Stream.make(0),
     b: Stream.make("")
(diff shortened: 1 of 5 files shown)


mikearnaldi:
feels a bit arbitrary and ad-hoc not sure how general this is

jessekelly881:
I mean, it's more general than the existing mergeEither.

mikearnaldi:
> I mean, it's more general than the existing mergeEither. The idea, however, is the same. It allows you to merge multiple streams while maintaining info on where an element in the new stream came from.

true, I guess I am in favour, unsure of the name maybe `mergeTagged` ? struct hints to a struct in output

jessekelly881:
Yeah.. I was wondering about the name as well. mergeTagged to me implies that the input is tagged. Maybe mergeWithTag

jessekelly881:
Yeah.. I was wondering about the name as well. I'll update it to mergeTagged

mikearnaldi:
mergeWithTag is kind of nice, let's have @tim-smart take the final decision on the name

tim-smart:
`mergeWithTag` sounds good.

.changeset/silver-tools-collect.md
```diff
@@ -0,0 +1,20 @@
+---
+"effect": patch
```
tim-smart, line 2:
```suggestion
"effect": minor
```

But too late now lol.

jessekelly881: Oops. Sorry. Why though? It's not breaking.

tim-smart: New features are minor for semver (once API stable)

jessekelly881: Gotcha! If the package is versioned below 1.0.0 does patch for feature still apply (e.g. the vitest package)?

PR #3703, opened 2024-09-30
Vitest prop testing

jessekelly881:
Adds property testing.

[https://www.npmjs.com/package/@fast-check/vitest](https://www.npmjs.com/package/@fast-check/vitest
)

```ts
import { it } from "@effect/vitest"

const realNumber = Schema.Finite.pipe(Schema.nonNaN())

it.prop("symmetry", [realNumber, realNumber], ([a, b]) => a + b === b + a)

it.effect.prop("symmetry", [realNumber, realNumber], ([a, b]) =>
  Effect.gen(function* () {
    yield* Effect.void
    return a + b === b + a
  })
)

it.scoped.prop(
  "should detect the substring",
  { a: Schema.String, b: Schema.String, c: Schema.String },
  ({ a, b, c }) =>
    Effect.gen(function* () {
      yield* Effect.scope
      return (a + b + c).includes(b)
    })
)
```

### DIFF
--- a/packages/vitest/src/index.ts
+++ b/packages/vitest/src/index.ts
@@ -4,6 +4,7 @@
 import type * as Duration from "effect/Duration"
 import type * as Effect from "effect/Effect"
 import type * as Layer from "effect/Layer"
+import type * as Schema from "effect/Schema"
 import type * as Scope from "effect/Scope"
 import type * as TestServices from "effect/TestServices"
 import * as V from "vitest"
@@ -39,6 +40,11 @@ export namespace Vitest {
   /**
    * @since 1.0.0
    */
+  export type SchemaObj = Array<Schema.Schema.Any> | { [K in string]: Schema.Schema.Any }
+
+  /**
+   * @since 1.0.0
+   */
   export interface Tester<R> extends Vitest.Test<R> {
     skip: Vitest.Test<R>
     skipIf: (condition: unknown) => Vitest.Test<R>
@@ -47,6 +53,21 @@ export namespace Vitest {
     each: <T>(
       cases: ReadonlyArray<T>
     ) => <A, E>(name: string, self: TestFunction<A, E, R, Array<T>>, timeout?: number | V.TestOptions) => void
+
+    /**
+     * @since 1.0.0
+     */
+    prop: <const S extends SchemaObj, A, E>(
+      name: string,
+      schemas: S,
+      self: TestFunction<
+        A,
+        E,
+        R,
+        [{ [K in keyof S]: Schema.Schema.Type<S[K]> }, V.TaskContext<V.RunnerTestCase<{}>> & V.TestContext]
+      >,
+      timeout?: number | V.TestOptions
+    ) => void
   }
 
   /**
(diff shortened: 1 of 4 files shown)


mikearnaldi:
why not the following?

```ts
it.effect.prop("should detect the substring", { a: Schema.String, b: Schema.String, c: Schema.String }, ({ a, b, c }) =>
    Effect.gen(function*() {
      return (a + b + c).includes(b)
    })
)
```

jessekelly881:
Yeah, I actually like that more.

jessekelly881:
@mikearnaldi Mind having another look when you get the chance?

mikearnaldi:
Looks good to me! I'll let @tim-smart and @gcanti comment / review

jessekelly881:
There's a weird issue with importing @effect/schema that I'm not quite seeing a solution to.

tim-smart (approved):

tim-smart:
Whoops wrong PR. I'll review this one shortly.

tim-smart:
Hmm we can't depend on /schema as it isn't part of the core package at this stage.

You can add it to a separate module however.

jessekelly881:
Not even as a devDependency(for every other package besides /vitest)?

tim-smart:
> Not even as a devDependency(for every other package besides /vitest)?

It just can't be used in the /vitest/index.ts module, as /effect dogfoods it.

So you can add it as /vitest/PropertyTesting, but will lose out on the chainable api.

mikearnaldi:
Then I would say let's hold off until we merge schema

jessekelly881:
Hmm. Well this is awkward.. jaja. I've updated it to hopefully fix the issue(hopefully). Still think we should wait?

mikearnaldi:
> Hmm. Well this is awkward.. jaja. I've updated it to hopefully fix the issue. Still think we should wait? What's the timeline for merging schema? If it's w/ 4.0 maybe it's worth merging this version and then updating the api once schema is merged?

Merging schema is not a breaking change for effect so most likely a matter or weeks

jessekelly881:
Gotcha. Makes sense to revert to the original version and wait then. I'll ping this again when Schema gets merged.

jessekelly881:
@tim-smart @mikearnaldi I rebased this against next-minor w/ schema merged into effect. Mind having another look?

packages/vitest/src/index.ts
```diff
@@ -36,6 +37,11 @@ export namespace Vitest {
     ): void
   }
 
+  /**
+   * @since 1.0.0
+   */
+  export type SchemaObj<A, E, R> = Array<Schema.Schema<A, E, R>> | { [K in string]: Schema.Schema<A, E, R> }
```
tim-smart, line 43: Don't think you need any generics here. Also I don't think you can assign Schema.Never to this.

jessekelly881: Arbitrary doesn't support Schema.Never so I think not being able to assign Schema.Never to it is fine. But, yeah, this can be Schema.All w/o the generics

https://github.com/Effect-TS/effect/blob/b836fc429a9aad36e7a063557d1448ed95398c85/packages/schema/src/Arbitrary.ts#L154

tim-smart (approved):

tim-smart:
Just need to rebase this again.

jessekelly881:
Done :)

tim-smart:
> Done :)

Rebase, not merge, as there are a bunch of duplicate commits atm.

jessekelly881:
There we go. Had to do some serious git magic to make that work jaja.

packages/vitest/src/internal.ts
```diff
@@ -16,6 +17,7 @@ import * as Scope from "effect/Scope"
 import * as TestEnvironment from "effect/TestContext"
 import type * as TestServices from "effect/TestServices"
 import * as Utils from "effect/Utils"
+import fc from "fast-check"
```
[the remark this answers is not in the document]
adrian-gierakowski: https://github.com/Effect-TS/effect/pull/4230

PR #3704, opened 2024-09-30
Accept TLS options for SSL configuration of PgClient

wewelll:
the PgClient should be able to accept a TLS connection object instead of a boolean for the `ssl` prop


wewelll:
done ! 

> Thanks! Could you run `pnpm changeset` and add a change log entry?

PR #3707, opened 2024-09-30
add node modules directory to `clean` script

fubhy:
I regularly need this when I come back to the repository after 1 or 2 weeks. Sometimes `pnpm` fails to properly fix up the node modules directory after an update to dependencies or the pnpm version(?) it seems.

### DIFF
--- a/scripts/clean.mjs
+++ b/scripts/clean.mjs
@@ -3,7 +3,7 @@ import * as Fs from "node:fs"
 
 const dirs = [".", ...Glob.sync("packages/*/")]
 dirs.forEach((pkg) => {
-  const files = [".tsbuildinfo", "docs", "build", "dist", "coverage"]
+  const files = [".tsbuildinfo", "docs", "build", "dist", "coverage", "node_modules"]
 
   files.forEach((file) => {
     if (pkg === "." && file === "docs") {


mikearnaldi:
this is "clean-self-destruct" lol

mikearnaldi:
maybe as an option? I when I run clean it usually doesn't require cleaning node_modules

fubhy:
Tbh. I wish `pnpm` had a full-reinstall option where it wipes all node modules across the workspace. I just went to check, it doesn't have it ... Strangely, because I run into this quite often due to the large range of projects I jump in and out of.

But yeah, we can also make that a separate command or an option.

mikearnaldi:
wondering why I never encounter the same, I think I never had to clean node_modules and my local repo is old old old, maybe nix + pnpm it's isolating it from the rest who knows

fubhy:
> wondering why I never encounter the same, I think I never had to clean node_modules and my local repo is old old old, maybe nix + pnpm it's isolating it from the rest who knows

Yah ... This is usually the kind of problem that more senior developers run into when they debug the difficult stuff.

mikearnaldi:
> > wondering why I never encounter the same, I think I never had to clean node_modules and my local repo is old old old, maybe nix + pnpm it's isolating it from the rest who knows
> 
> Yah ... This is usually the kind of problem that more senior developers run into when they debug the difficult stuff.

after senior developers there are 100x developers who just don't care and never have issues.

fubhy:
Woah, nevermind actually. I just found `npkill`.

```pnpx npkill```

fubhy:
> > > wondering why I never encounter the same, I think I never had to clean node_modules and my local repo is old old old, maybe nix + pnpm it's isolating it from the rest who knows
> > 
> > 
> > Yah ... This is usually the kind of problem that more senior developers run into when they debug the difficult stuff.
> 
> after senior developers there are 100x developers who just don't care and never have issues.

I was trolling friend

PR #3720, opened 2024-10-02
Fix Transactions in `sql-libsql`

### LINKED ISSUE #3709: Transactions in `sql-libsql` do not work

3.8.4

When working against a live Libsql database, transactions as currently implemented do not work.

Test code here:

https://github.com/Effect-TS/effect/compare/main...jkonowitch:effect:libsql-failing-transactions?expand=1

Tests should pass

### What do you see instead?

```
 DEV  v2.0.5 /home/jeff/dev/effect/packages/sql-libsql

 ❯ test/Client.test.ts (4) 2620ms
   ❯ Client (4) 2620ms
     ✓ should work
     ✓ should work with raw
     × withTransaction 1958ms
     × withTransaction rollback 2620ms
 ✓ test/Resolver.test.ts (6) 1847ms

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 2 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/Client.test.ts > Client > withTransaction
 FAIL  test/Client.test.ts > Client > withTransaction rollback
SqlError: Failed to execute statement
 ❯ catch src/LibsqlClient.ts:119:29
    117|         Effect.tryPromise({
    118|           try: () => db.execute({ sql, args: params as Array<any> }).then((results) => results.rows),
    119|           catch: (cause) => new SqlError({ cause, message: "Failed to execute statement" })
       |                             ^
    120|         })
    121| 
 ❯ ../effect/src/internal/core-effect.ts:1694:56

Caused by: LibsqlError: SQLITE_UNKNOWN: SQLite error: cannot commit - no transaction is active
 ❯ mapHranaError ../../node_modules/.pnpm/@libsql+client@0.12.0/node_modules/@libsql/client/lib-esm/hrana.js:268:16
 ❯ ../../node_modules/.pnpm/@libsql+client@0.12.0/node_modules/@libsql/client/lib-esm/http.js:76:23

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { span: undefined, code: 'SQLITE_UNKNOWN', rawCode: undefined }
Caused by: ResponseError: SQLite error: cannot commit - no transaction is active
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { span: undefined, code: 'SQLITE_UNKNOWN', proto: { message: 'SQLite error: cannot commit - no transaction is active', code: 'SQLITE_UNKNOWN' } }
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/2]⎯

 Test Files  1 failed | 1 passed (2)
      Tests  2 failed | 8 passed (10)
   Start at  16:15:08
   Duration  5.84s (transform 2.02s, setup 1.73s, collect 4.12s, tests 4.47s, environment 0ms, prepare 94ms)
```

### Additional information

_No response_

jkonowitch:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The `libsql` sdk uses a non-standard method of initiating transaction connections that is not compatible with `@effect/sql`. It expects to fully manage sending `BEGIN`, `COMMIT`, `ROLLBACK` statements as well as initiating new connections because under the hood it uses a custom websocket protocol. See 
* https://docs.turso.tech/sdk/ts/reference#interactive-transactions , 
* https://github.com/tursodatabase/libsql-client-ts/blob/e9db106651333b5111a640c1ea0d141a281d0aba/packages/libsql-client/src/hrana.ts

~Therefore, I believe we should treat `libsql` as we do `D1`, as a package that does not support transactions. Users can batch multiple statements together using the `batch` API if needed: https://docs.turso.tech/sdk/ts/reference#batch-transactions~

@thewilkybarkid realized that `withTransaction` can be easily overriden, so we are taking that approach.

## Related

- Closes #3709
- [Discord 🧵](https://discord.com/channels/795981131316985866/1134177528098078861/1290403328517476363)


thewilkybarkid:
I've just hit the same problem when using a libSQL Server (i.e. Turso); it works fine with a local SQLite (hence the tests passing). I've been looking at `sql-mssql` as it overrides `withTransaction`, but I'm unsure if doing that would be enough.

jkonowitch:
> I've just hit the same problem when using a libSQL Server (i.e. Turso); it works fine with a local SQLite (hence the tests passing). I've been looking at `sql-mssql` as it overrides `withTransaction`, but I'm unsure if doing that would be enough.

Oh very interesting. I didn't realize overrides were already an established pattern here. I mean, that provides some hope of a solution. Do you have any example code you've tried so far?

The other change I want to make to this package is to allow a live client to be provided instead of config. The reason being is that the Turso client is doing replication frame tracking as well, so if you want to use the `batch` API or otherwise interact with the database and get "read-your-writes semantics", you would need to share this client with other parts of your code.

```typescript
/**
 * @category models
 * @since 1.0.0
 */
export type LibsqlClientConfig =
  & {
    readonly spanAttributes?: Record<string, unknown> | undefined
    readonly transformResultNames?: ((str: string) => string) | undefined
    readonly transformQueryNames?: ((str: string) => string) | undefined
  }
  & ({
    _tag: "Config"
    /** The database URL.
     *
     * The client supports `libsql:`, `http:`/`https:`, `ws:`/`wss:` and `file:` URL. For more infomation,
     * please refer to the project README:
     *
     * https://github.com/libsql/libsql-client-ts#supported-urls
     */
    readonly url: string
    /** Authentication token for the database. */
    readonly authToken?: string | undefined
    /** etc... */
  } | {
    _tag: "LiveClient"
    liveClient: Libsql.Client
  })
  ```

**Comment by thewilkybarkid** [comment]
> Oh very interesting. I didn't realize overrides were already an established pattern here. I mean, that provides some hope of a solution. Do you have any example code you've tried so far?

Not yet, but it's a blocker for us at the moment so I'll be looking shortly to see if I can figure it out. (I'm new to LibSQL too!)
 
> The other change I want to make to this package is to allow a live client to be provided instead of config. The reason being is that the Turso client is doing replication frame tracking as well, so if you want to use the `batch` API or otherwise interact with the database and get "read-your-writes semantics", you would need to share this client with other parts of your code.

👍 Could make the LibSQL client available as a service too; probably outside the scope of this package though so maybe a separate `makeFromClient` constructor rather than differing config?

> Oh, and FYI I added the live libsql docker image to the tests now via `testcontainers` so we should not get any more false positive tests.

Fantastic! Might be worth running them against both LibSQL and SQLite as there seem to be more differences than I originally anticipated.

**Comment by thewilkybarkid** [comment]
I've hacked something together where transactions seem to work against both LibSQL Server and SQLite: https://github.com/thewilkybarkid/effect/commit/047ca253ee171b91ab6adba9e1a44ded97b8684a. There's probably plenty of gotchas though, and it doesn't support nested transactions/savepoints.

**Comment by tim-smart** [comment]
How does libsql handle nested transactions? I didn't see any mention of savepoints.

**Comment by thewilkybarkid** [comment]
I'm not sure the JS client does at the moment, can't see any references to it.

**Comment by jkonowitch** [comment]
Re savepoints - once you have initiated a transaction with the client, you proceed as normal and can execute `SAVEPOINT` and `ROLLBACK TO` commands as normal against the `tx` connection. I have tested this:

```typescript
import { createClient } from "@libsql/client";

const client = createClient({ url: 'http://localhost:8080' });

const tx = await client.transaction('write');

await tx.execute(`
-- Set a savepoint
SAVEPOINT sp1;
`);

await tx.execute(`
-- Insert a new user
INSERT INTO users (name) VALUES ('Alice');
`)

await tx.execute(`
  -- Set another savepoint
SAVEPOINT sp2;
`)

await tx.execute(`-- Insert another user
INSERT INTO users (name) VALUES ('Bob');
`)

await tx.execute(`-- Rollback to the first savepoint (sp1) to undo both inserts
ROLLBACK TO sp2;
`)
await tx.commit();
```
```
→  select * from users;
ID     NAME      
3      Alice
```

jkonowitch:
> I've hacked something together where transactions seem to work against both LibSQL Server and SQLite: [thewilkybarkid@047ca25](https://github.com/thewilkybarkid/effect/commit/047ca253ee171b91ab6adba9e1a44ded97b8684a). There's probably plenty of gotchas though, and it doesn't support nested transactions/savepoints.

This looks like a great start to me. Do you want to make a PR into my branch, have me copy it over, or open a new PR?

thewilkybarkid:
Happy for you to copy it over if you have time @jkonowitch!

jkonowitch:
> Happy for you to copy it over if you have time @jkonowitch!

Just did - added you as `--author` https://github.com/Effect-TS/effect/pull/3720/commits/46cdfb01f7bedcef377f1b61e2b3a924257e0b8c

I will add tests / implementation for savepoints.

jkonowitch:
@tim-smart @thewilkybarkid I think this is now ready for review.

tim-smart:
Thanks!

I ended up adding SqlClient.makeWithTransaction to ensure consistency with tracing etc.

thewilkybarkid:
Thanks both. 🙂

PR #3722, opened 2024-10-02
dx: add watch task to expose all compile errors to vscode, including not open files.

patroza:


fubhy:
I removed all our previous vscode tasks a while ago because they were unmaintained and completely broken. Which meant to me that noone was using them.

I don't mind re-adding some tasks for things like this. But the risk of these ending up in a broken state with noone noticing it for months is there.

patroza:
@fubhy I can imagine. This is a very simple and straightforward task however, and the first thing I do when opening the repo is hit cmd+shift+b :)

fubhy:
Ok. This comes with the obligatory weekly contribution from you going forward then to ensure this remains functional

PR #3724, opened 2024-10-03
Automatically wire active span from opentelemetry api

mikearnaldi:
`getActiveSpan` is set to optional to avoid breaking changes, discord discussion: https://discord.com/channels/795981131316985866/911666033885061240/1291311027144888383


mikearnaldi:
closing in favour of `Tracer.withSpanContext` given that active span is not propagated across async

PR #3743, opened 2024-10-07
add dotenv support to PlatformConfigProvider

sukovanej:


datner:
The name might carry association to some behaviors dotenv has (like variable expansion), theres also [dotenvx](https://dotenvx.com/) with its new mutli-env and encryption capabilities. I don't think this blocks the PR in any way, but we should think of the future plan with this. It might be better to just recommend people use `dotenvx` directly since from there the regular `ConfigProvider` would just work

sukovanej:
> The name might carry association to some behaviors dotenv has (like variable expansion)
this PR is somewhat a copy of dotenv and dotenv-expand codebases, so the variable expansion support is included

> theres also [dotenvx](https://dotenvx.com/) with its new mutli-env and encryption capabilities
I haven't used dotenvx myself, if by multi-env you mean providing different `.env.<suffix>` file, then that's also supported (the dotenv filename is an input of all the new functions); encryption is not, tho I guess it should be possible to implement a combinator on `ConfigProvider` dealing with encryption independently of the concrete `ConfigProvider` instance

> It might be better to just recommend people use dotenvx directly since from there the regular ConfigProvider would just work
I think both an in-code API and CLI are valid options, the advantage of an in-code approach is that I can more easily do whatever I want in terms of configuring the application, and with the proposed change it would be a matter of combining layers / effects. At the end of the day it's a question of whether I want to put the responsibility into the typescript code or the shell code, and there is unlikely a definite answer to that, but I prefer the part of my codebase where a type-checker and linter look over my shoulders.

tim-smart (changes requested):

packages/platform/src/PlatformConfigProvider.ts
```diff
@@ -111,3 +112,32 @@ export const layerFileTree = (options?: {
...
+ * @category constructors
+ * @since 1.0.0
+ */
+export const fromDotEnv: (
+  paths: string
+) => Effect.Effect<ConfigProvider.ConfigProvider, PlatformError, FileSystem.FileSystem> = internal.fromDotEnv
+
+/**
+ * Add the dotenv ConfigProvider to the environment, as a fallback to the current ConfigProvider.
+ * If the file is not found, a debug log is produced and empty layer is returned.
```
tim-smart, line 128: This statement doesn't seem to be correct atm.

sukovanej: Ouch, I removed the behaviour, nice catch. What do you think is better
- the current one, fail if the file doesn't exist
- the one described in the docstring

tim-smart: I think using .env only for dev is quite common, so a debug log + empty layer sounds good. You could add a `ensureFileExists` option to toggle the behaviour.

jessekelly881:
Why not just use: https://github.com/dotenvx/dotenvx?

```ts
dotenvx run -- tsx index.ts
```

I don't think I've ever seen a .env file used in production. And even if it is, using dotenvx would be better sense it supports encrypted .env files.

packages/platform/src/PlatformConfigProvider.ts
sukovanej: Agree, debug log it is. I think we don't need to make it configurable, this really feels like the most expected behaviour.

.changeset/neat-olives-learn.md
```diff
@@ -0,0 +1,23 @@
...
+Effect.gen(function* () {
+  const config = yield* Config.all({
+    api_url: Config.string("API_URL"),
+    api_key: Config.string("API_KEY")
+  })
+
+  console.log(`Api config: ${config}`)
+}).pipe(
+  PlatformConfigProvider.layerDotEnvAdd(".env"),
+  Effect.provide(NodeContext.layer)
```
tim-smart, line 21:
```suggestion
  Effect.provide(PlatformConfigProvider.layerDotEnvAdd(".env").pipe(
      Layer.provide(NodeContext.layer)
  )),
```

tim-smart (approved):

PR #3747, opened 2024-10-08
feat: add support for spanPrefix to HttpRpcRouter

patroza:


mikearnaldi:
we should probably make toHttpApp dual then

patroza:
> we should probably make toHttpApp dual then

how so? also `RpcRouter.toHandler` isn't.

mikearnaldi:
Adding an optional options arg is breaking, well actually it's breaking anyway

PR #3759, opened 2024-10-09
`@effect/sql-libsql` allow either a pre-built client or configuration options

jkonowitch:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR allows for a developer to provide _either_ a set of configuration options, as currently implemented, _or_ a client instance itself. It uses a discriminated union to detect what was passed in.

**Why?**

Libsql/Turso allows you to connect to either a cloud hosted or a local "embedded" replica, and provides ["read-your-writes semantics"](https://docs.turso.tech/features/embedded-replicas/introduction#read-your-writes) meaning that the client library tracks replication frames and ensures that the connection that wrote to the DB will immediately be able to read back that same data.

This depends on using the _same_ client instance. However, at the moment the client which is built inside of `@effect/sql-libsql` is not accessible outside. There are use cases, however, where it is desirable to use the client in other parts of the system outside of `@effect/sql`, but still guarantee that the entire application is reading from the same connection. The most important use case is when using the [`batch` API](https://docs.turso.tech/sdk/ts/reference#batch-transactions), a pattern not supported by `@effect/sql`, which allows you to commit an entire unit of work at once, by sending a set of SQL statements which will be executed as a transaction, without the overhead of opening an interactive transaction and sending commands one by one, with [the latency that entails](https://turso.tech/blog/batches-in-sqlite-838e0961).

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/dull-pumas-bow.md
+++ b/.changeset/dull-pumas-bow.md
@@ -1,0 +1,5 @@
+---
+"@effect/sql-libsql": minor
+---
+
+allow client instance to be passed in to LibsqlClient
--- a/.changeset/dull-pumas-bow.md
+++ b/.changeset/dull-pumas-bow.md
@@ -1,5 +1,5 @@
 ---
-"@effect/sql-libsql": minor
+"@effect/sql-libsql": patch
 ---
 
 allow client instance to be passed in to LibsqlClient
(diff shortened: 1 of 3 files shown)


tim-smart (approved):

.changeset/dull-pumas-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/sql-libsql": minor
```
tim-smart, line 2:
```suggestion
"@effect/sql-libsql": patch
```

PR #3760, opened 2024-10-09
add sql-clickhouse & clickhouse dialect

tim-smart:


anglinb:
👋 Hey Tim! Thank you implementing this! Was able to get up and running successfully! 

### Quick feedback: 

#### Consider `client.insert` instead of `client.command`
Would you be willing to consider using the `insert` api rather than the `command` api provided by `@clickhouse/client`? I'm not sure how this would differ from the existing `@effect/sql` apis but this seems to be the preferred way to insert data. When using command to insert, the query and parameters are passed via url parameters rather than the body. Additionally, the values statements are repeated relative to the number of rows inserted. (`INSERT INTO ... ({p1: String), ({p2: String})...`)  I don't believe search parameters are compressed whereas a POST request body can be compressed. So the command approach leads to more data usage, inability to compress the data over the wire and could potentially url length limits. (Additionally, [async inserts via command doesn't work in ClickHouse Cloud](https://github.com/ClickHouse/clickhouse-js/issues/339) 🙃 )

The `insert` api provided by `@clickhouse/client` bears a lot of resemblance to the `sql.insert` command provided by `@effect/sql` except that in the effect case, the writer of the query writes the `INSERT INTO table...` instead of passing a `table` parameter. Parsing the SQL feels pretty gross but maybe there is some escape hatch where client could opt in to this behavior. 

#### Allow callers to specify query_id
In ClickHouse queries can run for a very long time and unlike many database drivers queries are not automatically cancelled when the connection is interrupted. Moreover, it's actually an "endorsed" pattern to issue a command (like an `INSERT ... SELECT`), close the http connection, then poll the `system.processes` and `system.query_log` to observe the progress of the query. We use this functionality extensively for running backfills, ETLs, and data exports. Here's an example of what this looks like with the raw client: https://github.com/ClickHouse/clickhouse-js/blob/main/examples/long_running_queries_timeouts.ts 

Additionally, while you've added the `AbortController` implementation, the abort only effects the HTTP request itself, not the underlying query. I believe this is probably not the expected behavior based on how other drivers work. In order to successfully abort the query we need to signal the HTTP request to stop and then execute another command to actually kill the running process. `KILL QUERY WHERE query_id = '...'`. I'm not sure whose responsibility that should be (the caller or the `@effect/sql-clickhouse` library) but I think the continuation of the query after the fiber is suspended would catch people off guard. 

Thanks again for working on this! So excited to get this running in production!

PR #3777, opened 2024-10-13
add Redactable to Inspectable module

patroza:
I think a much better solution to https://github.com/Effect-TS/effect/pull/3721

### DIFF
--- a/packages/effect/src/Inspectable.ts
+++ b/packages/effect/src/Inspectable.ts
@@ -2,6 +2,7 @@
  * @since 2.0.0
  */
 
+import type { FiberRefs } from "./index.js"
 import { hasProperty, isFunction } from "./Predicate.js"
 
 /**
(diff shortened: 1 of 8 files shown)


packages/effect/src/Inspectable.ts
```diff
@@ -100,18 +105,42 @@ export const toStringUnknown = (u: unknown, whitespace: number | string | undefi
...
+ */
+export interface Redactable {
+  readonly [RedactableId]: (fiberRefs: FiberRefs.FiberRefs) => unknown
+}
+
+/**
+ * @since 3.10.0
+ * @category type ids
+ */
+export const RedactableId: unique symbol = Symbol.for("@effect/platform/Headers")
```
fubhy, line 141: 🫴

fubhy (dismissed):

packages/effect/src/Inspectable.ts
patroza: Im such a bad boi!

tim-smart (approved):

PR #3778, opened 2024-10-13
refactor: Alternative Effect.Service suggestion

patroza:
since we combine tags and default layers, we can go one step further and grab and pass the dependent services to the constructors.

NOTE: api is not yet finished, I would suggest we have overloads without dependencies as second argument.

at least I couldn't make it work to infer the dependencies and provide them within the same options object.


mikearnaldi:
closing as not planned

PR #3794, opened 2024-10-17
HttpApi second revision

tim-smart:


kattsushi:
It would be amazing if OpenAPI could generate schema types based on the schemas defined and used for failure responses and inputs in the endpoints for Swagger. Here’s an example:

Screenshot 2024-10-17 at 2 35 08 p m

PR #3830, opened 2024-10-23
add the literals as their own example when using Schema.Literal

jessekelly881:
It would be convenient if Schema.Literal set the examples annotation to the literals themselves.

```ts
import { Schema } from "effect";

export class Currency extends Schema.Literal("USD", "EUR").annotations({
  title: "Currency",
  examples: ["USD", "EUR"], // now set automatically
}) {}
```

### DIFF
--- a/.changeset/old-radios-unite.md
+++ b/.changeset/old-radios-unite.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+sets the literals themselves as the example annotation when using Schema.Literal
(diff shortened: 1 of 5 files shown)


gcanti:
👎  Doing this by default adds too much noise.

jessekelly881:
In the case of json schema generation it does. An idea might be to add the ability to exclude fields like examples when generating json schema docs. Especially when sending json schemas over the network you may want to exclude examples(for all schemas) as well as descriptions, etc.

PR #3851, opened 2024-10-28
add UriComponent schemas and encoding

titouancreach:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add schema (and encoding) for encodeUriComponent and decodeUriComponent. 

Very useful for storing data in the url with a Schema : 
```ts
import { Schema as S } from "effect"

const PaginationSchema = S.Struct({
  maxItemPerPage: S.Number,
  page: S.Number
})

const UrlSchema = S.compose(S.encodeUriComponent, S.parseJson(PaginationSchema))

const queryComponent = S.encodeSync(UrlSchema)({ maxItemPerPage: 10, page: 1 });


```

## Related

- Related Issue #
- Closes #


packages/effect/src/Encoding.ts
```diff
@@ -88,6 +88,23 @@ export const decodeHex = (str: string): Either.Either<Uint8Array, DecodeExceptio
  */
 export const decodeHexString = (str: string) => Either.map(decodeHex(str), (_) => Common.decoder.decode(_))
 
+/**
+ * Encodes a UTF-8 `string` into a URI component `string`.
+ *
+ * @category encoding
+ * @since 3.11.0
+ */
+export const encodeUriComponent = (str: string) => encodeURIComponent(str)
```
KhraksMamtsov, line 97: it must be wrapped as it throws [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#exceptions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#exceptions)

titouancreach: Thanks for the feedbacks, I have updated my code accordingly

PR #3859, opened 2024-10-29
Adds `withClickhouseSettings` to `@effect/sql-clickhouse`

anglinb:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds the `.withClickhouseSettings` method to `@effect/sql-clickhouse` following the pattern for `.withQueryId` to allow the caller to set per-query settings, like adding a "log_comment" to be stored in the query_log. 

## Related

- Related Issue https://github.com/Effect-TS/effect/pull/3760/commits/e48be4b2a55b8da55f13ad9ec893e2c8bdfd9843

### DIFF
--- a/packages/sql-clickhouse/src/ClickhouseClient.ts
+++ b/packages/sql-clickhouse/src/ClickhouseClient.ts
@@ -55,11 +55,11 @@ export interface ClickhouseClient extends Client.SqlClient {
   }
   readonly withClickhouseSettings: {
     (
-      settings: Clickhouse.BaseQueryParams["clickhouse_settings"]
+      settings: NonNullable<Clickhouse.BaseQueryParams["clickhouse_settings"]>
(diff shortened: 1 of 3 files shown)


packages/sql-clickhouse/src/ClickhouseClient.ts
```diff
@@ -231,28 +243,36 @@ export const make = (
           readonly values: Clickhouse.InsertValues<Readable, T>
           readonly format?: Clickhouse.DataFormat
         }) {
-          return FiberRef.getWith(currentQueryId, (queryId_) =>
```
anglinb, line 234: Wasn't sure about this nesting here so pls lmk if there's a prettier pattern to follow.

PR #3874, opened 2024-10-31
Fix WebSdk.layer not infering properly due to R and A type args being swapped

jbmusso:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

It looks like `R` and `A` type arguments were not swapped during the major breaking change of the Effect API a few moons ago.
This PR fixes this in `WebSdk.layer` so that inference isn't broken when building a layer when using an Effect.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/opentelemetry/src/WebSdk.ts
+++ b/packages/opentelemetry/src/WebSdk.ts
@@ -66,7 +66,7 @@ export const layerTracerProvider = (
  */
 export const layer: {
   (evaluate: LazyArg<Configuration>): Layer.Layer<Resource.Resource>
-  <R, E>(evaluate: Effect.Effect<Configuration, E, R>): Layer.Layer<Resource.Resource, E, R>
+  <E, R>(evaluate: Effect.Effect<Configuration, E, R>): Layer.Layer<Resource.Resource, E, R>
 } = (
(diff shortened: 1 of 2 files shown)


packages/opentelemetry/src/WebSdk.ts
```diff
@@ -66,9 +66,9 @@ export const layerTracerProvider = (
  */
 export const layer: {
   (evaluate: LazyArg<Configuration>): Layer.Layer<Resource.Resource>
-  <R, E>(evaluate: Effect.Effect<R, E, Configuration>): Layer.Layer<Resource.Resource, E, R>
+  <R, E>(evaluate: Effect.Effect<Configuration, E, R>): Layer.Layer<Resource.Resource, E, R>
```
patroza, line 69: nitpick but, but in an AER world, ER would be the correct order

tim-smart, line 69:
```suggestion
  <E, R>(evaluate: Effect.Effect<Configuration, E, R>): Layer.Layer<Resource.Resource, E, R>
```

jbmusso: Nice catch, fixed.

PR #3886, opened 2024-11-04
Avoid creation of subarray view in Base64.decode

fubhy:
@tim-smart It likely doesn't matter much but while we are here ... The outcome is the same but we avoid the creation of an additional typed array instance for the subarray view.


tim-smart:
Just needs a changeset

PR #3898, opened 2024-11-05
Fix broken basic auth base64 decoding

mattphillips:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Basic auth isn't working for the `HttpApiSecurity` as it is currently trying to base 64 decode the whole header which includes the `basic` word resulting in the parsed security credentials to always be empty.

I've quickly added this fix via Github so haven't written any tests or ran it locally but applying this patch to my compiled Effect node_modules fixes the bug. 

Please feel free to commit on-top of this, add tests or close in favour of another solution. 

## Related

- Related Issue: Haven't made one
- Closes N/A

### DIFF
--- a/packages/platform/src/HttpApiBuilder.ts
+++ b/packages/platform/src/HttpApiBuilder.ts
@@ -971,7 +971,7 @@ export const securityDecode = <Security extends HttpApiSecurity.HttpApiSecurity>
         password: Redacted.make("")
       } as any
       return HttpServerRequest.HttpServerRequest.pipe(
-        Effect.flatMap((request) => Encoding.decodeBase64String(request.headers.authorization ?? "")),
+        Effect.flatMap((request) => Encoding.decodeBase64String(request.headers.authorization?.split(" ")[1] ?? "")),
         Effect.match({
           onFailure: () => empty,
           onSuccess: (header) => {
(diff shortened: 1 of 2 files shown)


packages/platform/src/HttpApiBuilder.ts
```diff
@@ -971,7 +971,7 @@ export const securityDecode = <Security extends HttpApiSecurity.HttpApiSecurity>
         password: Redacted.make("")
       } as any
       return HttpServerRequest.HttpServerRequest.pipe(
-        Effect.flatMap((request) => Encoding.decodeBase64String(request.headers.authorization ?? "")),
+        Effect.flatMap((request) => Encoding.decodeBase64String(request.headers.authorization?.split(" ")[1] ?? "")),
```
fubhy, line 974: Nitpicky, but I'd probably use the same approach as for the Bearer token above:

```ts
(request.headers.authorization ?? "").slice(basicLen)
```

fubhy (changes requested):

packages/platform/src/HttpApiBuilder.ts
mattphillips: @fubhy I'm AFK at the moment and have no preference over stylistic changes so please feel free to amend this PR as you see best

tim-smart (approved):
:shipit:

PR #3902, opened 2024-11-05
Added `Permissions` module to `@effect/platform-browser`

SandroMaglione:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added `Permissions` module to `@effect/platform-browser` ([Permissions API](https://developer.mozilla.org/en-US/docs/Web/API/Permissions)).

### DIFF
--- a/packages/platform-browser/src/Permissions.ts
+++ b/packages/platform-browser/src/Permissions.ts
@@ -1,0 +1,52 @@
+/**
+ * Wrapper on the Permission API (`navigator.permissions`)
+ * with methods for querying status of permissions.
+ *
+ * @since 1.0.0
+ */
+
+import type { PlatformError } from "@effect/platform/Error"
+import type { Tag } from "effect/Context"
+import type * as Effect from "effect/Effect"
+import type * as Layer from "effect/Layer"
+import * as internal from "./internal/permissions.js"
+
+/**
+ * @since 1.0.0
+ * @category interface
+ */
+export interface Permissions {
+  /**
+   * Returns the state of a user permission on the global scope.
+   */
+  readonly query: <Name extends PermissionName>(
+    name: Name
+  ) => Effect.Effect<
+    // `name` is identical to the name passed to Permissions.query
+    // https://developer.mozilla.org/en-US/docs/Web/API/PermissionStatus
+    Omit<PermissionStatus, "name"> & { name: Name },
+    PlatformError
+  >
+}
+
+/**
+ * @since 1.0.0
+ * @category constructor
+ */
+export const make: (
+  impl: Permissions
+) => Permissions = internal.make
+
+/**
+ * @since 1.0.0
+ * @category tag
+ */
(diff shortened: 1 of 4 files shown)


packages/platform/src/Error.ts
```diff
@@ -106,6 +114,7 @@ export type SystemErrorReason =
   | "Unknown"
   | "WouldBlock"
   | "WriteZero"
+  | "InvalidState"
```
SandroMaglione, line 117: https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query#exceptions

packages/platform-browser/src/Permissions.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * Wrapper on the Permission API (`navigator.permissions`)
```
tim-smart, line 2: You can move the docs to the Permission interface itself.

packages/platform-browser/src/Permissions.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { Tag } from "effect/Context"
+import type * as Effect from "effect/Effect"
+import type * as Layer from "effect/Layer"
+import * as internal from "./internal/permissions.js"
+
+/**
+ * @since 1.0.0
+ * @category interface
+ */
+export interface Permissions {
```
tim-smart, line 18: This will need a `TypeId` to prevent it from being merged.

packages/platform-browser/src/Permissions.ts
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Returns the state of a user permission on the global scope.
+   */
+  readonly query: <Name extends PermissionName>(
+    name: Name
+  ) => Effect.Effect<
+    // `name` is identical to the name passed to Permissions.query
+    // https://developer.mozilla.org/en-US/docs/Web/API/PermissionStatus
+    Omit<PermissionStatus, "name"> & { name: Name },
+    PlatformError
```
tim-smart, line 28: I would add a separate `PermissionsError` here, as this module isn't really cross-platform.

tim-smart (changes requested):

SandroMaglione:
@tim-smart do we have a way to implement tests for browser APIs? Would it make sense to abstract `navigator` in a separate service for testing purposes?

packages/platform-browser/src/Permissions.ts
SandroMaglione: Should we add it also for `Clipboard`? I can open another PR for that

https://github.com/Effect-TS/effect/blob/5392047b3398cb4ae82e87ec1d77279c23661e8a/packages/platform-browser/src/Clipboard.ts#L15-L22

tim-smart: Yes, if you want to PR the change go for it.

tim-smart (approved):

PR #3907, opened 2024-11-06
Schema: BigDecimal Arbitrary restricted to the range of 0-18

arijoon:
## Type

- [x] Optimization
- [x] Bug

## Description

In the case of very large scales, the BigDecimal can no longer be formatted into a string, which breaks in the default encoding (if trying to serialize BigDecimal to a string).

this is extremely limting the usecase of the arbitrary as the default Decimal will serialize to a string

Additionally a negative scale is invalid. A decimal cannot have a negative number of decimal points therefore the scale must always be positive or 0 (bug)

### DIFF
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -7174,7 +7174,8 @@ const bigDecimalPretty = (): pretty_.Pretty<bigDecimal_.BigDecimal> => (val) =>
   `BigDecimal(${bigDecimal_.format(bigDecimal_.normalize(val))})`
 
 const bigDecimalArbitrary = (): LazyArbitrary<bigDecimal_.BigDecimal> => (fc) =>
-  fc.tuple(fc.bigInt(), fc.integer()).map(([value, scale]) => bigDecimal_.make(value, scale))
+  fc.tuple(fc.bigInt(), fc.integer({ min: 0, max: 18 }))
+    .map(([value, scale]) => bigDecimal_.make(value, scale))
 
 /**
  * @category BigDecimal constructors
--- a/.changeset/fast-readers-try.md
+++ b/.changeset/fast-readers-try.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Schema.BigDecimal Arbitrary's scale limited to the range 0-18


fubhy:
Ah yeah, that makes sense :-P

fubhy:
> ## Type
> * [x]  Optimization
> * [x]  Bug
> 
> ## Description
> In the case of very large scales, the BigDecimal can no longer be formatted into a string, which breaks in the default encoding (if trying to serialize BigDecimal to a string).
> 
> this is extremely limting the usecase of the arbitrary as the default Decimal will serialize to a string
> 
> Additionally a negative scale is invalid. A decimal cannot have a negative number of decimal points therefore the scale must always be positive or 0 (bug)

This is not correct though. A negative scale is indeed valid. The `scale` either shifts the decimal point to the right (adding zeros) if negative, or the the left if positive.

E.g:

```ts
BigDecimal.make(1234n, -5) // 123400000 (1234 * 10^5)
```

Regardless, I'm in favour of this change because it is likely closer to what users expect (decimal values and not large integers) when generating arbitraries.

Thoughts @gcanti ?

packages/effect/src/Schema.ts
```diff
@@ -7174,7 +7174,8 @@ const bigDecimalPretty = (): pretty_.Pretty<bigDecimal_.BigDecimal> => (val) =>
   `BigDecimal(${bigDecimal_.format(bigDecimal_.normalize(val))})`
 
 const bigDecimalArbitrary = (): LazyArbitrary<bigDecimal_.BigDecimal> => (fc) =>
-  fc.tuple(fc.bigInt(), fc.integer()).map(([value, scale]) => bigDecimal_.make(value, scale))
+  fc.tuple(fc.bigInt(), fc.integer({ min: 0, max: 18 }))
+    .map(([value, scale]) => bigDecimal_.make(value, scale))
```
fubhy, line 7178: Currently (both before and after the proposed change), this can generate non-normalized BigDecimal's of the form:

```ts
BigDecimal.make(1234000n, 0) // Would normalize to BigDecimal.make(1234n, -4)
```

I'm wondering if we should normalize the generated values.

```ts
BigDecimal.normalize(BigDecimal.make(value, scale))
```

fubhy: It's not a big deal becaues the module still renders / formats them correctly and also calculates correctly with them (of course). Normalization just matters for the internal representation and if you print the raw properties of the object.

gcanti (approved):

fubhy:
> In the case of very large scales, the BigDecimal can no longer be formatted into a string, which breaks in the default encoding (if trying to serialize BigDecimal to a string).

We will have to address this in a follow-up. The current `BigDecimal.format` implementation fails with `RangeError` or `OutOfMemory` because of the way it generates the zero padding. There's no way to solve this whilst keeping the zero padded notation for arbitrary length BigDecimals. Imho we should fall back to scientific notation beyond the `DEFAULT_PRECISION` (100 decimals).

arijoon:
> This is not correct though. A negative scale is indeed valid. The `scale` either shifts the decimal point to the right (adding zeros) if negative, or the the left if positive.

You're right, I did not consider this case because bigint can handle large numbers, hence assumed scaled is only used if decimal point exists, but it is true that it can use a negative scale for a more optimal storage of very large numbers

gcanti:
Thanks @arijoon @fubhy

PR #3911, opened 2024-11-07
Add scientific notation for BigDecimal formatting

fubhy:
Giulio is going to drive this over the finish line (property tests, fixes for `fromNumber`, etc.)


packages/effect/src/BigDecimal.ts
```diff
@@ -811,20 +811,71 @@ export const fromBigInt = (n: bigint): BigDecimal => make(n, 0)
...
+ */
+export const fromNumber: (n: number) => BigDecimal = unsafeFromNumber
+
+/**
+ * Creates a `BigDecimal` from a `number` value.
+ *
+ * It is not recommended to convert a floating point number to a decimal directly,
+ * as the floating point representation may be unexpected.
+ *
+ * Returns `None` if the number is not finite.
```
gcanti, line 852:
```suggestion
 * Returns `None` if the number is not finite (`NaN`, `+Infinity` or `-Infinity`).
```

gcanti (changes requested):

gcanti (approved):

KhraksMamtsov:
sorry - but it seems we need to remember to process such a case - `fromString("0.0000e2e1")` now it returns `{ _id: 'BigDecimal', value: '0', scale: -16 }`

fubhy:
> sorry - but it seems we need to remember to process such a case - `fromString("0.0000e2e1")` now it returns `{ _id: 'BigDecimal', value: '0', scale: -16 }`

That's an invalid input. When you say "we need to remember to process such a case" you mean "we need to throw in this case" right?

KhraksMamtsov:
I think it is necessary to return `Option.none()` for any incorrect input (Correct me if I'm wrong)
In any case, it's probably worth reflecting this behavior in the tests.

fubhy:
> I think it is necessary to return `Option.none()` for any incorrect input (Correct me if I'm wrong) In any case, it's probably worth reflecting this behavior in the tests.

Yes for sure. I was just trying to understand if you wanted a different kind of behavior (wasn't sure if you meant that this should be valid or not)

fubhy:
@KhraksMamtsov Happy with this adjustment?

PR #3922, opened 2024-11-10
bug: http response stream not interruptible

patroza:
Preparations:
- `cd packages/platform-node/examples`
- `tsx ./http-tag-router.ts`
- visit http://localhost:3000/users/stream in browser

Steps:
- send ctrl+c to the server process

Observed:
Process don't exit

Expected:
Process exits


patroza:
It looks like one way to do it, is via:
```ts
const GetUsers = UserRouter.useScoped((router) =>
  Effect.gen(function*() {
    const ps = yield* PubSub.unbounded<Uint8Array>()
    yield* Effect.addFinalizer(() => ps.shutdown) // doesn't help
```
but shouldn't the http server terminate the listening to the Stream instead?
the above would never allow streams to be finished, even if you wanted the request to be uninterruptible

fubhy:
Closing this inactive draft/WIP PR. Please open a fresh PR when the change is ready for review.

PR #3929, opened 2024-11-12
Prompt multi select

joepjoosten:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added prompt to select multiple choices


joepjoosten:
Scratchpad to see it in action:

```ts
import { Command, Prompt } from "@effect/cli"
import { NodeContext, NodeRuntime } from "@effect/platform-node"
import { Array, Console, Effect, pipe } from "effect"

const run = Command.make("example", {}, () =>
  Effect.gen(function*() {
    const choices = pipe(Array.range(1, 20), Array.map((n) => ({ title: `Option ${n}`, value: n })))
    const selectMulti = yield* Prompt.selectMulti({
      message: "select a option",
      choices
    })
    yield* Console.log(`You selected: ${selectMulti.join(", ")}`)
  })).pipe(
    Command.run({
      name: "Test",
      version: "1.0.0"
    })
  )

const main = Effect.suspend(() => run(globalThis.process.argv))

main.pipe(
  Effect.provide(NodeContext.layer),
  Effect.tapErrorCause(Effect.logError),
  NodeRuntime.runMain
)
```

IMax153:
@joepjoosten - this looks good to me! The only additional request I have is that we make this a patch release since we're only adding a feature, which is backward compatible.

joepjoosten:
Thanks!

Should i add min/max choices support? Then it can be used in the wizard when there is a repeated choice option?

IMax153:
Sure! That would be a cool addition :)

joepjoosten:
I'm also not really fond of using default text for the select all/none and inverse selection. This is not very international because there in English. I don't know if there are universal symbols for select all/none and inverse?

joepjoosten:
And nothing more difficult than naming things: "selectMulti", i'm still not happy with this name...

joepjoosten:
I've added the min max functionality, and renamed it to multiSelect

IMax153:
> The only additional request I have is that we make this a patch release since we're only adding a feature, which is backward compatible.

@joepjoosten - only one thing left to take care of before we can merge this in 👍

joepjoosten:
Sorry, i missed that comment. Is this what you mean? (see latest commit)

IMax153:
Thanks @joepjoosten !

PR #3935, opened 2024-11-13
`Data.Error#fromCause` constructor

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Data.Error#fromCause` and `Data.TaggedError#fromCause` have been added
`fromCause` takes all constructor options except property with name 'cause' and returns function which accepts specified type of cause (unknown by default) and creates an instance of the error.
Basically, it is convenient sugar for `(cause) => new MyError({ props: 1, cause })`.

```ts
import { Data } from "Effect"

class MyError extends Data.Error {}

class MySecondError extends Data.TaggedError("MySecondError")<{
  prop: number
  cause: MyError
}> {}

MyError.fromCause() // (cause: unknown) => MyError
MySecondError.fromCause({ prop: 2 }) // (cause: MyError) => MySecondError

Effect.try({
  try: () => {
    throw new Error("External Error")
  },
  catch: MyError.fromCause()
}).pipe(
  Effect.mapError(
    MySecondError.fromCause({ prop: 2 })
  )
)
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/breezy-dogs-press.md
+++ b/.changeset/breezy-dogs-press.md
@@ -1,0 +1,29 @@
+---
+"effect": minor
+---
+
+`Data.Error#fromCause` and `Data.TaggedError#fromCause` have been added
+`fromCause` takes all constructor options except property with name 'cause'
+and returns function which accept specified type of cause (unknown by default) and created an instantiate the error
+Basically, it is sugar for `(cause) => new MyError({ props: 1, cause })`
+
+```ts
+import { Data } from "Effect"
+
+class MyError extends Data.TaggedError("MyError") {}
+
+class MySecondError extends Data.TaggedError("MySecondError")<{
+  prop: number
+  cause: MyError
+}> {}
+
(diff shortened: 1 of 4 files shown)


mikearnaldi:
not sure about that, feels ad-hoc

KhraksMamtsov:
image

Well, I expected a more violent reaction XD
I still think it's convenient for poinfree error building
but it seems the community doesn't need it.

mikearnaldi:
> image
> 
> Well, I expected a more violent reaction XD I still think it's convenient for poinfree error building but it seems the community doesn't need it.

Pointfree is heavily discouraged (not to say forbidden) within effect

datner:
FWIW I like the idea, I think it's good to have api proposals from the API like these even if they don't pass 😄

PR #3936, opened 2024-11-13
allow DateTime.makeZoned to default to the local time zone

tim-smart:


packages/effect/src/DateTime.ts
```diff
@@ -474,15 +474,21 @@ export const unsafeMake = <A extends DateTime.Input>(input: A): DateTime.Preserv
  *
  * DateTime.unsafeMakeZoned(new Date(), { timeZone: "Europe/London" })
  */
-export const unsafeMakeZoned = (input: DateTime.Input, options: {
-  readonly timeZone: number | string | TimeZone
+export const unsafeMakeZoned = (input: DateTime.Input, options?: {
+  readonly timeZone?: number | string | TimeZone | undefined
   readonly adjustForTimeZone?: boolean | undefined
 }): Zoned => {
+  if (!options?.timeZone && isDateTime(input) && isZoned(input)) {
```
fubhy, line 481:
```suggestion
  if (options?.timeZone  === undefined && isDateTime(input) && isZoned(input)) {
```

fubhy (changes requested):

fubhy:
It would appear that this is BROKEN

PR #3938, opened 2024-11-13
Implement Effect.fn to define traced functions

mikearnaldi:
Allows:

```ts

const logExample = Effect.fn("logExample")(
  function*<N extends number>(n: N) {
    yield* Effect.annotateCurrentSpan("n", n)
    yield* Effect.logInfo(`got: ${n}`)
    yield* Effect.fail(new Error())
  },
  Effect.delay("1 second")
)

Effect.runFork(
  logExample(100).pipe(
    Effect.catchAllCause(Effect.logError)
  )
)
```

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -7192,7 +7192,10 @@ export namespace fn {
  */
 export const fn: (
   name: string,
-  options?: Omit<FunctionWithSpanOptions, "name">
+  options?: {
+    readonly options?: Omit<FunctionWithSpanOptions, "name">
+    readonly captureStackTrace?: boolean | undefined
+  }
 ) => fn.Gen & fn.NonGen = (name, options) =>
 // @ts-ignore
 (body: Function, ...pipeables: Array<any>) => {
@@ -7224,11 +7227,9 @@ export const fn: (
       }
       return res
     }
-    let opts: any
-    if (options) {
-      opts = { captureStackTrace, ...options }
-    } else {
-      opts = { captureStackTrace }
+    let opts: any = options?.captureStackTrace === false ? {} : { captureStackTrace }
+    if (options?.options) {
+      opts = { ...opts, ...options.options }
     }
     return withSpan(
       core.suspend(() => res(...args)),
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -25,6 +25,7 @@ import { dual } from "./Function.js"
 import type * as HashMap from "./HashMap.js"
 import type * as HashSet from "./HashSet.js"
 import type { TypeLambda } from "./HKT.js"
+import * as internalCause from "./internal/cause.js"
 import * as _console from "./internal/console.js"
 import { TagProto } from "./internal/context.js"
 import * as effect from "./internal/core-effect.js"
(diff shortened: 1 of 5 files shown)


packages/effect/src/Effect.ts
```diff
@@ -6912,3 +6912,323 @@ export declare namespace Service {
...
+        ? gen(() => internalCall(() => body.apply(self, args)))
+        : body.apply(self, args)
+      for (const x of pipeables) {
+        res = x(res)
+      }
+      return res
+    }
+    const opts: any = (options && "captureStackTrace" in options) ? options : { captureStackTrace, ...options }
+    return withSpan(
+      core.suspend(() => res(...args)),
```
tim-smart, line 7229:
```suggestion
    const opts: any = (options && "captureStackTrace" in options) ? options : { captureStackTrace, ...options }
    let effect: Effect<any, any, any>
    return withSpan(
      core.suspend(() => effect ??= res(...args)),
```

Could cache the evaluation here.

tim-smart (commented):
Seems good to go. Left one comment.

packages/effect/src/Effect.ts
mikearnaldi: `res` isn't assumed to be pure so caching would lead to potentially different results

mikearnaldi:
in theory if we make this compatible conditionally with every function it could be auto-applied via a babel-plugin, will think about it a bit more

packages/effect/src/Effect.ts
tim-smart: I was think to perserve the semantics of calling a function:

```ts
const someFn = Effect.fn("someFn")(...)

// currently
const effect = someFn(...)

// function will be called once per iteration
// when really it should only be called once
Effect.forever(effect)
```

mikearnaldi: not sure we can, if the function is specified using a generator the whole body may contain side-effects and repeating the effect is expected to repeat the side-effects

tim-smart: Ah yeah, we can only do it for the non-generator case. Just noticed a small bug where pipeables like `Effect.catchAllCause` would not catch defects from calling the function. This should fix it:

```diff
diff --git a/packages/effect/src/Effect.ts b/packages/effect/src/Effect.ts
index a2b4c2cc3..48f1057d6 100644
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -7215,20 +7215,18 @@ export const fn: (
     // @ts-ignore
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     const self = this
-    const res = (...args: Array<any>) => {
-      let res = isGeneratorFunction(body)
+    let effect: Effect<any, any, any>
+    try {
+      effect = isGeneratorFunction(body)
         ? gen(() => internalCall(() => body.apply(self, args)))
         : body.apply(self, args)
-      for (const x of pipeables) {
-        res = x(res)
-      }
-      return res
+    } catch (defect) {
+      effect = die(defect)
+    }
+    for (const x of pipeables) {
+      effect = x(effect)
     }
     const opts: any = (options && "captureStackTrace" in options) ? options : { captureStackTrace, ...options }
-    return withSpan(
-      core.suspend(() => res(...args)),
-      name,
-      opts
-    )
+    return withSpan(effect, name, opts)
   }
 }
```

mikearnaldi: hmm this has the same issue if a defect arises in a pipeable, wondering if we shouldn't just suspend everything

tim-smart: I think I have the behaviour correct now.

tim-smart (approved):

PR #3939, opened 2024-11-13
`HttpApi` annotation with additional schemas

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added the ability to annotate the `HttpApi` with additional schemas
Which will be taken into account when generating `components.schemas` section of `OpenApi` schema

```ts
import { HttpApi } from "@effect/platform"

HttpApi.empty.annotate(
  HttpApi.AdditionalSchemas,
  [
    Schema.Struct({
      contentType: Schema.String,
      length: Schema.Int
    }).annotations({
      identifier: "ComponentsSchema"
    })
  ]
)
/**
 {
  "openapi": "3.0.3",
  ...
  "components": {
    "schemas": {
      "ComponentsSchema": {...},
      ...
  },
  ...
  }
 */
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/platform/src/HttpApi.ts
+++ b/packages/platform/src/HttpApi.ts
@@ -349,3 +349,12 @@ const extractMembers = (
   }
   return members
 }
+
+/**
+ * @since 1.0.0
+ * @category tags
+ */
+export class ApiAdditionalSchemas extends Context.Tag("@effect/platform/HttpApi/ApiAdditionalSchemas")<
+  ApiAdditionalSchemas,
+  ReadonlySet<Schema.Schema.All>
+>() {}
--- a/packages/platform/src/HttpApi.ts
+++ b/packages/platform/src/HttpApi.ts
@@ -354,7 +354,7 @@ const extractMembers = (
  * @since 1.0.0
  * @category tags
  */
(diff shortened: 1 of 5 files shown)


packages/platform/src/HttpApi.ts
```diff
@@ -349,3 +349,12 @@ const extractMembers = (
   }
   return members
 }
+
+/**
+ * @since 1.0.0
+ * @category tags
+ */
+export class ApiAdditionalSchemas extends Context.Tag("@effect/platform/HttpApi/ApiAdditionalSchemas")<
```
tim-smart, line 357: I think we can just call this `AdditionalSchemas`.

packages/platform/src/HttpApi.ts
```diff
@@ -349,3 +349,12 @@ const extractMembers = (
...
+  ApiAdditionalSchemas,
+  ReadonlySet<Schema.Schema.All>
```
tim-smart, line 359: `ReadonlyArray` is fine. They will get de-duplicated anyway if the identifier clashes.

tim-smart (changes requested):

tim-smart (approved):

PR #3943, opened 2024-11-14
Add support for timezones in Cron

fubhy:


packages/effect/src/Cron.ts
```diff
@@ -276,58 +287,76 @@ export const next = (cron: Cron, now?: Date): Date => {
...
+  for (let i = 0; i < 10_000; i++) {
+    const parts = dateTime.toParts(current)
+
+    if (restrictMonths && !months.has(parts.month)) {
+      current = dateTime.setParts(current, {
+        month: dateTime.getPart(current, "month") + 1,
+        day: 1,
+        hours: 0,
+        minutes: 0
+      })
```
fubhy, line 311: I have not profiled this but it's probably not very efficient compared to the previous mutative implementation. Maybe it's possible to implement a version of `DateTime.mutate` that gives access to the actual `DateTime` object instead of the internal `Date` representation.

PR #3945, opened 2024-11-14
"format" property in `JSONSchema` and `OpenApiSpec`

KhraksMamtsov:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- `JsonSchema` interfaces have been updated
- added tests that confirm that the field `"format"` is reflected in the final `JSONSchema` and `OpenApi` spec.
- updated `UUID` schema

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/small-swans-juggle.md
+++ b/.changeset/small-swans-juggle.md
@@ -3,6 +3,9 @@
 "@effect/platform": patch
 ---
 
-- `JsonSchema` interfaces have been updated
-- added tests that confirm that the field `"format"` is reflected in the final `JSONSchema` and `OpenApi` spec.
-- updated `UUID` schema
+- JSONSchema module
+  - add `format?: string` optional field to `JsonSchema7String` interface
+- Schema module
+  - add custom json schema annotation to `UUID` schema including `format: "uuid"`
+- OpenApiJsonSchema module
+  - add `format?: string` optional field to `String` and ` Numeric` interfaces
--- a/.changeset/small-swans-juggle.md
+++ b/.changeset/small-swans-juggle.md
@@ -1,0 +1,8 @@
+---
+"effect": minor
+"@effect/platform": patch
+---
+
+- `JsonSchema` interfaces have been updated
+- added tests that confirm that the field `"format"` is reflected in the final `JSONSchema` and `OpenApi` spec.
+- updated `UUID` schema
--- a/packages/effect/src/JSONSchema.ts
+++ b/packages/effect/src/JSONSchema.ts
@@ -97,7 +97,6 @@ export interface JsonSchema7Numeric extends JsonSchemaAnnotations {
   exclusiveMinimum?: number
   maximum?: number
   exclusiveMaximum?: number
-  format?: string
 }
 
 /**
--- a/packages/effect/src/JSONSchema.ts
+++ b/packages/effect/src/JSONSchema.ts
@@ -85,6 +85,7 @@ export interface JsonSchema7String extends JsonSchemaAnnotations {
   minLength?: number
   maxLength?: number
   pattern?: string
+  format?: string
 }
 
 /**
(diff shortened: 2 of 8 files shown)


packages/effect/src/JSONSchema.ts
```diff
@@ -96,6 +97,7 @@ export interface JsonSchema7Numeric extends JsonSchemaAnnotations {
   exclusiveMinimum?: number
   maximum?: number
   exclusiveMaximum?: number
+  format?: string
```
gcanti, line 100: Isn't `"format"` only applicable to `"type": "string"`?

KhraksMamtsov: yes, you are right
But then how to implement adding `"format": "int32"` to the OpenApi Spec?
it might be worth adding an annotation to `HttpApiSchema.Format` and read it in the `OpenApiJsonSchema.make`?

.changeset/small-swans-juggle.md
```diff
@@ -0,0 +1,8 @@
+---
+"effect": minor
+"@effect/platform": patch
+---
+
+- `JsonSchema` interfaces have been updated
+- added tests that confirm that the field `"format"` is reflected in the final `JSONSchema` and `OpenApi` spec.
+- updated `UUID` schema
```
gcanti, line 8:
```suggestion
- JSONSchema module
  - add `format?: string` optional field to `JsonSchema7String` interface
- Schema module
  - add custom json schema annotation to `UUID` schema including `format: "uuid"`
- OpenApiJsonSchema module
  - add `format?: string` optional field to `String` and ` Numeric` interfaces
```

packages/effect/src/JSONSchema.ts
gcanti: > But then how to implement adding "format": "int32" to the OpenApi Spec?

Could you clarify? Are you asking about adding `"format": "int32"` to the output of `OpenApiJsonSchema.make` when it's called on `Schema.Int`?

```ts
import { OpenApiJsonSchema } from "@effect/platform";
import { Schema } from "effect";

console.log(OpenApiJsonSchema.make(Schema.Int));
/*
{
  '$schema': 'http://json-schema.org/draft-07/schema#',
  type: 'integer',
  description: 'an integer',
  title: 'Int'
}
*/
```

KhraksMamtsov: As far as I understand, the OpenAPI specification, unlike the JSONSchema, allows you to annotate numeric types with certain formats eg `int32` `float` ... https://swagger.io/specification/v3/#data-types

It is now assumed that the `.annotate({ jsonSchema: ... })` annotation will be used when generating the JSONSchema, and adding OpenAPI fields (format: 'int32') in this case would lead to the generation of an incorrect JSONSchema.

at the same time, JSONSchema annotations can be used to generate OpenAPI specifications as they are similar in many ways.
I see that there are already some annotations like `HttpApiSchema.AnnotationStatus` that affect the generated specification.
so I suggest adding other annotation `HttpApiSchema.AnnotationFormat` that will affect the behavior of `OpenApiJsonSchema.make`and take precedence over annotations for jsonSchema when calculating the 'format' field
```ts
Schema.Int.annotations({ [HttpApiSchema.AnnotationFormat]: 'int32' });
/*
{"type": "integer", ... , "format": "int32"}
*/
```

gcanti: > I see that there are already some annotations like HttpApiSchema.AnnotationStatus that affect the generated definition

Does it? I don't see any tests in https://github.com/Effect-TS/effect/blob/main/packages/platform/test/OpenApiJsonSchema.test.ts related to a "status" info. I'm not sure if the `HttpApiSchema` module has anything to do with `OpenApiJsonSchema.make`. I'm not very familiar with this part, I'll defer the review of the platform portion of this PR to @tim-smart

PR #3947, opened 2024-11-15
Allow data classes to be branded with an additional type parameter

mikearnaldi:


mikearnaldi:
cc @gcanti we should make a list of APIs that benefit from branding, basically all APIs that produce class instances

fubhy:
This PR is still marked draft and currently has merge conflicts. Could you confirm whether this is still intended to move forward and update it if so?

tim-smart:

PR #3966, opened 2024-11-20
allow setting a custom Redacted.toString

jessekelly881:
Allow setting a custom string when creating redacted values e.g. "REDACTED_TOKEN" which is logged as:

`level=INFO msg="permission granted" user=Perry token=REDACTED_TOKEN`

```ts
Redacted.make("abc", "REDACTED_TOKEN")
```

This is useful for keeping tract of redacted values between log statements and is a feature provided by a few other logging libraries.

### DIFF
--- a/.changeset/short-monkeys-compete.md
+++ b/.changeset/short-monkeys-compete.md
@@ -1,0 +1,13 @@
+---
+"effect": minor
+---
+
+allow users to provide a custom name for Redacted values on creation
+
+```ts
+Redacted.make("1234567890", "API_KEY")
+```
+
+example logged output:
+
+level=INFO msg="permission granted" user=Perry token=API_KEY
--- a/packages/effect/src/Redacted.ts
+++ b/packages/effect/src/Redacted.ts
@@ -65,12 +65,12 @@ export const isRedacted: (u: unknown) => u is Redacted<unknown> = redacted_.isRe
  * @example
  * import { Redacted } from "effect"
  *
- * const API_KEY = Redacted.make("1234567890")
+ * const API_KEY = Redacted.make("1234567890", "API_KEY")
  *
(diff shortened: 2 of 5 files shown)


jessekelly881:
A better approach might be to use a `ref = {}` prop on the Redacted object that actually points to the entry in the redactedRegistry so that the Redacted can be mapped using a `setString` fn.

KhraksMamtsov:
I started working on this idea yesterday)  I agree that it is convenient.
but I came to a slightly different implementation.

```ts
proto = {
...
  label: "redacted",
  toString() {
    return `<${this.label}>`
  },
  toJSON() {
    return `<${this.label}>`
  },
  [NodeInspectSymbol]() {
    return `<${this.label}>`
  },
...
}
export const make = <T>(
  value: T,
  options?: {
    label?: string | undefined
  } | undefined
): Redacted.Redacted<T> => {
  const redacted = Object.create(proto)
  if (options?.label !== undefined) {
    redacted.label = options?.label
  }
  redactedRegistry.set(redacted, value)
  return redacted
}
```
Could you please integrate this idea into this places as well? 
- `Schema.Redacted/RedactedFromSelf`
- `Config.redacted`
- `Options.redacted` in cli package

jessekelly881:
I think having a custom string is much better than forcing the "<>" wrapper. I was looking at the Config.redacted implementation as well and I think it makes more sense to use a Redacted.setString fn. But as far as I can tell that requires having a .ref prop that actually points to the value rather than using the Redacted instance itself as the ref.

fubhy:
Thanks for the PR. This still looks potentially actionable, but it currently has merge conflicts. Could you please rebase/update the branch and resolve the conflicts so maintainers can continue review?

tim-smart:

PR #3977, opened 2024-11-21
`HttpApiClient.group` & `HttpApiClient.endpoint`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`HttpApiClient.group` & `HttpApiClient.endpoint` have been added
This makes it possible to create `HttpApiClient` for some part of the `HttpApi`
This eliminates the need to provide all the dependencies for the entire `HttpApi` - but only those necessary for its specific part to work

- ~~I do not know how to avoid duplication in the runtime part (besides passing other hooks inside the `make` function)- if there are ideas, I will be glad to implement them~~
- renamed the `HttpApiGroup.ContextWithName` & `HttpApiGroup.Context` in the `HttpApiGroup.MiddlewareContextWithName` & `HttpApiGroup.Middleware` - it seems to be a more accurate naming
- `HttpApiGroup.ContextWithName` & `HttpApiGroup.Context` now return context without middlewares

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/platform/src/HttpApiClient.ts
+++ b/packages/platform/src/HttpApiClient.ts
@@ -6,6 +6,7 @@ import * as Effect from "effect/Effect"
 import { identity } from "effect/Function"
 import * as Option from "effect/Option"
 import * as ParseResult from "effect/ParseResult"
+import type * as Predicate from "effect/Predicate"
 import * as Schema from "effect/Schema"
 import type * as AST from "effect/SchemaAST"
 import type { Scope } from "effect/Scope"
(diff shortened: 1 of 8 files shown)


packages/platform/src/HttpApiClient.ts
```diff
@@ -187,11 +209,134 @@ export const make = <Groups extends HttpApiGroup.Any, ApiError, ApiR>(
...
+> => {
+  const client: Record<string, Record<string, any>> = {}
+  return makeClient(api, {
+    ...options,
+    onGroup({ group }) {
+      if (group.topLevel) return
+      client[group.identifier] = {}
+    },
+    onEndpoint({ endpoint, endpointFn, group }) {
+      ;(group.topLevel ? client : client[group.identifier])[endpoint.name] = endpointFn
```
tim-smart, line 247: It might make more sense to allow a predicate to be passed in.

KhraksMamtsov: I didn't understand - could you explain or suggest changes)
now this callback (`makeClient.onEndpoint`) serves to perform quite diverse tasks and I'm not sure that you can do with a predicate

tim-smart (approved):

PR #3996, opened 2024-11-25
switch to @types/bun

fubhy:


packages/sql-sqlite-bun/tsconfig.examples.json
```diff
@@ -12,6 +12,6 @@
     "tsBuildInfoFile": ".tsbuildinfo/examples.tsbuildinfo",
     "rootDir": "examples",
     "noEmit": true,
-    "types": ["bun-types"]
+    "types": ["bun"]
```
lishaduck, line 15:
```suggestion
```
@types is implicitly included, so this is redundant and even prevents the other @types/* packages from being included.

fubhy: No this is not redundant. This is a mono repo and we want to be explicit about which types are available to which scopes within the repo. We do not want to leak types into places where they should not be used. This is also why our `tsconfig.base.json` explicitly specifies an empty `types: []` array.

PR #3998, opened 2024-11-25
ensure fiber observers are cleared after exit to prevent memory leaks

tim-smart:


packages/effect/src/internal/fiberRuntime.ts
```diff
@@ -812,6 +812,7 @@ export class FiberRuntime<in out A, in out E = never> extends Effectable.Class<A
     for (let i = this._observers.length - 1; i >= 0; i--) {
       this._observers[i](exit)
     }
+    this._observers = []
```
davidgoli, line 815: That would certainly do it!

PR #4004, opened 2024-11-27
fix behavour of Stream.partition to match the types

tim-smart:


leonitousconforti:
```ts
import { Effect, Stream } from "effect";

const partition = Stream.partition(Stream.range(1, 10), (n) => n % 2 === 0);

const program = Effect.scoped(
    Effect.gen(function* () {
        const [evens, odds] = yield* partition;
        console.log(yield* Stream.runCollect(evens));
        console.log(yield* Stream.runCollect(odds));
    })
);

await Effect.runPromise(program);
```

Docs read: "Partition a stream using a predicate. The first stream will contain all element evaluated to true and the second one will contain all element evaluated to false."

But this logs:
{ _id: 'Chunk', values: [ 1, 3, 5, 7, 9 ] }
{ _id: 'Chunk', values: [ 2, 4, 6, 8, 10 ] }

I feel like this reversed the behavior and its was correct before?

PR #4010, opened 2024-11-27
Optimize cron

### LINKED ISSUE #3932: Optimize `Cron.next` algorithm

The algorithm for finding the next cron time can be optimized quite a bit. Currently, we start each iteration again from the top even if we have not advanced the day/month yet.

...

_No response_

fubhy:
This should *massively* improve the efficiency (number of loops) for the `Cron.next` lookup. With this, I think we should also be able to support seconds & milisecond granularity.

That said, it's already possible to compose schedules with second or milisecond ticks within an active Cron interval... So I'm not sure how valuable that'd be tbh.

Closes #3932

### DIFF
--- a/packages/effect/src/Cron.ts
+++ b/packages/effect/src/Cron.ts
@@ -338,11 +338,13 @@ const daysInMonth = (date: Date): number =>
  * @since 2.0.0
  */
 export const next = (cron: Cron, now?: DateTime.DateTime.Input): Date => {
+  const tz = Option.getOrUndefined(cron.tz)
   const zoned = dateTime.unsafeMakeZoned(now ?? new Date(), {
-    timeZone: Option.getOrUndefined(cron.tz)
+    timeZone: tz
   })
 
-  const adjustDst = dateTime.isUtc(zoned) ? constVoid : (current: Date) => {
+  const utc = tz !== undefined && dateTime.isTimeZoneNamed(tz) && tz.id === "UTC"
+  const adjustDst = utc ? constVoid : (current: Date) => {
     const adjusted = dateTime.unsafeMakeZoned(current, {
       timeZone: zoned.zone,
       adjustForTimeZone: true
--- a/packages/effect/src/Cron.ts
+++ b/packages/effect/src/Cron.ts
@@ -342,16 +342,16 @@ export const next = (cron: Cron, now?: DateTime.DateTime.Input): Date => {
     timeZone: Option.getOrUndefined(cron.tz)
   })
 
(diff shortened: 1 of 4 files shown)


fubhy:
@IMax153 has agreed to review this.

Screenshot 2024-11-27 at 19 45 23

tim-smart:
I think some more tests that cover off daylight savings, leap years etc. would be good.

IMax153:
What Tim said :)

packages/effect/test/Cron.test.ts
```diff
@@ -95,4 +132,34 @@ describe("Cron", () => {
...
+    const start = make("2024-03-31 00:00:00")
+    const sequence = Cron.sequence(parse("30 * * * *", berlin), start)
+    const next = () => sequence.next().value
+
+    deepStrictEqual(next(), make("2024-03-31 00:30:00"))
+    deepStrictEqual(next(), make("2024-03-31 01:30:00"))
+    deepStrictEqual(next(), make("2024-03-31 02:30:00"))
+    deepStrictEqual(next(), make("2024-03-31 03:30:00")) // Get's stuck here (time of transition)
+    deepStrictEqual(next(), make("2024-03-31 04:30:00")) // This is never reached
+  })
```
fubhy, line 164: You are right, @tim-smart ... It does not handle transitions across daylight savings times currently :(

fubhy:
> I think some more tests that cover off daylight savings, leap years etc. would be good.

Can you verify if this makes any sense now? I believe it does but I am not super confident ...

packages/effect/src/Cron.ts
```diff
@@ -285,66 +338,99 @@ export const match = (cron: Cron, date: DateTime.DateTime.Input): boolean => {
...
+  const zoned = dateTime.unsafeMakeZoned(now ?? new Date(), {
+    timeZone: Option.getOrUndefined(cron.tz)
+  })
 
-  const restrictMinutes = minutes.size !== 0
-  const restrictHours = hours.size !== 0
-  const restrictDays = days.size !== 0
-  const restrictMonths = months.size !== 0
-  const restrictWeekdays = weekdays.size !== 0
+  const adjustDst = dateTime.isUtc(zoned) ? constVoid : (current: Date) => {
```
tim-smart, line 345: .isUtc will always be false here

fubhy: Gotcha. The intent here was to check if the Cron instance is set to UTC. I adapted it to cover that now.

tim-smart (approved):

Issue #3986, opened 2024-11-23
Tools for working with FormData


MrOxMasTer:
### What is the problem this feature would solve?

FormData validation

### What is the feature you are proposing to solve the problem?

Use schemas to validate formData

### What alternatives have you considered?

https://www.npmjs.com/package/zod-form-data

MrOxMasTer:
I'm not so versed in Effect yet that I can come up with anything, but I guess you could probably do something like this:
```ts
import { Schema as S } from “effect”;

const TodoSchema = S.Struct({
    id: S.String,
    content: S.String
});

const TodoFormDataSchema = S.formData(Todo);
 const decoded = TodoFormDataSchema.decodeUnknownPromise(TodoFormDataSchema)

const action = async (formData: FormData) => {
    const decode = await decoded(formData) // -> { id: “...”, content: “...” };
 
}

await action();
```

MrOxMasTer:
You should add more types related to `FormData`, these are:
![image](https://github.com/user-attachments/assets/3f7abe28-86b3-426f-b643-71f13213d240)

```ts
const SchemaFormData = S.FormData(S.Struct({
  text: S.Text(),
  numeric: S.Numeric(),
  checkbox: S.Checkbox(),
  file: S.File(),
  repeatable: S.Repeatable(),
  repeatableOfType: S.RepeatableOfType()
}))
```

datner:
These exist under the [`Multipart`](https://effect-ts.github.io/effect/platform/Multipart.ts.html) module, and formdata has supporting API in a server and in a client as can be seen in `HttpBody`, `HttpClientRequest`, `HttpServerRequest`, `HttpServerResponse`, and various modules of the `HttpApi` family. All from `@effect/platform`.

Also, text is just `Schema.String`, or if pushed can be `Schema.NonEmptyTrimmedString` or `Schema.OptionFromNonEmptyTrimmedString`.
numeric is `Schema.NumberFromString`,
checkbox:
```ts
Schema.transformLiterals(["on", true], ["", false]).pipe(
  Schema.optionalWith({ default: () => false }),
);
```
File exists, only thing that is missing is `Repeatable` or more exactly support for `Schema.Array`

MrOxMasTer:
[...]

PR #4011, opened 2024-11-27
`FormData` and `FormDataFromSelf`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I tried to solve this issue https://github.com/Effect-TS/effect/issues/3986

```ts
import { Schema } from 'effect'

const fdSchema = S.FormData(S.Struct({
    num: S.NumberFromString
}))
```
the interesting part was that when encoding, information about an empty array is lost - I solved this using ast-based parsing - I'm not sure if this is the right or complete solution

## Related

- Related Issue # https://github.com/Effect-TS/effect/issues/3986
- Closes #

### DIFF
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -4513,7 +4513,7 @@ function getFieldsTypes(ast: AST.AST, fieldTypes: {
 function compileFormDataToObject(ast: AST.AST) {
   const fieldsTypes = getFieldsTypes(ast, { arrays: new Set(), regular: new Set() })
 
-  return (fd: FormData): Record<string, FormDataEntryValue | ReadonlyArray<FormDataEntryValue>> => {
+  return (fd: FormData): Record<string, string | Blob | ReadonlyArray<string | Blob>> => {
     const obj: Record<string, any> = {}
     fieldsTypes.arrays.forEach((arrayFieldKey) => (obj[arrayFieldKey] = []))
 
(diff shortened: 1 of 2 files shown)


IMax153:
cc @gcanti specifically for input here

.changeset/rare-ducks-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
+---
+
+`FormData` and `FormDataFromSelf` have benn added
```
patroza, line 5: `been`

packages/effect/src/Schema.ts
```diff
@@ -4466,6 +4466,198 @@ export const split = (separator: string): transform<typeof String$, Array$<typeo
...
+
+function compileFormDataToObject(ast: AST.AST) {
+  const fieldsTypes = getFieldsTypes(ast, { arrays: new Set(), strings: new Set() })
+
+  return (fd: FormData): Record<string, string | ReadonlyArray<string>> => {
+    const obj: Record<string, any> = {}
+    fieldsTypes.arrays.forEach((arrayFieldKey) => (obj[arrayFieldKey] = []))
+
+    fd.forEach((value, key) => {
+      if (typeof value !== "string") return
```
tim-smart, line 4521: I think we should also handle File / Blob instances.

KhraksMamtsov:
@gcanti do you have any thoughts on this?
it seems there is a request for this schema from the community, so I thought it would make sense to land it in effect

gcanti:
I think a good starting point would be to write a detailed proposal in the linked issue (as suggested by its label). It's unclear to me which use cases are being addressed, which array formats will be supported (dot notation? bracket notation? both? neither?), and why this should be a concern for Schema (I don't dismiss it, but it would be helpful to have evidence for this). For instance, for the sake of discussion, why not simply use something like https://github.com/fabian-hiller/decode-formdata and validate its output instead?

nikelborm:
Leaving it here, because I think it's related
https://github.com/inato/inato-form/

PR #4019, opened 2024-11-28
Fix issue where `identifier` is ignored, closes #4012

### LINKED ISSUE #4012: From Discord: Setting `identifier` on a schema with a refinement is a no-op
# Summary
In the chat, a user inquires whether it is intentional that setting the `identifier` on a schema in Effect Typescript is ignored when a refinement (`filter`) is applied to it. They provide an example where they try to annotate a schema with an `identifier`, but the generated JSON schema does not include the `identifier`, only the refinement property. Another community member asks for the version, and the library author responds, indicating that this behavior seems like a bug.

Key takeaways:
- There might be a bug in the Effect Typescript library where setting an `identifier` on a schema with refinements does not work as expected.
- The issue is observed in a specific scenario where annotations are applied after a refinement.
- The library author acknowledged the behavior as unexpected, suggesting it is unintended.

# Discord thread

https://discord.com/channels/795981131316985866/1311111474130976823

gcanti:
# Fix 1

JSONSchema: Correct the output order when generating a JSON Schema from a Union that includes literals and primitive schemas.

Before

```ts
import { JSONSchema, Schema } from "effect"

const schema = Schema.Union(Schema.Literal(1, 2), Schema.String)

console.log(JSON.stringify(JSONSchema.make(schema), null, 2))
/*
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "anyOf": [
    {
      "type": "string"
    },
    {
      "enum": [
        1,
        2
      ]
    }
  ]
}
*/
```

After

```ts
import { JSONSchema, Schema } from "effect"

const schema = Schema.Union(Schema.Literal(1, 2), Schema.String)

console.log(JSON.stringify(JSONSchema.make(schema), null, 2))
/*
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "anyOf": [
    {
      "enum": [
        1,
        2
      ]
    },
    {
      "type": "string"
    }
  ]
}
*/
```

# Fix 2

JSONSchema: Fix issue where `identifier` is ignored when a refinement is applied to a schema, closes #4012

Before

```ts
import { JSONSchema, Schema } from "effect"

const schema = Schema.NonEmptyString

console.log(JSON.stringify(JSONSchema.make(schema), null, 2))
/*
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "string",
  "description": "a non empty string",
  "title": "NonEmptyString",
  "minLength": 1
}
*/
```

After

```ts
import { JSONSchema, Schema } from "effect"

const schema = Schema.NonEmptyString

console.log(JSON.stringify(JSONSchema.make(schema), null, 2))
/*
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$ref": "#/$defs/NonEmptyString",
  "$defs": {
    "NonEmptyString": {
      "type": "string",
      "description": "a non empty string",
      "title": "NonEmptyString",
      "minLength": 1
    }
  }
}
*/
```

# Fix 3

JSONSchema: Use identifier with Class APIs to create a `$ref` instead of inlining the schema.

Before

```ts
import { JSONSchema, Schema } from "effect"

class A extends Schema.Class<A>("A")({
  a: Schema.String
}) {}

console.log(JSON.stringify(JSONSchema.make(A), null, 2))
/*
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "required": [
    "a"
  ],
  "properties": {
    "a": {
      "type": "string"
    }
  },
  "additionalProperties": false
}
*/
```

After

```ts
import { JSONSchema, Schema } from "effect"

class A extends Schema.Class<A>("A")({
  a: Schema.String
}) {}

console.log(JSON.stringify(JSONSchema.make(A), null, 2))
/*
{
  "$ref": "#/$defs/A",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$defs": {
    "A": {
      "type": "object",
      "required": [
        "a"
      ],
      "properties": {
        "a": {
          "type": "string"
        }
      },
      "additionalProperties": false
    }
  }
}
*/
```

# Fix 4

OpenApiJsonSchema: Use the experimental `JSONSchema.makeWithOptions` API for implementation.

# Fix 5

Add missing `jsonSchema` annotations to the following filters:

- `lowercased`
- `capitalized`
- `uncapitalized`
- `uppercased`

Before

```ts
import { JSONSchema, Schema } from "effect"

const schema = Schema.Struct({
  a: Schema.Uppercased
})

console.log(JSON.stringify(JSONSchema.make(schema), null, 2))
/*
throws:
Error: Missing annotation
details: Generating a JSON Schema for this schema requires a "jsonSchema" annotation
schema (Refinement): Uppercased
*/
```

After

```ts
import { JSONSchema, Schema } from "effect"

const schema = Schema.Uppercased

console.log(JSON.stringify(JSONSchema.make(schema), null, 2))
/*
Output:
{
  "$ref": "#/$defs/Uppercased",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$defs": {
    "Uppercased": {
      "type": "string",
      "description": "an uppercase string",
      "title": "Uppercased",
      "pattern": "^[^a-z]*$"
    }
  }
}
*/
```


patroza:
before:

```ts
export interface ISOCodeBrand extends Simplify<Brand.Brand<"ISOCode"> & NonEmptyString2kBrand> {
}

export type ISOCode = string & ISOCodeBrand

export const ISOCode = S.String
  .pipe(
    S.length(2),
    S.uppercased(),
    S.filter((s): s is ISOCode => /^[A-Z]{2}$/.test(s), {
      identifier: "ISOCode",
      title: "ISOCode",
      description: "a valid ISOCode like 'US'",
      jsonSchema: {}
    })
  )
```

causing in a nested schema:
```
Error: Missing annotation
at path: ["address"]["country"]
details: Generating a JSON Schema for this schema requires a "jsonSchema" annotation
schema (Refinement): an uppercase string
```

after:
```ts
export const ISOCode = S.String
  .pipe(
    S.length(2),
    S.uppercased({ jsonSchema: {} }), // here
    S.filter((s): s is ISOCode => /^[A-Z]{2}$/.test(s), {
      identifier: "ISOCode",
      title: "ISOCode",
      description: "a valid ISOCode like 'US'",
      jsonSchema: {}
    })
  )
```

patroza:
btw, the above schema rendering revealed that NullOr seems strangely represented:
`console.log(JSON.stringify(JSONSchema.make(S.NullOr(S.String)), undefined, 2))`

```json
{
  "anyOf": [
    {
      "type": "string"
    },
    {
      "enum": [
        null
      ]
    }
  ],
  "$schema": "http://json-schema.org/draft-07/schema#"
}
```

expected instead:
```json
{
  "anyOf": [
    {
      "type": "string"
    },
    {
      "type": "null"
    }
  ],
  "$schema": "http://json-schema.org/draft-07/schema#"
}
```

should I create a new issue?

gcanti:
@patroza It looks like the filters:

- `lowercased`
- `capitalized`
- `uncapitalized`
- `uppercased`

are missing the required JSON Schema annotations. I will fix this.

AFAIK `{"type":"null"}` is not a valid JSON Schema in **Draft 7** (however, it could be used when targeting Draft 2019-09 or OpenAPI)

packages/effect/src/JSONSchema.ts
```diff
@@ -541,69 +556,82 @@ const go = (
...
     case "Transformation": {
-      // Properly handle S.parseJson transformations by focusing on
-      // the 'to' side of the AST. This approach prevents the generation of useless schemas
-      // derived from the 'from' side (type: string), ensuring the output matches the intended
-      // complex schema type.
       if (isParseJsonTransformation(ast.from)) {
-        return go(ast.to, $defs, true, path)
+        switch (options.parseJsonStrategy) {
+          case "to":
+            return go(ast.to, $defs, handleIdentifier, path, options)
```
tim-smart, line 623: Hmm wouldn't this generate invalid schemas? If a schema is expecting a json "string" this would result in a parse error.

gcanti: @tim-smart the special case was introduced to address [https://github.com/Effect-TS/effect/issues/3086](https://github.com/Effect-TS/effect/issues/3086), but I suppose this behavior shouldn't be recursive as it is now. It should only apply once at the top level

PR #4027, opened 2024-11-30
Typecheck path components

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
I think in the next iteration it will be better to set the path and schema for it together in order to avoid storing useless `PathInput` in type parameters of `HttpApiEndpoint`

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/flat-moons-greet.md
+++ b/.changeset/flat-moons-greet.md
@@ -3,3 +3,28 @@
 ---
 
 typecheck path components
+
+```ts
+import { Schema } from "effect"
+import { HttpApiEndpoint } from "@effect/platform"
+
+const endpoint = HttpApiEndpoint.get("findById", "/:id")
+/**
+ * Type 'Struct<{}>' is not assignable to type
+ * '["Missing path components:", "id"]'
+ */
+endpoint.setPath(Schema.Struct({}))
+
+/**
+ * Type 'Struct<{ id: typeof NumberFromString; id2: typeof NumberFromString; }>'
+ * is not assignable to type
+ * '["Redundant path components:", "id2" | "id3"]'.
+ */
+endpoint.setPath(
+  Schema.Struct({
+    id: Schema.NumberFromString,
+    id2: Schema.NumberFromString,
+    id3: Schema.NumberFromString
+  })
+)
+```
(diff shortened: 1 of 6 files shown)


tim-smart:
Eventually I want to add something like this:

```ts

interface Endpoint<Name extends string, Path extends ReadonlyArray<Schema.Schema.Any>> {
  readonly name: Name
  readonly path: Path
}

type AnySchemaString = Schema.Schema<any, string, never> | Schema.Schema<any, string, any>

class HttpApiEndpoint {
  static get<Name extends string>(_name: Name): <const Path extends ReadonlyArray<AnySchemaString>>(
    strings: TemplateStringsArray,
    ...args: Path
  ) => Endpoint<Name, Path> {
    return (_strings, ..._args) => undefined as any
  }
}

export const endpoint: Endpoint<
  "findById",
  readonly [typeof Schema.NumberFromString]
> = HttpApiEndpoint.get("findById")`/api/${Schema.NumberFromString}`
```

Just needs to be explored further to see where the limitations are.

KhraksMamtsov:
Yes - it's very elegant
the only point is that I would use aliases rather than positioning for components.
aliases seem to be more intuitive and readable.
```ts
class HttpApiEndpoint {
  static get<Name extends string>(
    _name: Name,
  ): <
    const Path extends ReadonlyArray<
      Record<string /* represents alias */, AnySchemaString>
    >,
  >(
    strings: TemplateStringsArray,
    ...args: Path
  ) => Endpoint<Name, Path> {
    return (_strings, ..._args) => undefined as any;
  }
}

export const endpoint: Endpoint<
  "findById",
  readonly [typeof Schema.NumberFromString]
> = HttpApiEndpoint.get("findById")`/api/directory/${{
  directoryId: Schema.NumberFromString,
}}/file/${{ fileId: Schema.NumberFromString }}`;

// client
findById({
  path: { directoryId: 1, fileId: 2 },
  // instead of
  // path: [1, 2]
});

// server
handler: ({ path }) => {
  path.directoryId;
  path.fileId;
};
// instead of
// handler: ({ path }) => {
//   path[0]
//   path[1]
// };
```

I am sure that at the type level we can guarantee disjoint alias names and that there is only one schema field in the record

tim-smart:
https://github.com/Effect-TS/effect/pull/4035

PR #4034, opened 2024-11-30
repro for deadlock on worker forkScoped

patroza:
this is a repro for a deadlock that @schickling is experiencing.

`cd packages/platform-node`
`pnpm tsx examples/worker.ts`

Expected result: 
- worker closes, process exits with "Error: hello"

Observed result:
- worker closes, but nothing happens further

replace `forkScoped` with `forkDaemon`: everything is fine


tim-smart:
❤️

patroza:
fixed in https://github.com/Effect-TS/effect/pull/4052

PR #4041, opened 2024-12-01
fix docs for Stream.partition

tim-smart:


leonitousconforti:
Thanks, was just going to do this once I realized that it was just the docs that were wrong

PR #4051, opened 2024-12-02
Update isArray type declaration for T | T[]

forno:
Fix under pattern.

```typescript
const a: int | int[] = [0]

if (Array.isArray(a)) {
  const b = a; // it is int[] | unknown[]. But I expect int[]
}
```

## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Now declaration make `T[] | unknown[]` from `T | T[]`.
But fixed version make just `T[] | T[]` 🎉

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Array.ts
+++ b/packages/effect/src/Array.ts
@@ -502,8 +502,8 @@ export const scanRight: {
  * @since 2.0.0
  */
 export const isArray: {
-  (self: unknown): self is Array<unknown>
   <T>(self: T): self is Extract<T, ReadonlyArray<any>>
+  (self: unknown): self is Array<unknown>
 } = Array.isArray
 
 /**


mikearnaldi:
it breaks test, also unsure the type would be correct, what about `Array<A> | Array<Array<A>>`?

forno:
> it breaks test, also unsure the type would be correct, what about `Array<A> | Array<Array<A>>`?

You’re absolutely right—when handling something like `Array<A> | Array<Array<A>>`, it would indeed resolve to `Array<A>`, which I hadn’t considered before. Thank you for pointing that out.

I also noticed the test failures, and while I’m not entirely sure of the cause, it’s clear that this approach has introduced more unintended side effects than I expected.  

Would it be alright if I withdrew this pull request and instead re-framed it as an issue to gather more information and explore potential solutions? I feel that might be a more productive way forward without causing additional disruptions.  

Thank you again for your time and valuable feedback—I really appreciate it!

mikearnaldi:
> > it breaks test, also unsure the type would be correct, what about `Array<A> | Array<Array<A>>`?
> 
> You’re absolutely right—when handling something like `Array<A> | Array<Array<A>>`, it would indeed resolve to `Array<A>`, which I hadn’t considered before. Thank you for pointing that out.
> 
> I also noticed the test failures, and while I’m not entirely sure of the cause, it’s clear that this approach has introduced more unintended side effects than I expected.
> 
> Would it be alright if I withdrew this pull request and instead re-framed it as an issue to gather more information and explore potential solutions? I feel that might be a more productive way forward without causing additional disruptions.
> 
> Thank you again for your time and valuable feedback—I really appreciate it!

as you prefer, I don't have much insight and think the current version is ok

PR #4067, opened 2024-12-03
Fix template literal parser throw on string literals

Avaq:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Fix:

```ts
Schema.TemplateLiteralParser('asd')
```

...resulting in:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'replace')
    at Object.escape (./node_modules/effect/dist/cjs/RegExp.js:45:33)
    at Object.getTemplateLiteralCapturingRegExp (./node_modules/effect/dist/cjs/SchemaAST.js:1718:27)
    at Object.TemplateLiteralParser (./node_modules/effect/dist/cjs/Schema.js:492:18)
```

### DIFF
--- a/packages/effect/test/Schema/Schema/TemplateLiteralParser.test.ts
+++ b/packages/effect/test/Schema/Schema/TemplateLiteralParser.test.ts
@@ -20,6 +20,80 @@ schema (SymbolKeyword): symbol`)
     expect(schema.params).toStrictEqual(params)
   })
 
+  describe("string literal based schemas", () => {
+    it("decoding", async () => {
+      const schema = Schema.TemplateLiteralParser("foo")
+      await Util.expectDecodeUnknownSuccess(schema, "foo", ["foo"])
+    })
+
+    it("encoding", async () => {
+      const schema = Schema.TemplateLiteralParser("foo")
+      await Util.expectEncodeSuccess(schema, ["foo"], "foo")
+    })
+  })
+
+  describe("literal based schemas", () => {
+    it("decoding", async () => {
+      const schema = Schema.TemplateLiteralParser(Schema.Literal("foo"))
+      await Util.expectDecodeUnknownSuccess(schema, "foo", ["foo"])
+    })
+
+    it("encoding", async () => {
+      const schema = Schema.TemplateLiteralParser(Schema.Literal("foo"))
+      await Util.expectEncodeSuccess(schema, ["foo"], "foo")
+    })
+  })
+
+  describe("literal union based schemas", () => {
+    it("decoding", async () => {
+      const schema = Schema.TemplateLiteralParser(Schema.Literal("foo", "bar"))
+      await Util.expectDecodeUnknownSuccess(schema, "foo", ["foo"])
+      await Util.expectDecodeUnknownSuccess(schema, "bar", ["bar"])
+    })
+
+    it("encoding", async () => {
+      const schema = Schema.TemplateLiteralParser(Schema.Literal("foo", "bar"))
+      await Util.expectEncodeSuccess(schema, ["foo"], "foo")
+      await Util.expectEncodeSuccess(schema, ["bar"], "bar")
+    })
+  })
+
+  describe("union of literals based schemas", () => {
+    it("decoding", async () => {
+      const schema = Schema.TemplateLiteralParser(Schema.Union(Schema.Literal("foo")))
+      await Util.expectDecodeUnknownSuccess(schema, "foo", ["foo"])
+    })
+
+    it("encoding", async () => {
+      const schema = Schema.TemplateLiteralParser(Schema.Union(Schema.Literal("foo")))
+      await Util.expectEncodeSuccess(schema, ["foo"], "foo")
+    })
+  })
+
+  describe("complex literal schemas", () => {
+    it("decoding", async () => {
+      const schema = Schema.TemplateLiteralParser(
+        Schema.Union(Schema.Literal("foo", "bar"), Schema.Literal("baz")),
+        "cux"
+      )
+
+      await Util.expectDecodeUnknownSuccess(schema, "foocux", ["foo", "cux"])
+      await Util.expectDecodeUnknownSuccess(schema, "barcux", ["bar", "cux"])
+      await Util.expectDecodeUnknownSuccess(schema, "bazcux", ["baz", "cux"])
+    })
+
+    it("encoding", async () => {
+      const schema = Schema.TemplateLiteralParser(
+        Schema.Union(Schema.Literal("foo", "bar"), Schema.Literal("baz")),
+        "cux"
+      )
+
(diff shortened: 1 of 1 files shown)


packages/effect/src/Schema.ts
```diff
@@ -820,7 +820,17 @@ export const TemplateLiteralParser = <Params extends array_.NonEmptyReadonlyArra
     }
   }
   const from = TemplateLiteral(...encodedSchemas as any)
-  const re = AST.getTemplateLiteralCapturingRegExp(from.ast as AST.TemplateLiteral)
```
Avaq, line 823: The issue is that this type coercion doesn't tell the truth. The `TemplateLiteral` constructor has a code path where it returns an `AST.Literal` instead of an `AST.TemplateLiteral`.

packages/effect/src/SchemaAST.ts
```diff
@@ -2075,6 +2075,13 @@ export const getTemplateLiteralCapturingRegExp = (ast: TemplateLiteral): RegExp
   return new RegExp(pattern)
 }
 
+/**
+ * @since 3.?.?
+ */
+export const getLiteralCapturingRegExp = (ast: Literal): RegExp => {
+  return new RegExp(`^(${regexp.escape(String(ast.literal))})$`)
```
Avaq, line 2082: `ast.literal` is a broad type, not just `string`; wasn't sure what to do with that

packages/effect/test/Schema/Schema/TemplateLiteralParser.test.ts
```diff
@@ -20,6 +20,18 @@ schema (SymbolKeyword): symbol`)
     expect(schema.params).toStrictEqual(params)
   })
 
+  describe("string literal based schemas", () => {
+    it("decoding", async () => {
+      const schema = Schema.TemplateLiteralParser(Schema.Literal("foo"), "bar")
+      await Util.expectDecodeUnknownSuccess(schema, "foobar", ["foo", "bar"])
```
Avaq, line 26: While this test doesn't fail with `Cannot read properties of undefined` anymore, as it does in the parent commit, it still fails on validation.

packages/effect/src/SchemaAST.ts
```diff
@@ -2075,6 +2075,13 @@ export const getTemplateLiteralCapturingRegExp = (ast: TemplateLiteral): RegExp
   return new RegExp(pattern)
 }
 
+/**
+ * @since 3.?.?
```
Avaq, line 2079: Need to version this function

packages/effect/src/Schema.ts
```diff
@@ -820,7 +820,17 @@ export const TemplateLiteralParser = <Params extends array_.NonEmptyReadonlyArra
...
+
+  const re = from.ast instanceof AST.TemplateLiteral
+    ? AST.getTemplateLiteralCapturingRegExp(from.ast)
+    : from.ast instanceof AST.Literal
+    ? AST.getLiteralCapturingRegExp(from.ast)
+    : null
+
+  if (re === null) {
+    throw new Error("TemplateLiteral produced unexpected AST")
+  }
```
gcanti, line 832: You can use `AST.isLiteral` here, like:

```ts
const ast = from.ast as AST.TemplateLiteral | AST.Literal
const re = AST.isLiteral(ast) ? AST.getLiteralCapturingRegExp(ast) : AST.getTemplateLiteralCapturingRegExp(ast)
```

packages/effect/src/SchemaAST.ts
gcanti: Since this is a bug fix, let's go with version `3.11.3`.

packages/effect/test/Schema/Schema/TemplateLiteralParser.test.ts
gcanti: That's expected, a successful test would be:

```ts
await Util.expectDecodeUnknownSuccess(schema, "foobar", ["foobar"])
```

because the two literals are combined into `"foobar"`.

This is another bug, namely a type-level bug:

```ts

const schema = Schema.TemplateLiteralParser(Schema.Literal("foo"), "bar")

// type Type = readonly ["foo", "bar"] // <= this is wrong
type Type = typeof schema.Type
```

gcanti (changes requested):

packages/effect/test/Schema/Schema/TemplateLiteralParser.test.ts
gcanti: I don't think it's possible at the moment. At first glance, we would need to change how `TemplateLiteralSpan` is modeled. Specifically, we would need to modify this part: https://github.com/Effect-TS/effect/blob/ec6cb5daa843f76b4f9119b58c6a2074d7145138/packages/effect/src/SchemaAST.ts#L1096 to add support for `AST.Literal` types other than `StringKeyword | NumberKeyword`.

Avaq: What's adding to the weirdness is that the `encode` side passes the tests just fine _while using `["foo", "bar"]`_. So if we would "fix" the type error and let it `decode` to `["foobar"]`, then its own decode type isn't assignable to the thing it needs for encoding. So we would have to also change the encoding behaviour.

I believe it simply shouldn't be combining literal strings, at least not for the Parser.

gcanti: Let me investigate this further. If we need to modify `TemplateLiteralSpan`, the situation becomes a bit more complicated, as I'll need to adjust other parts of the code accordingly.

packages/effect/src/Schema.ts
Avaq: I updated it to this, and found out a little later that `AST.TemplateLiteral | AST.Literal` is _also incorrect_. There is yet another code path where `TemplateLiteral()` returns an `AST.Union`:

```ts
S.TemplateLiteralParser(S.Literal("foo", "bar"))
```

Avaq: Maybe we should bite the bullet and create `AST.getCapturingRegExp(ast: AST.AST)`

Avaq:
@gcanti I removed the commit that attempts to fix the bugs, and instead pushed a bunch of additional test cases that were failing, even with the fix in place.

The test cases represent how I expect the StringLiteralParser to work with various types of literal inputs.

packages/effect/test/Schema/Schema/TemplateLiteralParser.test.ts
Avaq, line 1: Note that none of the tests I added captured how this thing should behave when given literal numbers, booleans, nulls, etc.

gcanti:
> The test cases represent how I expect the StringLiteralParser to work with various types of literal inputs.

FYI all of them pass in my branch

PR #4076, opened 2024-12-04
SchemaAST: fix `TemplateLiteral` model

gcanti:
@Avaq it took me a while (touching the AST module is a delicate task...)


gcanti:
@Avaq I forgot to mention that you can try out this PR by running `pnpm add https://pkg.pr.new/Effect-TS/effect@5023249`

PR #4082, opened 2024-12-05
Allow use of the URL object with libSQL

thewilkybarkid:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Allows URL objects to be in `sql-libsql`, so that it's compatible with `Config.url()`.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/big-hairs-give.md
+++ b/.changeset/big-hairs-give.md
@@ -1,0 +1,5 @@
+---
+"@effect/sql-libsql": patch
+---
+
+Allow use of the URL object with libSQL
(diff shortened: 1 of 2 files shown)


thewilkybarkid:
Although I have just found that using a relative `file:` string (https://docs.turso.tech/local-development#sqlite) is parsed by `URL` as an absolute rather than relative URI, so it's not safe to use `Config.url()` for parsing those. (`file:some.db` ends up as `file:///some.db`)

PR #4099, opened 2024-12-09
Allow passing fast-check parameters to prop function

mikearnaldi:
```ts
it.effect.prop(
  "adds context",
  [realNumber],
  ([num]) =>
    Effect.gen(function* () {
      const foo = yield* Foo
      expect(foo).toEqual("foo")
      return num === num
    }),
  {},
  { numRuns: 200 }
)
```

### DIFF
--- a/.changeset/four-ghosts-decide.md
+++ b/.changeset/four-ghosts-decide.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Update fast-check to latest version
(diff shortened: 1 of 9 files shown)


jessekelly881:
It might make sense to combine the props instead of adding a second arg. Say, `{ timeout: 500, fastCheck: { numRuns: 200 } }`

mikearnaldi:
> It might make sense to combine the props instead of adding a second arg. Say, `{ timeout: 500, fastCheck: { numRuns: 200 } }`

agreed, updated

tim-smart (approved):

PR #4110, opened 2024-12-10
added String.truncate

### LINKED ISSUE #4103: add `String.truncate` utility

Being able to truncate text, which is quite useful in for instance `@effect/cli` apps. I'm using it to make sure that the `multiSelect` doesn't overflow and cause glitches when rendering.

`String.truncate(text: string, maxLength: number) => string`

I currently have a utility function for this. But I kind of expected `String` module to contain it. I feel like if Effect wants to be JavaScript's standard library, then it's a useful function to have.

jessekelly881:
Closes #4103

### DIFF
--- a/packages/effect/src/String.ts
+++ b/packages/effect/src/String.ts
@@ -768,7 +768,7 @@ export const truncate: {
         length: number
 
         /**
-         * The separator to use to truncate the string.
+         * The separator to use to truncate the string. Use " " to truncate evenly on a word boundary.
          */
         separator?: string
         /**
--- a/packages/effect/src/String.ts
+++ b/packages/effect/src/String.ts
@@ -749,7 +749,7 @@ const isLineBreak2 = (char0: string, char1: string): boolean => char0.charCodeAt
 const linesSeparated = (self: string, stripped: boolean): LinesIterator => new LinesIterator(self, stripped)
 
 /**
- * Truncates a string to a specified length and by default adds the omission string (e.g. "...") to the end of the string.
+ * Truncates a string to a specified length and by default adds the omission string "..." to the end of the string.
  * @since 3.12.0
  *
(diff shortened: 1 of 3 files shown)


packages/effect/src/String.ts
```diff
@@ -747,3 +747,93 @@ const isLineBreak = (char: string): boolean => {
...
+ * @since 3.12.0
+ *
+ * @example
+ * import { truncate } from "effect/String";
+ * truncate("Hello World!", 5); // "Hello..."
+ */
+export const truncate: {
+  (
+    str: string,
+    lengthOrOptions:
```
tim-smart, line 762: I think we should just have an options object instead of a union. Also the member should be `readonly`.

packages/effect/src/String.ts
```diff
@@ -747,3 +747,78 @@ const isLineBreak = (char: string): boolean => {
...
+      readonly length: number
+      readonly separator?: string
+      readonly omission?: string
+    }
+  ) => {
+    if (str.length <= options.length) {
+      return str
+    }
+
+    let subString = str.slice(0, options.length)
```
tim-smart, line 812: The omission length should be considered too.

tim-smart (changes requested):

packages/effect/src/String.ts
jessekelly881: How do you mean?

skoshx:
Friendly bump, will this get included sometime? This would be super helpful.

fubhy:

fubhy:
Thanks for the PR. Closing this as stale because the requested changes have been outstanding since Dec 2024 without author follow-up. If you still want to pursue this, please feel free to open a fresh PR addressing the omission-length behavior.

PR #4149, opened 2024-12-16
upgrade undici to 7.x and expose re-exports in Undici module

thewilkybarkid:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

https://github.com/nodejs/undici/releases/tag/v7.0.0

I could only find 1 change to match (removal of `throwOnError`).

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/serious-eels-rule.md
+++ b/.changeset/serious-eels-rule.md
@@ -2,4 +2,4 @@
 "@effect/platform-node": minor
 ---
 
-upgrade undici to 7.x and move to peerDependencies
+upgrade undici to 7.x and expose re-exports in Undici module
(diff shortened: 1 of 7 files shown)


mikearnaldi:
Might need a minor due to the dep upgrade - cc @tim-smart

mikearnaldi:
> Will move it to peer deps too as we are exposing undici types.

using peerDependencies means that any major release of Undici leads to a major release of platform, if we are re-exporting it might be better to just use a dependency like we do with fast-check in effect

tim-smart:
Will switch to re-exports

PR #4177, opened 2024-12-20
`Url` module

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Url` module has been introduced:

- immutable setters with dual-function api
- integration with `UrlParams`

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/platform/src/Url.ts
+++ b/packages/platform/src/Url.ts
@@ -1,9 +1,9 @@
 /**
  * @since 1.0.0
  */
+import * as Cause from "effect/Cause"
 import * as Either from "effect/Either"
 import { dual } from "effect/Function"
-import * as Error from "./Error.js"
 import * as UrlParams from "./UrlParams.js"
 
 /** @internal */
@@ -17,43 +17,44 @@ const immutableSetter = <M>(clone: (mutable: M) => M) =>
     result[property] = value
     return result
   })
+
 /** @internal */
 const immutableURLSetter = immutableSetter<URL>((url) => new URL(url))
+
 /**
  * @since 1.0.0
  * @category constructors
  */
 export const make: {
-  (url: string | URL, base?: string | URL | undefined): Either.Either<URL, Error.BadArgument>
+  (url: string | URL, base?: string | URL | undefined): Either.Either<URL, Cause.IllegalArgumentException>
 } = (url, base) =>
   Either.try({
     try: () => new URL(url, base),
     catch: (cause) =>
-      Error.BadArgument({
-        module: "Url",
-        method: "make",
-        message: cause instanceof globalThis.Error ? cause.message : "Invalid input"
-      })
+      new Cause.IllegalArgumentException(cause instanceof globalThis.Error ? cause.message : "Invalid input")
   })
+
 /**
  * @since 1.0.0
- * @category utils
+ * @category setters
  */
 export const setHash: {
   (hash: string): (url: URL) => URL
   (url: URL, hash: string): URL
 } = immutableURLSetter("hash")
+
 /**
  * @since 1.0.0
- * @category utils
+ * @category setters
  */
 export const setHost: {
   (host: string): (url: URL) => URL
   (url: URL, host: string): URL
 } = immutableURLSetter("host")
+
 /**
  * @since 1.0.0
- * @category utils
+ * @category setters
  */
 export const setHostname: {
   (hostname: string): (url: URL) => URL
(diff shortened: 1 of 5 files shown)


packages/platform/src/Url.ts
```diff
@@ -0,0 +1,150 @@
...
+    return result
+  })
+/** @internal */
+const immutableURLSetter = immutableSetter<URL>((url) => new URL(url))
+/**
+ * @since 1.0.0
+ * @category constructors
+ */
+export const make: {
+  (url: string | URL, base?: string | URL | undefined): Either.Either<URL, Error.BadArgument>
```
tim-smart, line 27: Cause.IllegalArgument is probably a better error type to use.

packages/platform/src/Url.ts
```diff
@@ -0,0 +1,150 @@
...
+<P extends keyof M>(property: P): {
+  (value: M[P]): (mutable: M) => M
+  (mutable: M, value: M[P]): M
+} =>
+  dual(2, (mutable: M, value: M[P]) => {
+    const result = clone(mutable)
+    result[property] = value
+    return result
+  })
+/** @internal */
```
tim-smart, line 20:
```suggestion
  })

/** @internal */
```

Some whitespace between apis would be good.

packages/platform/src/Url.ts
```diff
@@ -0,0 +1,150 @@
...
+    catch: (cause) =>
+      Error.BadArgument({
+        module: "Url",
+        method: "make",
+        message: cause instanceof globalThis.Error ? cause.message : "Invalid input"
+      })
+  })
+/**
+ * @since 1.0.0
+ * @category utils
```
tim-smart, line 40:
```suggestion
 * @category setters
```

tim-smart (changes requested):

KhraksMamtsov:
@tim-smart done

packages/platform/src/Url.ts
```diff
@@ -0,0 +1,160 @@
...
+
+/** @internal */
+const immutableURLSetter = immutableSetter<URL>((url) => new URL(url))
+
+/**
+ * @since 1.0.0
+ * @category constructors
+ */
+export const make: {
+  (url: string | URL, base?: string | URL | undefined): Either.Either<URL, Cause.IllegalArgumentException>
```
nikelborm, line 29: Hi! Wouldn't it be great to have a limit on what parameters this function accepts? I'm talking about the situation when we pass URL objects as both the first and the second parameters. I think it's a good idea to check at compile time that the first parameter `url` cannot be of type URL, when the second `base` is passed, using for example function overloads.
When you attempt to pass 2 URL objects, the second one just gets ignored. It seems like the first parameter can even be a string representation of the URL and as long as the first parameter can be successfully parsed to the URL object without errors, it will completely dominate over the second parameter. Because of that, it's safe to assume that passing 2 URL objects to the URL's constructor is most likely a mistake and it's best to point the developer to it by typescript.

```
> const githubURL = new URL('http://github.com/nikelborm?param1=123')
undefined
> const googleURL = new URL('http://google.com/pizza?param2=qwe')
undefined
> const someURL = new URL(githubURL, googleURL)
undefined
> someURL
URL {
  href: 'http://github.com/nikelborm?param1=123',
  origin: 'http://github.com',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'github.com',
  hostname: 'github.com',
  port: '',
  pathname: '/nikelborm',
  search: '?param1=123',
  searchParams: URLSearchParams { 'param1' => '123' },
  hash: ''
}
> const someURL2 = new URL('http://github.com/nikelborm?param1=123', 'http://google.com/pizza?param2=qwe')
undefined
> someURL2
URL {
  href: 'http://github.com/nikelborm?param1=123',
  origin: 'http://github.com',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'github.com',
  hostname: 'github.com',
  port: '',
  pathname: '/nikelborm',
  search: '?param1=123',
  searchParams: URLSearchParams { 'param1' => '123' },
  hash: ''
}
> new URL('github.com/nikelborm?param1=123', 'http://google.com/pizza?param2=qwe')
URL {
  href: 'http://google.com/github.com/nikelborm?param1=123',
  origin: 'http://google.com',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'google.com',
  hostname: 'google.com',
  port: '',
  pathname: '/github.com/nikelborm',
  search: '?param1=123',
  searchParams: URLSearchParams { 'param1' => '123' },
  hash: ''
}
> new URL('github.com/nikelborm?param1=123')
Uncaught TypeError: Invalid URL
    at new URL (node:internal/url:818:25) {
  code: 'ERR_INVALID_URL',
  input: 'github.com/nikelborm?param1=123'
}
```

KhraksMamtsov: Thank you)
I removed `URL` from make (because it seems like a rather useless case, in my opinion - because making `make(MyURL)` looks like `clone`) and renamed it to `fromString`.
I also added a `clone` function.

nikelborm (approved):

tim-smart (approved):

nikelborm:
BTW, shouldn't it be somehow programmatically connected to `Schema.URL`?

PR #4185, opened 2024-12-23
incorrect type declaration in LibsqlClient.layer

jessekelly881:
This is incorrect. I think it's left over from when layer became layerConfig

Removing the declaration reveals the real type. 

Screenshot 2024-12-23 at 3 20 18 PM


tim-smart:
👍🏻 Just needs a change set :)

PR #4196, opened 2024-12-26
Avoid putting symbols in global to fix incompatibility with Temporal Sandbox.

mikearnaldi:


packages/effect/src/GlobalValue.ts
```diff
@@ -46,6 +42,12 @@ const globalStore = (globalThis as any)[globalStoreId] as Map<unknown, any>
  * @since 2.0.0
  */
 export const globalValue = <A>(id: unknown, compute: () => A): A => {
+  if (!globalStore) {
+    // @ts-expect-error
+    globalThis[globalStoreId] = new Map()
```
tim-smart, line 47:
```suggestion
    globalThis[globalStoreId] ??= new Map()
```

Not sure if we want to overwrite if it already exists.

izakfilmalter:
This pr is lit. Thank you @mikearnaldi and @mjameswh for helping me and getting this resolved.

packages/effect/src/GlobalValue.ts
mikearnaldi: how can it already exist if `globalStore` is undefined?

tim-smart: It would in the case of a duplicated module.

PR #4212, opened 2025-01-05
Refine `Effect.validateAll` return type to use `NonEmptyArray` for errors

KhraksMamtsov:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/good-eyes-attack.md
+++ b/.changeset/good-eyes-attack.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+A more precise type was used in the `Effect.validateAll` return type.
--- a/.changeset/good-eyes-attack.md
+++ b/.changeset/good-eyes-attack.md
@@ -2,4 +2,6 @@
 "effect": patch
 ---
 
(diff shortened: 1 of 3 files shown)


.changeset/good-eyes-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+A more precise type was used in the `Effect.validateAll` return type.
```
gcanti, line 5:
```suggestion
Refine `Effect.validateAll` return type to use `NonEmptyArray` for errors.

This refinement is possible because `Effect.validateAll` guarantees that when the input iterable is non-empty, any validation failure will produce at least one error. In such cases, the errors are inherently non-empty, making it safe and accurate to represent them using a `NonEmptyArray` type. This change aligns the return type with the function's actual behavior, improving type safety and making the API more predictable for developers.
```

gcanti (approved):

PR #4228, opened 2025-01-09
fix "ParseError" conflict between `Cron` & `Schema`

fubhy:


fubhy:
This is a breaking change ofc. But it is also a bug fix. So good to merge as is imo.

PR #4230, opened 2025-01-10
fix missing fast-check in @effect/vitest deps

adrian-gierakowski:


packages/vitest/package.json
```diff
@@ -34,5 +34,8 @@
   "devDependencies": {
     "effect": "workspace:^",
     "vitest": "^2.1.4"
+  },
+  "dependencies": {
+    "fast-check": "^3.23.1"
```
fubhy, line 39: This would have to be a peerDependency

fubhy (changes requested):

packages/vitest/package.json
adrian-gierakowski: Cool. Do I need to add changset for this?

adrian-gierakowski: done

adrian-gierakowski: also pushed a changeset

fubhy: Wait a second... `fast-check` is already a dependency of `effect`. So this should not be needed altogether.

Instead, we might want to import `FastCheck` from `effect` within `vitest`.

E.g.

```ts
import * as FastCheck from "effect/FastCheck"
```

fubhy:
Superseded by #4232

packages/vitest/package.json
adrian-gierakowski: Good catch, that’s preferable

PR #4232, opened 2025-01-11
Use `fast-check` re-export

fubhy:


fubhy:
@adrian-gierakowski Can you check if the snapshot build of this works for you?

mikearnaldi (approved):

PR #4284, opened 2025-01-17
Add `Effect.transposeOption`, closes #3142

### LINKED ISSUE #3142: Option.mapEffect

```ts

export const mapEffect: {
	<B, A, E, R>(
	<B, A, E, R>(
		self: Option.Option<A>,
	Option.match(self, {
	}),
);
```

gcanti:
Converts an `Option` of an `Effect` into an `Effect` of an `Option`.

**Details**

This function transforms an `Option<Effect<A, E, R>>` into an
`Effect<Option<A>, E, R>`. If the `Option` is `None`, the resulting `Effect`
will immediately succeed with a `None` value. If the `Option` is `Some`, the
inner `Effect` will be executed, and its result wrapped in a `Some`.

**Example**

```ts
import { Effect, Option } from "effect"

//      ┌─── Option<Effect<number, never, never>>
//      ▼
const maybe = Option.some(Effect.succeed(42))

//      ┌─── Effect<Option<number>, never, never>
//      ▼
const result = Effect.transposeOption(maybe)

console.log(Effect.runSync(result))
// Output: { _id: 'Option', _tag: 'Some', value: 42 }
```

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -34,6 +34,7 @@ import * as defaultServices from "./internal/defaultServices.js"
 import * as circular from "./internal/effect/circular.js"
 import * as fiberRuntime from "./internal/fiberRuntime.js"
 import * as layer from "./internal/layer.js"
+import * as option_ from "./internal/option.js"
 import * as query from "./internal/query.js"
 import * as runtime_ from "./internal/runtime.js"
 import * as schedule_ from "./internal/schedule.js"
@@ -12811,7 +12812,7 @@ export const withParentSpan: {
  * ```
  *
  * @since 2.0.0
- * @category Optional Wrapping
+ * @category Optional Wrapping & Unwrapping
  */
 export const fromNullable: <A>(value: A) => Effect<NonNullable<A>, Cause.NoSuchElementException> = effect.fromNullable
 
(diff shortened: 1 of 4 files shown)


packages/effect/src/Effect.ts
```diff
@@ -12866,12 +12867,47 @@ export const fromNullable: <A>(value: A) => Effect<NonNullable<A>, Cause.NoSuchE
...
+ * // Output: { _id: 'Option', _tag: 'Some', value: 42 }
+ * ```
+ *
+ * @since 3.13.0
+ * @category Optional Wrapping & Unwrapping
+ */
+export const transposeOption = <A = never, E = never, R = never>(
+  self: Option.Option<Effect<A, E, R>>
+): Effect<Option.Option<A>, E, R> => {
+  return option_.isNone(self) ? succeedNone : map(self.value, option_.some)
```
tim-smart, line 12908:
```suggestion
  return option_.isNone(self) ? succeedNone : succeedSome(self.value)
```

If you are feeling fancy.

gcanti: it wouldn't typecheck, `succeedSome(self.value)` returns an `Effect<Option<Effect<A, E, R>>>` instead of `Effect<Option<A>, E, R>`

tim-smart: Ah yes, there is an Effect inside.

tim-smart (approved):

PR #4292, opened 2025-01-18
improve UnknownException error messages

### LINKED ISSUE #4221: From Discord: Change in `tryPromise` behavior when upgrading from 3.2.5 to latest version
# Summary
In the chat, a user named AlexDixon is discussing an issue encountered when upgrading the Effect Typescript library from version 3.2.5 to the latest version. Specifically, they noticed a change in the behavior of `tryPromise` when it is used with one argument: the error message isn't propagated in the latest version, while it was in the older version.

Michaelarnaldi responds by noting that without a catch, the error becomes an `UnknownException`, and they express uncertainty about why the message is no longer being carried.

**Key Takeaways:**
- There is a change in error propagation behavior in `tryPromise` between Effect library versions 3.2.5 and the latest.
- When not using a catch, errors might be transformed into `UnknownException`.
- The exact reason for the loss of the error message in new versions is unclear from the messages.

# Discord thread

https://discord.com/channels/795981131316985866/1326266797455970407

gcanti:
Behavior:

1. Default to "An unknown error occurred" if no message is provided.
2. **Use the `message` property from the cause, if it exists and is a string**.
3. Prioritize a provided custom message over the inherited or default message.

@mikearnaldi @tim-smart It seems the regression was introduced in this PR: https://github.com/Effect-TS/effect/pull/3295. I restored the previous code that inherits the message from the provided cause when possible and added tests for the `UnknownException` constructor.

### DIFF
--- a/.changeset/long-terms-sing.md
+++ b/.changeset/long-terms-sing.md
@@ -2,8 +2,24 @@
 "effect": patch
 ---
 
-Restore `UnknownException` error message handling behavior, closes #4221.
+Improve `UnknownException` error messages
 
-1. Default to "An unknown error occurred" if no message is provided.
-2. **Use the `message` property from the cause, if it exists and is a string**.
-3. Prioritize a provided custom message over the inherited or default message.
+`UnknownException` error messages now include the name of the Effect api that
+created the error.
+
+```ts
+import { Effect } from "effect"
+
+Effect.tryPromise(() => Promise.reject(new Error("The operation failed"))).pipe(
+  Effect.catchAllCause(Effect.logError),
+  Effect.runFork
+)
+
+// timestamp=2025-01-21T00:41:03.403Z level=ERROR fiber=#0 cause="UnknownException: An unknown error occurred in Effect.tryPromise
+//     at fail (.../effect/packages/effect/src/internal/core-effect.ts:1654:19)
+//     at <anonymous> (.../effect/packages/effect/src/internal/core-effect.ts:1674:26) {
+//   [cause]: Error: The operation failed
+//       at <anonymous> (.../effect/scratchpad/error.ts:4:24)
+//       at .../effect/packages/effect/src/internal/core-effect.ts:1671:7
+// }"
+```
--- a/.changeset/long-terms-sing.md
+++ b/.changeset/long-terms-sing.md
@@ -1,0 +1,9 @@
+---
+"effect": patch
+---
+
+Restore `UnknownException` error message handling behavior, closes #4221.
+
+1. Default to "An unknown error occurred" if no message is provided.
+2. **Use the `message` property from the cause, if it exists and is a string**.
+3. Prioritize a provided custom message over the inherited or default message.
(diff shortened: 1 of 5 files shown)


gcanti:
Repro provided in https://discord.com/channels/795981131316985866/795983589644304396/1326266797455970407 after this change:

```diff
import { Effect, pipe } from "effect"

pipe(
  Effect.tryPromise(() => Promise.reject(new Error("The operation failed for a well-known reason"))),
  Effect.runPromise
).catch((e) => console.error("\nWithout identity in catch", e))
/*
Output:
-Without identity in catch (FiberFailure) UnknownException: An unknown error occurred
+Without identity in catch (FiberFailure) UnknownException: The operation failed for a well-known reason
    at <anonymous> (.../effect/packages/effect/src/internal/core-effect.ts:1667:80)
*/

pipe(
  Effect.tryPromise({
    try: () => Promise.reject(new Error("The operation failed for a well-known reason")),
    catch: (e) => e
  }),
  Effect.runPromise
).catch((e) => console.error("\nWith identity in catch", e))
/*
Output:
With identity in catch (FiberFailure) Error: The operation failed for a well-known reason
    at try (.../index.ts:10:31)
    at .../effect/src/internal/core-effect.ts:1664:7
*/
```

tim-smart:
Hmm it used to render the `cause` with the original error, which made copying over the message redundant. We should try fix that regression first.

gcanti:
@tim-smart even after the changes introduced in https://github.com/Effect-TS/effect/pull/4294, should this still be considered a regression? The default message "An unknown error occurred" does not provide much useful context, relegating the original error message to the cause context, which might make debugging more cumbersome:

`effect@3.2.5`

```ts
import { Effect, pipe } from "effect"

pipe(
  Effect.tryPromise(() => Promise.reject(new Error("The operation failed for a well-known reason"))),
  Effect.runPromise
).catch((e) => console.error("\nWithout identity in catch", e))
/*
Output:
Without identity in catch (FiberFailure) UnknownException: The operation failed for a well-known reason
    at e (.../src/internal/core-effect.ts:1659:36)
*/
```

`main` branch

```ts
import { Effect, pipe } from "effect"

pipe(
  Effect.tryPromise(() => Promise.reject(new Error("The operation failed for a well-known reason"))),
  Effect.runPromise
).catch((e) => console.error("\nWithout identity in catch", e))
/*
Output:
Without identity in catch (FiberFailure) UnknownException: An unknown error occurred
    at <anonymous> (.../src/internal/core-effect.ts:1667:80) {
  [cause]: Error: The operation failed for a well-known reason
      at <anonymous> (.../index.ts:4:42)
      at .../src/internal/core-effect.ts:1664:7
}
*/
```

tim-smart:
I think the cause provide better context as it also includes the accurate stack trace.

The message would be better if we updated it to reflect the api it was generated from, i.e:

```
UnknownException: An unknown error occurred in Effect.tryPromise
```

PR #4298, opened 2025-01-20
Added type-level validation for the `Effect.Service`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -13014,7 +13014,7 @@ export const Tag: <const Id extends string>(id: Id) => <
   }
 
 /** @internal */
-type MissingSelfGeneric = `Missing \`Self\` generic - use \`class Self extends Service<Self>()...\``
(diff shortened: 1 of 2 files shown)


packages/effect/src/Effect.ts
```diff
@@ -13013,6 +13013,9 @@ export const Tag: <const Id extends string>(id: Id) => <
     return makeTagProxy(TagClass as any)
   }
 
+/** @internal */
+type MissingSelfGeneric = `Missing \`Self\` generic - use \`class Self extends Service<Self>()...\``
```
tim-smart, line 13017:
```suggestion
type MissingSelfGeneric = `Missing \`Self\` generic - use \`class Self extends Effect.Service<Self>()...\``
```

I think including the Effect namespace in the message is clearer.

tim-smart (changes requested):

PR #4302, opened 2025-01-20
Add hashmap tovalues

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This adds `HashMap.toValues()` that returns the values of a HashMap as an array instead of as an iterator like is the case for `HashMap.values()`

I took inspiration from `entries` vs `toEntries` for the naming. I don't know if there's another convention for this.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/three-bees-attack.md
+++ b/.changeset/three-bees-attack.md
@@ -1,5 +1,5 @@
 ---
-"effect": major
+"effect": minor
 ---
 
 Add HashMap.toValues getter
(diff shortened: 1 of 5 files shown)


vinassefranche:
Closing as I messed up with the changeset so the label is not the correct one. I'll re-open it properly

PR #4313, opened 2025-01-21
Schema: Fix `Duration` Encoding to Support All Duration Types

gcanti:
The primary purpose of the `Duration` schema is to encode durations for transmission, so it must handle all possible types, including:

- Finite durations (e.g., milliseconds).
- **Infinite durations** (e.g., `Duration.infinity`).
- **Nanosecond durations**.

**Example**

```ts
import { Duration, Schema } from "effect"

console.log(Schema.encodeSync(Schema.Duration)(Duration.millis(1000)))
// Output: { _tag: 'Millis', millis: 1000 }

console.log(Schema.encodeSync(Schema.Duration)(Duration.infinity))
// Output: { _tag: 'Infinity' }

console.log(Schema.encodeSync(Schema.Duration)(Duration.nanos(1000n)))
// Output: { _tag: 'Nanos', nanos: '1000' }
```

### DIFF
--- a/.changeset/six-mangos-think.md
+++ b/.changeset/six-mangos-think.md
@@ -1,0 +1,50 @@
+---
+"effect": patch
+---
+
+Schema: Update `Duration` Encoding to a Tagged Union Format.
+
+This changeset fixes the `Duration` schema to support all possible duration types, including finite, infinite, and nanosecond durations. The encoding format has been updated from a tuple (`readonly [seconds: number, nanos: number]`) to a tagged union.
+
+This update introduces a change to the encoding format. The previous tuple representation is replaced with a more expressive tagged union, which accommodates all duration types:
+
+```ts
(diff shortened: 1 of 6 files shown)


fubhy (commented):
This is a bug fix for sure but also a breaking change potentially affecting data at rest that was previously encoded with the previous format (`[seconds: number, nanos: number]`).

In the changeset (currently missing) we should be very clear about this.

mikearnaldi:
> This is a bug fix for sure but also a breaking change potentially affecting data at rest that was previously encoded with the previous format (`[seconds: number, nanos: number]`).
> 
> In the changeset (currently missing) we should be very clear about this.

A way of minimising the data-at-rest problem would be to allow the old encoded value as a valid value for decoding, so `Encoded = OldEncoded | NewEncoded`

PR #4316, opened 2025-01-21
Allow configuring OpenAI API URL and API key

### LINKED ISSUE #4271: AI openai client configuration

In order to support proxying requests to openai, allow configuration of client url and possibly make the apiToken optional (as it may be handled by the proxy).

https://github.com/Effect-TS/effect/blob/main/packages/ai/openai/src/OpenAiClient.ts#L67-L68

- an additional config var for url, with a fallback to the current value
- making the apiToken and authorisation header optional

_No response_

IMax153:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #4271
- Closes #4271

### DIFF
--- a/packages/ai/openai/src/OpenAiClient.ts
+++ b/packages/ai/openai/src/OpenAiClient.ts
@@ -55,7 +55,8 @@ export declare namespace OpenAiClient {
  * @category constructors
  */
 export const make = (options: {
-  readonly apiKey: Redacted.Redacted
+  readonly apiKey?: Redacted.Redacted | undefined
+  readonly apiUrl?: string | undefined
   readonly organizationId?: Redacted.Redacted | undefined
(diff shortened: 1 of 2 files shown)


packages/ai/openai/src/OpenAiClient.ts
```diff
@@ -165,7 +167,8 @@ export const layer = (options: {
  */
 export const layerConfig = (
   options: Config.Config.Wrap<{
-    readonly apiKey: Redacted.Redacted
+    readonly apiKey?: Redacted.Redacted
```
tim-smart, line 170: Should add undefined as a valid value

IMax153: 👍

PR #4317, opened 2025-01-21
Add HashMap.toValues and HashSet.toValues

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This adds `HashMap.toValues()` that returns the values of a `HashMap` as an array instead of as an iterator like is the case for `HashMap.values()`. Also add `HashSet.toValues()`.

I took inspiration from `entries` vs `toEntries` for the naming. I don't know if there's another convention for this.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/three-bees-attack.md
+++ b/.changeset/three-bees-attack.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Add HashMap.toValues getter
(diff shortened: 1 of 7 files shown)


vinassefranche:
> This would be useful for `HashSet` too.

indeed. I added it to the P.R.

tim-smart (approved):

PR #4320, opened 2025-01-22
[Fix] Cannot find name 'MissingSelfGeneric'.

KhraksMamtsov:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
I marked this type as internal by mistake and it disappeared in the resulting build. in this PR(https://github.com/Effect-TS/effect/pull/4298)

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -13013,8 +13013,7 @@ export const Tag: <const Id extends string>(id: Id) => <
     return makeTagProxy(TagClass as any)
   }
 
-/** @internal */
-type MissingSelfGeneric = `Missing \`Self\` generic - use \`class Self extends Effect.Service<Self>()...\``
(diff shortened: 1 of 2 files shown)


packages/effect/src/Effect.ts
```diff
@@ -13013,7 +13013,6 @@ export const Tag: <const Id extends string>(id: Id) => <
     return makeTagProxy(TagClass as any)
   }
 
-/** @internal */
 type MissingSelfGeneric = `Missing \`Self\` generic - use \`class Self extends Effect.Service<Self>()...\``
```
IMax153, line 13016: @KhraksMamtsov - I _think_ the type should also be exported but not 100% sure.

IMax153: Nevermind - we do something similar in `Schema` and don't export the type.

PR #4326, opened 2025-01-22
option: make Option.none and Option.some return Some<T> and None<T>

coffeeispower:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I was trying to use Option in a union where it is expected to be Some in some situations and None in others:

```ts
export interface AnonymousAuthenticationResult {
  readonly kind: 'anonymous';
  readonly user: None<User.User>;
  readonly session: None<Session.Session>;
  readonly workspace: None<Workspace.Workspace>;
}
export interface UserAuthenticationResult {
  readonly kind: 'user';
  readonly user: Some<User.User>;
  readonly session: Some<Session.Session>;
  readonly workspace: Option.Option<Workspace.Workspace>;
}
export interface WorkspaceAuthenticationResult {
  readonly kind: 'workspace';
  readonly user: None<User.User>;
  readonly session: None<Session.Session>;
  readonly workspace: Some<Workspace.Workspace>;
}
export type AuthenticationResult =
  | UserAuthenticationResult
  | WorkspaceAuthenticationResult
  | AnonymousAuthenticationResult;
```

However, since `Option.some` and `Option.none` return `Option<T>`, it becomes non-ergonomic to create these structures, since you have to cast manually to `Some` or `None`

```ts
const authResult: AuthenticationResult = {
  kind: "workspace",
  user: Option.none() as None<User.User>,
  session: Option.none() as None<Session.Session>,
  workspace: Option.some(workspace) as Some<Workspace.Workspace>
};
```

which doesn't make any sense because it is expected that `Option.some` always returns `Some`, and `Option.none` always returns `None`

### DIFF
--- a/packages/effect/src/Option.ts
+++ b/packages/effect/src/Option.ts
@@ -107,7 +107,7 @@ export interface OptionTypeLambda extends TypeLambda {
  * @category constructors
  * @since 2.0.0
  */
-export const none => option.none <A = never>(): None<A>
+export const none = option.none as <A = never>() => None<A>
 
 /**
  * Creates a new `Option` that wraps the given value.
--- a/packages/effect/src/Option.ts
+++ b/packages/effect/src/Option.ts
@@ -107,7 +107,7 @@ export interface OptionTypeLambda extends TypeLambda {
  * @category constructors
  * @since 2.0.0
  */
-export const none = <A = never>(): Option<A> => option.none
+export const none => option.none <A = never>(): None<A>
 
 /**
  * Creates a new `Option` that wraps the given value.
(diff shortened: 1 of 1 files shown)


tim-smart:
I would recommend making your state more explicit

```ts
export interface AnonymousAuthenticationResult {
  readonly kind: "anonymous"
}

export interface UserAuthenticationResult {
  readonly kind: "user"
  readonly user: User.User
  readonly session: Session.Session
  readonly workspace: Option.Option<Workspace.Workspace>
}

export interface WorkspaceAuthenticationResult {
  readonly kind: "workspace"
  readonly workspace: Workspace.Workspace
}

export type AuthenticationResult =
  | UserAuthenticationResult
  | WorkspaceAuthenticationResult
  | AnonymousAuthenticationResult
```

coffeeispower:
@tlm-smart but then in the api routes I have to check the kind. The reason i made all of the properties exist in all variants is because then I don't have to check the kind to access them. Like for example the workspace field, the api route should not care if it's kind == "workspace" or kind == "user", it should work the same way.

PR #4342, opened 2025-01-25
Add `Effect.whenLogLevel`

indietyp:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is the result from the following Discord discussion: https://discord.com/channels/795981131316985866/1332442902629978132, it implements a new function `Effect.whenLogLevel`, for which the specified effect is only executed whenever the log level is enabled.

I am unsure if the return type here is okay, currently it permits the effect to return a value, and wrap the return value in `Option.Option`. The other possibility would be to simply say the return must be (like for `Effect.tap`) to be `void`. I could imagine scenarios in which it could be helpful to get the return value, but also feel like that it could be abused in a sense, 95% of usecases would likely just `yield*` without considering the return value (but one could for example use the return value to determine if other diagnostics should be generated, etc.)

## Related

- Discord discussion: https://discord.com/channels/795981131316985866/1332442902629978132
- Related Rust feature: https://docs.rs/tracing/latest/tracing/macro.enabled.html

### DIFF
--- a/packages/effect/src/internal/core-effect.ts
+++ b/packages/effect/src/internal/core-effect.ts
@@ -950,8 +950,13 @@ export const logAnnotations: Effect.Effect<HashMap.HashMap<string, unknown>> = c
 
 /** @internal */
 export const whenLogLevel = dual<
-  (level: LogLevel.LogLevel | LogLevel.Literal) => <A, E, R>(effect: Effect.Effect<A, E, R>) => Effect.Effect<Option.Option<A>, E, R>,
-  <A, E, R>(effect: Effect.Effect<A, E, R>, level: LogLevel.LogLevel | LogLevel.Literal) => Effect.Effect<Option.Option<A>, E, R>
+  (
+    level: LogLevel.LogLevel | LogLevel.Literal
+  ) => <A, E, R>(effect: Effect.Effect<A, E, R>) => Effect.Effect<Option.Option<A>, E, R>,
+  <A, E, R>(
+    effect: Effect.Effect<A, E, R>,
+    level: LogLevel.LogLevel | LogLevel.Literal
+  ) => Effect.Effect<Option.Option<A>, E, R>
 >(2, (effect, level) => {
   const requiredLogLevel = typeof level === "string" ? LogLevel.fromLiteral(level) : level
 
--- a/packages/effect/src/internal/core-effect.ts
+++ b/packages/effect/src/internal/core-effect.ts
@@ -35,7 +35,6 @@ import * as core from "./core.js"
 import * as defaultServices from "./defaultServices.js"
 import * as doNotation from "./doNotation.js"
 import * as fiberRefsPatch from "./fiberRefs/patch.js"
-import * as fiberRuntime from "./fiberRuntime.js"
 import type { FiberRuntime } from "./fiberRuntime.js"
 import * as metricLabel from "./metric/label.js"
 import * as runtimeFlags from "./runtimeFlags.js"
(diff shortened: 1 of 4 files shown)


packages/effect/src/internal/core-effect.ts
```diff
@@ -948,6 +948,20 @@ export const logAnnotations: Effect.Effect<HashMap.HashMap<string, unknown>> = c
     core.currentLogAnnotations
   )
 
+/** @internal */
+export const whenLogLevel = dual<
+  (level: LogLevel.LogLevel) => <A, E, R>(effect: Effect.Effect<A, E, R>) => Effect.Effect<Option.Option<A>, E, R>,
+  <A, E, R>(effect: Effect.Effect<A, E, R>, level: LogLevel.LogLevel) => Effect.Effect<Option.Option<A>, E, R>
+>(2, (effect, level) => {
+  return core.withFiberRuntime((fiberState) => {
+    const currentLogLevel = FiberRef.currentMinimumLogLevel.pipe(fiberState.getFiberRef)
```
tim-smart, line 957: No need for the .pipe here

packages/effect/src/internal/core-effect.ts
```diff
@@ -948,6 +948,20 @@ export const logAnnotations: Effect.Effect<HashMap.HashMap<string, unknown>> = c
     core.currentLogAnnotations
   )
 
+/** @internal */
+export const whenLogLevel = dual<
+  (level: LogLevel.LogLevel) => <A, E, R>(effect: Effect.Effect<A, E, R>) => Effect.Effect<Option.Option<A>, E, R>,
```
tim-smart, line 953: I think we should support passing a string literal too.

tim-smart (changes requested):

packages/effect/src/internal/core-effect.ts
indietyp: fixed in [1dce660](https://github.com/Effect-TS/effect/pull/4342/commits/1dce6600254f45c64ae02f018aa704622dc416c4), I've also taken care of aligning the code used to filter with the one used in `FiberRuntime.log` just to keep it consistent (e.g. moving from `LogLevel.lessThanEqual` to `LogLevel.greaterThan` and doing an early return.

packages/effect/src/internal/core-effect.ts
indietyp: added in [1dce660](https://github.com/Effect-TS/effect/pull/4342/commits/1dce6600254f45c64ae02f018aa704622dc416c4)

packages/effect/src/internal/core-effect.ts
```diff
@@ -948,6 +948,30 @@ export const logAnnotations: Effect.Effect<HashMap.HashMap<string, unknown>> = c
...
+  ) => <A, E, R>(effect: Effect.Effect<A, E, R>) => Effect.Effect<Option.Option<A>, E, R>,
+  <A, E, R>(
+    effect: Effect.Effect<A, E, R>,
+    level: LogLevel.LogLevel | LogLevel.Literal
+  ) => Effect.Effect<Option.Option<A>, E, R>
+>(2, (effect, level) => {
+  const requiredLogLevel = typeof level === "string" ? LogLevel.fromLiteral(level) : level
+
+  return core.withFiberRuntime((fiberState) => {
+    const minimumLogLevel = fiberState.getFiberRef(FiberRef.currentMinimumLogLevel)
```
tim-smart, line 964:
```suggestion
    const minimumLogLevel = fiberState.getFiberRef(fiberRuntime.currentMinimumLogLevel)
```

Will need to use the internal API here

tim-smart (changes requested):

tim-smart:
Hmm will need to move the implementation into the fiberRuntime file to prevent the circular dependency.

indietyp:
> Hmm will need to move the implementation into the fiberRuntime file to prevent the circular dependency.

this should be resolved in [474f51c](https://github.com/Effect-TS/effect/pull/4342/commits/474f51ce0c8ae20e41c9e87c13152cb9513dd8e8). I've also fixed the external API to allow for literals in [8a1f47d](https://github.com/Effect-TS/effect/pull/4342/commits/8a1f47d42ac880fd1acde15b040858305d15d90a).

tim-smart (approved):
🚀

PR #4343, opened 2025-01-26
add ISO8601 duration formatting

### LINKED ISSUE #4319: Support ISO-8601 format for `Duration`

We support ISO formatting for `DateTime`. It would make sense to also support the same for `Duration`.

Support https://en.wikipedia.org/wiki/ISO_8601#Durations

_No response_

fubhy:
Should we add this? If so, I'd add a parser implementation too. The spec isn't clear on how to deal with months and years but I left them out anyways for now which is in line with our existing Duration implementation which also doesn't handle those.

Closes #4319


mikearnaldi:
I can't think of an argument against imho

fubhy:
> I can't think of an argument against imho

Alright, let's do it then!

gcanti (approved):

PR #3889, opened 2024-11-04
`URL` and `URLFromSelf` schemas

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

PR #4346, opened 2025-01-27
Updated zod to schema comparison: added mention of Schema.URL

nikelborm:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

`schema/comparisons.md` said that there is no alternative in `Effect` to `zod`'s `z.string().url()`, while @KhraksMamtsov added support in #3889

### DIFF
--- a/schema/comparisons.md
+++ b/schema/comparisons.md
@@ -221,7 +221,7 @@ S.String.pipe(S.maxLength(5))
 S.String.pipe(S.minLength(5))
 S.String.pipe(S.length(5))
 // S.string().email() // No equivalent
-// S.string().url() // No equivalent
+S.URL
 // S.string().emoji() // No equivalent
 S.UUID
 // S.string().nanoid() // No equivalent


schema/comparisons.md
```diff
@@ -221,7 +221,7 @@ S.String.pipe(S.maxLength(5))
 S.String.pipe(S.minLength(5))
 S.String.pipe(S.length(5))
 // S.string().email() // No equivalent
-// S.string().url() // No equivalent
+S.URL
```
gcanti, line 224: AFAIK `S.URL` is not equivalent, `z.string().url()` just validates a string:

```ts
import z from "zod"

console.log(typeof z.string().url().parse("https://example.com"))
// string
```

nikelborm: `Schema.Date` is also not an exact equivalent, but this fact didn't prevent it from being mentioned as an equivalent a few lines later: 
<details>
<summary><a href="https://github.com/nikelborm/effect/blob/999a5f60244fb80789d7021290cf6783cd975869/schema/comparisons.md#L286-L307">schema/comparisons.md L286-L307</a></summary>

### Dates

Zod

```ts
const date = z.string().date()

date.parse("2020-01-01") // pass
date.parse("2020-1-1") // fail
date.parse("2020-01-32") // fail
```

Schema

```ts
import { Schema as S } from "effect"

S.decodeUnknownSync(S.Date)("2020-01-01") // pass
S.decodeUnknownSync(S.Date)("2020-1-1") // pass
S.decodeUnknownSync(S.Date)("2020-01-32") // fail
```

---

</details>

```js
import z from 'zod';

console.log(typeof z.string().date().parse('2020-01-01'));
// string

console.log(typeof Schema.decodeUnknownSync(Schema.Date)('2020-01-01'));
// object
```

And I think that's fine that they're not exactly equivalent and still mentioned because there's a value in it. When I first saw that there's no equivalent of URL validation, I was really confused, spent time reimplementing `Schema.URL` and only after I was done, I noticed it actually exists and I created absolutely the same thing.

gcanti: it doesn't claim that they are equivalent, it shows what can be used in `Schema` when the input is a string. However, without proper description, that section is confusing and should be reorganized. Furthermore, since these docs were written, some things have changed (e.g., `Schema.URL` didn’t exist at the time). IMO, the entire document needs to be refactored

nikelborm: I assumed it claimed they were equivalent because it didn't claim otherwise. In cases where they're not equivalent, it was stated as:
```
### Times

No equivalent.

### IP addresses

No equivalent.
```

gcanti: yes, that's not your fault, the document needs to be restructured, it's currently confusing

nikelborm:
Thank you @gcanti ☺️❤️

PR #4357, opened 2025-01-28
Type level support (Standard-Schema)

mikearnaldi:
I tried any kind of conditional over R to implement the spec directly but my intuition that it would break variance is correct. The only working solution I found is to implement it like `SomeSchema.standatd()` but at this point it could simply be a function like anything else so `Schema.standard(SomeSchema)`


gcanti:
Closing in favour of https://github.com/Effect-TS/effect/pull/4359

PR #4362, opened 2025-01-29
@effect/rpc 3rd revision

### LINKED ISSUE #3142: Option.mapEffect

tim-smart:


packages/rpc/src/Rpc.ts
```diff
@@ -10,6 +10,7 @@ import * as Predicate from "effect/Predicate"
 import * as Schema from "effect/Schema"
 import type * as AST from "effect/SchemaAST"
 import type { Stream } from "effect/Stream"
+import type { RpcSchema } from "./index.js"
```
fubhy, line 13: Wrong import path

fubhy (dismissed):

tsconfig.json
```diff
@@ -7,8 +7,6 @@
     { "path": "packages/ai/openai" },
     { "path": "packages/cli" },
     { "path": "packages/cluster" },
-    { "path": "packages/cluster-browser" },
-    { "path": "packages/cluster-node" },
```
fubhy, line 11: When we deprecate these, let's push a deprecation notice to npmjs.com for these packages. And also do that for the other packages we have deprecated in the past (especially @effect/schema, which is still seeing more downloads than it should).

tim-smart: They are being rewritten in another branch

PR #4363, opened 2025-01-29
Support for StandardSchemaV1 in Micro

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
I think it makes sense to support the StandardSchema Specification in the Micro module.
Since the use of Schema negates the reason for using Micro, the use of Standard Schema opens up opportunities for utilizing more lightweight alternatives

## Related

- Related Issue #
- Closes #


mikearnaldi:
Unsure how I feel about this, especially given Micro is a "temporary" module until we get to 4.0 where Effect in itself will be small

KhraksMamtsov:
> Effect in itself will be small

I'm glad to hear that) 🚀
but I still think it's worth supporting this feature, besides, It will probably take some time to develop a new version.
In any case, the module is experimental and I think those who use it understand the risks.

mikearnaldi:
> > Effect in itself will be small
> 
> I'm glad to hear that) 🚀 but I still think it's worth supporting this feature, besides, It will probably take some time to develop a new version. In any case, the module is experimental and I think those who use it understand the risks.

Effect doesn't really integrate with anything external in core, this would be a first, wouldn't be much different to have a `fetch` wrapper for example, but we don't in core, we have dedicated packages

KhraksMamtsov:
Okay — then I'll close the PR.

PR #4368, opened 2025-01-30
Add `Embeddings` service to the AI integration packages

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/wise-jeans-chew.md
+++ b/.changeset/wise-jeans-chew.md
@@ -1,0 +1,41 @@
+---
+"@effect/ai-openai": patch
+"@effect/ai": patch
+---
+
+Support creation of embeddings from the AI integration packages.
+
+For example, the following program will create an OpenAI `Embeddings` service
+that will aggregate all embedding requests received within a `500` millisecond
+window into a single batch.
+
+```ts
+import { Embeddings } from "@effect/ai"
+import { OpenAiClient, OpenAiEmbeddings } from "@effect/ai-openai"
+import { NodeHttpClient } from "@effect/platform-node"
+import { Config, Effect, Layer } from "effect"
+
+// Create the OpenAI client
+const OpenAi = OpenAiClient.layerConfig({
+  apiKey: Config.redacted("OPENAI_API_KEY")
+}).pipe(Layer.provide(NodeHttpClient.layerUndici))
+
+// Create an embeddings service for the `text-embedding-3-large` model
+const TextEmbeddingsLarge = OpenAiEmbeddings.layerDataLoader({
+  model: "text-embedding-3-large",
+  window: "500 millis",
+  maxBatchSize: 2048
+}).pipe(Layer.provide(OpenAi))
+
+// Use the generic `Embeddings` service interface in your program
+const program = Effect.gen(function*() {
+  const embeddings = yield* Embeddings.Embeddings
+  const result = yield* embeddings.embed("The input to embed")
+})
+
+// Provide the specific implementation to use
+program.pipe(
+  Effect.provide(TextEmbeddingsLarge),
+  Effect.runPromise
+)
+```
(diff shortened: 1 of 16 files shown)


tim-smart (dismissed):
You need to yield here

packages/ai/ai/src/Embeddings.ts
```diff
@@ -0,0 +1,152 @@
...
+export const makeDataLoader = (options: {
+  readonly embedMany: (input: ReadonlyArray<string>) => Effect.Effect<Array<Embeddings.Result>, AiError>
+  readonly window: Duration.DurationInput
+  readonly maxBatchSize?: number
+}) =>
+  Effect.gen(function*() {
+    const scope = yield* Effect.scope
+
+    const resolver = makeBatchedResolver(options.embedMany)
+    const resolverDelayed = dataLoader(resolver, {
```
tim-smart, line 135: Here

packages/ai/ai/src/Embeddings.ts
```diff
@@ -0,0 +1,152 @@
...
+      maxBatchSize: options.maxBatchSize
+    })
+
+    function embed(input: string) {
+      return Effect.request(new EmbeddingRequest({ input }), resolverDelayed).pipe(
+        Effect.withSpan("Embeddings.embed", {
+          captureStackTrace: false,
+          attributes: { input }
+        })
+      ).pipe(Scope.extend(scope))
```
tim-smart, line 146: You shouldn't need this anymore

packages/ai/ai/src/Embeddings.ts
```diff
@@ -0,0 +1,152 @@
...
+    const resolverDelayed = yield* dataLoader(resolver, {
+      window: options.window,
+      maxBatchSize: options.maxBatchSize
+    })
+
+    function embed(input: string) {
+      return Effect.request(new EmbeddingRequest({ input }), resolverDelayed).pipe(
+        Effect.withSpan("Embeddings.embed", {
+          captureStackTrace: false,
+          attributes: { input }
```
tim-smart, line 144: Probably not a good idea to attach the input to the span. It could be quite large.

.changeset/wise-jeans-chew.md
```diff
@@ -0,0 +1,41 @@
...
+import { NodeHttpClient } from "@effect/platform-node"
+import { Config, Effect, Layer } from "effect"
+
+// Create the OpenAI client
+const OpenAi = OpenAiClient.layerConfig({
+  apiKey: Config.redacted("OPENAI_API_KEY")
+}).pipe(Layer.provide(NodeHttpClient.layerUndici))
+
+// Create an embeddings service for the `text-embedding-3-large` model
+const TextEmbeddingsLarge = OpenAiEmbeddings.layerDataLoader({
```
floydspace, line 24: will it batch if I use
```ts
(documents: Array<string>) => Effect.forEach(documents, embedding.embed, { concurrency: "unbounded" })
```
and `OpenAiEmbeddings.layer`?

packages/ai/openai/src/OpenAiEmbeddings.ts
```diff
@@ -0,0 +1,143 @@
...
+      return client.client.createEmbedding({
+        input,
+        model: options.model,
+        ...parentConfig,
+        ...localConfig
+      })
+    }),
+    Effect.map((response) =>
+      response.data.map(({ embedding, index }) => ({
+        embeddings: embedding as Array<number>,
```
floydspace, line 63: should we define `EmbeddingVector` data type?

.changeset/wise-jeans-chew.md
IMax153: You would need to add batching true to the forEach options as under the hood it's using RequestResolver.batched. 

Unless you use the data loader variant, in which case it will always batch as much as it can during the specified window (up to the max batch size).

packages/ai/openai/src/OpenAiEmbeddings.ts
IMax153: I don't see a real reason to do so unless I'm missing something?

I'm just casting from a read only array to a mutable one here.

PR #4371, opened 2025-01-31
Add `LoggerProvider` support to `@effect/opentelemetry`

jpowersdev:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add `LoggerProvider` and `LogRecordProcessor` support to `@effect/opentelemetry`, as well as a logger layer (using `Logger.addEffect` to run alongside existing loggers) that will emit logs using the registered `LogRecordProcessor`.


jpowersdev:
Working on tests and docs before removing draft status

jpowersdev:
I tried to follow the existing structure as best I could, but please let me know if anything needs to be modified.

packages/opentelemetry/src/internal/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+  (value, key) => [`logSpan.${key}`, `${value}ms`] as const
+)
+
+/** @internal */
+const formatAnnotations = Record.map(
+  (value) => Predicate.isString(value) ? value : Inspectable.format(value)
+)
+
+/** @internal */
+const formatMessage = Match.type<unknown>().pipe(
```
tim-smart, line 87: I think instead we can reuse this

https://github.com/Effect-TS/effect/blob/main/packages%2Fopentelemetry%2Fsrc%2Finternal%2Ftracer.ts#L310

Using the Match module is a bit too much here.

There are probably other functions that could be moved to a utils module

packages/opentelemetry/src/internal/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+      ...formatLogSpans(structured.spans),
+      ...formatAnnotations(structured.annotations)
+    }
+
+    otelLogger.emit({
+      body: formatMessage(structured.message),
+      severityText: structured.logLevel,
+      severityNumber: options.logLevel.ordinal,
+      timestamp: options.date,
+      observedTimestamp: new Date(),
```
tim-smart, line 70: This should use the clock service instead.

packages/opentelemetry/src/Logger.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @since 1.0.0
+ * @category constructors
+ */
+export const logger: Effect<Logger.Logger<unknown, void>, never, OtelLoggerProvider> = internal.make
+
+/**
+ * @since 1.0.0
+ * @category layers
+ */
+export const layerLoggerProvider = (
```
tim-smart, line 45: This should be a direct export instead of a wrapper function

tim-smart (changes requested):

packages/opentelemetry/src/internal/logger.ts
jpowersdev: Pulled some stuff to `internal/utils.ts`

packages/opentelemetry/src/Logger.ts
jpowersdev: Done! Could you go into your reasoning a little bit? I'm unsure what justifies a direct export vs a wrapper function in these scenarios.

tim-smart: It is just un-necessary indirection if the signature is the same.

tim-smart (approved):

PR #4374, opened 2025-01-31
Make runtime run sync dual

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This makes Runtime.runSync dual so we both of these work:
- `Runtime.runSync(runtime)(Effect.succeed(1))`
- `Runtime.runSync(Effect.succeed(1), runtime)`

## Related

- Related Issue #
- Closes #


vinassefranche:
I've limited the pull request to only `Runtime.runSync` for now but if it's accepted, I can add the same for:
- `Runtime.runSyncExit`
- `Runtime.runPromise*`
- the same functions on `ManagedRuntime`

tim-smart:
I the data first signature should be:

```ts
(runtime, effect) => result
```

vinassefranche:
> I think the data first signature should be:
> 
> ```ts
> (runtime, effect) => result
> ```

@tim-smart if I'm not mistaken, I can't have `(runtime, effect) => result` without needing to update the data last signature to `(effect): (runtime) => result` which would be a breaking change.

Is there any way to make `dual` do what you suggest?

tim-smart:
There are a couple places we don't use `dual` for this reason.

vinassefranche:
> There are a couple places we don't use `dual` for this reason.

OK. I'll try to do that then. Thanks!

vinassefranche:
@tim-smart I gave it a try. Tell me if it looks fine to you!

vinassefranche:
@tim-smart I merged the next-minor branch yesterday because the rebase was too tedious so I think it's why the automatic rebase does not work.
Do you want me to make it clean? The number of commit of the pull request is wrong but the number of changed files is correct so it should be fine to review it like this.

tim-smart:
If you could do a clean rebase that would be appreciated :)

vinassefranche:
> If you could do a clean rebase that would be appreciated :)

@tim-smart this is done

tim-smart:
Don't worry too much, as long as you aren't doing merge commits it's fine. 

I will try to look at this PR later today.

tim-smart:
Yeah I think we would want all of the apis to have a consistent signature, so if you want to make them all dual :)

.changeset/yellow-lemons-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
+---
+
+Make Runtime.runSync, Runtime.runSyncExit, Runtime.runPromise and Runtime.runPromiseExit dual
```
tim-smart, line 5:
```suggestion
Make Runtime.run* apis dual
```

tim-smart (approved):

PR #4376, opened 2025-01-31
deps: update vitest to `v3`

FedericoBiccheddu:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Update `vitest` to latest available major version, `3.0`.


FedericoBiccheddu:
Open as draft as some test are failing, one of them (`test/CommandExecutor.test.ts`) fails even on `main` and need to understand why.

FedericoBiccheddu:
> Open as draft as some test are failing, one of them ([`test/CommandExecutor.test.ts`](https://github.com/Effect-TS/effect/blob/d1a5b43acaf56c060defd95cc2078466e6aee561/packages/platform-node-shared/test/CommandExecutor.test.ts)) fails even on `main` and need to understand why.

The problem was trivial: I don't have `bash` installed on my host; trying with `sh` made the test pass!

mikearnaldi:
wondering if this should be a `major` given the `3.x` bump on `vitest`

FedericoBiccheddu:
> wondering if this should be a `major` given the `3.x` bump on `vitest`

I initially thought a minor bump was ok for two reasons:
- AFAIK `@effect/vitest` is not stable; other packages, like platform, rpc, etc have undergone massive refactors that resulted in BC, which were expected. Unless the approach has changed or I'm missing important information about the effect team's stance on BC, I'm *personally* fine w/ this being a minor bump instead of a v`1.*`, as it aligns with the principles of other effect packages from a consumer PoV;
- the consumer API has not changed.

However, upon further reflection, `@effect/vitest` has a peer dep on `vitest` and users may have tests running on `2.x` version without relying on `@effect/test`; this minor bump could be a BC for them hindering upgrades to other effect packages, right?

codingismy11to7:
> However, upon further reflection, @effect/vitest has a peer dep on vitest and users may have tests running on 2.x version without relying on @effect/test; this minor bump could be a BC for them hindering upgrades to other effect packages, right?

i thought about that too, but when i looked at the vitest 3.0 release notes it didn't look like a particularly onerous upgrade?

also, people who are depending on < 1.0 stuff (like myself) kinda need to be prepared for things like that, imo

IMax153:
> i thought about that too, but when i looked at the vitest 3.0 release notes it didn't look like a particularly onerous upgrade?

It doesn't matter if the upgrade is "onerous" or not - if there are breaking changes in Vitest 3.0 that will affect the external API or functionality of `@effect/vitest`, a minor version bump of `@effect/vitest` (which allows for breaking changes before there is a major) or a major version bump is most appropriate.

I don't have a strong preference and am fine with just a minor.

codingismy11to7:
> It doesn't matter if the upgrade is "onerous" or not - if there are breaking changes in Vitest 3.0 that will affect the external API or functionality of `@effect/vitest`, a minor version bump of `@effect/vitest` (which allows for breaking changes before there is a major) or a major version bump is most appropriate.

um...since there's no major release, either is appropriate, as you just said. so what do we determine by? in my mind, how much it affects downstream users...so onerousness. if vitest3 changed behavior so badly that it becomes an onerous upgrade, then effect vitest should bump major. if it presents basically no changes, and effect vitest isn't at 1.0, then minor. so I disagree, but hey nbd.

anyway, @FedericoBiccheddu any chance of a rebase?

FedericoBiccheddu:
Done @codingismy11to7.

I'm still wondering why the `check` command is failing even though the patch for `vitest-websocket-mock` is correctly applied; I'd appreciate some help with this issue :pray:

codingismy11to7:
cool, I'll take a look later today when I get to my desk. not any kind of expert, but happy to be a second set of eyes

codingismy11to7:
cool, I'll take a look later today when I get to my desk. not any kind of expert, but happy to be a second set of eyes

codingismy11to7:
@FedericoBiccheddu just remove the `vitest-websocket-mock.patch` - it's breaking the check

image

codingismy11to7:
[change.patch](https://github.com/user-attachments/files/18815511/change.patch)

fubhy:
@FedericoBiccheddu Thanks. I think this one warrants a follow-up to get rid of the deprecation warnings:

```
Using an object as a third argument is deprecated. Vitest 4 will throw an error if the third argument is not a timeout number. Please use the second argument for options. See more at https://vitest.dev/guide/migration
```

Do you want to tackle that or shall I?

FedericoBiccheddu:
> @FedericoBiccheddu Thanks. I think this one warrants a follow-up to get rid of the deprecation warnings:
> 
> ```
> Using an object as a third argument is deprecated. Vitest 4 will throw an error if the third argument is not a timeout number. Please use the second argument for options. See more at https://vitest.dev/guide/migration
> ```
> 
> Do you want to tackle that or shall I?

Done @fubhy!

Feel free to overwrite it if doesn't adhere to the team standards or whatever reason or leave suggestions :smile:

fubhy:
@FedericoBiccheddu Yeah, this one is closed. It would have to be in a new pull request.

fubhy:
Opened here: https://github.com/Effect-TS/effect/pull/4468

Yedfree (approved):

PR #4402, opened 2025-02-05
Fix CLI -h to display default values (#4400)

### LINKED ISSUE #4400: From Discord: Default Value Not Displayed in CLI Usage for `Args.withDefault`
# Summary
A user, rawbeannn, reported an issue where the default value specified using `Args.withDefault` in the Effect Typescript CLI library does not display in the help (`-h`/`--help`) usage documentation. They observed that the default value is only shown if it is wrapped inside an `Option`, but when used this way, it prints the internal details of the `Option` object instead of just the value.

Maxwellbrown clarified that while the first behavior (not displaying defaults meant to be primitive) is intentional, because defaults can also be complex objects, the second behavior (displaying the internal details of an `Option` object) appears to be a bug. 

Key takeaways:
- The CLI library is designed to handle both primitive and complex objects as defaults, thus defaults are not always shown directly.
- There is a recognized bug in how the library displays defaults when using `Option`, where it shows the object's internal structure instead of just the content.

# Discord thread

https://discord.com/channels/795981131316985866/1336395893053718700

Delusion2056:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Args` and `Options` in `@effect/cli` did not correctly display default values in the help output when `withDefault` was used. If the fallback value was wrapped in `Option.some`, the internal object details were printed instead of the expected default value.

Fix attempted is:
- Coerce primitive fallback values to strings for proper display.
- Serialize non-primitive fallback values to (compact-style) JSON using the `effect/Inspectable` module.
- This ensures that both `Args` and `Options` display expected default values in their help output.

A few basic test cases were added for both `Args` and `Options`. Snapshots used due to deep partial object matching getting messy without custom matchers. (Unless I missed something.)

## Related

- Closes #4400

### DIFF
--- a/.changeset/quiet-boats-smile.md
+++ b/.changeset/quiet-boats-smile.md
@@ -1,0 +1,5 @@
+---
+"@effect/cli": patch
+---
+
+Fix CLI help output to correctly display default values
--- a/packages/cli/src/internal/args.ts
+++ b/packages/cli/src/internal/args.ts
@@ -8,9 +8,11 @@ import * as Console from "effect/Console"
 import * as Effect from "effect/Effect"
 import * as Either from "effect/Either"
 import { dual, pipe } from "effect/Function"
+import * as Inspectable from "effect/Inspectable"
 import * as Option from "effect/Option"
 import * as ParseResult from "effect/ParseResult"
 import { pipeArguments } from "effect/Pipeable"
+import * as Predicate from "effect/Predicate"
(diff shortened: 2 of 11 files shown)


IMax153 (approved):
This change looks great @Delusion2056 ! Thank you so much for the PR and the help with this :)

PR #4419, opened 2025-02-07
Fix `Context.Tag` unification

KhraksMamtsov:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/loud-cheetahs-remain.md
+++ b/.changeset/loud-cheetahs-remain.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix Context.Tag unification
--- a/packages/effect/dtslint/Unify.ts
+++ b/packages/effect/dtslint/Unify.ts
@@ -1,3 +1,4 @@
+import type * as Context from "effect/Context"
 import type * as Deferred from "effect/Deferred"
 import type * as Effect from "effect/Effect"
 import * as Either from "effect/Either"
@@ -46,6 +47,12 @@ Unify.unify(Math.random() > 0 ? Either.right(10) : Either.left("ok"))
 hole<
   Unify.Unify<
     Stream.Stream<0, 1, 2> | Stream.Stream<"a", "b", "c">
+  >
+>()
+// $ExpectType Tag<0, 1> | Tag<"a", "b">
+hole<
+  Unify.Unify<
+    Context.Tag<0, 1> | Context.Tag<"a", "b">
   >
 >()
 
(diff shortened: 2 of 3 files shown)


KhraksMamtsov:
@mikearnaldi 
I investigated the issue and concluded that TypeScript fails to infer types, for example, in the case of `Layer.succeed` used with `this` in class body(It looks like a shortcoming on Typescript's side because it works if the function is used outside the class body).

The issue can be resolved by using type parameters directly instead of inferring them from the provided tag — this approach is used in `Micro`(https://github.com/Effect-TS/effect/blob/main/packages/effect/src/Micro.ts#L1835) and other modules.
```ts
// before
export const succeed: {
  <T extends Context.Tag<any, any>>(tag: T): (resource: Context.Tag.Service<T>) => Layer<Context.Tag.Identifier<T>>
  <T extends Context.Tag<any, any>>(
    tag: T,
    resource: Context.Tag.Service<T>
  ): Layer<Context.Tag.Identifier<T>>
} = internal.succeed
// after 
export const succeed: {
  <I, S>(tag: Context.Tag<I, S>): (resource: S) => Layer<I>
  <I, S>(tag: Context.Tag<I, S>, resource: S): Layer<I>
} = internal.succeed
```

This approach seems more correct since type parameters are invariant, and union tags cannot be unified into a tag of unions.

However, this would be a breaking change, though it might be reasonable to treat it as a bug fix.
What do you think? I'm not sure whether the use case for unification is common enough to justify fixing it at the cost of a breaking change. On the other hand, I believe this change is theoretically correct.

mikearnaldi:
I can't remember why we are using `T extends Context.Tag<any, any>` for tags, I am sure there was a reason but it's evading me.

In principle I agree with you that it would be more correct to use generics

mikearnaldi:
> I can't remember why we are using `T extends Context.Tag<any, any>` for tags, I am sure there was a reason but it's evading me.
> 
> In principle I agree with you that it would be more correct to use generics

Ok I remembered, it was to make sure the service shape was inferred from the tag and not the other way around, we can use NoInfer though which at the time wasn't part of the codebase

mikearnaldi:
Updated usage in https://github.com/Effect-TS/effect/pull/4422

KhraksMamtsov:
It looks good

KhraksMamtsov:
@mikearnaldi I rebased the branch – it's ready to merge.

PR #4459, opened 2025-02-16
remove duplicate code in stm

fubhy:
We would've seen this earlier if we'd be using Skott ;-) /cc @antoine-coulon


antoine-coulon:
Say less @fubhy, message received loud and clear ⏳

PR #4462, opened 2025-02-16
update pnpm and node version

fubhy:


flake.nix
```diff
@@ -19,7 +19,7 @@
             bun
             corepack
             deno
-            nodejs
+            nodejs-slim_22
```
fubhy: Slim should be fine (we don't need `npm`)

package.json
```diff
@@ -85,6 +85,19 @@
...
+    "ignoredBuiltDependencies": [
+      "cpu-features",
+      "esbuild",
+      "lmdb",
+      "msgpackr-extract",
+      "msw",
+      "protobufjs",
+      "ssh2",
+      "workerd"
+    ],
```
fubhy, line 98: This is a new feature coming from `pnpm` v10. It doesn't look like we need any of these to have their install scripts (build native deps) run.

.github/workflows/check.yml
```diff
@@ -70,8 +70,6 @@ jobs:
       - name: Test
         run: pnpm vitest --shard ${{ matrix.shard }}
         if: matrix.runtime == 'Node'
-        env:
-          NODE_OPTIONS: --max_old_space_size=8192
```
fubhy, line 74: This had no effect but also the random test fails seem to be appearing way less recently.

PR #4463, opened 2025-02-16
fix jsdoc codemod

fubhy:
@IMax153 This is definitely broken atm. Can you take a look? I'm not sure how it is *intended* to work.


.github/workflows/check.yml
```diff
@@ -13,18 +13,6 @@ concurrency:
...
-    name: Build
-    runs-on: ubuntu-latest
-    timeout-minutes: 10
-    steps:
-      - uses: actions/checkout@v4
-      - name: Install dependencies
-        uses: ./.github/actions/setup
-      - run: pnpm codegen
-      - name: Check source state
-        run: git add packages/*/src && git diff-index --cached HEAD --exit-code packages/*/src
```
fubhy, line 26: We can eliminate this, if we consider `pnpm codemod` a pre-publishing step only.

.github/workflows/check.yml
```diff
@@ -62,7 +49,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install dependencies
         uses: ./.github/actions/setup
-      - run: pnpm codemod
```
fubhy, line 65: Removing it these workflows then also speeds up CI

scripts/codemod.mjs
```diff
@@ -1,22 +1,17 @@
...
 import * as Path from "node:path"
 
-const packageJsonPath = Path.resolve("package.json")
-const packageJson = JSON.parse(Fs.readFileSync(packageJsonPath, "utf-8"))
-const workspaces = Glob.globSync(packageJson["workspaces"])
-const packages = workspaces.map((workspace) => workspace.replace("packages/", ""))
-const pattern = `packages/{${packages.join(",")}}/src/**/*.ts`
-
+// Look up files in all workspace packages including those nested in sub-packages (e.g. `packages/ai/openapi`).
+const pattern = "packages/{*,*/*}/src/**/*.ts"
```
fubhy, line 7: We have other places in our config where the nested dirs `packages/ai/*` are not considered properly.

fubhy: This is somewhat burdensome to maintain because we have to keep that in mind everytime we add a new subdir package group. Instead, we should probably just simplify this as I did here. It's unlikely that this will cause false-positives.

PR #4465, opened 2025-02-17
add Tracer Span.addLinks, for dynamically linking spans

tim-smart:


mikearnaldi:
maybe `Effect.spanAddLink()`

tim-smart:
> maybe `Effect.spanAddLink()`, or even `Effect.link()`

We already have:
- Effect.linkSpan
- Effect.annotateSpan
- Effect.annotateCurrentSpan

So we just need an api for linking to the current span (not regional)

Effect.linkCurrentSpan or Effect.linkToCurrentSpan were some options.

mikearnaldi:
> * Effect.linkSpan

we don't have `Effect.linkSpan` we have `Effect.linkSpans`, we will need to re-think naming in 4.0 as it's very unclear from the name what `Effect.linkSpans` does, probably something like `Effect.addLinkToChildSpans` or similar

PR #4479, opened 2025-02-19
Add `Array.modifyFirst` and `Array.modifyFirstOption`

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add `Array.modifyFirst` and `Array.modifyFirstOption`
 ```ts
import { Array } from "effect"

const numbers = [1, 2, 3, 4]
const result = Array.modifyFirst(numbers, n => n > 2, (n) => n * 2)
assert.deepStrictEqual(result, [1, 2, 6, 4])
```

## Related

- Related Issue #
- Closes #


tim-smart:
Might be worth investigating object options, it is quite hard to read otherwise. I'll let @gcanti weigh in here :)

vinassefranche:
> Might be worth investigating object options, it is quite hard to read otherwise. I'll let @gcanti weigh in here :)

I did the test and I think that it's clearer too. Not completely sure about the `condition` and `updateFunction` names though. Feel free to tell me which names fit better and I'll update them.

vinassefranche:
@gcanti any view on this P.R.?

gcanti:
@vinassefranche sorry for the delay, IMO this is a bit too specific of a use case to justify adding a dedicated API

vinassefranche:
> @vinassefranche sorry for the delay, IMO this is a bit too specific of a use case to justify adding a dedicated API

Alright @gcanti. I understand the point. Thanks for taking the time to answer!

PR #4481, opened 2025-02-20
chore: add missing type assertions

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

As I mentioned in https://github.com/Effect-TS/effect/pull/4467#discussion_r1958752870, few type assertions went missing. Adding them here.

In the future TSTyche will be able to detect similar issues. I was using not yet released patch while working on this PR.


.changeset/perfect-cats-work.md
```diff
@@ -0,0 +1,2 @@
+---
```
gcanti, line 1: No changeset needed (this is an internal change)

mrazauskas: Ah.. Sorry. The bot was complaining so I though this might be required.

PR #4485, opened 2025-02-20
chore: add `references` for type test projects

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR adds `references` for type test projects (their `tsconfig.json` files).

I was trying out TSTyche’s `checkSourceFiles` option and noticed few errors. For example, `packages/effect/tsconfig.src.json` has `types: ["node"]` set or `packages/platform/tsconfig.src.json` has `exactOptionalPropertyTypes: false` set, but these options were ignored while checking `effect` or `@effect/platform` source files. Seems like package were not treated as separate projects. Adding `references` solves the problem.

The `checkSourceFiles` option will be enable by default in the next major TSTyche release. That is why I wanted to try it out.

## Context

To retrieve type information TSTyche wraps around language service, which allows opening / checking files one by one. Even if two test files belong to the same project / program, this approach allows reporting errors only in a single test file. The problem is that source files are not checked. That is why the option was added.

Another aspect: language service handles `references`. In contrary `tsc --noEmit` can’t do that.

### DIFF
--- a/tstyche.config.json
+++ b/tstyche.config.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "https://tstyche.org/schemas/config.json",
+  "checkSourceFiles": true,
+  "rejectAnyType": true,
+  "testFileMatch": [
+    "packages/*/dtslint/**/*.tst.*"
+  ]
+}
--- a/packages/effect/dtslint/tsconfig.json
+++ b/packages/effect/dtslint/tsconfig.json
@@ -1,6 +1,9 @@
 {
   "extends": "../../../tsconfig.base.json",
   "include": ["."],
+  "references": [
+    { "path": "../tsconfig.src.json" },
+  ],
   "compilerOptions": {
     "incremental": false,
     "composite": false,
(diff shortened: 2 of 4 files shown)


tstyche.config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "https://tstyche.org/schemas/config.json",
+  "checkSourceFiles": true,
+  "rejectAnyType": true,
```
mrazauskas, line 4: Protects agains accidental `any`s. For instance a missing import can be silence with `// @ts-expect-error` and it would become `any`. Assignability tests would be happily passing, but `rejectAnyType` does catch this. Will be unable by default in the next major.

tstyche.config.json
```diff
@@ -0,0 +1,8 @@
...
+  "testFileMatch": [
+    "packages/*/dtslint/**/*.tst.*"
```
mrazauskas, line 6: After running `pnpm build` lots of files are emitted. This pattern does not walk into `build`, `dist`, `src` or `test` directories. On my computer adding this pattern save around `0.3` seconds at startup of TSTyche.

gcanti (approved):

PR #4497, opened 2025-02-23
refactor of @effect/cluster packages

tim-smart:


packages/cluster/TODO.md
mattiamanzati, line 1: Leftover I think 😉

safareli:
Hi, I saw this video a while ago:
- [Durable Workflows with **Effect** Cluster by Mattia Manzati (Effect Days 2024)](https://www.youtube.com/watch?v=EcyrFrJFMWY)
and was looking to try it out, but after some investigation noticed this pr which removed the workflo/activity parts that was demoed in this video.


were this constructs renamed in this pr or removed all together in favor of some different construct?

PR #4498, opened 2025-02-24
chore: simplify `test-types` script

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

With this PR I would like to propose several simplifications of the `test-types` script.

#### `test-types`

First I would like to suggest rename the script from `dtslint` to `test-types`. This way it sounds similar to `test`, but for type-level tests.

#### `--target`

It also works similar. For instance, you can run `dtslint semigroup --watch`. But in current setup errors will be logged and the runner will not be very responsive. Try running `test-types semigroup --watch` from this branch and you will see the difference.

TSTyche is able to handle watch mode with several versions of TypeScript, but that is an expensive operation. All four `typescript` modules will be loaded into memory and four language services created.

This is the reason I suggest keeping the command simple and moving `--target` to CI. Currently one can opt-out by passing `--target currrent`, but isn’t it more optimal keeping `--target` as opt-in? For instance, `test-types semigroup --target 5.6 --watch`.

#### `--recursive`, `--parallel`

`--recursive` and `--parallel` look suboptimal for my eye. With `--recursive` (without `--target`) three instances of TSTyche are created, three `typescript` modules loaded and three language services created. Removing `--recursive` creates only a single language service and this allows a lot of caching as well (for instance, module resolutions).

With `--parallel` output becomes harder to ready. (Running JavaScript tests in parallel would give similar result.) Just try running `test-types` from this branch:

Screenshot 2025-02-24 at 10 04 45

Version of TypeScript and path to landed TSConfig are easy to find. But with `--recursive --parallel` the output gets noisy and paths of TSConfig and test file are hard to follow:

Screenshot 2025-02-24 at 10 11 44

---

It can also be that I have missed something. Just wanted to draw your attention.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install dependencies
         uses: ./.github/actions/setup
       - run: pnpm check
-      - run: pnpm dtslint
+      - run: pnpm test-types --target '>=5.4'
 
   lint:
     name: Lint
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Run the following commands to ensure your changes do not introduce any issues:
 - `pnpm circular`: Check for any circular dependencies in imports.
 - `pnpm lint`: Ensure the code adheres to our coding standards.
   - If you encounter style issues, use `pnpm lint-fix` to automatically correct some of these.
-- `pnpm dtslint`: Run type-level tests. Tests are written using [tstyche](https://tstyche.org/).
+- `pnpm test-types`: Run type-level tests. Tests are written using [tstyche](https://tstyche.org/).
 - `pnpm docgen`: Ensure the documentation generates correctly and reflects any changes made.
 
 ### Document Your Changes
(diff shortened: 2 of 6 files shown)


fubhy:
Especially in core-constrained environments like CI (we use the standard single core runners), this should be preferable indeed.

Locally, the argument around "wastefulness" is usually not as relevant and all you care for is whether the whole thing finishes faster with parallelized processing or with a single threat that benefits from the shared bootstrapping and shared memory.

But overall I think this makes sense. We do the same with the type checker, the test suite, etc. (letting the tool handle the workspace as a whole).

mrazauskas:
Seems like this setup spins slightly faster in CI, or? Might be I checked at wrong places.

I think, the output looks much better. Can be that my eye is used to this form. Everything can be improved, of course.

fubhy:
> Seems like this setup spins slightly faster in CI, or? Might be I checked at wrong places.
> 
> I think, the output looks much better. Can be that my eye is used to this form. Everything can be improved, of course.

Yeah I agree. I think this is preferable over what we have atm.

gcanti (approved):

PR #4500, opened 2025-02-24
feat: add Record.findFirst

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds `Record.FindFirst`
 ```ts
import { Record, Option } from "effect"

const record = { a: 1, b: 2, c: 3 }
const result = Record.findFirst(record, (value, key) => value > 1 && key !== "b")
assert.deepStrictEqual(result, Option.some(["c", 3]))
```

## Related

- Related Issue #
- Closes #

### CODE REVIEW

File: packages/effect/dtslint/Record.tst.ts [hunk 1]
```diff
@@ -589,4 +589,20 @@ describe("Record", () => {
     expect(Record.intersection(string$structAB, { b: 2 }, (a, _) => a))
       .type.toBe<Record<"b", number>>()
   })
+
+  it("findFirstEntry", () => {
+    expect(Record.findFirstEntry(string$numbersOrStrings, (a, _) => predicateNumbersOrStrings(a)))
+      .type.toBe<Option.Option<[string, string | number]>>()
+    expect(pipe(string$numbersOrStrings, Record.findFirstEntry((a, _) => predicateNumbersOrStrings(a))))
+      .type.toBe<Option.Option<[string, string | number]>>()
+    expect(Record.findFirstEntry(string$numbersOrStrings, (a, _) => Predicate.isString(a)))
```
vinassefranche, line 598: This expect fails with Typescript 5.4.5. If you have any tip on how to fix it, I'm all ears!

vinassefranche:
@gcanti any view on this one?

IMax153:
@vinassefranche i think this one also may be a bit too specific of a method. 

We really appreciate the PR but we try to be careful to avoid an explosion of combinators in each model for highly specific use cases where we can. Hope that makes sense.

vinassefranche:
> @vinassefranche i think this one also may be a bit too specific of a method.
> 
> We really appreciate the PR but we try to be careful to avoid an explosion of combinators in each model for highly specific use cases where we can. Hope that makes sense.

@IMax153 Thanks for your answer!
There's no find method at all in the Record module, if I'm not mistaken, while there is a `HashMap.findFirst` that exists. I don't find it too much to have one method to search for an entry in the Record module. I understand your point and would have closed the P.R. if there was already a `findFirst` though.
Maybe the name feels too specific though. `Record.findFirst` to mirror `HashMap` might be better.
WDYT?

IMax153:
> There's no find method at all in the Record module, if I'm not mistaken

An absolutely fair point. I think if we change this to `Record.findFirst` to be more consistent with other modules then I'm fine with keeping it and allowing others to review.

Thanks @vinassefranche !

vinassefranche:
> > There's no find method at all in the Record module, if I'm not mistaken
> 
> An absolutely fair point. I think if we change this to `Record.findFirst` to be more consistent with other modules then I'm fine with keeping it and allowing others to review.
> 
> Thanks @vinassefranche !

Thanks @IMax153 I'll do that quick then!

packages/effect/dtslint/Record.tst.ts
vinassefranche: it seems like we don't test on all typescript versions anymore

vinassefranche:
@IMax153 it's done!

IMax153:
Great! I'll also ask @gcanti to take a look.

gcanti:
I'd remove the two overloadings that use a mapping function instead of a predicate / refinement:

```ts
<K extends string | symbol, V, V2>(
    f: (value: NoInfer<V>, key: NoInfer<K>) => Option.Option<V2>
  ): (self: Record<K, V>) => Option.Option<V2>

<K extends string | symbol, V, V2>(
    self: Record<K, V>,
    f: (value: NoInfer<V>, key: NoInfer<K>) => Option.Option<V2>
  ): Option.Option<V2>
```

vinassefranche:
> I'd remove the two overloadings that use a mapping function instead of a predicate / refinement:
> 
> ```ts
> <K extends string | symbol, V, V2>(
>     f: (value: NoInfer<V>, key: NoInfer<K>) => Option.Option<V2>
>   ): (self: Record<K, V>) => Option.Option<V2>
> 
> <K extends string | symbol, V, V2>(
>     self: Record<K, V>,
>     f: (value: NoInfer<V>, key: NoInfer<K>) => Option.Option<V2>
>   ): Option.Option<V2>
> ```

@gcanti I removed them.
There's a type issue for typescript 5.4. Do you have an idea on what to do with it?

gcanti:
@vinassefranche That's because TypeScript 5.4 is not able to infer the refinement when it's written inline:

```ts
expect(Record.findFirst(string$numbersOrStrings, (a, _) => Predicate.isString(a)))
  .type.toBe<Option.Option<[string, string]>>()
expect(pipe(string$numbersOrStrings, Record.findFirst((a, _) => Predicate.isString(a))))
  .type.toBe<Option.Option<[string, string]>>()
```

You can fix it by passing the refinement directly:

```ts
expect(Record.findFirst(string$numbersOrStrings, Predicate.isString))
  .type.toBe<Option.Option<[string, string]>>()
expect(pipe(string$numbersOrStrings, Record.findFirst(Predicate.isString)))
  .type.toBe<Option.Option<[string, string]>>()
```

vinassefranche:
> @vinassefranche That's because TypeScript 5.4 is not able to infer the refinement when it's written inline:
> 
> ```ts
> expect(Record.findFirst(string$numbersOrStrings, (a, _) => Predicate.isString(a)))
>   .type.toBe<Option.Option<[string, string]>>()
> expect(pipe(string$numbersOrStrings, Record.findFirst((a, _) => Predicate.isString(a))))
>   .type.toBe<Option.Option<[string, string]>>()
> ```
> 
> You can fix it by passing the refinement directly:
> 
> ```ts
> expect(Record.findFirst(string$numbersOrStrings, Predicate.isString))
>   .type.toBe<Option.Option<[string, string]>>()
> expect(pipe(string$numbersOrStrings, Record.findFirst(Predicate.isString)))
>   .type.toBe<Option.Option<[string, string]>>()
> ```

Amazing! Thanks @gcanti !

gcanti (approved):

PR #4511, opened 2025-02-25
Add DateTime.nowAsDate

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add `DateTime.nowAsDate` creator

This is handy in codebases that use Date to model dates

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/DateTime.ts
+++ b/packages/effect/src/DateTime.ts
@@ -439,7 +439,7 @@ export const now: Effect.Effect<Utc> = Internal.now
  * })
  * ```
  */
-export const nowAsDate: Effect.Effect<Date> = Effect.map(Internal.now, (dateTime) => toDate(dateTime))
+export const nowAsDate: Effect.Effect<Date> = Internal.nowAsDate
 
 /**
  * Get the current time using `Date.now`.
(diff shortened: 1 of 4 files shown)


packages/effect/src/DateTime.ts
```diff
@@ -425,6 +425,22 @@ export const makeZonedFromString: (input: string) => Option.Option<Zoned> = Inte
...
+ * @example
+ * ```ts
+ * import { DateTime, Effect } from "effect"
+ *
+ * Effect.gen(function* () {
+ *   const now = yield* DateTime.nowAsDate
+ * })
+ * ```
+ */
+export const nowAsDate: Effect.Effect<Date> = Effect.map(Internal.now, (dateTime) => toDate(dateTime))
```
tim-smart, line 442: I would use the clock service directly, no need for the indirection.

vinassefranche: Indeed. I did the change

packages/effect/test/DateTime.test.ts
```diff
@@ -438,4 +438,12 @@ describe("DateTime", () => {
         strictEqual(dt.toJSON(), "2023-12-31T12:00:00.000Z")
       }))
   })
+  describe("nowAsDate", () => {
+    it.effect("should mutate the date", () =>
```
tim-smart, line 442:
```suggestion
    it.effect("should return the current Date", () =>
```

tim-smart (approved):

PR #4512, opened 2025-02-25
Make TestClock.setTime accept a DateTime.Input

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add `TestClock.setDate` helper

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/sixty-starfishes-grow.md
+++ b/.changeset/sixty-starfishes-grow.md
@@ -2,4 +2,4 @@
 "effect": minor
 ---
 
-Make TestClock.setTime accept a Date input
+Make TestClock.setTime accept a DateTime.Input
--- a/.changeset/sixty-starfishes-grow.md
+++ b/.changeset/sixty-starfishes-grow.md
@@ -2,4 +2,4 @@
 "effect": minor
 ---
 
-Add TestClock.setDate helper
+Make TestClock.setTime accept a Date input
--- a/.changeset/sixty-starfishes-grow.md
+++ b/.changeset/sixty-starfishes-grow.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Add TestClock.setDate helper
--- a/packages/effect/test/Schedule.test.ts
+++ b/packages/effect/test/Schedule.test.ts
@@ -17,7 +17,6 @@ import {
   ScheduleIntervals
 } from "effect"
 import { constVoid } from "effect/Function"
-import * as DateTime from "effect/internal/dateTime"
 import { assertTrue, deepStrictEqual, strictEqual } from "effect/test/util"
 import * as TestClock from "effect/TestClock"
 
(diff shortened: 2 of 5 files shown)


tim-smart:
Hmm not sure it is worth it, compared to `TestClock.setTime(date.getTime())`

vinassefranche:
> Hmm not sure it is worth it, compared to `TestClock.setTime(date.getTime())`

@tim-smart your call. We have this small util in our code base as it's pretty handy but I get that it's not incredibly useful:
```ts
export const setTestClockToDate = (date: Date) =>
  TestClock.setTime(date.getTime());
```

vinassefranche:
> > Hmm not sure it is worth it, compared to `TestClock.setTime(date.getTime())`
> 
> @tim-smart your call. We have this small util in our code base as it's pretty handy but I get that it's not incredibly useful:
> 
> ```ts
> export const setTestClockToDate = (date: Date) =>
>   TestClock.setTime(date.getTime());
> ```

To push a little for the addition of this, the main advantage is that you don't have to think about which `get` Date function to use before passing it to `TestClock.setTime`. There are many `get` function and I'm often confused by them. `TestClock.setDate` would remove the questioning.

tim-smart:
I think just changing the .setTime signature to accept DateTime.Input would be better.

vinassefranche:
> I think just changing the .setTime signature to accept DateTime.Input would be better.

Ah yes, good idea! I can try to do that!

vinassefranche:
> I think just changing the .setTime signature to accept DateTime.Input would be better.

@tim-smart this is done!

vinassefranche:
Ok, I misunderstood your comment, sorry....
and thanks for taking over!

vinassefranche:
I updated the tests and both tests were doing the exact same things. Feel free to get rid of some of the three tests if you feel it's too much

packages/effect/test/Schedule.test.ts
```diff
@@ -894,11 +894,12 @@ const checkRepetitions = <Env>(schedule: Schedule.Schedule<number, number, Env>)
     return [actual, expected] as const
   })
 }
+const MAX_DATE_TIMESTAMP = 8640000000000000
 export const run = <A, E, R>(
   effect: Effect.Effect<A, E, R>
 ): Effect.Effect<A, E, R> => {
   return Effect.fork(effect).pipe(
-    Effect.tap(() => TestClock.setTime(Number.POSITIVE_INFINITY)),
+    Effect.tap(() => TestClock.setTime(MAX_DATE_TIMESTAMP)),
```
vinassefranche, line 902: `TestClock.setTime(Number.POSITIVE_INFINITY)` return a fail of `Invalid date`
`8640000000000000` is the max date.
Should I export `maxEpochMillis` from the internal DateTime and use it here?

vinassefranche: I implemented it. Feel free to revert the commit if you think it's a bad idea

.changeset/sixty-starfishes-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
+---
+
+Make TestClock.setTime accept a Date input
```
tim-smart, line 5:
```suggestion
Make TestClock.setTime accept a DateTime.Input
```

PR #4515, opened 2025-02-26
chore: test types against TypeScript’s nightly releases

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

To draw your attention, passing `--target '>=5.4'` to TSTyche means that newly released TypeScript versions will be picked as well. For instance, `5.8` will be used as soon as it will be available on `npm`. This means that an arbitrary PR might be failing, because of unrelated reasons.

I think it is worth to run type tests agains TypeScript’s `next` version which is released daily. This can serve as an early warning solution.

Feel free to adjust the details. A good question is: how to setup notifications? GitHub Actions [documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) states:

> Notifications for scheduled workflows are sent to the user who last modified the cron syntax in the workflow file.

Hm.. Not sure about that. Should be some smarter way to setup notifications.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,17 @@
+name: TypeScript's Nightly
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 12 * * *"
+
+jobs:
+  types:
+    name: Types
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        uses: ./.github/actions/setup
+      - run: pnpm test-types --target next


fubhy:
I am all for this but just running a nightly workflow is not going to be super useful without giving it any visibility. Noone is going to monitor the workflow result.

Test failures should be surfaced somewhere with more visibility for this to be useful.
Suggestions?

fubhy:
> Perhaps a solution could be to use [`jayqi/failed-build-issue-action`](https://github.com/jayqi/failed-build-issue-action) that could open new issue when the workflow fails?
> 
> I can open a PR. Here is usage example: https://github.com/jestjs/jest/blob/28b4469e690d9c006a328b67f275362842a995f3/.github/workflows/nightly.yml#L51. And here is how the issue looks like: [jestjs/jest#14993](https://github.com/jestjs/jest/issues/14993)
> 
> An alternative could be built a new TSTyche specific action. Ideally it could open an issue with more details about the failure.

Both sound good. The latter might a bit overkill for us only but if it is useful for tstyche in general go for it! Sounds useful imo.

PR #4528, opened 2025-02-27
chore: notify on failed TypeScript's nightly type test runs

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Following up #4515

As it was discussed, this PR improves the logic of TypeScript's nightly type test runs. I have added few lines that will open an issue, if the nightly runs would fail.

An example repo with the same code here: https://github.com/tstyche/ts-nightly-runs-example. And you can see an example issue as well: https://github.com/tstyche/ts-nightly-runs-example/issues/2

References:

- `actions/github-script`, provides globals and executes the code: https://github.com/actions/github-script
- `octokit.rest.issues.create`, can also take `labels` and `assignees`: https://octokit.github.io/rest.js/v21/#issues

---

At first I was playing with [`jayqi/failed-build-issue-action`](https://github.com/jayqi/failed-build-issue-action), but it required more or less the same amount of lines like`actions/github-script`.


mrazauskas:
Sorry, about force push. I mixed up repos. The code got simplified to use `context` from `actions/github-script` instead of GHA context variables.

.github/workflows/nightly.yml
```diff
@@ -1,10 +1,12 @@
...
-
+name: TypeScript's nightly
 on:
   workflow_dispatch:
   schedule:
     - cron: "0 12 * * *"
 
+permissions:
+  issues: write
+
```
IMax153, line 9: I think my preference would be to follow the semantics we have for our other workflows, and set the global permissions to none:

```
permissions: {}
```

and apply the required permissions to the jobs that need them.

Obviously here we only have one job, but as a best practice I feel we should keep it consistent.

Also pinging @fubhy here for additional review.

mrazauskas: Agreed. I simply didn’t think deep enough. The code was failing while I was testing it and these two lines made it work. Hence in this context they looked like a solution, not like a problem at all.

mrazauskas: Thanks for the insight. I have just pushed an update. Before doing that I also updated the example repo to be sure that all work as expected.

fubhy (approved):
Great! Thanks

PR #4539, opened 2025-03-01
Bug fix: change interruptible order in platform shutdown flow

schickling:
I was facing some process shutdown issues where the timeout seems to have blocked the shutdown of the process. After some investigation (thanks @fubhy) we realized that changing the order of the `Effect.interruptible` like the following fixed the issue:

```ts
      return Effect.as(
        Scope.addFinalizer(
          scope,
          Effect.suspend(() => {
            worker.postMessage([1])
            return Deferred.await(exitDeferred)
          }).pipe(
            Effect.timeout(5000),
            Effect.interruptible,
            Effect.catchAllCause(() => Effect.sync(() => worker.terminate()))
          )
        ),
        worker
      )
```

### DIFF
--- a/packages/platform-bun/src/internal/worker.ts
+++ b/packages/platform-bun/src/internal/worker.ts
@@ -19,8 +19,8 @@ const platformWorkerImpl = Worker.makePlatform<globalThis.Worker>()({
             worker.postMessage([1])
             return Deferred.await(closeDeferred)
           }).pipe(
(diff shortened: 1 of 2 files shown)


IMax153:
@schickling - thanks for opening this PR! However, it needs a changeset - you can generate one with `pnpm changeset`. A patch should be sufficient.

tim-smart (approved):

PR #4545, opened 2025-03-01
Avoid inclusion of `_tag` in tool call parameters

### LINKED ISSUE #4505: AiToolkit JsonSchema generation & Tagged Requests
Currently AiTools are defined using a TaggedRequest which is great because it defines the expected return type of the tool along with the expected payload. The only downside to using a TaggedRequest is that the `_tag` parameter is included in the JSONSchema that gets generated and sent along with the AI Completion request. 

This generally makes the schema unnecessarily complex and from my experience may cause some weaker (and not so weak!) models to struggle generating the correct data to fit the schema. 

This is unnecessary because the tool call is already returned by the tool name that should match the `_tag` parameter. There is [already a workaround in place](https://github.com/Effect-TS/effect/blob/cc5588df07f9103513547cb429ce041b9436a8bd/packages/ai/ai/src/Completions.ts#L421-L434) to inject the tag back into the data before decoding the data anyway which I think should just be the default.

I foresee this becoming a larger problem with the rise of [MCP Servers](https://modelcontextprotocol.io/introduction) which I am working on an [Effect implementation of](https://github.com/ghardin1314/effect-mcp). This schema will be exposed across some transport layer where the toolkit provider has no control over which model will be consuming the schema. 

I am not sure the best way to implement this though. Is there a way to extract the `payload` schema from the `TaggedRequest` as just an untagged schema (this would be useful for me in other places too). Or should we just strip the `_tag` from the AST [before generating the JSONSchema](https://github.com/Effect-TS/effect/blob/cc5588df07f9103513547cb429ce041b9436a8bd/packages/ai/ai/src/Completions.ts#L322C1-L331C2)

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #4505
- Closes #4505

### DIFF
--- a/.changeset/wicked-starfishes-bow.md
+++ b/.changeset/wicked-starfishes-bow.md
@@ -1,0 +1,5 @@
+---
+"@effect/ai": patch
+---
+
+Avoid inclusion of `_tag` in tool call JSONSchema
--- a/packages/ai/ai/src/Completions.ts
+++ b/packages/ai/ai/src/Completions.ts
@@ -315,7 +315,7 @@ const convertTool = <A, I, R>(
 ) => ({
   name,
   description: getDescription(schema.ast),
-  parameters: makeJsonSchema(schema.ast),
+  parameters: makeJsonSchema(AST.omit(schema.ast, ["_tag"])),
   structured
 })
 


IMax153:
@ghardin1314 - can you test out this PR and see if it solves your use case? I tested locally and it's working as I expect, but I'd also like your confirmation that this solves your issue.

You can test it out by installing the relevant packages from their snapshots: https://github.com/Effect-TS/effect/pull/4545/checks?check_run_id=38045911970

IMax153:
Going to go ahead and merge this in lieu of confirmation that it resolves the user's issue.

PR #4567, opened 2025-03-09
Add ParseOptions argument to standardSchemaV1

rehos:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Add optional ParseOptions argument to standardSchemaV1. This will make form validation in form libraries like [@tanstack/form](https://tanstack.com/form/latest) more usable (i.e. display all errors to the user instead of just the first)

### DIFF
--- a/.changeset/five-melons-leave.md
+++ b/.changeset/five-melons-leave.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add ParseOptions argument to standardSchemaV1
--- a/.changeset/five-melons-leave.md
+++ b/.changeset/five-melons-leave.md
@@ -2,4 +2,6 @@
 "effect": patch
 ---
 
-Add ParseOptions argument to standardSchemaV1
+Schema: `standardSchemaV1` now returns all errors by default and supports custom options.
+
+The `standardSchemaV1` now returns **all validation errors** by default (`ParseOptions = { errors: "all" }`). Additionally, it now accepts an optional `overrideOptions` parameter, allowing you to customize the default parsing behavior as needed.
--- a/packages/effect/test/Schema/Schema/standardSchemaV1.test.ts
+++ b/packages/effect/test/Schema/Schema/standardSchemaV1.test.ts
@@ -252,7 +252,7 @@ describe("standardSchemaV1", () => {
       a: Schema.NonEmptyString,
       b: Schema.NonEmptyString
     })
-    const standardSchema = Schema.standardSchemaV1(schema, { errors: "all" })
+    const standardSchema = Schema.standardSchemaV1(schema)
     expectSyncSuccess(standardSchema, {
       a: "a",
       b: "b"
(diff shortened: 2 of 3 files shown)


packages/effect/test/Schema/Schema/standardSchemaV1.test.ts
```diff
@@ -2,6 +2,7 @@ import type { StandardSchemaV1 } from "@standard-schema/spec"
 import { Context, Effect, ParseResult, Predicate, Schema } from "effect"
 import { assertTrue, deepStrictEqual, strictEqual } from "effect/test/util"
 import { describe, it } from "vitest"
+import { b } from "vitest/dist/chunks/suite.BJU7kdY9.js"
```
gcanti, line 5: this seems like an incorrect automatic import

gcanti (changes requested):
I wonder if we should make `{ errors: "all" }` the default. AFAIK, the spec doesn't specify how many issues should be raised by default.

packages/effect/test/Schema/Schema/standardSchemaV1.test.ts
rehos: Sorry I will fix that

rehos:
I would prefer the default with `{ errors: "all" }`. Should we do something like:

```ts
export const standardSchemaV1 = <A, I>(
  schema: Schema<A, I, never>,
  overrideOptions?: AST.ParseOptions // Changed
): StandardSchemaV1<I, A> => {
  const decodeUnknown = ParseResult.decodeUnknown(schema, { errors: "all" }) // Changed
  return {
    "~standard": {
      version: 1,
      vendor: "effect",
      validate(value) {
        const scheduler = new scheduler_.SyncScheduler()
        const fiber = Effect.runFork(
          Effect.matchEffect(decodeUnknown(value, overrideOptions), { // Changed
            onFailure: makeStandardFailureFromParseIssue,
            onSuccess: (value) => Effect.succeed({ value })
          }),
          { scheduler }
        )
        scheduler.flush()
        const exit = fiber.unsafePoll()
        if (exit) {
          return makeStandardResult(exit)
        }
        return new Promise((resolve) => {
          fiber.addObserver((exit) => {
            resolve(makeStandardResult(exit))
          })
        })
      }
    }
  }
}

**Comment by gcanti** [comment]
> Should we do something like:

Yes 👍

**Comment by rehos** [comment]
@gcanti I have changed the default to all errors with the possibility override the parse options. Also added an extra test for the override use case with`{ errors: "first" }`. And the incorrect import has been fixed.

File: .changeset/five-melons-leave.md [hunk 2]
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add ParseOptions argument to standardSchemaV1
```
gcanti, line 5:
```suggestion
Schema: `standardSchemaV1` now returns all errors by default and supports custom options.

The `standardSchemaV1` now returns **all validation errors** by default (`ParseOptions = { errors: "all" }`). Additionally, it now accepts an optional `overrideOptions` parameter, allowing you to customize the default parsing behavior as needed.
```

gcanti (approved):

gcanti:
Thanks @rehos

rehos:
Thank you @gcanti for your quick response and feedback.

PR #4574, opened 2025-03-10
fix: unnecessary `as` in `all` function in the Effect `Either` module

giuliobracci:
## Type

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Removed unnecessary `as` in `all` function from the `Either` module. Currently the `input` is correctly inferred by the compiler as an `Iterable<Either<any, any>>`.  I believe the `as` in this case was a leftover of a previous implementation.

I also tried to remove the `@ts-expect-error` at the start of the function, but I did not manage to fix the type error.

## Related

- Related Issue #
- Closes #


giuliobracci:
Thanks for merging!

PR #4586, opened 2025-03-11
fix: unnecessary `as` type cohercion in `Either.fromNullable`

giuliobracci:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [X] Optimization
- [ ] Documentation Update

## Description

Fixes unnecessary `as` in `Either.fromNullable`. Covered the change with type tests.

### DIFF
--- a/packages/effect/dtslint/Either.tst.ts
+++ b/packages/effect/dtslint/Either.tst.ts
@@ -137,6 +137,38 @@ describe("Either", () => {
     ).type.toBe<Either.Either<number, "b">>()
   })
 
+  it("fromNullable", () => {
+    const nullableString = hole<string | null>()
+    const nullableObject = hole<{ a: string } | undefined>()
+
+    expect(
+      Either.fromNullable(
+        nullableString,
+        () => new Error()
+      )
+    ).type.toBe<Either.Either<string, Error>>()
+
+    expect(
+      pipe(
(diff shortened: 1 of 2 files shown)


packages/effect/dtslint/Either.tst.ts
```diff
@@ -137,6 +137,38 @@ describe("Either", () => {
     ).type.toBe<Either.Either<number, "b">>()
   })
 
+  it("fromNullable", () => {
+    const nullableString: string | null = "A"
+    const nullableObject: { a: string } | undefined = { a: "A" }
```
gcanti, line 142:
```suggestion
    const nullableString = hole<string | null>()
    const nullableObject = hole<{ a: string } | undefined>()
```

gcanti: It's always better to avoid using true constant definitions in type-level tests (TypeScript often retains the actual value as its type instead of applying the type annotation)

gcanti (changes requested):

packages/effect/dtslint/Either.tst.ts
giuliobracci: @gcanti Thank you for the tip.

gcanti (approved):

PR #4597, opened 2025-03-13
Add Effect.transposeMapOption

### LINKED ISSUE #3142: Option.mapEffect

LaureRC:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add `Effect.traverseOption`. Today in our codebase we're using a helper to do that, but I really think it would be a nice addition that would be useful to others. 

I'm aware that we could use the `Effect.transposeOption`:
```ts
pipe(
  Effect.succeed(Option.some('value')), 
  Effect.map(Option.map(someEffectfulOperation)), 
  Effect.flatMap(Effect.transposeOption)
)
```
But this is simpler and more relevant for all the use cases where I need it in my work:
```ts
pipe(
  Effect.succeed(Option.some('value')),
  Effect.flatMap(Effect.traverseOption(someEffectfulOperation))
)
```
So I'd still push for this addition, hence my move to try to open a PR for it! :)

For the naming of this method, in our codebase we used `optionTraverseEffect` but given the discussions in issue #3142, it seems that adding this method in the Effect module and using the term `traverse` would be the way to go.
Also, if you agree with this addition and have ideas on improving the documentation part of this feature I'm interested as I'm not fully happy with my final version.

## Related

- Related Issue #3142
- Closes #3142

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -13084,7 +13084,7 @@ export const transposeOption = <A = never, E = never, R = never>(
  * //          ▼
  * const noneResult = yield* pipe(
  *   Option.none(),
- *   Effect.traverseOption(() => Effect.succeed(42)) // will not be executed
+ *   Effect.transposeMapOption(() => Effect.succeed(42)) // will not be executed
  * )
  * console.log(Effect.runSync(noneResult))
  * // Output: { _id: 'Option', _tag: 'None' }
@@ -13093,7 +13093,7 @@ export const transposeOption = <A = never, E = never, R = never>(
  * //          ▼
  * const someSuccessResult = yield* pipe(
  *   Option.some(42),
- *   Effect.traverseOption((value) => Effect.succeed(value * 2))
+ *   Effect.transposeMapOption((value) => Effect.succeed(value * 2))
  * )
  * console.log(Effect.runSync(someSuccessResult))
  * // Output: { _id: 'Option', _tag: 'Some', value: 84 }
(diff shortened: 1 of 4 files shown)


gcanti:
I would go with `Effect.transposeMapOption` as it follows the same pattern as `filterMap`: applying `map` first, then `transpose`, just like `filterMap` applies `map` first, then `filter`. Additionally, both APIs would appear when you start typing `transpo...`.

packages/effect/src/Effect.ts
```diff
@@ -13068,6 +13068,49 @@ export const transposeOption = <A = never, E = never, R = never>(
...
+ *
+ * //          ┌─── Effect<Option<number>, never, never>>
+ * //          ▼
+ * const someSuccessResult = yield* pipe(
+ *   Option.some(42),
+ *   Effect.traverseOption((value) => Effect.succeed(value * 2))
+ * )
+ * console.log(Effect.runSync(someSuccessResult))
+ * // Output: { _id: 'Option', _tag: 'Some', value: 84 }
+ *
```
gcanti, line 13100: Missing closing fence ```.  

Note that the examples are type-checked. Once you add the closing fence, you should see an error when generating the docs. You can verify this locally by running `pnpm docgen` inside the `effect` package directory.

PR #4606, opened 2025-03-14
Fix issue with generic filters when generating arbitraries, closes #4605

### LINKED ISSUE #4605: From Discord: Regression in Arbitrary Generation with Schema.Date and Filter after Upgrade
# Summary
In the conversation, a user named `_adrian_g` reported an issue after upgrading the Effect library. They noticed a regression causing an exception when generating arbitraries using `Schema.Date` with an applied `filter`. They've provided a reproduction link to demonstrate the issue.

`gcanti` confirmed the reproduction of the issue by posting a code snippet. This snippet attempts to create an arbitrary using `Schema.Date` with a `filter`, resulting in a `TypeError` because a property `noInvalidDate` is undefined.

Key takeaways:
1. An upgrade to the Effect library introduced a regression affecting `Schema.Date` when used with `filter`.
2. The regression specifically causes a `TypeError` due to an undefined property when generating arbitraries.
3. Users encountering this issue may need a workaround or await a fix from the library maintainers.

# Discord thread

https://discord.com/channels/795981131316985866/1350179735572971520

gcanti:
Previously, applying a `filter` to a schema when generating arbitraries could cause a `TypeError` due to missing properties. This fix ensures that arbitraries are generated correctly when filters are used.

**Before**

```ts
import { Arbitrary, Schema } from "effect"

const schema = Schema.BigIntFromSelf.pipe(Schema.filter(() => true))

Arbitrary.make(schema)
// TypeError: Cannot read properties of undefined (reading 'min')
```

**After**

```ts
import { Arbitrary, Schema } from "effect"

const schema = Schema.BigIntFromSelf.pipe(Schema.filter(() => true))

const result = Arbitrary.make(schema) // Works correctly
```

### DIFF
--- a/.changeset/famous-ladybugs-talk.md
+++ b/.changeset/famous-ladybugs-talk.md
@@ -1,0 +1,28 @@
+---
+"effect": patch
+---
+
+Fix issue with generic filters when generating arbitraries, closes #4605.
+
+Previously, applying a `filter` to a schema when generating arbitraries could cause a `TypeError` due to missing properties. This fix ensures that arbitraries are generated correctly when filters are used.
+
+**Before**
+
+```ts
+import { Arbitrary, Schema } from "effect"
+
+const schema = Schema.BigIntFromSelf.pipe(Schema.filter(() => true))
+
+Arbitrary.make(schema)
+// TypeError: Cannot read properties of undefined (reading 'min')
+```
+
+**After**
+
+```ts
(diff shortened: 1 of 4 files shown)


joaopaulomoraes:
> const schema = Schema.BigIntFromSelf.pipe(Schema.filter(() => true))

`const schema = Schema.BigIntFromSelf.pipe(Schema.filter(Boolean))`

PR #4623, opened 2025-03-21
add jsonrpc serialization to rpc

tim-smart:


packages/rpc/src/RpcSerialization.ts
```diff
@@ -73,6 +75,152 @@ export const ndjson: Effect.Effect<RpcSerialization["Type"]> = Effect.sync(() =>
...
+      },
+      encode: (response_) => {
+        const response = response_ as RpcMessage.FromServerEncoded | RpcMessage.FromClientEncoded
+
+        let jsonRpcMessage: JsonRpcMessage
+        if (response._tag === "Request") {
+          jsonRpcMessage = {
+            jsonrpc: "2.0",
+            method: response.tag,
+            params: response.payload,
```
fubhy, line 145: Generally for the RPC module, `payload` should be:

a) optional
b) not just `Schema.Struct`

E.g. JSON RPC commonly uses positional tuples for payloads and for some methods, the payload can also be optional entirely. In these cases, the `params` key should not be set (but most implementations will accept an empty array or empty object anyways).

packages/rpc/src/RpcSerialization.ts
```diff
@@ -73,6 +75,152 @@ export const ndjson: Effect.Effect<RpcSerialization["Type"]> = Effect.sync(() =>
...
 })
 
+/**
+ * @since 1.0.0
+ * @category serialization
+ */
+export const jsonrpc: Effect.Effect<RpcSerialization["Type"]> = Effect.sync(() => {
+  const decoder = new TextDecoder()
+  return RpcSerialization.of({
+    contentType: "application/json-rpc",
```
fubhy, line 85: Some JSON RPC server will NOT accept `application/json-rpc` (as ridiculous as that sounds).

In these cases, we could either require the user to override the header in their client or alternatively we'd have to change how the muxing currently behaves when e.g. `application/json` is set and instead make that more explicitly configurable here (I think I'd prefer that).

packages/rpc/src/RpcSerialization.ts
```diff
@@ -73,6 +75,152 @@ export const ndjson: Effect.Effect<RpcSerialization["Type"]> = Effect.sync(() =>
...
+        return [rpcMessage]
+      },
+      encode: (response_) => {
+        const response = response_ as RpcMessage.FromServerEncoded | RpcMessage.FromClientEncoded
+
+        let jsonRpcMessage: JsonRpcMessage
+        if (response._tag === "Request") {
+          jsonRpcMessage = {
+            jsonrpc: "2.0",
+            method: response.tag,
```
fubhy, line 144: I don't like using the `tag` for the method here. Some JSON RPC methods can have awkward names that do not translate well to function names.

fubhy (changes requested):

packages/rpc/src/RpcSerialization.ts
```diff
@@ -73,6 +75,152 @@ export const ndjson: Effect.Effect<RpcSerialization["Type"]> = Effect.sync(() =>
...
+ * @category serialization
+ */
+export const jsonrpc: Effect.Effect<RpcSerialization["Type"]> = Effect.sync(() => {
+  const decoder = new TextDecoder()
+  return RpcSerialization.of({
+    contentType: "application/json-rpc",
+    supportsBigInt: false,
+    unsafeMake: () => ({
+      decode: (bytes) => {
+        const decoded: JsonRpcMessage = JSON.parse(typeof bytes === "string" ? bytes : decoder.decode(bytes))
```
fubhy, line 89: Some JSON RPC server support batch requests.

fubhy (changes requested):

PR #4630, opened 2025-03-21
Handle arbitrary length StreamChunkPart

### LINKED ISSUE #4629: `@effect/ai-openai` discards meaningful content with multiple `StreamChunkPart`

effect@3.14.1, @effect/ai-openai@0.16.2

Try accessing `.text` or `.asAiResponse.text` on a `StreamChunk` that contains both a "Usage" part and a "Content" part. These fields will appear empty, discarding the meaningful content in the chunk.

Playground: https://effect.website/play#266ba839c1c9

I expect any "Content" parts present in the chunk to appear in the `StreamChunk.text` and `StreamChunk.asAiResponse.text`

### What do you see instead?

Instead the `StreamChunk.text` and `StreamChunk.asAiResponse.text` appear empty 

### Additional information

This is necessary for using Google's Gemini models via the OpenAI compatibility API

https://ai.google.dev/gemini-api/docs/openai

When `{ stream_options: { include_usage: true } }` is provided to the API, the response stream looks something like this:

```
data: {"choices":[{"delta":{"content":"Hello","role":"assistant"},"index":0}],"created":1742591997,"model":"gemini-2.0-flash","object":"chat.completion.chunk","usage":{"completion_tokens":0,"prompt_tokens":10,"total_tokens":10}}

data: {"choices":[{"delta":{"content":" there! How","role":"assistant"},"index":0}],"created":1742591997,"model":"gemini-2.0-flash","object":"chat.completion.chunk","usage":{"completion_tokens":0,"prompt_tokens":10,"total_tokens":10}}

data: {"choices":[{"delta":{"content":" can I help you today?\n","role":"assistant"},"finish_reason":"stop","index":0}],"created":1742591997,"model":"gemini-2.0-flash","object":"chat.completion.chunk","usage":{"completion_tokens":11,"prompt_tokens":8,"total_tokens":19}}

data: [DONE]
```

Notice that each segment contains both "choices" and "usage". By comparison, OpenAI's endpoint returns one or the other, with a single "usage" segment at the end of the stream.

However, the current implementation of `@effect/ai-openai` makes some unsafe assumptions that there is only one part per StreamChunk.

Here:

https://github.com/Effect-TS/effect/blob/9a88bb9404ea6001b70d1aab62ce4630d5c56159/packages/ai/openai/src/OpenAiClient.ts#L287

And here:

https://github.com/Effect-TS/effect/blob/9a88bb9404ea6001b70d1aab62ce4630d5c56159/packages/ai/openai/src/OpenAiClient.ts#L299

By changing these to be resilient to arbitrary length `this.parts`, Google's Gemini models will work with the OpenAI client.

mattrossman:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This changes `OpenAiClient.StreamChunk.text` and `OpenAiClient.StreamChunk.asAiResponse` getters to respect arbitrary length `this.parts` instead of only considering the first part.

After patching this locally, I was able to use Google's Gemini models via [their OpenAI compatibility API.
](https://ai.google.dev/gemini-api/docs/openai)

### Areas to Review:
I'm not sure what is expected when there's multiple "Content" parts, all I know is that Gemini delivers one "Usage" and one "Content" part in the same chunk. For now I've `.join()`-ed all content parts into a single string.

I'm not certain why the `asAiReponse` implementation produces an empty string of content for the case when no parts are present (versus just providing an empty parts array).

https://github.com/Effect-TS/effect/blob/9a88bb9404ea6001b70d1aab62ce4630d5c56159/packages/ai/openai/src/OpenAiClient.ts#L293-L298

I've preserved this behavior in case, though I imagine it could be removed.

## Related

- Closes #4629

### DIFF
--- a/packages/ai/openai/src/OpenAiClient.ts
+++ b/packages/ai/openai/src/OpenAiClient.ts
@@ -291,13 +291,6 @@ export class StreamChunk extends Data.Class<{
    * @since 1.0.0
    */
   get asAiResponse(): AiResponse.AiResponse {
-    if (this.parts.length === 0) {
-      return AiResponse.AiResponse.fromText({
-        role: AiRole.model,
-        content: ""
-      })
-    }
-
     const aiResponseParts: Array<AiResponse.Part> = []
 
     for (let i = 0; i < this.parts.length; i++) {
@@ -314,6 +307,13 @@ export class StreamChunk extends Data.Class<{
           }))
           break
       }
+    }
+
+    if (aiResponseParts.length === 0) {
+      return AiResponse.AiResponse.fromText({
+        role: AiRole.model,
+        content: ""
+      })
     }
 
     return new AiResponse.AiResponse({
--- a/packages/ai/openai/src/OpenAiClient.ts
+++ b/packages/ai/openai/src/OpenAiClient.ts
@@ -318,7 +318,7 @@ export class StreamChunk extends Data.Class<{
 
     return new AiResponse.AiResponse({
       role: AiRole.model,
-      parts: Chunk.fromIterable(aiResponseParts)
+      parts: Chunk.unsafeFromArray(aiResponseParts)
     })
   }
 }
--- a/packages/ai/openai/src/OpenAiClient.ts
+++ b/packages/ai/openai/src/OpenAiClient.ts
@@ -284,7 +284,8 @@ export class StreamChunk extends Data.Class<{
    * @since 1.0.0
    */
   get text(): Option.Option<string> {
-    return this.parts[0]?._tag === "Content" ? Option.some(this.parts[0].content) : Option.none()
+    const contentParts = this.parts.filter((part) => part._tag === "Content").map((part) => part.content)
+    return contentParts.length > 0 ? Option.some(contentParts.join("")) : Option.none()
   }
   /**
    * @since 1.0.0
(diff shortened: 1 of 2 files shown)


packages/ai/openai/src/OpenAiClient.ts
```diff
@@ -284,37 +284,42 @@ export class StreamChunk extends Data.Class<{
...
             params: part.arguments
           }))
-        })
-      case "Usage":
-        return AiResponse.AiResponse.empty
+          break
+      }
+    }
+
+    if (aiResponseParts.length === 0) {
```
IMax153, line 312: I agree that we should leave this logic in here for now, at least until I've gotten some test cases written for `AiResponse`.

IMax153 (approved):

packages/ai/openai/src/OpenAiClient.ts
```diff
@@ -284,37 +284,42 @@ export class StreamChunk extends Data.Class<{
    * @since 1.0.0
    */
   get text(): Option.Option<string> {
-    return this.parts[0]?._tag === "Content" ? Option.some(this.parts[0].content) : Option.none()
+    const contentParts = this.parts.filter((part) => part._tag === "Content").map((part) => part.content)
+    return contentParts.length > 0 ? Option.some(contentParts.join("")) : Option.none()
```
tim-smart, line 288: I think instead of joining them all it should take the first matching text chunk.

I think if you configure openai to return multiple choices then you can have more than one part.

tim-smart (changes requested):

packages/ai/openai/src/OpenAiClient.ts
mattrossman: Thanks for the context, I've pushed a change. Let me know if I should change the `asAiResponse()` getter below that to do something similar (e.g. not sure if I should only allow one TextPart OR one ToolCallPart? Or one of both? Or one TextPart and arbitrary ToolCallPart?).

https://github.com/Effect-TS/effect/blob/4513ae8172796ac81dfc583156a48c033666c879/packages/ai/openai/src/OpenAiClient.ts#L293-L310

For now I'm leaving it as is since AiResponse seems appears designed to handle multiple text parts:

https://github.com/Effect-TS/effect/blob/9a88bb9404ea6001b70d1aab62ce4630d5c56159/packages/ai/ai/src/AiResponse.ts#L187-L197

packages/ai/openai/src/OpenAiClient.ts
```diff
@@ -284,37 +284,42 @@ export class StreamChunk extends Data.Class<{
...
+    if (aiResponseParts.length === 0) {
+      return AiResponse.AiResponse.fromText({
+        role: AiRole.model,
+        content: ""
+      })
     }
+
+    return new AiResponse.AiResponse({
+      role: AiRole.model,
+      parts: Chunk.fromIterable(aiResponseParts)
```
tim-smart, line 321:
```suggestion
      parts: Chunk.unsafeFromArray(aiResponseParts)
```

Can re-use the array without copying here.

tim-smart (approved):

PR #4648, opened 2025-03-25
Schema: `standardSchemaV1` now includes the schema, closes #4494

### LINKED ISSUE #4494: `Schema.standardSchemaV1` should include the schema itself

3.13.2

```ts
import { JSONSchema, Schema } from 'effect'

const Person = Schema.Struct({
  name: Schema.optionalWith(Schema.NonEmptyString, { exact: true })
})

// TypeError: Cannot read properties of undefined (reading '_tag')
console.log(JSONSchema.make(Schema.standardSchemaV1(Person)))
```

Returns the json schema normally

### What do you see instead?

`TypeError: Cannot read properties of undefined (reading '_tag')`

### Additional information

_No response_

gcanti:
This update fixes an issue where passing `Schema.standardSchemaV1(...)` directly to `JSONSchema.make` would throw a `TypeError`. The schema was missing from the returned object, causing the JSON schema generation to fail.

Now `standardSchemaV1` includes the schema itself, so it can be used with `JSONSchema.make` without issues.

**Example**

```ts
import { JSONSchema, Schema } from "effect"

const Person = Schema.Struct({
  name: Schema.optionalWith(Schema.NonEmptyString, { exact: true })
})

const standardSchema = Schema.standardSchemaV1(Person)

console.log(JSONSchema.make(standardSchema))
/*
{
  '$schema': 'http://json-schema.org/draft-07/schema#',
  '$defs': {
    NonEmptyString: {
      type: 'string',
      description: 'a non empty string',
      title: 'nonEmptyString',
      minLength: 1
    }
  },
  type: 'object',
  required: [],
  properties: { name: { '$ref': '#/$defs/NonEmptyString' } },
  additionalProperties: false
}
*/
```

### DIFF
--- a/.changeset/shy-pots-dream.md
+++ b/.changeset/shy-pots-dream.md
@@ -1,0 +1,40 @@
+---
+"effect": patch
+---
+
+Schema: `standardSchemaV1` now includes the schema, closes #4494.
+
+This update fixes an issue where passing `Schema.standardSchemaV1(...)` directly to `JSONSchema.make` would throw a `TypeError`. The schema was missing from the returned object, causing the JSON schema generation to fail.
+
+Now `standardSchemaV1` includes the schema itself, so it can be used with `JSONSchema.make` without issues.
+
+**Example**
+
+```ts
+import { JSONSchema, Schema } from "effect"
+
+const Person = Schema.Struct({
+  name: Schema.optionalWith(Schema.NonEmptyString, { exact: true })
+})
+
+const standardSchema = Schema.standardSchemaV1(Person)
+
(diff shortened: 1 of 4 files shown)


mmkal:
@gcanti thank you for this. Worth noting that this ended up breaking the [documented trpc `effect` parser support](https://trpc.io/docs/server/validators#with-effect) since trpc currently checks `if (typeof parser === 'function')` before checking `if ('~standard' in parser)`.

I think it makes sense to fix on the trpc side though, see https://github.com/trpc/trpc/pull/6634.

PR #4649, opened 2025-03-25
fixed typo in jsdocs

jessekelly881:
Typo "fibe" in jsdocs


fubhy (commented):
I think this is the wrong fix.

This should probably read "Vibe".

PR #4667, opened 2025-03-29
docs(effect/HashSet): fix `HashSet.md` api docs.

suddenlyGiovanni:
Fixed `HashSet.md` api docs, previously broken by a parsing issue with some JsDoc tags: 
- `@example`
- `@linkcode`

e.g. check the current broken state of https://effect-ts.github.io/effect/effect/HashSet.ts.html#fromiterable 

This PR addresses all these issues.

## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/brave-dolphins-sort.md
+++ b/.changeset/brave-dolphins-sort.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix: `HashSet.md` api docs; previously broken by issue with Docgen JsDoc parser.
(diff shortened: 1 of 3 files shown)


gcanti:
FYI, you can move examples directly into the description instead of relying on `@example` tags. I'm gradually adopting this approach in other modules (like `Effect`) so that I can include a title for each example (e.g. `**Example** (...title)`).

PR #4675, opened 2025-03-31
`Match.tagsTupleExhaustive` has been added

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Match.tagsTupleExhaustive` has been added

```ts
import { Either, Match, Option } from "effect"

Match.value([Option.some(1), Either.left(33)]).pipe(
  Match.tagsTupleExhaustive({
    NoneLeft: (none, left) => {}, // (none: None<number>, left: Left<number, never>) => void
    NoneRight: (none, right) => {}, // (none: None<number>, left: Right<number, never>) => void
    SomeLeft: (some, left) => {}, // (none: Some<number>, left: Left<number, never>) => void
    SomeRight: (some, right) => {} // (none: Some<number>, left: Right<number, never>) => void
  })
)
```

It also preserves rendering of union member in type hints.
This is a special sugar for tagged sun-types designed to make it easier to work with a pattern in which the user wants to process all possible combinations of union members in a flat manner.

```ts
export const queenBeeState = (swarm: Swarm) => {
  const whiteSurroundedQueenBee = getSurroundedQueenBee(swarm, Side.White);
  const blackSurroundedQueenBee = getSurroundedQueenBee(swarm, Side.Black);

  return Match.value(
    distributive([whiteSurroundedQueenBee, blackSurroundedQueenBee])
  ).pipe(
    Match.when([{ _tag: "Some" }, { _tag: "Some" }], ([white, black]) =>
      QueenBeeState.BothSurrounded({ cells: [white.value, black.value] })
    ),
    Match.when([{ _tag: "None" }, { _tag: "Some" }], ([_white, black]) =>
      QueenBeeState.OneSurrounded({ cell: black.value })
    ),
    Match.when([{ _tag: "Some" }, { _tag: "None" }], ([white, _black]) =>
      QueenBeeState.OneSurrounded({ cell: white.value })
    ),
    Match.when([{ _tag: "None" }, { _tag: "None" }], ([_white, _black]) =>
      QueenBeeState.Free()
    ),
    Match.exhaustive
  );
};
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/tall-hotels-tap.md
+++ b/.changeset/tall-hotels-tap.md
@@ -1,0 +1,18 @@
+---
+"effect": minor
+---
+
+`Match.tagsTupleExhaustive` has been added
+
+```ts
+import { Either, Match, Option } from "effect"
+
+Match.value([Option.some(1), Either.left(33)]).pipe(
+  Match.tagsTupleExhaustive({
+    NoneLeft: (none, left) => {}, // (none: None<number>, left: Left<number, never>) => void
+    NoneRight: (none, right) => {}, // (none: None<number>, left: Right<number, never>) => void
+    SomeLeft: (some, left) => {}, // (none: Some<number>, left: Left<number, never>) => void
+    SomeRight: (some, right) => {} // (none: Some<number>, left: Right<number, never>) => void
+  })
+)
+```
--- a/packages/effect/dtslint/Match.tst.ts
+++ b/packages/effect/dtslint/Match.tst.ts
@@ -493,6 +493,49 @@ describe("Match", () => {
     )(value)
   })
 
+  it("tagsTupleExhaustive", () => {
+    pipe(
+      Match.type<[Value, Value]>(),
+      Match.tagsTupleExhaustive({
+        AA: (A1, A2) => {
+          expect(A1).type.toBe<{ _tag: "A"; a: number }>()
+          expect(A2).type.toBe<{ _tag: "A"; a: number }>()
+          return [A1.a, A2.a] as const
+        },
+        AB: (A, B) => {
+          expect(A).type.toBe<{ _tag: "A"; a: number }>()
+          expect(B).type.toBe<{ _tag: "B"; b: number }>()
+          return [A.a, "B"] as const
+        },
+        BA: (B, A) => {
+          expect(A).type.toBe<{ _tag: "A"; a: number }>()
+          expect(B).type.toBe<{ _tag: "B"; b: number }>()
+          return ["B", A.a] as const
+        },
+        BB: (B1, B2) => {
+          expect(B1).type.toBe<{ _tag: "B"; b: number }>()
+          expect(B2).type.toBe<{ _tag: "B"; b: number }>()
+          return ["B", "B"] as const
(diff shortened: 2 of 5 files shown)


KhraksMamtsov:
But maybe would be better to introduce `Tagged.ts` module with functions for `type Tagged = {_tag: string}`?
@mikearnaldi what do you think?

KhraksMamtsov:
@mikearnaldi  Hi! I'm sorry if I'm distracting you, but could you give me some feedback? What do you think about this feature?

KhraksMamtsov:
On one hand, using multiple concatenated tags as keys seems a bit strange. 
But I think the fact that tags are strings can be used to achieve more compact code.

tim-smart:
It feels a little too specific to be included. You can do thing like this already which gets pretty close:

```ts
import { Either, Match, Option } from "effect"

Match.value([Option.some(1), Either.left(33)]).pipe(
  Match.when([Option.isSome, Either.isLeft], ([some, left]) => {}),
  Match.when([Option.isNone, Either.isLeft], ([none, left]) => {}),
  Match.when([Option.isSome, Either.isRight], ([some, right]) => {}),
  Match.when([Option.isNone, Either.isRight], ([none, right]) => {}),
  Match.orElseAbsurd,
)
```

KhraksMamtsov:
I agree – but still, this option doesn't ensure completeness (`Match.exhaustive` doesn't work) and is more verbose.

Maybe it makes sense to include a helper like this then? With it, `Match.exhaustive` will work.
I think we could simply call it `distibutive` and overload it to work with both tuples and structs – what do you think?

https://effect.website/play#335756a0cfb9

```ts
import { Either, Match, Option } from "effect"

export type Distributive<
  I extends ReadonlyArray<any>,
  Result extends ReadonlyArray<any> = []
> = I extends readonly [infer T, ...infer Rest extends ReadonlyArray<any>]
  ? [T] extends [never] ? Distributive<Rest, [...Result, T]>
  : T extends infer TMember ? Distributive<Rest, [...Result, TMember]>
  : never
  : Result

export const distributiveTuple = <const T extends ReadonlyArray<any>>(
  value: T
): Distributive<T> => {
  return value as any as Distributive<T>
}

Match.value(distributiveTuple([Option.some(1), Either.left(33)])).pipe(
  Match.when([Option.isSome, Either.isLeft], ([some, left]) => {}),
  Match.when([Option.isNone, Either.isLeft], ([none, left]) => {}),
  Match.when([Option.isSome, Either.isRight], ([some, right]) => {}),
  Match.when([Option.isNone, Either.isRight], ([none, right]) => {}),
  Match.exhaustive
)
```

tim-smart:
Probably what really needs changing then is improving the exhaustiveness checks.

KhraksMamtsov:
@tim-smart 
What do you mean? Should we always consider the input data in a distributive way?

tim-smart:
The match exclusion logic should be updated so this is exhaustive by default.

PR #4688, opened 2025-04-02
Re-export `default` from Undici

OliverJAsh:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This fixes a type error in the following code:

```ts
import Undici from '@effect/platform-node/Undici';
new Undici.cacheStores.MemoryCacheStore()
```

### DIFF
--- a/packages/platform-node/src/Undici.ts
+++ b/packages/platform-node/src/Undici.ts
@@ -6,4 +6,5 @@
  * @since 1.0.0
  * @category undici
  */
+export { default } from "undici"
 export * from "undici"
--- a/packages/platform-node/src/Undici.ts
+++ b/packages/platform-node/src/Undici.ts
@@ -2,9 +2,9 @@
  * @since 1.0.0
  */
 
+export { default } from "undici"
 /**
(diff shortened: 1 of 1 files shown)


packages/platform-node/src/Undici.ts
```diff
@@ -2,6 +2,7 @@
  * @since 1.0.0
  */
 
+export { default } from "undici"
```
OliverJAsh, line 5: How should I define `@since`?

tim-smart: You can copy paste the jsdocs from the other export

OliverJAsh: Done, although I'm not sure it's correct to say the default export has been available since 1.0.0?

tim-smart: While platform is api unstable we are using 1.0.0 as the since tag for everything.

tim-smart (approved):
🚀

PR #4709, opened 2025-04-10
Add a timeout option to HttpClientRequest

thewilkybarkid:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a `timeout` option to `HttpClientRequest`, which results in a `Transport` `RequestError` if it's exceeded. I've erred on adding to the request rather than having it alongside to cater for cases where the client has been composed.

I'm sure that some of the implementation could be improved. 😄

## Related

- Thread https://discord.com/channels/795981131316985866/1359836164005302392
- Related Issue #
- Closes #

### DIFF
--- a/.changeset/beige-birds-attack.md
+++ b/.changeset/beige-birds-attack.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": minor
+---
+
+Add a timeout option to HttpClientRequest
--- a/packages/platform/src/HttpClient.ts
+++ b/packages/platform/src/HttpClient.ts
@@ -358,7 +358,7 @@ export const makeWith: <E2, R2, E, R>(
  */
 export const make: (
   f: (
-    request: ClientRequest.HttpClientRequest,
+    request: Omit<ClientRequest.HttpClientRequest, "timeout">,
     url: URL,
     signal: AbortSignal,
     fiber: RuntimeFiber<ClientResponse.HttpClientResponse, Error.HttpClientError>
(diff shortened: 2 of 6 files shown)


mikearnaldi:
why not using plain Effect.timeout?

thewilkybarkid:
I'm looking to turn these into `RequestError`s, which you can't (easily) with `Effect.timeout*`.

As mentioned in https://discord.com/channels/795981131316985866/1359836164005302392/1360164805025206322, composing the `HttpClient` with one that times out instead with an API like:

```ts
HttpClient.get('http://example.com/'),
HttpClient.withTimeout('1 second'),
Effect.andThen(...)
```

would work too.

tim-smart:
You can add timeouts with customs errors like this:

```ts
Effect.gen(function*() {
  const client = (yield* HttpClient.HttpClient).pipe(
    HttpClient.transform((effect, request) =>
      Effect.timeoutFail(effect, {
        duration: 1000,
        onTimeout: () => new Error(`${request.method} ${request.url} timed out`)
      })
    )
  )
})
```

fubhy:
Closing due to inactivity and merge conflicts. Maintainers pointed out that this can already be modeled with `HttpClient.transform` and `Effect.timeoutFail`; if a built-in timeout API is still desired, please open a fresh PR against current `main`/`next-minor` with the proposed API clarified.

PR #4718, opened 2025-04-12
feat(effect): Introduce new algebras modules for `NaturalNumber` and `Integers`

suddenlyGiovanni:
This pull request introduces new modules for working with numeric types and enhances the existing `Number` module. It includes changes to the standard library, internal implementation, and test cases.

### New Modules and Enhancements:

* **Standard Library**: Added `Integer` and `NaturalNumber` modules for handling whole numbers and non-negative integers, respectively. Refactored the `Number` module to align with these new modules.

### Internal Implementation:

* **Internal Number Module** (`packages/effect/src/internal/number.ts`):
  - Introduced `IntegerConstructor` and `NaturalNumberConstructor` for type-safe numeric operations.
  - Added various mathematical operations including `sum`, `subtract`, `multiply`, `negate`, `unsafeDivide`, `divide`, `increment`, `decrement`, `sumAll`, `multiplyAll`, `remainder`, `pow`, `square`, and `cube`.
  - Implemented error handling for division operations with `DivisionByZeroError`.

## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Related

- Related Issue #
- Closes #


suddenlyGiovanni:
Todo: need to rebase on next-minor. Some changes are not mine.

mikearnaldi:
A few notes:
1) I am not sure we should address this domain, I'll need some practical use cases beyond theory
2) Number doesn't represent a Real number, it's a floating point, there are ways to represent real numbers in programming languages but I am unsure how efficient it can be
3) we should probably call NaturalNumber simply Natural
4) Integer unless it is backed by arbitrary precision isn't going to be lawful
5) Generally speaking mathematically sound implementations are far from trivial

I think the most important thing is to decide if we even want this (point 1). cc @gcanti @IMax153 @tim-smart

suddenlyGiovanni:
> A few notes:
> 
> 1. I am not sure we should address this domain, I'll need some practical use cases beyond theory
> 2. Number doesn't represent a Real number, it's a floating point, there are ways to represent real numbers in programming languages but I am unsure how efficient it can be
> 3. we should probably call NaturalNumber simply Natural
> 4. Integer unless it is backed by arbitrary precision isn't going to be lawful
> 5. Generally speaking mathematically sound implementations are far from trivial
> 
> I think the most important thing is to decide if we even want this (point 1). cc @gcanti @IMax153 @tim-smart

Thanks for the feedback ❤️ .

Besides point `1)`, which is primarily up to you all to decide if it is worth being part of the "standard library, "I have many open questions related to the module API design that would need further analysis and discussion.

That said, I'll wait for you all to make a decision before annoying you further with lower-level details.

mikearnaldi:
Unless we have concrete use cases that are common enough it won't be accepted

tim-smart:
I feel like Number, BigInt & BigDecimal are more than enough :)

IMax153:
While I certainly appreciate the effort put in by @suddenlyGiovanni on this PR, I also am not sure we need a dedicated Number module. 

I kind of agree with Tim that the native APIs are "enough".

suddenlyGiovanni:
Thank you all for the feedback 🙇🏻 . 
I understand all your concerns and pov.

I will move these utils in the user space and see what happens.

I'm planning to spin off out of this PR 
- small TsDoc documentation/improvement for the existing `Number.ts` module at the module level and for each missing symbol. 
- small `Number.test.ts` improvements
- fix to misleading `Number.unsafeDivide` documentation [throw condition](https://github.com/Effect-TS/effect/pull/4718/files#diff-ed2e25e59741d9074cce2b7eae159f74372b63781cf95e27802f70db41ed48a5L116) .  

On the side of the effect/website, 
I'd like to add the documentation section for the existing `Number` module under the `Data types` section. I think users are unaware of its existence and are not leveraging it even when they would benefit. 

What do you think?

PR #4724, opened 2025-04-13
fix(ConfigProvider.fromEnv): use import.meta.env instead of process.env

coffeeispower:
## Type

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The reason for this is that this allows using environment variables in the browser with vite. Vite has a feature where all environment variables that have a `VITE_` prefix are exposed in `import.meta.env`. And even in node.js, `import.meta.env` is a alias to `process.env` so i don't see any reason why `ConfigProvider.fromEnv` only supported `process.env`

### DIFF
--- a/packages/effect/src/internal/configProvider.ts
+++ b/packages/effect/src/internal/configProvider.ts
@@ -108,8 +108,7 @@ export const fromEnv = (
   const makePathString = (path: ReadonlyArray<string>): string => pipe(path, Arr.join(pathDelim))
   const unmakePathString = (pathString: string): ReadonlyArray<string> => pathString.split(pathDelim)
 
-  const getEnv = () =>
-    typeof process !== "undefined" && "env" in process && typeof process.env === "object" ? process.env : {}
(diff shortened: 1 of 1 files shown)


mikearnaldi:
Unfortunately `import.meta.env` is an es-module only feature, it doesn't work in CJS

coffeeispower:
This is a difficult problem because import.meta is a syntax error when running in CJS for some f reason, i think we can do some weird hacks like dynamically importing another file using the import() function which just exports import.meta so it works for both EJS and CJS

mikearnaldi:
> This is a difficult problem because import.meta is a syntax error when running in CJS for some f reason, i think we can do some weird hacks like dynamically importing another file using the import() function which just exports import.meta so it works for both EJS and CJS.

in 4.0 we'll be publishing a single ESM module given that now node LTS supports sync loading of ESM modules so it won't be a problem any longer

PR #4726, opened 2025-04-13
add AiModels service to AnthropicClient.layerConfig

dearlordylord:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The problem, more specifically, is that the guide https://effect.website/docs/ai/introduction/ won't work if I simply substitute OpenAI with Anthropic. The change should be backward-compatible (only adds a ROut)

Meanwhile I expect ROut of Anthropic and OpenAI be the same, more so ROut of .layer and .layerConfig

### DIFF
--- a/.changeset/strange-hairs-remain.md
+++ b/.changeset/strange-hairs-remain.md
@@ -1,0 +1,5 @@
+---
+"@effect/ai-anthropic": minor
+---
+
+AnthropicClient.layerConfig has the same ROut as AnthropicClient.layer - specifically, AiModels.AiModels was missing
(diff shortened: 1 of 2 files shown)


.changeset/strange-hairs-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/ai-anthropic": minor
```
tim-smart, line 2:
```suggestion
"@effect/ai-anthropic": patch
```

Issue #3473, opened 2024-08-18
IndexedDB support in browser platform package


schickling:
### What is the problem this feature would solve?

Would be great if the `@effect/platform-browser` package would provide a module for [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API).

### What is the feature you are proposing to solve the problem?

.

fubhy:
Closing this as stale since there is no current maintainer-owned action on the issue and the related IndexedDB PR is closed. This can be revisited with a fresh proposal or implementation target if there is renewed interest.

Issue #2647, opened 2024-04-28
Add indexedDbKeyValueStore to @effect/platform-browser


mattiamanzati:
### What is the problem this feature would solve?

Right now the platform-browser packages exposes session storage and local storage.
Unfortunately those seems not to be an option while running inside a webworker or a sharedworker.

### What is the feature you are proposing to solve the problem?

We could support indexeddb which is available both in a UI tab, and in a webworker context.

jessekelly881:
I have an implementation of this here https://github.com/jessekelly881/effect-idb. I could probably create a PR to add this.

schickling:
Would be super useful. Maybe @myobie could take a look at the eventual PR as well, as he knows all the nuances and pitfalls related to IDB so we make sure to not wrap it in a leaky way.

jessekelly881:
+1 my wrapper is basically just duct tape and hope lol.

fubhy:
Closing as stale for now. This can be reopened if there is current maintainer ownership for adding IndexedDB support to platform-browser.

PR #4739, opened 2025-04-15
Added modules for IndexedDb

SandroMaglione:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Closes:
- https://github.com/Effect-TS/effect/issues/3473
- https://github.com/Effect-TS/effect/issues/2647

## Description

Add modules for [IndexedDb](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API). This PR includes the following new modules (in order of dependency between each other):
- `IndexedDbTable`: define a table (with `Schema`)
- `IndexedDbVersion`: collects tables in a database version
- `IndexedDbMigration`: collects a sequence of versions and their schema + data migrations
- `IndexedDbDatabase`: service that initialises the database and executes the migration
- `IndexedDbQuery`: api to interact with the database

This also adds `BrowserIndexedDb` to `@effect/platform-browser` with a layer to create an instance of `IndexedDb` from the browser `window`.

> Initial full API example [included in the PR](https://github.com/Effect-TS/effect/blob/acb681740f4db337f982132b413311590d94274a/packages/platform-browser/examples/indexedDb.ts)

## Notes
- I added the modules inside `@effect/platform` since they are technically "generic" (not dependent on the browser `window`)
- Currently the migration doesn’t automatically create object stores and indexes (even if it would be possible). That's because the user may want to read the previous data for data migrations. Note that IndexedDb requires to delete the previous object store and create a new one whenever an index changes
- Migrations and transactions cannot have async effects inside them. That's a "limitation" of IndexedDb. [From the docs](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#adding_data_to_the_database):
> Transactions are tied very closely to the event loop. If you make a transaction and return to the event loop without using it then the transaction will become inactive.

- There are more features that can be added (e.g. reactivity), but I would leave them as additions for later

> [!NOTE]
> I plan to add more examples and `README` after the PR review, to avoid having to rewrite it all because of API changes

***

## Notes for review
The modules can be reviewed in the following order (each directly depends/references the previous module):
1. `IndexedDbTable`
2. `IndexedDbVersion`
3. `IndexedDbMigration`
4. `IndexedDbDatabase`
5. `IndexedDbQuery`

`IndexedDbTable` and `IndexedDbVersion` don't contain any logic, they are just data containers (schemas).

`IndexedDbMigration` contains code to execute migrations (e.g. `createObjectStore`/`deleteObjectStore`).

`IndexedDbDatabase` is mainly used to initialise the database inside its `layer`.

`IndexedDbQuery` is the main service used to extract a type safe API and execute queries. The code that interacts with IndexedDb is mainly inside `internal/indexedDbQuery.ts`.

`IndexedDb` is a service to provide a valid instance of IndexedDb (with `BrowserIndexedDb` exporting a `window` layer).


packages/platform/test/IndexedDbDatabase.test.ts
```diff
@@ -0,0 +1,232 @@
...
+      Effect.gen(function*() {
+        yield* api.createObjectStore("todo")
+        yield* api.createIndex("todo", "titleIndex")
+        yield* api.from("todo").insert({ id: 1, title: "test", completed: false })
+      }))
+    {}
+
+    return Effect.gen(function*() {
+      const { makeApi, use } = yield* IndexedDbQuery.IndexedDbApi
+      const api = makeApi(Db)
```
tim-smart, line 54: This doesn't feel quite right yet. Maybe a field on the IndexedDbVersion would be better.

packages/platform/src/IndexedDbDatabase.ts
```diff
@@ -0,0 +1,282 @@
...
+                  })
+                  const toApi = internal.makeTransactionProto({
+                    database,
+                    IDBKeyRange,
+                    source: migration.toVersion,
+                    transaction
+                  })
+
+                  return prev.then(() =>
+                    Effect.runPromise(migration.execute(fromApi, toApi)).catch(
```
tim-smart, line 227: Instead of .runPromise, you can resume with the migration.execute

tim-smart:
I think this should just live in platform-browser, I can't see us adding implementations for any other platform.

packages/platform/test/IndexedDbDatabase.test.ts
SandroMaglione: I moved `IndexedDbApi` inside `IndexedDbVersion`. Do you think is possible to include the api directly inside the `IndexedDbVersion` interface? Not sure because it's a context with a generic parameter

packages/platform/src/IndexedDbDatabase.ts
SandroMaglione: What do you mean here? I moved the error handling inside `mapError`

tim-smart: Something like this:

```patch
diff --git a/packages/platform-browser/src/IndexedDbDatabase.ts b/packages/platform-browser/src/IndexedDbDatabase.ts
index 703da5533..43cae93f3 100644
--- a/packages/platform-browser/src/IndexedDbDatabase.ts
+++ b/packages/platform-browser/src/IndexedDbDatabase.ts
@@ -176,7 +176,7 @@ export const layer = <
             oldVersion = event.oldVersion
 
             if (transaction === null) {
-              resume(
+              return resume(
                 Effect.fail(
                   new IndexedDbDatabaseError({
                     reason: "TransactionError",
@@ -184,92 +184,68 @@ export const layer = <
                   })
                 )
               )
-            } else {
-              transaction.onabort = (event) => {
-                resume(
-                  Effect.fail(
-                    new IndexedDbDatabaseError({
-                      reason: "Aborted",
-                      cause: event
-                    })
-                  )
-                )
-              }
+            }
 
-              transaction.onerror = (event) => {
-                resume(
-                  Effect.fail(
-                    new IndexedDbDatabaseError({
-                      reason: "TransactionError",
-                      cause: event
-                    })
-                  )
+            transaction.onabort = (event) => {
+              resume(
+                Effect.fail(
+                  new IndexedDbDatabaseError({
+                    reason: "Aborted",
+                    cause: event
+                  })
                 )
-              }
+              )
+            }
 
-              migrations.slice(oldVersion).reduce((prev, untypedMigration) => {
-                if (untypedMigration._tag === "Migration") {
-                  const migration = untypedMigration as IndexedDbMigration.IndexedDbMigration.AnyMigration
-                  const fromApi = internal.makeTransactionProto({
-                    database,
-                    IDBKeyRange,
-                    source: migration.fromVersion,
-                    transaction
-                  })
-                  const toApi = internal.makeTransactionProto({
-                    database,
-                    IDBKeyRange,
-                    source: migration.toVersion,
-                    transaction
+            transaction.onerror = (event) => {
+              resume(
+                Effect.fail(
+                  new IndexedDbDatabaseError({
+                    reason: "TransactionError",
+                    cause: event
                   })
+                )
+              )
+            }
 
-                  return prev.then(() =>
-                    Effect.runPromise(
-                      migration.execute(fromApi, toApi).pipe(
-                        Effect.mapError((cause) =>
-                          resume(
-                            Effect.fail(
-                              new IndexedDbDatabaseError({
-                                reason: "UpgradeError",
-                                cause
-                              })
-                            )
-                          )
-                        )
-                      )
-                    )
-                  )
-                } else if (untypedMigration._tag === "Initial") {
-                  const migration = untypedMigration as IndexedDbMigration.IndexedDbMigration.AnyInitial
-                  const api = internal.makeTransactionProto({
-                    database,
-                    IDBKeyRange,
-                    source: migration.version,
-                    transaction
-                  })
+            const effect = Effect.forEach(migrations.slice(oldVersion), (untypedMigration) => {
+              if (untypedMigration._tag === "Migration") {
+                const migration = untypedMigration as IndexedDbMigration.IndexedDbMigration.AnyMigration
+                const fromApi = internal.makeTransactionProto({
+                  database,
+                  IDBKeyRange,
+                  source: migration.fromVersion,
+                  transaction
+                })
+                const toApi = internal.makeTransactionProto({
+                  database,
+                  IDBKeyRange,
+                  source: migration.toVersion,
+                  transaction
+                })
+                return migration.execute(fromApi, toApi)
+              } else if (untypedMigration._tag === "Initial") {
+                const migration = untypedMigration as IndexedDbMigration.IndexedDbMigration.AnyInitial
+                const api = internal.makeTransactionProto({
+                  database,
+                  IDBKeyRange,
+                  source: migration.version,
+                  transaction
+                })
+                return migration.execute(api)
+              }
+              return Effect.dieMessage("Invalid migration")
+            }, { discard: true })
 
-                  return prev.then(() =>
-                    Effect.runPromise(
-                      migration.execute(api).pipe(
-                        Effect.mapError((cause) =>
-                          resume(
-                            Effect.fail(
-                              new IndexedDbDatabaseError({
-                                reason: "UpgradeError",
-                                cause
-                              })
-                            )
-                          )
-                        )
-                      )
-                    )
-                  )
-                } else {
-                  resume(Effect.dieMessage("Invalid migration"))
-                  return Promise.resolve()
-                }
-              }, Promise.resolve())
-            }
+            resume(effect.pipe(
+              Effect.mapError((cause) =>
+                new IndexedDbDatabaseError({
+                  reason: "UpgradeError",
+                  cause
+                })
+              ),
+              Effect.as(database)
+            ))
           }
 
           request.onsuccess = (event) => {
```

packages/platform/test/IndexedDbDatabase.test.ts
tim-smart: Something like:

```ts
const api = yield* Db
// or
const api = yield* Db.api
```

SandroMaglione: Done

packages/platform/src/IndexedDbDatabase.ts
SandroMaglione: `resume` with the successful value must be called from `onsuccess`, calling it inside `onupgradeneeded` causes the transaction to close (indexed db quirks). I left a single `runPromise` keeping the `Effect.forEach`

SandroMaglione: Done

packages/platform-browser/src/IndexedDb.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Instance of IndexedDb from the `window` object.
+ *
+ * @since 1.0.0
+ * @category constructors
+ */
+export const layerWindow: Layer.Layer<IndexedDb, ConfigError.ConfigError> = Layer.effect(
+  IndexedDb,
+  Effect.fromNullable(window).pipe(
```
tim-smart, line 56: This could be simplified to somthing like:

```ts
Effect.suspend(() => {
  if (window.indexedDB && window.IDBKeyRange) {
    ...
```

packages/platform-browser/src/IndexedDbDatabase.ts
```diff
@@ -0,0 +1,263 @@
...
+}
+
+const makeProto = <Id extends string>(options: {
+  readonly identifier: Id
+  readonly version: number
+  readonly database: globalThis.IDBDatabase
+  readonly IDBKeyRange: typeof globalThis.IDBKeyRange
+}): IndexedDbDatabase.Service<Id> => {
+  function IndexedDb() {}
+  Object.setPrototypeOf(IndexedDb, Proto)
```
tim-smart, line 114: You don't need the function here. You could probably just use a plain object.

packages/platform-browser/src/IndexedDbDatabase.ts
```diff
@@ -0,0 +1,263 @@
...
+                  IDBKeyRange,
+                  tables: migration.version.tables,
+                  transaction
+                })
+                return migration.execute(api)
+              }
+              return Effect.dieMessage("Invalid migration")
+            }, { discard: true })
+
+            Runtime.runPromise(runtime)(effect.pipe(
```
tim-smart, line 242: Do you need to use runSync here?

SandroMaglione: `runSync` seems to block `onupgradeneeded` from exiting (the tests timeout), while `runPromise` doesn’t

packages/platform-browser/src/IndexedDbMigration.ts
```diff
@@ -0,0 +1,166 @@
...
+ * @since 1.0.0
+ * @category type ids
+ */
+export type ErrorTypeId = typeof ErrorTypeId
+
+/**
+ * @since 1.0.0
+ * @category errors
+ */
+export const IndexedDbMigrationError = internal.IndexedDbMigrationError
```
tim-smart, line 39: This needs an explicit type as internal types are stripped

packages/platform-browser/src/IndexedDbMigration.ts
```diff
@@ -0,0 +1,166 @@
...
+   * @category models
+   */
+  export type AnyInitial = Initial<IndexedDbVersion.IndexedDbVersion.AnyWithProps, any>
+}
+
+/**
+ * @since 1.0.0
+ * @category constructors
+ */
+export const make = internal.makeInitialProto
```
tim-smart, line 166: This needs an explicit type as internal types are stripped

packages/platform-browser/src/IndexedDbQuery.ts
```diff
@@ -0,0 +1,418 @@
...
+      IndexedDbVersion.IndexedDbVersion.Tables<Source>
+    > = never,
+    Index extends IndexFromTable<Source, Table> = never
+  > {
+    [Symbol.iterator](): Effect.EffectGenerator<Effect.Effect<number>>
+
+    readonly [TypeId]: TypeId
+    readonly from: From<Source, Table>
+    readonly index?: Index
+    readonly only?: internal.ExtractIndexType<Source, Table, Index>
```
tim-smart, line 177: You can't use internal types in public modules

packages/platform-browser/src/IndexedDbVersion.ts
```diff
@@ -0,0 +1,119 @@
...
+ * @category interface
+ */
+export interface IndexedDbVersion<
+  out Tables extends IndexedDbTable.IndexedDbTable.AnyWithProps = never
+> extends Pipeable {
+  new(_: never): {}
+
+  readonly [TypeId]: TypeId
+  readonly tables: HashMap.HashMap<string, Tables>
+  readonly api: Effect.Effect<
```
tim-smart, line 38: I think `getQueryBuilder` is a more obvious name, and the `IndexedDbQuery` module should be called `IndexedDbQueryBuilder`.

And a `IndexedDbQueryBuilder` returns a `IndexedDbQuery`.

packages/platform-browser/test/IndexedDbDatabase.test.ts
```diff
@@ -0,0 +1,229 @@
...
+        id: Schema.Number,
+        title: Schema.String,
+        completed: Schema.Boolean
+      }),
+      { keyPath: "id", indexes: { titleIndex: "title" } }
+    )
+
+    const Db = IndexedDbVersion.make(Table)
+
+    class Migration extends IndexedDbMigration.make(Db, (api) =>
```
tim-smart, line 39: It feels like `IndexedDbMigration` should actually be `IndexedDbDatabase`, and the `IndexedDbDatabase` tag can become more of an implementation detail.

Then `IndexedDbDatabase` could have a `getQueryBuilder` api for getting a query builder for the latest version.

packages/platform-browser/src/IndexedDbQuery.ts
SandroMaglione: I moved the internal types inside the `namespace`

packages/platform-browser/src/IndexedDbVersion.ts
SandroMaglione: Renamed

packages/platform-browser/test/IndexedDbDatabase.test.ts
SandroMaglione: It works, removed `IndexedDbMigration` and unified all inside `IndexedDbDatabase`. A migration is created from `IndexedDbDatabase` and it has both a `getQueryBuilder` and `layer` api

packages/platform-browser/src/IndexedDbDatabase.ts
```diff
@@ -0,0 +1,191 @@
...
+ * @since 1.0.0
+ * @category type ids
+ */
+export type ErrorTypeId = typeof ErrorTypeId
+
+/**
+ * @since 1.0.0
+ * @category errors
+ */
+export const IndexedDbDatabaseError = internal.IndexedDbDatabaseError
```
tim-smart, line 41: This needs an explicit type.

packages/platform-browser/src/IndexedDbDatabase.ts
```diff
@@ -0,0 +1,191 @@
...
+  export interface Transaction<
+    Source extends IndexedDbVersion.IndexedDbVersion.AnyWithProps = never
+  > extends Pipeable, Omit<IndexedDbQueryBuilder.IndexedDbQueryBuilder<Source>, "transaction"> {
+    readonly transaction: globalThis.IDBTransaction
+
+    readonly createObjectStore: <
+      A extends IndexedDbTable.IndexedDbTable.TableName<
+        IndexedDbVersion.IndexedDbVersion.Tables<Source>
+      >
+    >(table: A) => Effect.Effect<globalThis.IDBObjectStore, internal.IndexedDbDatabaseError>
```
tim-smart, line 61: internal types are being used here

packages/platform-browser/src/IndexedDbQueryBuilder.ts
```diff
@@ -0,0 +1,570 @@
...
+ * @since 1.0.0
+ * @category type ids
+ */
+export type TypeId = typeof TypeId
+
+/**
+ * @since 1.0.0
+ * @category type ids
+ */
+export const ErrorTypeId: unique symbol = internal.ErrorTypeId
```
tim-smart, line 28: You can't use internal types here

packages/platform-browser/src/IndexedDbQueryBuilder.ts
```diff
@@ -0,0 +1,570 @@
...
+ * @since 1.0.0
+ * @category type ids
+ */
+export type ErrorTypeId = typeof ErrorTypeId
+
+/**
+ * @since 1.0.0
+ * @category errors
+ */
+export type IndexedDbQueryError = typeof internal.IndexedDbQueryError
```
tim-smart, line 40: You can't use internal types here

packages/platform-browser/src/IndexedDbQueryBuilder.ts
```diff
@@ -0,0 +1,570 @@
...
+  }
+
+  /**
+   * @since 1.0.0
+   * @category models
+   */
+  export interface ClearAll<
+    Source extends IndexedDbVersion.IndexedDbVersion.AnyWithProps = never
+  > {
+    [Symbol.iterator](): Effect.EffectGenerator<Effect.Effect<void>>
```
tim-smart, line 289: What is the Symbol.iterator for?

packages/platform-browser/src/IndexedDbDatabase.ts
SandroMaglione: Updated the error definition for internal/exported value and type

packages/platform-browser/src/IndexedDbQueryBuilder.ts
SandroMaglione: Updated error definition

packages/platform-browser/src/IndexedDbQueryBuilder.ts
SandroMaglione: I simplified `clearAll` and `clear` to just return an `Effect` instead of an intermediate `interface`

SandroMaglione:
@tim-smart next steps for this?

tim-smart:
When I get some time I'll do another review

tim-smart:
Did a bit more work on this to try clean things up. Still needs work:

- support for schema classes with auto-incrementing keys, maybe add `IndexedDb.AutoIncrement` schema
- insert / update needs to use Schema.encode
- needs tests for schemas with context

I'm going to be working on some other things for a while, so feel free to give these a go

nikelborm (commented):
Maybe it worth adding some open-telemetry spans information, as all of the SQL-family packages have?

SandroMaglione:
@tim-smart added encode for modify operations and schema context test.

Not sure what you mean with `support for schema classes with auto-incrementing keys`. As far as I understood, IndexedDb `autoIncrement` is a boolean, and it can only be applied to one field (the primary key). That's why the current API has `autoIncrement` as an option. Were you thinking about something different?

nikelborm:
By the way, @SandroMaglione, I  accidentally found in the effect codebase this:
https://github.com/Effect-TS/effect/blob/main/packages/experimental/src/EventJournal.ts#L358

And I think it's worth your attention, maybe you could insert your implementation here

tim-smart:
> By the way, @SandroMaglione, I accidentally found in the effect codebase this: https://github.com/Effect-TS/effect/blob/main/packages/experimental/src/EventJournal.ts#L358
> 
> And I think it's worth your attention, maybe you could insert your implementation here

I would prefer to not add the overhead of a query builder to the EventJournal queries, same with an implementation of KeyValueStore.

SandroMaglione:
@tim-smart I still don’t get what needs to be updated.

> .insert queries should be updated to use the constructor type of the table schema

Is that not already the case?
https://github.com/Effect-TS/effect/blob/4787f125b4efd8bde21eec31e17107b40ecf31b3/packages/platform-browser/src/IndexedDbQueryBuilder.ts#L182-L184

> make the autoincrement key optional even if it is marked as required on the table

What do you mean here? `autoIncrement` is already optional when creating a table with `make`, and specifying the value of the key is also optional when using `.insert`. Also, I am not sure what the TODO below means (if relevant).
https://github.com/Effect-TS/effect/blob/4787f125b4efd8bde21eec31e17107b40ecf31b3/packages/platform-browser/test/IndexedDbQueryBuilder.test.ts#L63-L70

> It should also work with Schema.Class

Is not the case already? What's the difference in implementation when using `Schema.Class`?
https://github.com/Effect-TS/effect/blob/4787f125b4efd8bde21eec31e17107b40ecf31b3/packages/platform-browser/test/IndexedDbQueryBuilder.test.ts#L38-L42

tim-smart:
Ideally the auto-increment key would not be optional, so it would require removing the autoincrement key from the schema and types for inserts.

SandroMaglione:
@tim-smart would add this changes to `effect-smol` help to move this forward?

fubhy:

PR #4740, opened 2025-04-15
add Effect native OtlpMetrics & OtlpLogger

tim-smart:


packages/opentelemetry/src/OtlpLogger.ts
```diff
@@ -0,0 +1,243 @@
...
+    default:
+      return ESeverityNumber.SEVERITY_NUMBER_UNSPECIFIED
+  }
+}
+
+/**
+ * Numerical value of the severity, normalized to values described in Log Data Model.
+ */
+const enum ESeverityNumber {
+  /** Unspecified. Do NOT use as default */
```
pmanole-volvocars, line 217: Maybe add an exclamation mark at the end.

PR #4745, opened 2025-04-15
chore: upgrade TSTyche

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR upgrades TSTyche to version 4. (It is still in beta, because I plan to implement few more features and that will break programatic usage. Nothing breaking is planned for CLI usage.)

The highlight of TSTyche 4 are the ability matchers: `.toBeCallableWith()`, `.toBeConstructableWith()`, etc. Detailed release notes can be found here: https://tstyche.org/releases/tstyche-4

In this PR many `// @ts-expect-error` are replaced with `.not.toBeCallableWith()`. There are much more `// @ts-expect-error` still left around. I will handle them later. The idea was to show you what is possible and to ask for feedback. Also these tests were a great playground to try out new features.

### How it Works?

Implementation of the ability matchers is somewhat similar to [`ts-blank-space`](https://github.com/bloomberg/ts-blank-space). After retrieving diagnostics for a file, TSTyche creates a virtual copy of it with erased assertions and asks TypeScript for diagnostics again.

An assertion like this:

```ts
expect(Duration.decode).type.toBeCallableWith("10 unknown")
```

Becomes:

```ts
;     (Duration.decode)                      ("10 unknown")
```

Any found error are reported (or silenced if `.not` is included):

Screenshot 2025-04-15 at 14 52 27


packages/effect/dtslint/Array.tst.ts
```diff
@@ -355,10 +351,8 @@ describe("Array", () => {
       })
     )).type.toBe<Array<string | number>>()
 
-    // @ts-expect-error: wrong `_item` type
-    Array.filter(numbersOrStrings, (_item: string) => true)
-    // @ts-expect-error: wrong `_item` type
-    pipe(numbersOrStrings, Array.filter((_item: string) => true))
+    expect(Array.filter).type.not.toBeCallableWith(numbersOrStrings, (_item: string) => true)
+    expect(Array.filter<string | number, string | number>).type.not.toBeCallableWith((_item: string) => true)
```
mrazauskas, line 355: Type arguments must be passed to `Array.filter` to reproduce the same behaviour. Perhaps there are better solutions?

The following does not work, because the error originates from `Array.filter`. It is reported as a problem.

```ts
expect(pipe).type.not.toBeCallableWith(numbersOrStrings, Array.filter((_item: string) => true))
```

The assertion can be moved to the argument list of `pipe()`. This will error as well, because `pipe()` cannot that `expect()` as an argument:

```ts
pipe(numbersOrStrings, expect(Array.filter).type.not.toBeCallableWith((_item: string) => true))
```

packages/effect/dtslint/Schedule.tst.ts
```diff
@@ -24,12 +24,8 @@ describe("Schedule", () => {
       (x: string | number) => Console.log(x)
     )).type.toBe<Schedule.Schedule<number, unknown, never>>()
 
-    Schedule.once.pipe(
-      Schedule.as<number | string>(1),
-      Schedule.tapOutput(
-        // @ts-expect-error: Type 'number' is not assignable to type 'string'
-        (s: string) => Console.log(s.trim())
-      )
+    expect(Schedule.tapOutput<void, never, string | number>).type.not.toBeCallableWith(
```
mrazauskas, line 27: Same as above.

tstyche.config.json
```diff
@@ -1,7 +1,5 @@
 {
   "$schema": "https://tstyche.org/schemas/config.json",
-  "checkSourceFiles": true,
-  "rejectAnyType": true,
```
mrazauskas, line 4: These now are enabled by default.

packages/effect/dtslint/Array.tst.ts
mrazauskas: By the way, I think `// @ts-` comments must be removed in the virtual file. They will hide errors in erased expressions and matchers will not work.

Cases like this one could use the ignore comment. And then the first solution above works as expected.

mrazauskas: A helper function could be an elegant solution. Instead of making noise here, I opened https://github.com/tstyche/tstyche/issues/458

gcanti:
This is awesome, thanks for working on this @mrazauskas

mrazauskas:
@gcanti Could you take a look, please.

There are only 31 instances of `// @ts-expect` left. These are either not related with callability or too complex to be covered now. This change is large enough. Perhaps it is worth to land it and to iterate.

If `beta` is an issue, I am ready to publish `rc`.

gcanti (approved):

gcanti:
> This change is large enough. Perhaps it is worth to land it and to iterate

agreed

> If beta is an issue, I am ready to publish rc.

no problem, I'm fine with a beta, there's no rush, we are happy to help you out with the next release of `tstyche`

mrazauskas:
Wonderful! Thanks so much.

PR #4764, opened 2025-04-19
Allow rpc middleware to wrap execution

mattiamanzati:
## Type

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

With this change, a new variant of an RpcMiddleware will be added that allows to wrap execution of requests.
This will allow things like providing scoped resources, accessing the exit result per-request alongside request type metadata, etc...

### DIFF
--- a/packages/rpc/src/RpcMiddleware.ts
+++ b/packages/rpc/src/RpcMiddleware.ts
@@ -39,14 +39,26 @@ export interface RpcMiddleware<Provides, E> {
  * @since 1.0.0
  * @category models
  */
-export interface RpcMiddlewareWrap {
+export interface RpcMiddlewareWrap<Provides, E> {
   (options: {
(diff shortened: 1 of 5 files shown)


packages/rpc/src/RpcMiddleware.ts
```diff
@@ -35,6 +35,19 @@ export interface RpcMiddleware<Provides, E> {
   }): Effect.Effect<Provides, E>
 }
 
+/**
+ * @since 1.0.0
+ * @category models
+ */
+export interface RpcMiddlewareWrap {
```
mattiamanzati, line 42: @tim-smart Opted out for a completely different interface, would you prefer a typelevel parameter into RpcMiddleware instead and perform conditional arguments over there instead?

tim-smart (approved):

PR #4778, opened 2025-04-23
kill path aliases, use src exports

fubhy:
The plan here is to:

**For development:**
- [x] Use `exports` to point to source modules for development
- [x] Eliminate tsconfig path aliases
- [x] Eliminate vitest path aliases
- [x] Move effect test utils under `@effect/vitest` package

**For publishing (update `@effect/build-utils`):**
- [x] Generate `exports` for publishing based on these instead of generating them with `glob`
- [x] Generate `typesVersions` from those exports too
- [x] Generate the `main`, `module` and `types` package.json fields based on the presence of a `.` export

This should simplify the package maintenance because we now just have to maintain the project references and not worry about path aliasing anymore.

It also makes package exports publishing decisions much more explicit (we specify them manually) and gives us more flexibility there.


IMax153 (approved):
I'm very much in favor of this, but would like @tim-smart and @mikearnaldi to weigh in as well.

tim-smart:
It feels like forgetting an export will be a common error. Maybe we generate most of them like we do now?

mattiamanzati (approved):

fubhy:
> It feels like forgetting an export will be a common error. Maybe we generate most of them like we do now?

@tim-smart does the recent change address your concern? We now support glob patterns there following the node module resolition algorithm.

tim-smart:
Do we still need seperate tsconfig.build.json from tsconfig.src.json?

It seems like they are mostly the same.

fubhy:
> Do we still need seperate tsconfig.build.json from tsconfig.src.json?
> 
> It seems like they are mostly the same.

Very good point. We don't need that anymore.

tim-smart:
image

Getting a build error locally

fubhy:
> Getting a build error locally
Yeah, you'll likely have to flush your node_modules and tsbuildinfo files throughout.

tim-smart (approved):
You have to remove the node_modules from the sub-folders too, for the build to work.

PR #4784, opened 2025-04-23
Possibility to handle remaining errors in `catchTags`

akoenig:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Sometimes, it’s necessary to recover from specific errors in `catchTags`, while still handling all other possible errors in a general way.

This PR is based on [this Discord thread](https://discord.com/channels/795981131316985866/1364558789801672734) and introduces a possible function as a second parameter to `catchTags`, allowing you to handle all the other errors that are not explicitly specified.

An example:

```ts
import { Data, Effect } from "effect"

class A extends Data.TaggedError("A") {}
class B extends Data.TaggedError("B") {}
class C extends Data.TaggedError("C") {}
class D extends Data.TaggedError("D") {}

  const types = [A, B, C, D];
  const Exception = types[Math.floor(Math.random() * types.length)];

  return yield* new Exception
}).pipe(
  Effect.catchTags(
    {
      A: (a) => Effect.succeed(a._tag),
      B: (b) => Effect.succeed(b._tag),
    },
    remaining => Effect.succeed(`remaining: ${remaining._tag}`) // C | D
  )
);
```

### DIFF
--- a/.changeset/soft-bears-play.md
+++ b/.changeset/soft-bears-play.md
@@ -1,0 +1,31 @@
+---
+"effect": minor
+---
+
+Possibility to recover from unhandled errors in `catchTags`.
+
+This addition introduces the possibility to define a function as a second parameter that gets the error type which was unhandled in the map before.
+
+```ts
+import { Data, Effect } from "effect"
+
+class A extends Data.TaggedError("A") {}
+class B extends Data.TaggedError("B") {}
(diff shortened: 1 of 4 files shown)


tim-smart:

PR #4787, opened 2025-04-24
Update eslint-plugin and enable no-import-from-barrel rule

mattiamanzati:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR updates the eslint plugin (so that now also natively supports eslint9's style of config).
It also enables, only for sources folders, the rule that will prevent imports like #4779 
Enabling this rule revealed a similar issue inside the @effect/sql package.
To keep the rule fast, it just does static analysis (to avoid enable of type info).
So packageNames has to be manually provided in the rule config.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,3 +1,4 @@
+import * as effectEslint from "@effect/eslint-plugin"
 import { fixupPluginRules } from "@eslint/compat"
 import { FlatCompat } from "@eslint/eslintrc"
 import js from "@eslint/js"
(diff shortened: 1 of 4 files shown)


tim-smart (approved):

eslint.config.mjs
```diff
@@ -135,5 +136,20 @@ export default [
...
+    files: ["packages/*/src/**/*"],
+    rules: {
+      "@effect/no-import-from-barrel-package": [
+        "error",
+        {
+          packageNames: [
+            "effect",
+            "@effect/platform",
+            "@effect/sql"
+          ]
```
fubhy, line 150: Why is this only enabled for these packages?

PR #4788, opened 2025-04-24
`Brand.unbranded` getter

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

TS can't narrow the type of a tuple element when accessed via a branded index — which results in a wider result type.
It is possible to cast using `Brand.Brand.Unbranded`, but it's inconvenient — so I propose adding a corresponding utility function.
image

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/calm-zebras-relate.md
+++ b/.changeset/calm-zebras-relate.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+`Brand.unbranded` getter has been added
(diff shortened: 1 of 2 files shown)


KhraksMamtsov:
@tim-smart, would you mind taking a look? It seems straightforward.)

PR #4794, opened 2025-04-27
fix(Metric): summary quantile value when overshoots

IGassmann:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This fixes an issue within summary metric.

When calculating the value for a quantile, if the first chunk of elements of the same value (`10` x 6) was great than the target rank, it would return `Option.none()` since `current_1 ` hasn't yet been set to some value.

### DIFF
--- a/.changeset/itchy-cows-march.md
+++ b/.changeset/itchy-cows-march.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix summary metric’s quantile value calculation
--- a/packages/effect/src/internal/metric/hook.ts
+++ b/packages/effect/src/internal/metric/hook.ts
@@ -391,7 +391,8 @@ const resolveQuantile = (
     }
     // Split into two chunks - the first chunk contains all elements of the same
     // value as the chunk head
-    const sameHead = Arr.span(rest_1, (n) => n <= rest_1[0])
+    const headValue = Arr.headNonEmpty(rest_1) // Get head value since rest_1 is non-empty
+    const sameHead = Arr.span(rest_1, (n) => n === headValue)
     // How many elements do we want to accept for this quantile
     const desired = quantile_1 * sampleCount_1
     // The error margin
@@ -417,12 +418,14 @@ const resolveQuantile = (
       rest_1 = rest_2
       continue
     }
-    // If we have too many elements, select the previous value and hand back the
-    // the rest as leftover
+    // If consuming this chunk leads to too many elements (rank is too high)
     if (candConsumed > desired + allowedError) {
+      const valueToReturn = Option.isNone(current_1)
+        ? Option.some(headValue)
+        : current_1
       return {
         quantile: quantile_1,
-        value: current_1,
+        value: valueToReturn,
         consumed: consumed_1,
         rest: rest_1
       }
--- a/packages/effect/test/Metric.test.ts
+++ b/packages/effect/test/Metric.test.ts
@@ -577,6 +577,30 @@ describe("Metric", () => {
         strictEqual(result2.count, 1)
         strictEqual(result3.count, 1)
       }))
+    it.effect("should return correct quantile when first chunk overshoots", () =>
+      Effect.gen(function*() {
+        const name = nextName()
+        // Samples: [10 (x6), 20, 30, 40, 50] (10 samples)
+        // Target rank for 0.5 quantile = 0.5 * 10 = 5
+        // Allowed error = (0.01 / 2) * 5 = 0.025. Range [4.975, 5.025]
+        // First chunk: 6 * 10. candConsumed = 6. 6 > 5.025
+        const samples = [10, 10, 10, 10, 10, 10, 20, 30, 40, 50]
+        const summary = Metric.summary({
+          name,
+          maxAge: Duration.minutes(1),
+          maxSize: 15,
+          error: 0.01,
+          quantiles: [0.5]
+        })
+
+        yield* Effect.forEach(samples, (value) => Metric.update(summary, value), { discard: true })
+
+        const result = yield* Metric.value(summary)
+
+        const medianQuantileValue = result.quantiles[0][1]
+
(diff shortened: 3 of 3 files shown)


IMax153:
@IGassmann - this PR looks good to me, thank you for your attention to detail!

As a side note, in the work I've done on `Metric` for Effect 4.0, I've _substantially_ simplified calculation of snapshots for summary metrics, and I've also removed the error margin option given that with a rank-based approach for quantile calculation it doesn't make much sense anyway.

If you're curious or would like to provide feedback, take a look [here](https://github.com/Effect-TS/effect-smol/blob/683306b068392435f89d14c39563107904d8fdf5/src/Metric.ts#L521).

IGassmann:
> @IGassmann - this PR looks good to me, thank you for your attention to detail!
> 
> As a side note, in the work I've done on `Metric` for Effect 4.0, I've _substantially_ simplified calculation of snapshots for summary metrics, and I've also removed the error margin option given that with a rank-based approach for quantile calculation it doesn't make much sense anyway.
> 
> If you're curious or would like to provide feedback, take a look [here](https://github.com/Effect-TS/effect-smol/blob/683306b068392435f89d14c39563107904d8fdf5/src/Metric.ts#L521).

Hey @IMax153! :) Thanks for merging in my patch!

Nice to see that you've simplified the calculation logic!

Glad to share some feedback:

1. There seems to be a bug with the condition of the [while loop that constructs the `builder` array](https://github.com/Effect-TS/effect-smol/blob/683306b068392435f89d14c39563107904d8fdf5/src/Metric.ts#L486C3-L557C4). The loop iterates from `⁠i = 0` up to ⁠`this.#maxSize - 2`. It misses checking the very last element in the ⁠observations array (at index `⁠this.#maxSize - 1)`. I guess it should be `while (i < this.#maxSize)`
2. If we have a large `this.#maxSize` value—say 100K—but only a few observations have been made, for example, 5, and we snapshot the summary, the [while loop that constructs the `builder` array](https://github.com/Effect-TS/effect-smol/blob/683306b068392435f89d14c39563107904d8fdf5/src/Metric.ts#L486C3-L557C4) will iterate for 100K times unnecessarily even though we only need it to iterate 5 times.
3. The very first observation is written to index `1` instead of `0` in the `observations` array because `⁠head` is incremented before calculating the ⁠`target` value. Shouldn't the incrementation happen after writing the observation?
4. The ⁠observe method updates `⁠min`, `⁠max`, ⁠`sum`, and ⁠`count` based on all observations ever seen since the metric was created. However, the ⁠`snapshot` function calculates ⁠`quantiles` based only on the observations currently in the buffer and within the ⁠`maxAge`. This means the ⁠`quantiles` represent a time/size-limited window, while ⁠`count`, `⁠sum`, `⁠min`, `⁠max` represent the entire history. This can be confusing for users. I would either modify ⁠`snapshot` to also calculate ⁠`min`, `⁠max`, ⁠`sum`, and `⁠count` only from the filtered ⁠`samples` array it uses for `quantiles` so that all returned values pertain to the same window, or keep it as it is, but clearly document that ⁠`quantiles` are windowed estimates, while ⁠`count`, `⁠sum`, ⁠`min`, ⁠`max` are global aggregates since the start. 
5. From what I understand, this quantile calculation algorithm is approximate (e.g. rare but significant values could enter the observations buffer and be evicted before the next ⁠snapshot, potentially skewing results temporarily or being missed entirely). Considering this, it might be worth documenting this reality. Here's some docs (probably too verbose), I've generated for some inspiration:

```md
**Accuracy and Limitations**

This metric estimates quantiles based on a **sliding window** or **fixed-size sample** of recent observations, determined by the `maxSize` and `maxAge` parameters. It provides an *approximation* of the quantiles for the data observed within that specific window.

**Important Considerations:**

*   **No Guaranteed Error Bounds:** Unlike mathematically rigorous streaming quantile algorithms (e.g., Greenwald-Khanna, t-digest), this implementation does **not** provide formal guarantees on the maximum error (i.e., no ε-approximation). The actual accuracy depends entirely on how representative the sampled window is.
*   **Dependence on Parameters:**
    *   `maxSize`: Limits the number of data points stored. A small `maxSize` might not capture the full distribution, especially if data varies significantly or has important outliers.
    *   `maxAge`: Filters out older data points. If the relevant time frame for your quantile calculation is longer than `maxAge`, the result will only reflect the more recent data.
*   **Window Representativeness:** The core assumption is that the `maxSize` most recent points (within `maxAge`) are sufficient to approximate the desired quantiles for that window. This may not hold true if:
    *   The window size (`maxSize` or the number of points within `maxAge`) is too small relative to the data variability.
    *   Significant values or distribution shifts occur but the corresponding data points are quickly overwritten (due to `maxSize`) or expire (due to `maxAge`) before a quantile calculation (`snapshot`).
*   **Use Case:** This implementation is suitable when you need a computationally simpler, memory-bounded quantile estimate focused on recent data, and you can tolerate potential inaccuracies due to the sampling nature.

**Guidance:** Choose `maxSize` and `maxAge` carefully based on your specific data characteristics, expected data volume, the desired recency of the estimate, and your tolerance for approximation errors. Monitor results and adjust parameters if needed. If guaranteed error bounds are required, consider using a different quantile estimation algorithm.
```

IMax153:
@IGassmann - thank you for your detailed review. Here are some follow ups to your comments:

1. Yes, you're correct. I've fixed this in [this PR]()
2. You're also correct about this, but how would one track when to stop adding to the observed values? We don't know the number of samples that will end up in the summary when a `snapshot` is taken, so all we really have to go on is `maxSize`. 
3. Also fixed in the above PR.
4. I don't think this is correct. The idea of Effect's metrics is to mirror to some degree the logic Prometheus uses. For summary metrics, Prometheus exposes these values as essentially counters that track the cumulative value (see the [docs](https://prometheus.io/docs/concepts/metric_types/#summary)).
5. It is definitely an approximation. I'd like to improve the calculation in the future to be honest, but that's something for the future.

PR #4796, opened 2025-04-27
allow catching multiple different tags in Effect.catchTag

tim-smart:


tim-smart:
In my IDE I'm seeing a typescript inference issue, where I have to update the error arg name for the type to update.

PR #4812, opened 2025-05-01
allow safe `mapEffect` in type signature

fubhy:


packages/cli/src/Args.ts
```diff
@@ -334,8 +334,8 @@ export const map: {
  * @category mapping
  */
 export const mapEffect: {
-  <A, B>(f: (a: A) => Effect<B, HelpDoc, FileSystem | Path | Terminal>): (self: Args<A>) => Args<B>
-  <A, B>(self: Args<A>, f: (a: A) => Effect<B, HelpDoc, FileSystem | Path | Terminal>): Args<B>
+  <A, B>(f: (a: A) => Effect<B, HelpDoc | never, FileSystem | Path | Terminal>): (self: Args<A>) => Args<B>
```
tim-smart, line 337: A union with never doesn't change the type. An effect with never in the error channel should already be assignable to this.

fubhy: In my defense, it's early morning over here.

PR #4814, opened 2025-05-01
improve Effect.filter* types to exclude candidates in fallback functions

tim-smart:


.changeset/hot-bats-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
+---
+
+improve Effect.filter\* types to exludes candidates in fallback functions
```
tim-smart, line 5:
```suggestion
improve Effect.filter\* types to exclude candidates in fallback functions
```

PR #4827, opened 2025-05-06
Fix `StreamAdapter` error forwarding

mlegenhausen:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

If a `Stream` fails the error is not correctly handled by `StreamAdapter` which results in a stream that never emits an error. See 9c15bad28766e77c7a2fd34fc2fab3ce3ba292c2.

This fix now correctly calls the `destroy` function on the `Readable` which initializes an appropriate `callback` which when called emits the `error` event on the `Readable`.

### DIFF
--- a/.changeset/plenty-gorillas-nail.md
+++ b/.changeset/plenty-gorillas-nail.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform-node-shared": patch
+---
+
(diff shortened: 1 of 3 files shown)


tim-smart (approved):
👍🏼

PR #4828, opened 2025-05-06
Add `Schedule.CurrentIterationMetadata` Reference 

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
`Schedule.CurrentIterationMetadata` has been added

## Related
https://discord.com/channels/795981131316985866/795983589644304396/threads/1369205653343703070

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Schedule.ts
+++ b/packages/effect/src/Schedule.ts
@@ -141,7 +141,7 @@ export declare namespace Schedule {
  */
 export interface ScheduleDriver<out Out, in In = unknown, out R = never> extends Schedule.DriverVariance<Out, In, R> {
   readonly state: Effect.Effect<unknown>
-  readonly iterationMeta: Ref.Ref<Chunk.Chunk<IterationMetadata>>
+  readonly iterationMeta: Ref.Ref<IterationMetadata>
   readonly last: Effect.Effect<Out, Cause.NoSuchElementException>
   readonly reset: Effect.Effect<void>
   next(input: In): Effect.Effect<Out, Option.Option<never>, R>
@@ -2214,5 +2214,5 @@ export interface IterationMetadata {
  */
 export const CurrentIterationMetadata: Context.Reference<
   CurrentIterationMetadata,
-  Chunk.Chunk<IterationMetadata>
+  IterationMetadata
 > = internal.CurrentIterationMetadata
--- a/packages/effect/src/Schedule.ts
+++ b/packages/effect/src/Schedule.ts
@@ -141,7 +141,7 @@ export declare namespace Schedule {
  */
 export interface ScheduleDriver<out Out, in In = unknown, out R = never> extends Schedule.DriverVariance<Out, In, R> {
   readonly state: Effect.Effect<unknown>
-  readonly lastIterationInfo: Ref.Ref<Option.Option<IterationInfo>>
+  readonly iterationMeta: Ref.Ref<Chunk.Chunk<IterationMetadata>>
   readonly last: Effect.Effect<Out, Cause.NoSuchElementException>
   readonly reset: Effect.Effect<void>
   next(input: In): Effect.Effect<Out, Option.Option<never>, R>
@@ -2191,7 +2191,7 @@ export const zipWith: {
  * @since 3.15.0
  * @category models
  */
-export interface LastIterationInfo {
+export interface CurrentIterationMetadata {
   readonly _: unique symbol
 }
 
@@ -2199,16 +2199,20 @@ export interface LastIterationInfo {
  * @since 3.15.0
  * @category models
  */
-export interface IterationInfo {
-  readonly duration: Duration.Duration
-  readonly iteration: number
+export interface IterationMetadata {
+  readonly input: unknown
+  readonly recurrence: number
+  readonly start: number
+  readonly now: number
+  readonly elapsed: Duration.Duration
+  readonly elapsedSincePrevious: Duration.Duration
 }
 
 /**
  * @since 3.15.0
  * @category models
  */
-export const LastIterationInfo: Context.Reference<
-  LastIterationInfo,
-  Option.Option<IterationInfo>
+export const CurrentIterationMetadata: Context.Reference<
+  CurrentIterationMetadata,
+  Array<IterationMetadata>
 > = internal.LastIterationInfo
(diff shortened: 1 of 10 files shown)


packages/effect/src/internal/schedule.ts
```diff
@@ -45,6 +45,12 @@ export const ScheduleDriverTypeId: Schedule.ScheduleDriverTypeId = Symbol.for(
   ScheduleDriverSymbolKey
 ) as Schedule.ScheduleDriverTypeId
 
+/** @internal */
+export const LastIterationInfo = Context.Reference<Schedule.LastIterationInfo>()(
+  "effect/Schedule/LastIterationInfo",
+  { defaultValue: () => Option.none<Schedule.IterationInfo>() }
```
tim-smart, line 51: The default could be Duration.zero / iteration: 0

KhraksMamtsov:
@tim-smart 
the only thing I can't figure out is how to pass this reference into the `Stream.scheduleWith` function.
Other than that, I think I'm done.

packages/effect/src/Schedule.ts
```diff
@@ -140,6 +141,7 @@ export declare namespace Schedule {
  */
 export interface ScheduleDriver<out Out, in In = unknown, out R = never> extends Schedule.DriverVariance<Out, In, R> {
   readonly state: Effect.Effect<unknown>
+  readonly iterationMeta: Ref.Ref<Chunk.Chunk<IterationMetadata>>
```
tim-smart, line 144: I don't think it should be a Chunk / Array, as Schedule.forever would then have a memory leak.

mikearnaldi: good point

KhraksMamtsov: Great remark – I've redone it – now it's only the meta of the latest iteration.

KhraksMamtsov:
> the only thing I can't figure out is how to pass this reference into the `Stream.scheduleWith` function.

I intentionally left one non-working test so we wouldn't forget about it.

tim-smart:
> > the only thing I can't figure out is how to pass this reference into the `Stream.scheduleWith` function.
> 
> I intentionally left one non-working test so we wouldn't forget about it.

I don't think it make sense here, as `Stream.scheduleWith` works on a per-element level rather than wrapping the entire stream.

KhraksMamtsov:
@tim-smart  
I removed this test

packages/effect/src/internal/schedule.ts
```diff
@@ -1886,11 +1938,17 @@ const repeatOrElseEffectLoop = <A, E, R, R1, B, C, E2, R2>(
...
-        onSuccess: (value) => repeatOrElseEffectLoop(self, driver, orElse, value)
+    onSuccess: (b) => {
+      const provideLastIterationInfo = effect.provideServiceEffect(
+        CurrentIterationMetadata,
+        ref.get(driver.iterationMeta)
+      )
+      const selfWithLastIterationInfo = provideLastIterationInfo(self)
+      return core.matchEffect(selfWithLastIterationInfo, {
+        onFailure: (error) => provideLastIterationInfo(orElse(error, Option.some(b))),
+        onSuccess: (value) => repeatOrElseEffectLoop(selfWithLastIterationInfo, driver, orElse, value)
```
tim-smart, line 1949:
```suggestion
        onSuccess: (value) => repeatOrElseEffectLoop(self, driver, orElse, value)
```

Otherwise the effect definition could get infinitely large.

packages/effect/src/internal/schedule.ts
```diff
@@ -1981,16 +2039,22 @@ const retryOrElse_EffectLoop = <A, E, R, R1, A1, A2, E2, R2>(
...
+      return core.matchEffect(driver.next(e), {
         onFailure: () =>
           pipe(
             driver.last,
             core.orDie,
-            core.flatMap((out) => orElse(e, out))
+            core.flatMap((out) => provideLastIterationInfo(orElse(e, out)))
           ),
-        onSuccess: () => retryOrElse_EffectLoop(self, driver, orElse)
+        onSuccess: () => retryOrElse_EffectLoop(provideLastIterationInfo(self), driver, orElse)
```
tim-smart, line 2055:
```suggestion
        onSuccess: () => retryOrElse_EffectLoop(self, driver, orElse)
```

packages/effect/src/internal/schedule.ts
```diff
@@ -2030,11 +2094,19 @@ const scheduleFrom_EffectLoop = <In, E, R, R2, Out>(
...
+): Effect.Effect<Out, E, R | R2> => {
+  const provideLastIterationInfo = effect.provideServiceEffect(
+    CurrentIterationMetadata,
+    ref.get(driver.iterationMeta)
+  )
+  return core.matchEffect(driver.next(initial), {
+    onFailure: () => core.orDie(provideLastIterationInfo(driver.last)),
+    onSuccess: () => {
+      const selfWithLastOptions = provideLastIterationInfo(self)
+      return core.flatMap(selfWithLastOptions, (a) => scheduleFrom_EffectLoop(selfWithLastOptions, a, driver))
```
tim-smart, line 2106:
```suggestion
      return core.flatMap(selfWithLastOptions, (a) => scheduleFrom_EffectLoop(self, a, driver))
```

packages/effect/src/internal/schedule.ts
```diff
@@ -2030,11 +2094,19 @@ const scheduleFrom_EffectLoop = <In, E, R, R2, Out>(
...
-  core.matchEffect(driver.next(initial), {
-    onFailure: () => core.orDie(driver.last),
-    onSuccess: () => core.flatMap(self, (a) => scheduleFrom_EffectLoop(self, a, driver))
+): Effect.Effect<Out, E, R | R2> => {
+  const provideLastIterationInfo = effect.provideServiceEffect(
+    CurrentIterationMetadata,
+    ref.get(driver.iterationMeta)
+  )
+  return core.matchEffect(driver.next(initial), {
+    onFailure: () => core.orDie(provideLastIterationInfo(driver.last)),
```
tim-smart, line 2103: Don't think you need to provide it here?

tim-smart (approved):

cmbntr:
```import { Duration, Effect, Option } from "effect";

Effect.gen(function*() {
  const [t] = yield* Effect.timed(Effect.void.pipe(Effect.repeat({ times: 1_000_000 })));
  yield* Effect.logInfo(Duration.formatIso(t).pipe(Option.getOrThrow));
}).pipe(Effect.runPromise);

PR #4836, opened 2025-05-07
Fix input type for Match.valueTags and Match.typeTags

thewilkybarkid:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Ensures that the input has a tag.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/large-llamas-live.md
+++ b/.changeset/large-llamas-live.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix input type for Match.valueTags and Match.typeTags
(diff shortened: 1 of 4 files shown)


tim-smart:
Ideally these apis will work with types that also don't contain tags, such as unions with a mix of types.

fubhy:
Thanks for the PR. A maintainer noted that these APIs should ideally still support unions that include untagged members. Could you confirm whether you want to revise the approach to preserve mixed-union support, or whether you think the stricter tag constraint is the intended behavior?

thewilkybarkid:
That sounds like a slightly different API, with a separate `orElse` option.

PR #4844, opened 2025-05-09
Http API streaming payloads and streaming responses

leonitousconforti:
Thoughts on adding streaming payloads and streaming responses to http api?

## Type

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

implementing payload streams and response streams for http api

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/platform/src/HttpApiEndpoint.ts
+++ b/packages/platform/src/HttpApiEndpoint.ts
@@ -8,6 +8,7 @@ import * as Option from "effect/Option"
 import { type Pipeable, pipeArguments } from "effect/Pipeable"
 import * as Predicate from "effect/Predicate"
 import * as Schema from "effect/Schema"
+import type { Stream } from "effect/Stream"
 import type * as Types from "effect/Types"
 import type * as HttpApiMiddleware from "./HttpApiMiddleware.js"
 import * as HttpApiSchema from "./HttpApiSchema.js"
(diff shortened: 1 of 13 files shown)


packages/platform/src/HttpApiEndpoint.ts
```diff
@@ -728,10 +759,18 @@ const Proto = {
...
+      readonly contentLength?: number | undefined
+      readonly etag?: string | undefined
+    }
+  ) {
     return makeProto({
       ...this,
-      payloadSchema: Option.some(schema)
+      payloadSchema: Option.some(payload)
     })
   },
```
leonitousconforti, line 775: Thoughts on adding a `stream` option to setPayload? Is this something you think could make it in with more refinement?

danimbrogno:
This would be great. Just picked up effect and am trying to write a backend API that makes calls to an LLM. 

It would be great to be able to forward along streaming responses from the LLM to my API client.

leonitousconforti:
I don't think this is ready yet, there are things that I would like to change, such as adding an NDJson stream body/responses which takes a schema and annotates it with the stream typeid and change the current stream one (which just assumes Uint8Array) to be raw stream.

Unfortunately, my house burned down last Sunday and so I don't expect to be able to polish this up anytime soon. If someone else wants to do that polishing (raw and NDJson streams with schemas) that would be welcomed

PR #4847, opened 2025-05-09
Schema: TaggedError no longer crashes when the `message` field is explicitly defined

gcanti:
If you define a `message` field in your schema, `TaggedError` will no longer add its own `message` getter. This avoids a stack overflow caused by infinite recursion.

Before

```ts

class Todo extends Schema.TaggedError<Todo>()("Todo", {
  message: Schema.optional(Schema.String)
}) {}

// ❌ Throws "Maximum call stack size exceeded"
console.log(Todo.make({}))
```

After

```ts
// ✅ Works correctly
console.log(Todo.make({}))
```

### DIFF
--- a/.changeset/tall-bobcats-rest.md
+++ b/.changeset/tall-bobcats-rest.md
@@ -1,0 +1,27 @@
+---
+"effect": patch
+---
+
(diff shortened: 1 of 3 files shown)


dmmulroy:
What a tricky bug! Thanks for the quick fix @gcanti!

PR #4855, opened 2025-05-11
Secure Swagger UI with safe serialization and escaping

gcanti:
Enhance OpenAPI documentation handling by adding safe serialization and HTML escaping functions. This prevents script injection and ensures valid JSON output in the Swagger UI

### DIFF
--- a/packages/platform/src/HttpApiSwagger.ts
+++ b/packages/platform/src/HttpApiSwagger.ts
@@ -6,36 +6,9 @@ import type { Layer } from "effect/Layer"
 import { Api } from "./HttpApi.js"
 import { Router } from "./HttpApiBuilder.js"
 import * as HttpServerResponse from "./HttpServerResponse.js"
+import * as Html from "./internal/html.js"
 import * as internal from "./internal/httpApiSwagger.js"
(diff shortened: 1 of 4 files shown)


gcanti:
RE https://x.com/MichaelArnaldi/status/1921434824914448739

packages/platform/src/HttpApiSwagger.ts
```diff
@@ -25,16 +57,19 @@ export const layer = (options?: {
 <head>
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
-  <title>${spec.info.title} Documentation</title>
+  <title>${htmlEscape(spec.info.title)} Documentation</title>
   <style>${internal.css}</style>
 </head>
 <body>
   <div id="swagger-ui"></div>
+  <script id="swagger-spec" type="application/json">
```
tim-smart, line 65: What's the purpose of moving it to a seperate script tag?

gcanti: IIRC a type="application/json" tag is never executed, so we don't accidentally run any code

PR #4856, opened 2025-05-11
remove use of isGeneratorFunction in Effect.fn

tim-smart:


fubhy:
This still looks plausible, but the patch no longer applies cleanly to current main. Could you rebase/update it and confirm this is still desired?

PR #4864, opened 2025-05-12
Add `Array.countBy`

vinassefranche:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add `Array.countBy`

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Array.ts
+++ b/packages/effect/src/Array.ts
@@ -1122,17 +1122,18 @@ export const findLast: {
  * @since 3.16.0
  */
 export const countBy: {
-  <A>(predicate: (a: NoInfer<A>, i: number) => boolean): (self: Array<A>) => number
-  <A>(self: Array<A>, predicate: (a: A, i: number) => boolean): number
+  <A>(predicate: (a: NoInfer<A>, i: number) => boolean): (self: Iterable<A>) => number
+  <A>(self: Iterable<A>, predicate: (a: A, i: number) => boolean): number
 } = dual(
   2,
   <A>(
-    self: Array<A>,
+    self: Iterable<A>,
     f: (a: A, i: number) => boolean
   ): number => {
     let count = 0
-    for (let i = 0; i < self.length; i++) {
-      const a = self[i]
+    const as = fromIterable(self)
+    for (let i = 0; i < as.length; i++) {
+      const a = as[i]
       if (f(a, i)) {
         count++
       }
--- a/packages/effect/src/Array.ts
+++ b/packages/effect/src/Array.ts
@@ -1107,7 +1107,7 @@ export const findLast: {
 )
 
 /**
- * Counts all the element of the given iterable that pass the given predicate
+ * Counts all the element of the given array that pass the given predicate
  *
  * **Example**
  *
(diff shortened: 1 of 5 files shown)


packages/effect/src/Array.ts
```diff
@@ -1106,6 +1106,42 @@ export const findLast: {
   }
 )
 
+/**
+ * Counts all the element of the given iterable that pass the given predicate
+ *
+ * **Example**
+ *
+ * ```ts
```
tim-smart, line 1114: If you could add this to the changelog too thanks :)

vinassefranche: sure. I just did it!

KhraksMamtsov:
Maybe it makes sense to place this in the Iterable module?

vinassefranche:
> Maybe it makes sense to place this in the `Iterable` module? Because the `Array<>` is not used in the implementation of the function
> 
> ```ts
> import { Iterable, pipe } from "effect"
> 
> pipe(
>   [1, 2, 3, 4, 5],
>   Iterable.filter(n => n % 2 === 0),
>   Iterable.size,
>   console.log
> ) // 2
> ```

@KhraksMamtsov This is a good point. I wonder if it's not more discoverable to have `Array.countBy` as I think we tend to use the `Array` module more easily than the `Iterable` module but I could implement it on the `Iterable` module and just re-export the function from the `Array` module typed with an `Array<A>` instead of with an `Iterable<A>`

KhraksMamtsov:
I still think that the implementation in the Array module should be based on fromIterable and iterating over the array using a for loop with index access, rather than using for...of as in Iterable.


By using a classic for loop with index access in the Array module, you achieve different performances and make it clear that, under the hood, the iterable is being treated as an array.

vinassefranche:
> I still think that the implementation in the Array module should be based on fromIterable and iterating over the array using a for loop with index access, rather than using for...of as in Iterable.
> 
> By using a classic for loop with index access in the Array module, you achieve different performances and make it clear that, under the hood, the iterable is being treated as an array.

@KhraksMamtsov it [this](https://github.com/Effect-TS/effect/pull/4864/commits/6d542a9798bad768e50b85bfc9c6050ae3c6e4a8) what you had in mind?

KhraksMamtsov:
> @KhraksMamtsov it [this](https://github.com/Effect-TS/effect/pull/4864/commits/6d542a9798bad768e50b85bfc9c6050ae3c6e4a8) what you had in mind?
Yes – but it's still possible to accept an `Iterable` and convert it into an array using `fromIterable`, as it's done in some functions of this module.

vinassefranche:
> @vinassefranche Yes, I think this is the correct way to do it. Many functions in this module are based on `fromIterable`, and I believe this is the expected behavior.

Alright. I just did the change!

vinassefranche:
@KhraksMamtsov do you think the pull request is fine now?

KhraksMamtsov:
> @KhraksMamtsov do you think the pull request is fine now?

Yes, lgtm

tim-smart (approved):

PR #4881, opened 2025-05-14
Use function.name as spanName in `Effect.fn`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/violet-flies-bake.md
+++ b/.changeset/violet-flies-bake.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+use function.name as spanName in Effect.fn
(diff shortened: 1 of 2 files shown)


mikearnaldi:
not specifying a name means you don't want a span, that's the behaviour of `Effect.fn`, with this the behaviour would be invalidated, also generally we never use identifiers that are not available when code is minified

KhraksMamtsov:
Got it. Thanks for the clarification.

PR #4885, opened 2025-05-15
update dev dependencies

fubhy:
Updating eslint here seems to have surfaced some light mistakes introduced during EffectAdapter removal refactoring


tim-smart (approved):
Thanks janitor

PR #4911, opened 2025-05-22
feat: Function Parameters for Effect.Service

TylorS:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

POC to adding overloads to all Effect.Service's `effect` and `scoped` variants to accept function parameters

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -13760,21 +13760,6 @@ export declare namespace Service {
     : Record<PropertyKey, any> & { readonly _tag?: Key }
 
   /**
-   * @since 3.16.0
-   */
-  export type LayerNoDeps<Self, Make> = Layer.Layer<Self, MakeError<Make>, MakeContext<Make>>
-
-  /**
-   * @since 3.16.0
-   */
-  export type Layer<Self, Make> = Layer.Layer<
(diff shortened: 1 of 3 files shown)


packages/effect/src/Effect.ts
```diff
@@ -13652,17 +13660,42 @@ export const Service: <Self = never>() => [Self] extends [never] ? MissingSelfGe
...
-          return layerCache ??= layer.fromEffect(TagClass, map(maker.effect, (_) => new this(_)))
-        }
-      })
+      const isFunction = typeof maker.effect === "function"
+
+      if (isFunction) {
+        Object.defineProperty(TagClass, layerName, {
+          get(this: any) {
+            return function (this: typeof TagClass) {
+              return layerCache ??= layer.fromEffect(TagClass, map(maker.effect.apply(null, arguments), (_) => new this(_)))
```
tim-smart, line 13669: This would return the same Layer for different args.

It also leads into the memoization issue. Potentially a MutableHashMap could be used to try to keep a stable reference per argument set, but it won't catch everything so not sure about this.

PR #4924, opened 2025-05-25
`Effect.fn`: fix type inference for `Effect` suptypes in `NonGen` case

KhraksMamtsov:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

https://effect.website/play#f3982ee79375

```ts
const fn = Effect.fn(() => Effect.succeed(123), () => Option.some('test'))
const before = fn()
//    ^? Option.Option<string>
console.log("isOption:", Option.isOption(result)) // false
console.log("isEffect:", Effect.isEffect(result)) // true
// after
const after = fn()
//    ^? Effect.Effect<string, NoSuchElementException, never>
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -253,6 +253,11 @@ export declare namespace Effect {
    * @category Effect Type Extractors
    */
   export type Success<T extends Effect<any, any, any>> = [T] extends [Effect<infer _A, infer _E, infer _R>] ? _A : never
+  /**
+   * @since 3.15.5
+   * @category Effect Type Extractors
+   */
+  export type AsEffect<T extends Effect<any, any, any>> = {} & Effect<Success<T>, Error<T>, Context<T>>
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/effect/src/Effect.ts
```diff
@@ -253,6 +253,11 @@ export declare namespace Effect {
    * @category Effect Type Extractors
    */
   export type Success<T extends Effect<any, any, any>> = [T] extends [Effect<infer _A, infer _E, infer _R>] ? _A : never
+  /**
+   * @since 3.15.5
+   * @category Effect Type Extractors
+   */
+  export type AsEffect<T extends Effect<any, any, any>> = {} & Effect<Success<T>, Error<T>, Context<T>>
```
tim-smart, line 260:
```suggestion
  export type AsEffect<T extends Effect<any, any, any>> = [T] extends [Effect<infer _A, infer _E, infer _R>] ? Effect<_A, _E, _R> : never
```

Does this work? Will be more direct this way.

tim-smart:
Does fnUntraced need it too?

KhraksMamtsov:
> Does fnUntraced need it too?

No

KhraksMamtsov:
> probably better to have it there too so it can unify the return types
I’ve looked into `Effect.fnUntraced` — as far as I can tell, everything works correctly. 
I don’t understand what you’re referring to — if you can clarify more specifically, I’ll take care of it.

PR #4927, opened 2025-05-26
fix package internal imports

fubhy:
@mattiamanzati any idea why these still happen with the recent eslint rule you did?


fubhy:
> Yeah, the rule banned import from barrel files, this seems more an import from fqn module name instead of local. Maybe we should ban "@package/" in each src folder of each package?

Yeah. Basically, inside of `@effect/platform`, there should be no imports of `@effect/platform` or `@effect/platform/*`

PR #4971, opened 2025-06-03
Add `failWith` and `failWithM`

vendethiel:
These are inspired by the Haskell package `errors`'s module `Control.Error.Util`, which offers these two functions:

```haskell
failWith :: Applicative m => e -> Maybe a -> ExceptT e m a
failWithM :: Applicative m => e -> m (Maybe a) -> ExceptT e m a
```

Since our `m` is `Effect` in both cases, we end up with these signatures in Effect:

```typescript
failWith: <A, E, R>(self: Option.Option<A>, f: () => E): Effect<A, E, R>
failWithM: <E1, A, E2, R>(self: Effect<Option.Option<A>, E1>, f: () => E2): Effect<A, E1 | E2, R>
```

## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Steal `failWith` and `failWithM` from the Haskell ecosystem

-----

### Notes
I don't think I understand the test failures. They do work if I run `pnpm test`, but `pnpm check` brings about type errors related to `TestServices`. I don't see in the tests exactly what's wrong, help would be appreciated :).

I wrote some basic documentation and tried to follow the style used in the rest of the doc comments I saw, but I'm (very) open to comments on how to improve that.

Thanks to @gcanti on Discord.

Thanks.

### DIFF
--- a/.changeset/blue-ducks-lay.md
+++ b/.changeset/blue-ducks-lay.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add `failWith` and `failWithM` to `Effect`
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -13359,6 +13359,62 @@ export const transposeMapOption = dual<
 >(2, (self, f) => option_.isNone(self) ? succeedNone : map(f(self.value), option_.some))
 
 /**
+ * Returns a function that turns an `Option` and a failure `E` into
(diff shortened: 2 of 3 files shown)


vendethiel:
One thing also, the `.mapErrior(option.asEffect()` handles the `failWith` case, but there’s no `flatMap` + `mapError` builtin for the `Effect<Option<A>>` case, is there?

PR #5003, opened 2025-06-06
forbid __proto__ as name in do notation functions

beezee:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

If you use do notation to assign a value to the name 'proto' - it is silently dropped from the accumulated object at runtime. This PR excludes 'proto' from the usable names that can be assigned with do notation.

You can find a minimal reproduction of the issue here: https://effect.website/play/#ec84afe37206

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -7884,12 +7884,12 @@ export const Do: Effect<{}> = effect.Do
  */
 export const bind: {
   <N extends string, A extends object, B, E2, R2>(
-    name: Exclude<N, keyof A>,
+    name: Exclude<N, effect.ForbiddenKeys<A>>,
     f: (a: NoInfer<A>) => Effect<B, E2, R2>
   ): <E1, R1>(self: Effect<A, E1, R1>) => Effect<{ [K in N | keyof A]: K extends keyof A ? A[K] : B }, E2 | E1, R2 | R1>
   <A extends object, N extends string, E1, R1, B, E2, R2>(
     self: Effect<A, E1, R1>,
-    name: Exclude<N, keyof A>,
+    name: Exclude<N, effect.ForbiddenKeys<A>>,
     f: (a: NoInfer<A>) => Effect<B, E2, R2>
   ): Effect<{ [K in N | keyof A]: K extends keyof A ? A[K] : B }, E1 | E2, R1 | R2>
 } = effect.bind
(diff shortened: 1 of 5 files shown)


packages/effect/src/Effect.ts
```diff
@@ -7884,12 +7884,12 @@ export const Do: Effect<{}> = effect.Do
  */
 export const bind: {
   <N extends string, A extends object, B, E2, R2>(
-    name: Exclude<N, keyof A>,
+    name: Exclude<N, effect.ForbiddenKeys<A>>,
```
mikearnaldi: You'll need a publicly exported type, outside of `internal` otherwise this type won't exist in the compiled dts

mikearnaldi (changes requested):

packages/effect/src/Effect.ts
beezee: Thanks for the fast feedback @mikearnaldi ! I've got this in the works with corresponding updates for other uses of internal doNotation (Either/Option/etc)

Should have another commit up tomorrow

beezee: All set!

packages/effect/test/Stream/do-notation.test.ts
```diff
@@ -37,6 +47,21 @@ describe("do notation", () => {
...
+    expectRight(
+      pipe(
+        Stream.succeed(1),
+        Stream.bindTo("a"),
+        (x) =>
+          pipe(
+            x,
+            // @ts-expect-error
+            Stream.bind("__proto__", ({ a }) => Stream.succeed(a + 1))
+          ) as Stream.Stream<{ a: number; __proto__: number }, unknown, never>,
```
beezee, line 56: At present, if the name param to Stream.bind is inferred as never, the reader type infers as unknown instead of never, which breaks the call to expectRight.

This is already true if you attempt to shadow a name using Stream.bind today, illustrated here: https://effect.website/play/#71d73fbd5124

packages/effect/src/Micro.ts
```diff
@@ -4056,24 +4066,24 @@ export const bindTo: {
  */
 export const bind: {
   <N extends string, A extends Record<string, any>, B, E2, R2>(
-    name: N,
+    name: Exclude<N, ForbiddenKeys<{}>>,
```
beezee: Micro currently allows shadowing, leaving that as is.

gcanti:
actually simply replacing `Object.assign` with the object-spread operator resolved the issue https://github.com/Effect-TS/effect-smol/pull/158

beezee:
@gcanti works for me. I wasn't sure what baggage __proto__ might carry that Object.assign was protecting us from, so I changed the types to match current behavior instead of changing behavior.

I'll switch for your change and update my tests.

beezee:
All set ty @gcanti !

PR #5064, opened 2025-06-21
Remove global state index by version, make version mismatch a waning message

mikearnaldi:


packages/effect/src/internal/fiberRuntime.ts
```diff
@@ -1370,11 +1370,15 @@ export class FiberRuntime<in out A, in out E = never> extends Effectable.Class<A
...
-              return core.dieMessage(
-                `Cannot execute an Effect versioned ${
-                  (cur as core.Primitive)[core.EffectTypeId]._V
-                } with a Runtime of version ${version.getCurrentVersion()}`
-              )
+              const level = this.getFiberRef(core.currentVersionMismatchErrorLogLevel)
+              if (level._tag === "Some") {
+                const effectVersion = (cur as core.Primitive)[core.EffectTypeId]._V
+                this.log(
+                  `Executing an Effect versioned ${effectVersion} with a Runtime of version ${version.getCurrentVersion()}, you may want to dedupe the effect dependencies, you can use the laguage service plugin to detect this at compile time: https://github.com/Effect-TS/language-service`,
```
fubhy:
```suggestion
                  `Executing an Effect versioned ${effectVersion} with a Runtime of version ${version.getCurrentVersion()}, you may want to dedupe the effect dependencies, you can use the language service plugin to detect this at compile time: https://github.com/Effect-TS/language-service`,
```

fubhy: We could add a section about this to our documentation and then link to it from here.

fubhy:
@mikearnaldi we should probably make this only log once.

mikearnaldi:
> @mikearnaldi we should probably make this only log once.

not sure I agree, there could be multiple version mismatch issues

fubhy:
Ok. It could be a feature. Hah. ... 

console.log in the browser can be quite slow. Especially if you have the debugger open. Could be an incentive for ppl to fix it :-D

.changeset/huge-sites-make.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
+---
+
+Remove global state index by version, make version mismatch a waning message
```
tim-smart:
```suggestion
Remove global state index by version, make version mismatch a warning message
```

packages/effect/src/FiberRef.ts
```diff
@@ -338,6 +338,13 @@ export const currentMaxOpsBeforeYield: FiberRef<number> = core.currentMaxOpsBefo
  */
 export const unhandledErrorLogLevel: FiberRef<Option.Option<LogLevel.LogLevel>> = core.currentUnhandledErrorLogLevel
 
+/**
+ * @since 2.0.0
```
tim-smart:
```suggestion
 * @since 3.17.0
```

packages/effect/src/Layer.ts
```diff
@@ -1061,6 +1061,17 @@ export const setUnhandledErrorLogLevel: (level: Option.Option<LogLevel>) => Laye
     fiberRuntime.fiberRefLocallyScoped(core.currentUnhandledErrorLogLevel, level)
   )
 
+/**
+ * @since 2.0.0
```
tim-smart:
```suggestion
 * @since 3.17.0
```

tim-smart (approved):

PR #5106, opened 2025-06-26
chore: upgrade TSTyche

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

I have just shipped TSTyche 4.1. Now it collects `// @ts-expect-error` directives (in the type test files only), requires to provide a message after the directive and checks it as well. The following:

```ts
// @ts-expect-error: Argument of type 'boolean' is not assignable to parameter of type 'string'
doSomething(false);
```

fails with:

Screenshot 2025-06-26 at 12 45 41

---

As you know, the downside of `// @ts-expect-error` is that it hides *any* amount of *any* errors in the following line. That does not work with type tests.

I think, it is better to handle `// @ts-expect-error` instead of having `.toRaiseError()` matcher. There are no red squiggles when the directive is used. Therefore I plan to remove `.toRaiseError()` in the future.

---

To ignore a directive, append the `!` character after it: `// @ts-expect-error!`. For instance, this is useful in case if there is an upstream bug.

That is explained when the message fragment is not provided:

Screenshot 2025-06-26 at 12 58 51

### DIFF
--- a/tstyche.config.json
+++ b/tstyche.config.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://tstyche.org/schemas/config.json",
+  "checkSuppressedErrors": true,
   "testFileMatch": [
     "packages/*/dtslint/**/*.tst.*"
   ]
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "playwright": "^1.52.0",
     "prettier": "^3.5.3",
     "rimraf": "^6.0.1",
-    "tstyche": "^4.0.0",
+    "tstyche": "^4.1.0",
     "tsx": "^4.19.4",
     "typescript": "^5.8.3",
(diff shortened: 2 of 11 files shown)


tstyche.config.json
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "https://tstyche.org/schemas/config.json",
+  "checkSuppressedErrors": true,
```
mrazauskas, line 3: Behind a flag for now, but will be enable by default in the future.

gcanti:
@mrazauskas nice addition, thanks!

mrazauskas:
@gcanti Thanks! Do you mean addiction? (;

Lint is fixed now.

PR #5132, opened 2025-07-02
Allow verbose printing of cause

mikearnaldi:


packages/effect/src/internal/cause.ts
```diff
@@ -978,26 +980,30 @@ const locationRegex = /\((.*)\)/g
...
+      )
     }
-    out.push(
-      lines[i]
-        .replace(/at .*effect_instruction_i.*\((.*)\)/, "at $1")
-        .replace(/EffectPrimitive\.\w+/, "<anonymous>")
-    )
   }
 
   if (span) {
```
dmaretskyi, line 1009: I believe this needs to be updated as well, to insert spans at a specific position in the stack when verbose=true

packages/effect/src/internal/cause.ts
```diff
@@ -1035,14 +1041,14 @@ const prettyErrorStack = (message: string, stack: string, span?: Span | undefine
 export const spanSymbol = Symbol.for("effect/SpanAnnotation")
 
 /** @internal */
-export const prettyErrors = <E>(cause: Cause.Cause<E>): Array<PrettyError> =>
+export const prettyErrors = <E>(cause: Cause.Cause<E>, verbose = true): Array<PrettyError> =>
```
dmaretskyi, line 1044: This function is exported as `Cause.prettyErrors`, so this change does change the default behaviour of it

fubhy:
Closing because this no longer applies cleanly to the target branch. Please open a fresh PR if the change is still relevant.

PR #5133, opened 2025-07-02
Support extracting tool call results from `AiResponse.WithToolCallResults`

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Allows extracting the results of specific tool calls from `AiResponse.WithToolCallResults`:

```ts
import { AiResponse, AiTool } from "@effect/ai"

const tool = AiTool.make("GetDadJoke", {
  description: "Get a hilarious dad joke from the ICanHazDadJoke API",
  success: Schema.String,
  failure: Schema.Never,
  parameters: {
    searchTerm: Schema.String.annotations({
      description: "The search term to use to find dad jokes"
    })
  }
})

declare const response: AiResponse.WithToolCallResults<typeof tool>

const result = response.getToolCallResult("GetDadJoke")
//    ^? Option<string>
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/large-dots-nail.md
+++ b/.changeset/large-dots-nail.md
@@ -1,0 +1,5 @@
+---
+"@effect/ai": patch
+---
+
+Support extracting tool call results from `AiResponse.WithToolCallResults`
--- a/packages/ai/ai/src/AiInput.ts
+++ b/packages/ai/ai/src/AiInput.ts
@@ -378,6 +378,10 @@ export class ToolCallResultPart extends Schema.TaggedClass<ToolCallResultPart>(
    */
   id: ToolCallId,
   /**
+   * The name of the tool that was called.
+   */
+  name: Schema.String,
+  /**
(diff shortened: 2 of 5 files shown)


elliotdauber (commented):
would it be possible for the ai to call the same tool multiple times? i dont think ive ever seen that, but i guess in theory it's possible? if so, another option would be to return an array instead of an option

IMax153:
> would it be possible for the ai to call the same tool multiple times? i dont think ive ever seen that, but i guess in theory it's possible? if so, another option would be to return an array instead of an option

That's a good question - according to ChatGPT the answer _seems_ to be no. Given the APIs are unstable, we could experiment with this and see if folks run into issues in the wild.
CleanShot 2025-07-02 at 13 12 45

elliotdauber:
cool sounds good

PR #5134, opened 2025-07-02
add `Random` implementation with fixed values

fubhy:


packages/effect/src/internal/random.ts
```diff
@@ -87,3 +88,73 @@ const swap = <A>(buffer: Array<A>, index1: number, index2: number): Array<A> =>
...
+      return min + (hash % (max - min))
+    })
+  }
+
+  shuffle<A>(elements: Iterable<A>): Effect.Effect<Chunk.Chunk<A>> {
+    return shuffleWith(elements, (n) => this.nextIntBetween(0, n))
+  }
+}
+
+export const fixed = <T extends Arr.NonEmptyArray<any>>(values: T): Random.Random => new FixedRandomImpl(values)
```
mikearnaldi, line 161: missing internal

fubhy: The `/* @internal */` annotation is technically only necessary in public modules

Issue #5069, opened 2025-06-22
Clarify length requirements in tuple documentation


cyberixae:
### What is the type of issue?

Documentation is confusing

### What is the issue?

The `Tuple` module contains functions like `map`, that work on tuples of arbitrary length, but also functions like `mapBoth`, that only work on tuples with exactly two items.

I think it would be good to clarify the length requirement in prose.

* `map` "Transforms each element of a tuple using the given function"
* `mapBoth` "Transforms both elements of a 2-tuple using the given functions"

I'm not sure if it makes sense for these two functions to be in the same module to begin with. The `mapBoth` function could perhaps be moved to a separate module called `Pair`, or `Tuple2` in case similar functions are to be expected for other tuple lengths. I guess some of the 2-tuple specific functions could alternatively be updated to work with arbitrary length tuples.

### Where did you find it?

https://effect-ts.github.io/effect/effect/Tuple.ts.html

aymenhmaidiwastaken:
Hey! I'd like to work on this. The distinction between 2-tuple-only functions (`getFirst`, `getSecond`, `mapBoth`, `mapFirst`, `mapSecond`, `swap`) and arbitrary-length tuple functions (`make`, `map`, `appendElement`, `at`, `getEquivalence`, `getOrder`, `isTupleOf`, `isTupleOfAtLeast`) is easy to miss from the current docs.

I'll update the JSDoc comments for the 2-tuple-specific functions to explicitly call out the 2-element constraint, and add a note to the module-level doc clarifying the two categories. No code changes — just documentation improvements.

cyberixae:
The Number chapter introduction seems to have a table. Perhaps a similar table could be added to tuple chapter.

See https://effect-ts.github.io/effect/effect/Number.ts.html

PR #5158, opened 2025-07-03
Clarify Tuple length requirements

cyberixae:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

Previously tuples of specific length were only referred to as "tuple" in the documentation. This pull request replaces "tuple" with "2-tuple" in situations where a tuple of length 2 is required.

## Related

- Related Issue #5069

### DIFF
--- a/packages/effect/src/Tuple.ts
+++ b/packages/effect/src/Tuple.ts
@@ -34,7 +34,7 @@ export interface TupleTypeLambda extends TypeLambda {
 export const make = <A extends ReadonlyArray<any>>(...elements: A): A => elements
 
 /**
- * Return the first element of a tuple.
+ * Return the first element from a tuple with two elements.
  *
  * @example
  * ```ts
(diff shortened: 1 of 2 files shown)


packages/effect/src/Tuple.ts
```diff
@@ -34,7 +34,7 @@ export interface TupleTypeLambda extends TypeLambda {
 export const make = <A extends ReadonlyArray<any>>(...elements: A): A => elements
 
 /**
- * Return the first element of a tuple.
+ * Return the first element of a 2-tuple.
```
tim-smart:
```suggestion
 * Return the first element from a tuple with two elements.
```

I think this reads better :)

cyberixae: PR updated!

tim-smart (approved):

PR #5164, opened 2025-07-04
Make `Effect.runPromise` and `Effect.runSync` Transparently Re-throw Original Errors

dmaretskyi:
## Motivation

This PR updates `Effect.runPromise` and `Effect.runSync` to re-throw the original errors from failed effects without wrapping them in internal structures like `FiberFailure`. This aligns error handling more closely with standard JavaScript behavior and improves ergonomics when integrating Effect into existing codebases.

Why This Matters:

- **Better interop with existing tools and environments**:
  - In Node.js, unhandled rejections are shown using the original error.
  - In test runners like Vitest or Jest, test failures show up as the original error, making debugging easier.
  - In browser consoles, the original error stack and message are preserved, making the experience consistent with non-effect code.
- **Improved transparency in mixed codebases**:
In apps that are adopting effect gradually, it's common to mix effect and non-effect code. If the effect fails due to an error thrown in some non-effect code inside the effect block, that exact error should propagate outward — as if the effect layer wasn’t there at all. This makes the effect block transparent and helps avoid confusion, especially for newcomers.

## Description

- If the `Cause` contains a single failure or defect, we re-throw the original error directly.
- If the `Cause` contains multiple errors, we throw an AggregateError containing all the causes.
- For users who need to inspect the full `Cause` structure, `Effect.runPromiseExit` and `Effect.runSyncExit` continue to expose the full details.
- `FiberFailure` error class becomes redundant and is removed.
- Errors that are re-thrown have their stacks cleaned from the Effect-runtime internals and Effect spans inserted in the correct place. This is similar to what `PrettyError` already does.
- Errors will have the full stack-trace available: including the code calling `Effect.runSync`/`Effect.runPromise` the Effect function stack as well as any vanilla JS functions called from inside Effect.

## TODO

- [ ] Clean-up error stack processing code and unify with existing code in `cause.ts`.
- [ ] Update the docs.
- [ ] Update failing tests.

### DIFF
--- a/packages/effect/src/Runtime.ts
+++ b/packages/effect/src/Runtime.ts
@@ -1,7 +1,6 @@
 /**
  * @since 2.0.0
  */
-import type { Cause } from "./Cause.js"
 import type * as Context from "./Context.js"
 import type * as Effect from "./Effect.js"
 import type * as Exit from "./Exit.js"
@@ -9,7 +8,6 @@ import type * as Fiber from "./Fiber.js"
 import type * as FiberId from "./FiberId.js"
 import type * as FiberRef from "./FiberRef.js"
 import type * as FiberRefs from "./FiberRefs.js"
-import type { Inspectable } from "./Inspectable.js"
 import * as internal from "./internal/runtime.js"
 import type { Pipeable } from "./Pipeable.js"
 import type * as RuntimeFlags from "./RuntimeFlags.js"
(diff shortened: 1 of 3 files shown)


tim-smart:
This is something we are looking at for effect 4, as it is a breaking change :)

fubhy:
Closing as stale. This changes public error/runtime behavior and was noted as Effect 4 work, but there has been no activity or active review path for over 6 months. If this is still desired for the current codebase, please open a fresh PR aligned with the current Effect 4/runtime design and include the needed tests and changeset.

PR #5169, opened 2025-07-05
add tool call meta to handlers

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Allows for accessing tool call metadata within the tool call handler:

```ts
const DadJokeToolHandlers = DadJokeTools.toLayer(
  Effect.gen(function*() {
    const icanhazdadjoke = yield* ICanHazDadJoke
    return {
      GetDadJoke: ({ searchTerm }, meta) => {
        // At the moment, the `meta` contains the tool name and id, but
        // can be further augmented in the future
        console.log(meta)
        return icanhazdadjoke.search(searchTerm)
      }
    }
  })
).pipe(Layer.provide(ICanHazDadJoke.Default))

```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/ai/ai/src/McpServer.ts
+++ b/packages/ai/ai/src/McpServer.ts
@@ -632,7 +632,7 @@ export const registerToolkit: <Tools extends AiTool.Any>(toolkit: AiToolkit.AiTo
     yield* registry.addTool({
       tool: mcpTool,
       handle(payload) {
-        return built.handle(tool.name as any, payload).pipe(
+        return built.handle("", tool.name as any, payload).pipe(
           Effect.provide(context as Context.Context<any>),
           Effect.match({
             onFailure: (error) =>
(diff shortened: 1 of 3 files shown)


packages/ai/ai/src/McpServer.ts
```diff
@@ -632,7 +632,7 @@ export const registerToolkit: <Tools extends AiTool.Any>(toolkit: AiToolkit.AiTo
     yield* registry.addTool({
       tool: mcpTool,
       handle(payload) {
-        return built.handle(tool.name as any, payload).pipe(
+        return built.handle("", tool.name as any, payload).pipe(
```
IMax153, line 635: In particular, this feels bad to me.

mattiamanzati: Wouldn’t the rpc request id be fine here?

IMax153: It could be if that's what we would like

fubhy:

PR #5174, opened 2025-07-06
feat(platform): add recursive option to FileSystem.watch

### LINKED ISSUE #2986: Add "recursive" option to `FileSystem.watch`

This would make it easier to recursively watch changes in a directory with the Node platform, rather than only changes at the top level of the watched directory.

Adding the `recursive` option to the FileSystem API. See [the Node.js documentation](https://nodejs.org/docs/latest/api/fs.html#fswatchfilename-options-listener) for information about this option.

Wrapping the Node.js-native `fs.watch` API manually—effectively just copying [the Effect implementation](https://github.com/Effect-TS/effect/blob/13e3dcf9f8a4180c7df3a7eb6e08e8735df18999/packages/platform-node-shared/src/internal/fileSystem.ts#L526-L573) and adding support for the option.

schickling:
## Summary

This PR implements the feature requested in #2986 to add a `recursive` option to `FileSystem.watch` for watching changes in subdirectories.

## Changes

- Added `WatchOptions` interface with an optional `recursive` boolean property
- Updated `FileSystem.watch` method signature to accept optional `WatchOptions` parameter  
- Modified the Node.js implementation to pass the recursive option to `fs.watch`
- Added comprehensive tests for recursive watching functionality
- Updated the fs-watch example to demonstrate recursive watching

## Implementation Details

The recursive option is passed directly to Node.js's native `fs.watch` API. As per Node.js documentation, recursive watching is only supported on macOS and Windows. On other platforms, the option is ignored.

## Test Plan

- [x] Added test to verify the watch API accepts the recursive option
- [x] Added test to verify recursive watching detects changes in subdirectories
- [x] Added test to verify non-recursive watching ignores subdirectory changes
- [x] All existing tests pass
- [x] Type checking passes
- [x] Linting passes

Closes #2986

🤖 Generated with Claude Code

### DIFF
--- a/packages/platform-node-shared/src/internal/fileSystem.ts
+++ b/packages/platform-node-shared/src/internal/fileSystem.ts
@@ -535,11 +535,11 @@ const utimes = (() => {
 
 // == watch
 
-const watchNode = (path: string) =>
+const watchNode = (path: string, options?: FileSystem.WatchOptions) =>
   Stream.asyncScoped<FileSystem.WatchEvent, Error.PlatformError>((emit) =>
     Effect.acquireRelease(
       Effect.sync(() => {
-        const watcher = NFS.watch(path, {}, (event, path) => {
+        const watcher = NFS.watch(path, { recursive: options?.recursive }, (event, path) => {
           if (!path) return
           switch (event) {
             case "rename": {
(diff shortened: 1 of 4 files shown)


packages/platform-node-shared/src/internal/fileSystem.ts
```diff
@@ -578,12 +578,16 @@ const watchNode = (path: string) =>
...
-const watch = (backend: Option.Option<Context.Tag.Service<FileSystem.WatchBackend>>, path: string) =>
+const watch = (
+  backend: Option.Option<Context.Tag.Service<FileSystem.WatchBackend>>,
+  path: string,
+  options?: FileSystem.WatchOptions
+) =>
   stat(path).pipe(
     Effect.map((stat) =>
       backend.pipe(
         Option.flatMap((_) => _.register(path, stat)),
```
tim-smart: The options should also be passed to the backend.

PR #5205, opened 2025-07-13
feat(@effect/vitest): modernize Vitest integration with AbortSignal support

schickling:
## Summary
- Modernizes the @effect/vitest package to use Vitest 3.2 with native AbortSignal support
- Integrates AbortSignal at the Effect runtime level for proper test cancellation
- Ensures Effect-based tests can be cleanly cancelled when they timeout, fail with --bail flag, or are interrupted

## Changes
- **Dependency Update**: Bump Vitest from ^3.0.0 to ^3.2.0 to access latest AbortSignal API
- **Runtime Integration**: Use `Effect.runPromise(effect, { signal })` for native cancellation support
- **Test Coverage**: Add comprehensive test cases demonstrating AbortSignal integration and type safety

## Test Plan
- [x] Added new test cases verifying AbortSignal is properly passed to test context
- [x] Tests verify AbortSignal type safety and availability
- [x] Existing tests continue to pass
- [x] Integration follows Effect patterns and conventions

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.changeset/vitest-abort-signal-modernization.md
+++ b/.changeset/vitest-abort-signal-modernization.md
@@ -1,0 +1,11 @@
+---
+"@effect/vitest": minor
+---
+
+Modernize Vitest integration with AbortSignal support
+
+- Bump Vitest dependency from ^3.0.0 to ^3.2.0 to support latest features
+- Add support for Vitest 3.2's AbortSignal API for proper test cancellation
+- Effect-based tests now respect AbortSignal for clean cancellation when tests timeout, fail with --bail flag, or are interrupted via Ctrl+C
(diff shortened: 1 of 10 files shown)


schickling:
@tim-smart seems like there are some new type-level incompatibilities with the new vitest version. Would appreciate some guidance on how to fix those.

mikearnaldi:
Hmm, the issue here is that `vitest` added a field named `scoped` to the api and this conflicts with our `scoped` method. Updating would be a breaking change.

@tim-smart / @IMax153 for 4.0 we should rethink how we extend the vitest api, unsure it's a good idea to add methods

tim-smart (approved):
In 4 we could export an explicit interface, and namespace the original vitest apis.

PR #5211, opened 2025-07-14
Removed some unnecessary single-arg pipe calls

mattiamanzati:
## Description

Cleaned up some single-arg pipe calls.

### DIFF
--- a/packages/cluster/test/Sharding.test.ts
+++ b/packages/cluster/test/Sharding.test.ts
@@ -103,7 +103,7 @@ describe.concurrent("Sharding", () => {
         }, 0)
       }).pipe(
         Effect.provide(TestShardingWithoutRunners.pipe(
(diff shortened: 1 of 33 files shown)


packages/cluster/test/Sharding.test.ts
```diff
@@ -103,7 +103,7 @@ describe.concurrent("Sharding", () => {
         }, 0)
       }).pipe(
         Effect.provide(TestShardingWithoutRunners.pipe(
-          Layer.provide(Layer.effect(
+          Layer.provide(Layer.scoped(
```
tim-smart, line 106: Was it leaking a scope?

mattiamanzati: Yup it was, got catched by the LSP

tim-smart (approved):

PR #5237, opened 2025-07-18
feat: Add Struct.entries

dmaretskyi:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds `Struct.entries` as a better-typed alternative to `Object.entries`.
The new function preserves the key type.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/curvy-sloths-tickle.md
+++ b/.changeset/curvy-sloths-tickle.md
@@ -1,5 +1,5 @@
 ---
-"effect": patch
+"effect": minor
 ---
 
 Add `Struct.entries` function
--- a/.changeset/curvy-sloths-tickle.md
+++ b/.changeset/curvy-sloths-tickle.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add `Struct.entries` function
--- a/packages/effect/src/Struct.ts
+++ b/packages/effect/src/Struct.ts
@@ -221,23 +221,24 @@ export const get =
 export const keys = <T extends {}>(o: T): Array<(keyof T) & string> => Object.keys(o) as Array<(keyof T) & string>
 
 /**
- * Retrieves the object entries in a typed manner
+ * Retrieves the entries (key-value pairs) of an object, where keys are strings,
+ * in a type-safe manner. Symbol keys are excluded from the result.
  *
  * @example
  * ```ts
  * import * as assert from "node:assert"
  * import { Struct } from "effect"
  *
- * const value = {
- *   a: 1,
- *   b: 2
- * }
+ * const c = Symbol("c")
+ * const value = { a: "foo", b: 1, [c]: true }
  *
  * const entries: Array<["a" | "b", number]> = Struct.entries(value)
  *
- * assert.deepStrictEqual(entries, [["a", 1], ["b", 2]])
+ * assert.deepStrictEqual(entries, [["a", "foo"], ["b", 1]])
  * ```
  *
- * @since 3.16.17
+ * @since 3.17.0
  */
-export const entries = <R>(obj: R): [keyof R, R[keyof R]][] => Object.entries(obj as any) as [keyof R, any][]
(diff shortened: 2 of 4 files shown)


.changeset/curvy-sloths-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
```
gcanti:
```suggestion
"effect": minor
```

packages/effect/src/Struct.ts
```diff
@@ -219,3 +219,25 @@ export const get =
...
+ * }
+ *
+ * const entries: Array<["a" | "b", number]> = Struct.entries(value)
+ *
+ * assert.deepStrictEqual(entries, [["a", 1], ["b", 2]])
+ * ```
+ *
+ * @since 3.16.17
+ */
+export const entries = <R>(obj: R): [keyof R, R[keyof R]][] => Object.entries(obj as any) as [keyof R, any][]
```
gcanti: we should exclude symbol keys

```ts
/**
 * Retrieves the entries (key-value pairs) of an object, where keys are strings,
 * in a type-safe manner. Symbol keys are excluded from the result.
 *
 * @example
 * ```ts
 * import * as assert from "node:assert"
 * import { Struct } from "effect"
 *
 * const c = Symbol("c")
 * const value = { a: "foo", b: 1, [c]: true }
 *
 * const entries: Array<["a" | "b", number]> = Struct.entries(value)
 *
 * assert.deepStrictEqual(entries, [["a", "foo"], ["b", 1]])
 * ```
 *
 * @since 3.17.0
 */
export const entries = <const R>(obj: R): Array<[keyof R & string, R[keyof R & string]]> =>
  Object.entries(obj as any) as any
```

the since tag should be 3.17.0

packages/effect/test/Struct.test.ts
```diff
@@ -120,4 +120,8 @@ describe("Struct", () => {
     strictEqual(pipe({ a: 1 }, Struct.get("a")), 1)
     strictEqual(pipe({}, Struct.get("a")), undefined)
   })
+
+  it("entries", () => {
+    deepStrictEqual(Struct.entries({ a: 1, b: 2 }), [["a", 1], ["b", 2]])
+  })
```
gcanti: I would change the runtime test accordingly

```ts
  it("entries", () => {
    const c = Symbol("c")
    // should not include symbol keys
    deepStrictEqual(Struct.entries({ a: "a", b: 1, [c]: 2 }), [["a", "a"], ["b", 1]])
  })
```

gcanti (changes requested):
I would also add a type-level test (in `packages/effect/dtslint/Struct.tst.ts`)

```ts
  it("entries", () => {
    const c = Symbol("c")
    const value = { a: "a", b: 1, [c]: 2 }
    // should not include symbol keys
    expect(Struct.entries(value)).type.toBe<Array<["a" | "b", string | number]>>()
    // when the object is passed as a parameter, the values should be narrowed
    expect(Struct.entries({ a: "a", b: 1, [c]: 2 })).type.toBe<Array<["a" | "b", "a" | 1]>>()
  })
```

packages/effect/src/Struct.ts
```diff
@@ -219,3 +219,25 @@ export const get =
...
+ *
+ * @example
+ * ```ts
+ * import * as assert from "node:assert"
+ * import { Struct } from "effect"
+ *
+ * const c = Symbol("c")
+ * const value = { a: "foo", b: 1, [c]: true }
+ *
+ * const entries: Array<["a" | "b", number]> = Struct.entries(value)
```
gcanti:
```suggestion
 * const entries: Array<["a" | "b", string | number]> = Struct.entries(value)
```

gcanti (changes requested):

PR #5255, opened 2025-07-22
Fix problematic prompt pasting

### LINKED ISSUE #1810: Copy-pasting in Prompt field only writes one character
When I paste some text in a prompt field, it only writes the first character

kitlangton:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

All keypress events (including rapid-fire paste events) now get pushed into an eventsQueue immediately as they occur.

## Related

- Closes #1810

### DIFF
--- a/packages/platform-node-shared/src/internal/terminal.ts
+++ b/packages/platform-node-shared/src/internal/terminal.ts
@@ -1,27 +1,26 @@
 import * as Error from "@effect/platform/Error"
 import * as Terminal from "@effect/platform/Terminal"
 import * as Effect from "effect/Effect"
+import * as Exit from "effect/Exit"
 import * as Layer from "effect/Layer"
+import * as Mailbox from "effect/Mailbox"
 import * as Option from "effect/Option"
-import * as Queue from "effect/Queue"
+import * as RcRef from "effect/RcRef"
 import * as readline from "node:readline"
 
 const defaultShouldQuit = (input: Terminal.UserInput) =>
   input.key.ctrl && (input.key.name === "c" || input.key.name === "d")
 
 /** @internal */
-export const make = (
+export const make = Effect.fnUntraced(function*(
   shouldQuit: (input: Terminal.UserInput) => boolean = defaultShouldQuit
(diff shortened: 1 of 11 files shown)


packages/platform-node-shared/src/internal/terminal.ts
```diff
@@ -3,124 +3,98 @@ import * as Terminal from "@effect/platform/Terminal"
...
-      // The columns property can be undefined if stdout was redirected
-      columns: Effect.sync(() => output.columns || 0),
+      columns,
       readInput,
       readLine,
       display
     })
   })
 
-/** @internal */
```
tim-smart, line 122: You can leave the internal comments here, so ts can strip away the internal types.

packages/platform-node-shared/src/internal/terminal.ts
```diff
@@ -3,124 +3,98 @@ import * as Terminal from "@effect/platform/Terminal"
...
 import * as Layer from "effect/Layer"
 import * as Option from "effect/Option"
+import * as Queue from "effect/Queue"
 import * as readline from "node:readline"
 
-const defaultShouldQuit = (input: Terminal.UserInput): boolean =>
+const defaultShouldQuit = (input: Terminal.UserInput) =>
   input.key.ctrl && (input.key.name === "c" || input.key.name === "d")
 
-/** @internal */
```
tim-smart, line 11: Another one.

packages/platform-node-shared/src/internal/terminal.ts
kitlangton: Doh! That was a mistake :) I'll fix

kitlangton: Updated :)

packages/platform-node-shared/src/internal/terminal.ts
kitlangton: Updated :)

PR #5257, opened 2025-07-22
fix: properly form otlp urls

jrmdayn:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Using `baseUrl` with a trailing `/` leads to nothing being exported. When using a `URL` object + `toString()` method there is always a trailing slash..
```ts
Otlp.layer({ baseUrl: (yield* Config.url('ENDPOINT')).toString() }) // will end up exporting to `http://endpoint//v1/traces`
```

To fix this we should always use the `URL` module to construct urls..

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/huge-olives-sink.md
+++ b/.changeset/huge-olives-sink.md
@@ -2,4 +2,4 @@
 "@effect/opentelemetry": patch
 ---
 
-properly form otlp urls
+use URL constructor for joining url's in Otlp.layer
--- a/.changeset/huge-olives-sink.md
+++ b/.changeset/huge-olives-sink.md
@@ -1,0 +1,5 @@
+---
+"@effect/opentelemetry": patch
+---
+
+properly form otlp urls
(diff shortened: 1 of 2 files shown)


.changeset/huge-olives-sink.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/opentelemetry": patch
+---
+
+properly form otlp urls
```
tim-smart, line 5: use URL constructor for joining url's in Otlp.layer

tim-smart:
```suggestion
use URL constructor for joining url's in Otlp.layer
```

tim-smart (approved):

johtso:
This breaks for base urls that have a path at the end, we're trying to join an absolute path to it (starts with a slash) which nukes whatever path was in the base url #5333

PR #5266, opened 2025-07-23
feat: onFailure callback for SuspendOnFailure

jrmdayn:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When using `SuspendOnFailure=true` annotation for a Workflow, it is difficult to know that the workflow is suspended because a failure happened. By providing an effectful callback in the annotation, we can for example log the `Cause` and `executionId` in a descriptive message explaining how to resume the workflow once the error has been fixed:

```ts
const MyWorkflow = Workflow.make({})
  .annotate(Workflow.SuspendOnFailure, {
    onFailure: ({ cause, instance }) =>
      Console.log(
        "Workflow is suspended because a failure occurred. Once it is fixed, you can resume the workflow using the executionId provided in this log.",
        { executionId: instance.executionId, workflow: instance.workflow.name, cause }
      )
  })
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/famous-masks-cover.md
+++ b/.changeset/famous-masks-cover.md
@@ -1,0 +1,5 @@
+---
+"@effect/workflow": patch
+---
+
+Provide an option to pass a callback when a failure occurs before suspending when using SuspendOnFailure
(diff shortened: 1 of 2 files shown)


tim-smart:
I think instead we can add a cause to the instance when it fails in this way, and then you can just register a finalizer that does the logging.

tim-smart:
Closing in favour of https://github.com/Effect-TS/effect/pull/5271

jrmdayn:
> I think instead we can add a cause to the instance when it fails in this way, and then you can just register a finalizer that does the logging.

Works great! thanks Tim

PR #5267, opened 2025-07-23
Add Number.polarity

cyberixae:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds `Number.polarity` that can be used to distinguish between zero and negative zero.

### DIFF
--- a/.changeset/eight-paws-chew.md
+++ b/.changeset/eight-paws-chew.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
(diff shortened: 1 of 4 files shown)


cyberixae:
I was planning to use this utility to detect negative zero. However, after studying the subject, I now think it makes more sense to use `Object.is(x, -0)` comparison instead.

PR #5276, opened 2025-07-25
fix(tests): add posibility to chain runif with only, closes #5199

### LINKED ISSUE #5199: it.effect.runIf(...).only is not available

_No response_

```ts
// Works -- native vitest APIs
it.runIf(process.env.ANTHROPIC_API_KEY).only('streaming', async ({ expect }) => {
  // TODO
});

// Type error
it.effect.runIf(process.env.ANTHROPIC_API_KEY).only('streaming', Effect.fn(function* ({ expect }) {
  // TODO
}));
```

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

MatthD:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- Add the possibility to chain .only after `runIf` as you could with vitest

## Related

- Related Issue #5199
- Closes #5199

### DIFF
--- a/packages/vitest/src/index.ts
+++ b/packages/vitest/src/index.ts
@@ -86,13 +86,10 @@ export namespace Vitest {
     | Array<Schema.Schema.Any | FC.Arbitrary<any>>
     | { [K in string]: Schema.Schema.Any | FC.Arbitrary<any> }
 
-  /**
-   * @since 1.0.0
-   */
   export interface Tester<R> extends Vitest.Test<R> {
     skip: Vitest.Test<R>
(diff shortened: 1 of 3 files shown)


fubhy:
Thanks for the PR. This still looks like a valid gap, but the branch is currently blocked by CI and the test adds a committed `.only`, which cannot be merged. Could you update the test so it verifies chainability without leaving a focused test, restore the dropped `@since` JSDoc, and add a changeset if this should release as a package fix?

PR #5282, opened 2025-07-27
Improve inference of Metric.mapInput for use in Effect.pipe(...)

beezee:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Metric.mapInput` is extremely useful for plucking values or running pure computation from lexical scope built in an `Effect.bind` pipe to send into a Metric, but the current typings prevent inference of the contramapping function. This means that with a large lexical scope object which changes from one `bind` to the next, you have to manually define potentially large and always ephemeral types as part of the `mapInput` `f` argument signature.

I've illustrated the before and after in this playground: https://effect.website/play/#a9714d08c979

If it's preferable to define an alias with the inferable type, I'm happy to make the change. If this is undesirable altogether, I can always use my own implementation as in the playground example.

Figured this is a strong enough use case for `mapInput` that it might make sense to support in the library.

### DIFF
--- a/packages/effect/src/Metric.ts
+++ b/packages/effect/src/Metric.ts
@@ -152,6 +152,10 @@ export const make: MetricApply = internal.make
  * @category mapping
  */
 export const mapInput: {
+  <Type, In, Out, In2>(
+    self: Metric<Type, In, Out>,
+    f: (input: In2) => In
+  ): <E, R>(x: Effect.Effect<In2, E, R>) => Effect.Effect<In2, E, R>
   <In, In2>(f: (input: In2) => In): <Type, Out>(self: Metric<Type, In, Out>) => Metric<Type, In2, Out>
   <Type, In, Out, In2>(self: Metric<Type, In, Out>, f: (input: In2) => In): Metric<Type, In2, Out>
 } = internal.mapInput
--- a/packages/effect/src/Metric.ts
+++ b/packages/effect/src/Metric.ts
@@ -152,10 +152,6 @@ export const make: MetricApply = internal.make
  * @category mapping
  */
 export const mapInput: {
-  <Type, In, Out, In2>(
-    self: Metric<Type, In, Out>,
-    f: (input: In2) => In
-  ): <E, R>(x: Effect.Effect<In2, E, R>) => Effect.Effect<In2, E, R>
   <In, In2>(f: (input: In2) => In): <Type, Out>(self: Metric<Type, In, Out>) => Metric<Type, In2, Out>
   <Type, In, Out, In2>(self: Metric<Type, In, Out>, f: (input: In2) => In): Metric<Type, In2, Out>
 } = internal.mapInput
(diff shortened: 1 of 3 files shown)


beezee:
forgot to git add the changeset, my bad! coming right up.

beezee:
Woops, I definitely used lint and lint-fix, maybe before I staged tests? Fixing now

beezee:
@tim-smart thanks for kicking this off the first time, sorry for the missed spot! Just re-ran lint-fix local, and lint has passed before amending commit and pushing.

packages/effect/src/Metric.ts
```diff
@@ -152,6 +152,10 @@ export const make: MetricApply = internal.make
  * @category mapping
  */
 export const mapInput: {
+  <Type, In, Out, In2>(
```
tim-smart: I think we will want a separate api for this, otherwise you will get back a function instead of a new metric by default.

Maybe `Metric.updateMap`?

tim-smart: Actually, Metric.trackSuccessWith already exists :)

tim-smart (changes requested):

packages/effect/src/Metric.ts
beezee: Thanks for pointing that out @tim-smart ! I think that the `A extends In2` on trackSuccessWith is still causing inference to fail. 

I'm on my phone, so I don't know if it's just a playground performance issue in my browser, but I tested here: https://effect.website/play/#3080ee0cadf4

Would you be opposed if I repurpose this PR to change the type of trackSuccessWith to remove the `A extends` and match the overload I had added originally to mapInput here which handles inference correctly?

beezee: Fwiw I believe in this context, you want inference to pick up and fully determine In2. Is there an advantage to calling out explicitly that it is a super type of what should actually be inferred that I am missing?

tim-smart: Yes I think it could be

```ts
export const trackSuccessWith: {
  <Type, In, Out, A>(
    metric: Metric<Type, In, Out>,
    f: (value: Types.NoInfer<A>) => In
  ): <A, E, R>(self: Effect.Effect<A, E, R>) => Effect.Effect<A, E, R>
  <A, E, R, Type, In, Out>(
    self: Effect.Effect<A, E, R>,
    metric: Metric<Type, In, Out>,
    f: (value: Types.NoInfer<A>) => In
  ): Effect.Effect<A, E, R>
} = internal.trackSuccessWith
```

beezee: ty @tim-smart I made those changes and updated tests to confirm. Looks good!

beezee: Curious along similar lines, any reason not to replace `unknown` with `E` in `trackDefectWith` ?

```
declare const trackDefectWith: {
  <Type, In, Out>(
    metric: Metric<Type, In, Out>,
    f: (defect: unknown) => In
  ): <A, E, R>(self: Effect.Effect<A, E, R>) => Effect.Effect<A, E, R>
  <A, E, R, Type, In, Out>(
    self: Effect.Effect<A, E, R>,
    metric: Metric<Type, In, Out>,
    f: (defect: unknown) => In
  ): Effect.Effect<A, E, R>
}
```

tim-smart: Defects are not typed, so unknown makes sense there.

packages/effect/src/Metric.ts
```diff
@@ -677,14 +677,14 @@ export const trackSuccess: {
...
-    f: (value: In2) => In
-  ): <A extends In2, E, R>(self: Effect.Effect<A, E, R>) => Effect.Effect<A, E, R>
-  <A extends In2, E, R, Type, In, Out, In2>(
+    f: (value: Types.NoInfer<A>) => In
+  ): <E, R>(self: Effect.Effect<A, E, R>) => Effect.Effect<A, E, R>
+  <A, E, R, Type, In, Out>(
     self: Effect.Effect<A, E, R>,
     metric: Metric<Type, In, Out>,
-    f: (value: In2) => In
+    f: (value: A) => In
```
tim-smart:
```suggestion
    f: (value: Types.NoInfer<A>) => In
```

tim-smart (approved):

PR #5290, opened 2025-07-28
remove error case from `Command.isRunning` type signature

fubhy:
@tim-smart any reason why this should be fallible?


fubhy:
Closing as stale because the platform behavior question was never resolved and there has been no activity for more than 6 months. If this is still desired, please open a fresh PR with current evidence that all supported implementations are infallible.

PR #5291, opened 2025-07-28
feat: add support for Optimistic Concurrency via `versionColumn` option

patroza:


fubhy:
Thanks for the PR. Could you add tests covering `versionColumn` behavior and clarify the intended semantics, especially what should happen on a stale version, what column types are supported, and why the new version is always generated with `Uuid.v4()`?

tim-smart:

PR #5293, opened 2025-07-28
improve: cause error output

patroza:
When throwing an Error classically, extra fields of the error are printed, this is however not the case with effect pretty printing causes.

This PR tries to enhance the error info, when `renderErrorCause: true` is passed.


fubhy:

PR #5301, opened 2025-07-29
Fix infinite loop when ConfigProvider.fromJson is called with nested arrays

RedPhortex:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Fixes a bug where the program would get stuck in a while loop when ConfigProvider.fromJson was called with a nested array, due to configPathToString expecting KeyIndex components to always follow KeyName components.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/silent-ducks-wear.md
+++ b/.changeset/silent-ducks-wear.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-fix ConfigProvider.fromJson being called with nested arrays, by handling consecutive KeyIndex components in configPathToString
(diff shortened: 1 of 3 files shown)


fubhy:
> *This was generated by AI during triage.*

Closing this as stale since it has had no activity for over 6 months and has no active review path or maintainer owner. The underlying nested-array ConfigProvider case may still be worth fixing, but this PR is no longer active enough to keep open.

PR #5305, opened 2025-07-30
introduce ReadonlyTag as a Tag that can only be used to read a service from requirements

mikearnaldi:
Introduce ReadonlyTag as the covariant side of a tag, enables:

```ts
import type { Context } from "effect"

export class MyRequirement extends Effect.Service<MyRequirement>()(
  "MyRequirement",
  { succeed: () => 42 }
) {}

export class MyUseCase extends Effect.Service<MyUseCase>()("MyUseCase", {
  dependencies: [MyRequirement.Default],
  effect: Effect.gen(function* () {
    const requirement = yield* MyRequirement
    return Effect.fn("MyUseCase.execute")(function* () {
      return requirement()
    })
  })
}) {}

export function effectHandler<I, Args extends Array<any>, A, E, R>(
  service: Context.ReadonlyTag<I, (...args: Args) => Effect.Effect<A, E, R>>
) {
  return Effect.fn("effectHandler")(function* (...args: Args) {
    const execute = yield* service
    yield* execute(...args)
  })
}

export const program = effectHandler(MyUseCase)
```

Discord discussion: https://discord.com/channels/795981131316985866/795983589644304396/1400218469160714261

### DIFF
--- a/.changeset/little-singers-wash.md
+++ b/.changeset/little-singers-wash.md
@@ -1,5 +1,5 @@
 ---
-"effect": patch
+"effect": minor
 ---
 
 Introduce ReadonlyTag as the covariant side of a tag, enables:


.changeset/little-singers-wash.md
```diff
@@ -0,0 +1,36 @@
+---
+"effect": patch
```
tim-smart: minor release?

PR #5313, opened 2025-08-01
use approximate years when using Duration.format

ilovedesert001:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
use approximate years when using Duration.format

### DIFF
--- a/.changeset/whole-ants-kiss.md
+++ b/.changeset/whole-ants-kiss.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+use approximate years when using Duration.format


ilovedesert001:
>  at the cost of less precision

but is it possible to make full precision human-readable string (Months , Years.) ? 

you are right, with "≈" and cost of precision, maybe this change is too much for base effect library.

I currently use "Duration.format" in my local simple scripts. if I were to use it in more formal setting (like a frontend client that real user see) I would use a full date (duration) library that supports internationalization and other necessities.

for now, I think I put this in my project utils.

PR #5320, opened 2025-08-02
feat: add Scope support as context for RPC Middleware

patroza:
exposes the ~~fact that rpc requests run in a Scope~~, and that middlware can subscribe to that.

apparently this is only the case on the http server protocol, so we need to update the other protocols to support something similar.


fubhy:
Closing because this draft no longer applies cleanly to the target branch and has been inactive. Please open a fresh PR if the change is still relevant.

PR #5322, opened 2025-08-03
non-greedy matching for Schema.String in Schema.TemplateLiteralParser

beezee:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR changes the regex used for Schema.String in TemplateLiteralParser from greedy to non-greedy (`[\\s\\S]*` -> `[\\s\\S]*?`). Greedy matching can consume too much input and prevent following spans from matching. The non-greedy version allows downstream segments to parse correctly. 

Minimal repro is shown in this playground: https://effect.website/play#a54077a17e06

Note the minimal repro avoids NumberFromString (case exercised in the added test) to avoid unsupported span error when extracting the regex via TemplateLiteral.

## Related

This was discussed briefly here: https://discord.com/channels/795981131316985866/1344319931625246770

### DIFF
--- a/.changeset/wide-candles-wait.md
+++ b/.changeset/wide-candles-wait.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Use non-greedy matching for Schema.String in Schema.TemplateLiteralParser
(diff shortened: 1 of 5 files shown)


beezee:
@gcanti my mistake, I clearly missed some tests! Updating now...

gcanti:
@beezee no problem, great fix, I have already applied it in v4

PR #5331, opened 2025-08-04
align RPC middleware api

patroza:
Align RPC middleware api; only the `wrap` variant remains, and `provides` and `requires` are supported as second generic argument `<Self, Config>`

### DIFF
--- a/packages/rpc/src/RpcGroup.ts
+++ b/packages/rpc/src/RpcGroup.ts
@@ -95,6 +95,8 @@ export interface RpcGroup<in out R extends Rpc.Any> extends Pipeable {
   ): Layer.Layer<
     Rpc.ToHandler<R>,
     EX,
+    | (R extends Rpc.Rpc<infer _A, infer _B, infer _C, infer _D, infer _E, infer _F> ? _F :
+      never)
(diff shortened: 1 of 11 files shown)


packages/rpc/src/RpcGroup.ts
```diff
@@ -95,6 +95,8 @@ export interface RpcGroup<in out R extends Rpc.Any> extends Pipeable {
   ): Layer.Layer<
     Rpc.ToHandler<R>,
     EX,
+    | (R extends Rpc.Rpc<infer _A, infer _B, infer _C, infer _D, infer _E, infer _F> ? _F :
```
patroza, line 98: @tim-smart could you please have a look here and below, if I change the order here, or move the `_F` infer to `HandlersContext`, and I remove the `.middleware(SomethingElseElseMiddleware)` from `UsersRpcs`, it resolves to `never`:o

fubhy:
Closing as stale. This PR has had no activity since Aug 2025 and no active maintainer review path after the open API/type-inference question. If this RPC middleware API change is still wanted, please open a fresh PR against current main with the current design and tests.

PR #5355, opened 2025-08-08
`HttpServerResponse.redirect` defaults to 302 Found status

nounder:
`HttpServerResponse.redirect` defaults to a 301 status, which is risky:
- Search engines treat 301 as a permanent move and may rerank pages.
- Browsers cache 301s and stop requesting those URLs.

This behavior is why frameworks like Rails, Laravel, Django, and Phoenix use 302 instead—it’s safer and avoids surprises.

Since the platform package is still pre-1.0, we should switch the default to 302 to align with common practices and avoid unexpected behavior.

### DIFF
--- a/.changeset/some-bottles-own.md
+++ b/.changeset/some-bottles-own.md
@@ -1,5 +1,5 @@
 ---
-"@effect/platform": minor
+"@effect/platform": patch
 ---
 
 Use `302 Found` status in `HttpServerResponse.redirect` as default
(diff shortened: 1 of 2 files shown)


nounder:
Here's one case when 301 browser caching my cause problem:

CleanShot 2025-08-08 at 20 44 45@2x

Because browser caches 301, it never executes server logic which leads to inconstant cookie state across server-client.

.changeset/some-bottles-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/platform": minor
```
tim-smart:
```suggestion
"@effect/platform": patch
```

tim-smart (approved):

Issue #4148, opened 2024-12-16
CatchIf - predicate that is always true or false does not Exclude<...> error types correctly


akomm:
### What version of Effect is running?

3.11.6

### What steps can reproduce the bug?

Code is based on [Effect-TS Example for catchif](https://effect.website/docs/error-management/expected-errors/#catchif)

[Open the code below in TS-Playground](https://www.typescriptlang.org/play/?ts=5.8.0-dev.20241212#code/JYWwDg9gTgLgBAbzgUQGaoKYGMYBo4BKAhgHYAmEIcAvnKlJXAEQbrYxMBQnWANkQGcBcABIwYYZFAZREnOHCgYiFErwCecAPowiAczgBeZmIlSZXatz6DhANSK9gZIjGAQS56HIVKVHjW1dA2MmBycXNw8vKEtuAHp4hWS4QBhSQABSTJQ2HAAeARgoYBI9fFNJaW8AHzhw51d3T0qofBIMADcMKAA+TkSUhUAe0h4PArgwBj0oIipjNEwcADo9DBIAClQAVxIcRoAqODWASh84LFH4EgBGIzh1YAxeMgPickpFtoAPGHkzi7gSAAmW73R7PQikCggD4Yb6-YCoQ7XOC5OAABkWAFYTghfgpQU8DvN2ItUERgLw1m0AO6icQVGTHI6-KwKBFI4GojHY074h6E7ILGCk8mUmm1Rz1KJNRlHZkKVmKDAwTZQEjMASUDBKgSbXgcTjUZl9JIDDJZAQACwgerIcAARtriXkCkUSq0Ol0PZ0eg7NvBgMJnTB8oViqU6WZmnAanVIo0Yt6ur1+gNhucSGMlOcfRg7cYJhApjNFmBgGAMGsTSkAHoAfl+wcWWFcWEtAElUFWUmsujIToZunA+9BFjp9EZDKFyjEmDGaiOoGPgpPQnGGtFmkxcHjDgOh03dVgsBg82smARsBAfeG6AwqDOt-K4MyjkA)
```ts
import { Effect, Random } from "effect"

class HttpError {
  readonly _tag = "HttpError"
}

class ValidationError {
  readonly _tag = "ValidationError"
}

//      ┌─── Effect<string, HttpError | ValidationError, never>
//      ▼
const program = Effect.gen(function* () {
  const n1 = yield* Random.next
  const n2 = yield* Random.next
  if (n1 < 0.5) {
[...]

PR #5360, opened 2025-08-09
Made Schema.brand not any-fy piped schemas

nikelborm:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When `Schema.brand` was extracted into a variable, other schemas, piped through that variable, resulted in `Schema.Any` type.

### Before

```ts
const UserIdBrandSchema = Schema.brand("UserId")
const UserIdSchema = pipe(Schema.Number, UserIdBrandSchema)
//    ^? Schema.brand<Schema.Schema.Any, "UserId">
```

### After

```ts
const UserIdBrandSchema = Schema.brand("UserId")
const UserIdSchema = pipe(Schema.Number, UserIdBrandSchema)
//    ^? Schema.brand<typeof Schema.Number, "UserId">
```

### DIFF
--- a/.changeset/flat-buttons-design.md
+++ b/.changeset/flat-buttons-design.md
@@ -1,0 +1,21 @@
+---
+"effect": patch
+---
+
+Made Schema.brand keep type of schema being piped into it even if brand schema was extracted into a variable. Type was previously Schema.any
+
+Before:
+
+```ts
+const UserIdBrandSchema = Schema.brand("UserId")
+const UserIdSchema = pipe(Schema.Number, UserIdBrandSchema)
+//    ^? Schema.brand<Schema.Schema.Any, "UserId">
+```
+
+After:
+
+```ts
+const UserIdBrandSchema = Schema.brand("UserId")
+const UserIdSchema = pipe(Schema.Number, UserIdBrandSchema)
+//    ^? Schema.brand<typeof Schema.Number, "UserId">
+```
--- a/packages/effect/dtslint/Schema/Brand.tst.ts
+++ b/packages/effect/dtslint/Schema/Brand.tst.ts
@@ -25,4 +25,17 @@ describe("SchemaBrand", () => {
       })
     )
   })
+
+  it("should not override types with any when extracted into a constant", () => {
+    const UserIdBrandSchema = Schema.brand("UserId")
+
+    const UserIdSchema = pipe(Schema.Number, UserIdBrandSchema)
+
+    type IUserId = Schema.Schema.Type<typeof UserIdSchema>
+
+    type IsAny<T> = 0 extends 1 & T ? true : false
+
+    expect<IsAny<IUserId>>().type.toBe<false>()
+    expect<IUserId>().type.toBeAssignableTo<number>()
+  })
 })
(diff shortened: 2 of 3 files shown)


nikelborm:
# 🧇

nikelborm:
Hello @gcanti!

May I please ask you for a few minutes of your time dedicated to the review of this tiny PR? I would greatly appreciate your attention and time, if you have the capacity and available resources for this effort.

If you would like me to add it to `effect-smol` too, I can do it too

gcanti:
@nikelborm actually I have been thinking about this for a few days, since it would be useful (for all APIs, not just for `brand`).

As for the implementation in this PR, it doesn't play well with annotations:


```ts
import { pipe, Schema } from "effect"

const UserIdBrandSchema = Schema.brand("UserId", {
  examples: ["a"] // UserIdBrandSchema should only apply to strings now
})

const UserIdSchema = pipe(
  Schema.Number, // no error
  UserIdBrandSchema
)

const UserIdSchema2 = pipe(
  Schema.Number, // no error
  Schema.brand("UserId", {
    examples: ["a"]
  })
)
```

nikelborm:
> for all APIs, not just for brand

I noticed that too. The same thing is in `Schema.compose` too.

```ts
const addAsdField = Schema.compose(Schema.Struct({
  asd: Schema.String
}))

// addAsdField: (from: Schema.Schema.Any) => Schema.transform<Schema.Schema.Any, Schema.Struct<{
//     asd: typeof Schema.String;
// }>>

// Has hardcoded Schema.Schema.Any here


const TestSchema = Schema.asSchema(pipe(
  Schema.Struct({sdf: Schema.Number}),
  addAsdField
))

// TestSchema: Schema.Schema<
//   { readonly asd: string; },
//   any,
//   unknown
// >
```

So it's generally applicable that there are 2 kinds of ways of writing generics. I noticed that some generics provide better inference in pipes, and some outside of pipes, and the user has to either choose, or there should be some clever thing that automatically determines if it is inside the pipe or outside. That's what I'm talking about: https://github.com/Effect-TS/effect/issues/4148#issuecomment-3189090610

So in the end, could you please help me better understand how I interpret your response?

I have a few interpretations in mind:

1. This PR won't fit well into the project just by itself, even if I resolve the problem with annotations, and you want me to find a way to fix it generally for every place where this kind of bug occurs.
2. Same as above, but you want to fix the problem yourself, because you know the codebase better
3. I have to fix the problem with annotations, because that's the only blocker for now, preventing the PR from being merged
4. You want to start a discussion on how this problem could be solved generally
5. You want me to close the PR and wait because it may/will be fixed by effect-smol

nikelborm:
Also, I figured there's even yet another thing that could be done:

Writing tests to catch that case.

```ts
import { pipe, Schema } from "effect"

const UserIdBrandSchema = Schema.brand("UserId", {
  examples: ["a"] // UserIdBrandSchema should only apply to strings now
})

const UserIdSchema = pipe(
  Schema.Number, // no error
  UserIdBrandSchema
)

const UserIdSchema2 = pipe(
  Schema.Number, // no error
  Schema.brand("UserId", {
    examples: ["a"]
  })
)
```

Because the CI says it's all clear, and it wasn't caught.

fubhy:
Closing because this patch currently fails type checks and has an unresolved maintainer-identified type-safety regression around annotations. The underlying inference issue may still be worth solving, but it needs a fresh PR that preserves the baseline annotation error behavior.

PR #5377, opened 2025-08-11
feat: added allowed tools

timurrakhimzhan:
## Type

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR allows to indicate a subset of a toolkit, that is going to be passed to LLM.


packages/ai/ai/src/AiLanguageModel.ts
```diff
@@ -66,7 +66,8 @@ export interface IdentifiedSchema<A, I, R> extends Schema.Schema<A, I, R> {
  * @category Models
  */
 export type ToolChoice<Tool extends AiTool.Any> = "auto" | "none" | "required" | {
-  readonly tool: Tool["name"]
+  readonly tool?: Tool["name"]
+  readonly oneOf?: Array<Tool["name"]>
 }
```
IMax153: This should become a separate item in the `ToolChoice` union. This is because the `{ tool: Tool["name"] }` member is meant for selecting one specific tool, whereas the `oneOf` member is for ensuring only a subset of tools are allowed.

```ts
export type ToolChoice<Tool extends AiTool.Any> = "auto" | "none" | "required" | {
  readonly tool: Tool["name"]
} | {
  readonly oneOf: Array<Tool["name"]>
}
```

packages/ai/ai/src/AiLanguageModel.ts
```diff
@@ -391,7 +392,11 @@ export const make: (
           }
           modelOptions.toolChoice = toolChoice
           const actualToolkit = Effect.isEffect(toolkit) ? yield* toolkit : toolkit
+          const hasUnallowedTools = typeof toolChoice === "object" && Array.isArray(toolChoice.oneOf)
```
IMax153: This check can probably just be `typeof toolChoice === "object" && "oneOf" in toolChoice`.

packages/ai/ai/src/AiLanguageModel.ts
timurrakhimzhan: I have made `ToolChoice` a union.
Regarding the provider integrations, as far as I checked the code, it must not affect provider integrations at all.

timurrakhimzhan:
> I'm not 100% sure but you should check / test if adding the additional member to `ToolChoice` necessitates any changes in the provider integrations.

packages/ai/ai/src/AiLanguageModel.ts
timurrakhimzhan: Great change, thank you

timurrakhimzhan:
@IMax153 I had to add `mode` property alongside with `oneOf`, so that user can switch between `auto` and `required` mode.

PR #5397, opened 2025-08-18
Remove dependency on `@opentelemetry/semantic-conventions`

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

The OpenTelemetry packages have issues with ESM builds. This PR removes the `@opentelemetry/semantic-conventions` dependency from all packages except the `@effect/opentelemetry` integration package in favor of statically declared attribute names / values.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/platform/src/internal/httpClient.ts
+++ b/packages/platform/src/internal/httpClient.ts
@@ -1,4 +1,3 @@
-import * as OtelSemConv from "@opentelemetry/semantic-conventions"
 import * as Cause from "effect/Cause"
 import * as Context from "effect/Context"
 import * as Effect from "effect/Effect"
@@ -27,6 +26,17 @@ import * as TraceContext from "../HttpTraceContext.js"
 import * as UrlParams from "../UrlParams.js"
 import * as internalRequest from "./httpClientRequest.js"
 import * as internalResponse from "./httpClientResponse.js"
+
+const ATTR_HTTP_REQUEST_HEADER = (key: string): string => `http.request.header.${key}`
+const ATTR_HTTP_REQUEST_METHOD = "http.request.method"
+const ATTR_HTTP_RESPONSE_HEADER = (key: string): string => `http.response.header.${key}`
+const ATTR_HTTP_RESPONSE_STATUS_CODE = "http.response.status_code"
+const ATTR_SERVER_ADDRESS = "server.address"
+const ATTR_SERVER_PORT = "server.port"
+const ATTR_URL_FULL = "url.full"
+const ATTR_URL_PATH = "url.path"
+const ATTR_URL_SCHEME = "url.scheme"
+const ATTR_URL_QUERY = "url.query"
 
 /** @internal */
 export const TypeId: Client.TypeId = Symbol.for(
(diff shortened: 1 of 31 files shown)


packages/platform/src/internal/httpClient.ts
```diff
@@ -28,6 +27,17 @@ import * as UrlParams from "../UrlParams.js"
...
+const ATTR_HTTP_REQUEST_HEADER = (key: string): string => `http.request.header.${key}`
+const ATTR_HTTP_REQUEST_METHOD = "http.request.method"
+const ATTR_HTTP_RESPONSE_HEADER = (key: string): string => `http.response.header.${key}`
+const ATTR_HTTP_RESPONSE_STATUS_CODE = "http.response.status_code"
+const ATTR_SERVER_ADDRESS = "server.address"
+const ATTR_SERVER_PORT = "server.port"
+const ATTR_URL_FULL = "url.full"
+const ATTR_URL_PATH = "url.path"
+const ATTR_URL_SCHEME = "url.scheme"
+const ATTR_URL_QUERY = "url.query"
```
colelawrence, line 39: Question: I'm wondering if it would be appropriate to include an inline comments above each of those const declaration groups that links to the relevant documentation online where someone could find out what all semantic keys are derived from.

IMax153: We are debating how to best handle semantic OTEL attributes. For now, this will unblock the bugs we are facing with ESM builds but are looking at a larger strategy. Once we have one in place we will document accordingly. I avoided doing it here to prevent drift.

IMax153:
> In the end, as a thing to consider, maybe it will be a good idea to create an `@effect/otel-conventions` package

We were actually just discussing something like this internally, but instead of creating a new package, keeping everything in our existing opentelemetry package given the OTEL deps are peer dependencies there. May make sense as a separate package but we really haven't come to a conclusion yet. 

The OTEL folks look like they autogenerate a lot of the semantic conventions code from the spec so it's probably not too difficult to maintain.  

@nikelborm if you are interested maybe you could spec something out in the existing OTEL package and we can separate it out if we decide it make sense?

nikelborm:
Could you please elaborate on what kinds of ESM problems `@opentelemetry/semantic-conventions` has?

> spec something out in the existing OTEL package

By that, do you mean in the upstream [repo of `@opentelemetry/semantic-conventions` package](https://github.com/open-telemetry/opentelemetry-js)?

IMax153:
> Could you please elaborate on what kinds of ESM problems `@opentelemetry/semantic-conventions` has?

Our users have consistently run into a variety of issues when building due to OTEL dependencies coming from our packages due to the way the libraries are published. It's highly specific to build environment. We would rather get rid of this friction point moving forward. 

I'll let @mikearnaldi provide more context on the concrete issues as he has dealt with more of this than I have.

nikelborm:
I found a few PRs that promise to fix the situation with ESM.

- https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2954
- https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2955

And they seem to be quite active. Will you be willing to reconsider removing dependency on `@opentelemetry/semantic-conventions` (or otel-family packages in general) after they are merged?

IMax153:
Yea I was referring to our `@effect/opentelemetry` package.

PR #5400, opened 2025-08-18
CLI: Support default selection in Prompt.multiSelect

kitlangton:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds per-choice default selection to `Prompt.multiSelect` by honoring `selected: true` on `SelectChoice`. The existing behavior remains unchanged for choices without `selected`, and for single select prompts. Meta actions (Select All/None/Inverse) operate from the seeded state. Includes a unit test that verifies preselected values are returned when submitting immediately.

### DIFF
--- a/.changeset/cli-multiselect-selected-defaults.md
+++ b/.changeset/cli-multiselect-selected-defaults.md
@@ -1,0 +1,11 @@
+---
+"@effect/cli": patch
+---
+
+cli: multiSelect supports per-choice default selection via `selected: true`; single select honors one default
+
+- Add `selected?: boolean` to `Prompt.SelectChoice`
+- Seed initial selection in `Prompt.multiSelect` from choices marked `selected: true`
+- Allow `Prompt.select` to honor a single `selected: true` choice and throw an error if multiple defaults are provided
+- Add tests covering both behaviors
+
--- a/packages/cli/src/internal/prompt/select.ts
+++ b/packages/cli/src/internal/prompt/select.ts
@@ -225,7 +225,22 @@ export const select = <const A>(options: Prompt.Prompt.SelectOptions<A>): Prompt
     maxPerPage: 10,
     ...options
   }
-  return InternalPrompt.custom(0, {
+  // Validate and seed initial index from any choice marked selected: true
+  let initialIndex = 0
+  let seenSelected = -1
+  for (let i = 0; i < opts.choices.length; i++) {
+    const choice = opts.choices[i] as Prompt.Prompt.SelectChoice<A>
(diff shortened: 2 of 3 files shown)


packages/cli/src/Prompt.ts
```diff
@@ -416,6 +416,10 @@ export declare namespace Prompt {
      * Whether or not this select option is disabled.
      */
     readonly disabled?: boolean
+    /**
+     * Whether this option should be selected by default (only used by MultiSelect).
+     */
+    readonly selected?: boolean
```
IMax153, line 422: The logic looks good, but I'm not a fan of exposing the `selected` prop to the non-multi-select prompt if it's not going to be used.

In theory, we could allow this in the select prompt too, and blow up if multiple options are selected.

The other option would be to specialize the choices of a multi-select to ones that have the `selected` prop. 

```ts
interface MultiSelectChoice<A> extends SelectChoice<A> {
  readonly selected?: boolean | undefined
}

interface MultiSelectOptions<A> {
  // ... other existing props...
  readonly choices: ReadonlyArray<MultiSelectChoice<A>>
}
```


Then the multi select signature could be:
```ts
/** @internal */
export const multiSelect = <const A>(
  options: Omit<Prompt.Prompt.SelectOptions<A>, "choices"> & Prompt.Prompt.MultiSelectOptions<A>
): Prompt.Prompt<Array<A>>
```

kitlangton: I think it would make sense to allow it for non-multi-selects as well. We could either blow up or simply select the first selected option. What do you prefer?

And if we blow up, just throw an error, right? I've seen that in other CLI constructors if you pass it some no-no values.

IMax153: I can't remember if there's a semantic way to "blow up" in `Prompt` (i.e. we could fail or die but I can't remember the structure of the API), but if there is then I'd prefer the blow up option vs. silently accepting an erroneous config.

kitlangton: added the single-select support and the blowing up.

.changeset/cli-multiselect-selected-defaults.md
```diff
@@ -0,0 +1,11 @@
+---
+"@effect/cli": minor
```
IMax153: Since this is a forward-compatible change we can make it a patch.

kitlangton: Fixed :)

PR #5404, opened 2025-08-19
CLI: add Prompt.spinner

kitlangton:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added a spinner to `Prompt` that animates a spinner while an effect is running.

```ts
const user = yield* Prompt.spinner("Fetching user…", fetchUser)
```

```ts
⠴ Fetching user…
```

The message is updated when the effect completes successfully or fails.

```ts
✔ Loaded Ada (ID: 42)
✖ Processing failed: Network timeout
```

### DIFF
--- a/packages/cli/examples/spinner.ts
+++ b/packages/cli/examples/spinner.ts
@@ -1,0 +1,75 @@
+import * as Prompt from "@effect/cli/Prompt"
+import { NodeRuntime, NodeTerminal } from "@effect/platform-node"
+import { Console, Effect } from "effect"
+
+// Demonstration of success, failure, and custom final messages
+const program = Effect.gen(function*() {
+  // Success case with custom success message
+  const user = yield* Prompt.spinner(
+    Effect.sleep("1200 millis").pipe(Effect.as({ id: 42, name: "Ada" })),
+    {
+      message: "Fetching user…",
+      onSuccess: (user: { id: number; name: string }) => `Loaded ${user.name} (ID: ${user.id})`
+    }
+  )
+  yield* Console.log(`User: ${JSON.stringify(user)}`)
+
+  // Failure case with custom error message and proper error handling
+  yield* Prompt.spinner(
+    Effect.sleep("800 millis").pipe(Effect.zipRight(Effect.fail(new Error("Network timeout")))),
+    {
+      message: "Processing data…",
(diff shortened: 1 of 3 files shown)


emilwidlund:
Love it! 🎉

packages/cli/examples/spinner.ts
```diff
@@ -0,0 +1,73 @@
+import * as Prompt from "@effect/cli/Prompt"
+import { NodeRuntime, NodeTerminal } from "@effect/platform-node"
+import { Console, Effect } from "effect"
+
+// Demonstration of success, failure, and custom final messages
+const program = Effect.gen(function*() {
```
IMax153, line 6: Looks like there are a bunch of compilation errors in the example here.

packages/cli/src/internal/prompt/spinner.ts
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * A spinner that renders while `effect` runs and prints ✔/✖ on completion.
+ *
+ * @internal
+ */
+export const spinner = <A, E, R>(
+  message: string,
+  effect: Effect.Effect<A, E, R>,
+  options?: SpinnerOptions<E, A>
+): Effect.Effect<A, E, R | Terminal.Terminal> =>
```
IMax153: Let's make this signature dual - I can help with that if you're unsure, sometimes getting the signature right with an optional third param can be tricky.

packages/cli/src/internal/prompt/spinner.ts
```diff
@@ -0,0 +1,164 @@
...
+  Effect.acquireUseRelease(
+    // acquire
+    Effect.gen(function*() {
+      const terminal = yield* Terminal.Terminal
+
+      // Hide cursor while active
+      yield* Effect.orDie(terminal.display(CURSOR_HIDE))
+
+      const indexRef = yield* Ref.make(0)
+      const exitRef = yield* Ref.make<Option.Option<Exit.Exit<A, E>>>(Option.none())
```
IMax153: Since these are only ever accessed / modified within the generator, they can be plain mutable variables.


```suggestion
      const index = 0
      const exit: Exit.Exit<A, E> | undefined  = undefined
```

IMax153 (changes requested):

PR #5420, opened 2025-08-21
chore(test): remove `assertType`

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR removes the `assertType()` utility which is used in few paces in run time tests. Seems like the utility is redundant, because type-level tests can cover all the cases and would be tested agains different TypeScript versions.

Before removing the assertions I was checked carefully if they exist in type tests already. Two cases were missing. So I tried to copy the code and adapt it for type testing.

### DIFF
--- a/packages/effect/dtslint/Option.tst.ts
+++ b/packages/effect/dtslint/Option.tst.ts
@@ -6,21 +6,21 @@ declare const string: Option.Option<string>
 declare const numberOrString: Option.Option<string | number>
 
 declare const primitiveNumber: number
-declare const primitiveNumerOrString: string | number
+declare const primitiveNumberOrString: string | number
 declare const predicateNumbersOrStrings: Predicate.Predicate<number | string>
 
 describe("Option", () => {
   it("liftPredicate", () => {
     expect(
-      Option.liftPredicate(primitiveNumerOrString, Predicate.isString)
+      Option.liftPredicate(primitiveNumberOrString, Predicate.isString)
     ).type.toBe<Option.Option<string>>()
     expect(
-      pipe(primitiveNumerOrString, Option.liftPredicate(Predicate.isString))
+      pipe(primitiveNumberOrString, Option.liftPredicate(Predicate.isString))
     ).type.toBe<Option.Option<string>>()
 
     expect(
       Option.liftPredicate(
-        primitiveNumerOrString,
+        primitiveNumberOrString,
         (n): n is number => {
           expect(n).type.toBe<string | number>()
           return typeof n === "number"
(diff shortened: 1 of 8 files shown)


packages/effect/test/Match.test.ts
```diff
@@ -571,16 +570,12 @@ describe("Match", () => {
   })
 
   it("instanceOf", () => {
-    // These type-level tests cannot be moved to a dtslint test because
-    // of the difference in the way the type is retrieved in ts 5.7
```
mrazauskas, line 575: This is now possible with `// @tstyche if`.

packages/effect/test/Effect/error-handling.test.ts
```diff
@@ -269,7 +268,6 @@ describe("Effect", () => {
         Effect.exit
       )
       deepStrictEqual(result, Exit.fail({ _tag: "ErrorB" as const }))
-      satisfies<true>(assertType<Exit.Exit<ErrorA, ErrorB>>()(result))
```
mrazauskas, line 272: Moved to `Effect.tst.ts`.

packages/effect/test/Effect/structural.test.ts
```diff
@@ -4,53 +4,41 @@ import * as Effect from "effect/Effect"
...
 import * as Option from "effect/Option"
-import { assertType, satisfies } from "../utils/types.js"
 
 describe("Effect", () => {
   describe("all", () => {
     it.effect("should work with one array argument", () =>
       Effect.gen(function*() {
         const res = yield* (Effect.all([Effect.succeed(0), Effect.succeed(1)]))
         deepStrictEqual(res, [0, 1])
-        satisfies<true>(assertType<[number, number]>()(res))
```
mrazauskas, line 15: Seems like all assertions in this file were already moved to `Effect.tst.ts`.

packages/effect/test/Option.test.ts
```diff
@@ -473,16 +473,12 @@ describe("Option", () => {
   })
 
   it("all/ tuple", () => {
-    assertType<Option.Option<[number, string]>>(Option.all([Option.some(1), Option.some("hello")]))
```
mrazauskas, line 476: Seems like `Option.tst.ts` already has assertions equivalent to the ones removed.

packages/effect/dtslint/Option.tst.ts
```diff
@@ -6,21 +6,21 @@ declare const string: Option.Option<string>
 declare const numberOrString: Option.Option<string | number>
 
 declare const primitiveNumber: number
-declare const primitiveNumerOrString: string | number
+declare const primitiveNumberOrString: string | number
```
mrazauskas, line 9: Just a typo.

gcanti:
Thanks @mrazauskas !

p.s.

Would you mind taking a look at the type-level tests of Effect v4? I'm not sure what happened in the meantime, but something must have changed in how we configured the monorepo, and meanwhile someone removed the task that is now failing: https://github.com/Effect-TS/effect-smol/pull/424  
(**only if you have time, no rush at all**)

PR #5432, opened 2025-08-26
feat: expose raw sqlite client in `sql-sqlite-bun` and `sql-sqlite-node`

Leka74:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR is intended to expose the internal database client for `sql-sqlite-bun` and `sql-sqlite-node` packages. My primary reason for doing this was the need to feed the `better-auth` library with the raw database client, which currently is not available.

I think this can be afterwards added to the other sql packages.

Usage:
```ts
const sql = yield* SqliteClient.SqliteClient
const rawClient = yield* sql.rawClient
yield* Effect.sync(() => rawClient.exec("CREATE TABLE test (id INTEGER PRIMARY KEY, name TEXT)"))
yield* Effect.sync(() => rawClient.exec("INSERT INTO test (name) VALUES ('hello')"))
const rows = yield* Effect.try(() => rawClient.prepare("SELECT * FROM test").all())
assert.deepStrictEqual(rows, [{ id: 1, name: "hello" }])
```
Or in my use-case:
```ts
const make = Effect.gen(function*() {
  const sql = yield* SqliteClient.SqliteClient;
  const rawClient = yield* sql.rawClient;

  const auth = yield* Effect.try(() =>
    betterAuth({
      database: rawClient,
      emailAndPassword: {
        enabled: true,
      },
    }),
  );

  return BetterAuthClient.of(auth);
});
```

## Related

- Related Issue N/A
- Closes N/A

### DIFF
--- a/.changeset/crazy-icons-rush.md
+++ b/.changeset/crazy-icons-rush.md
@@ -1,0 +1,36 @@
+---
+"@effect/sql-sqlite-node": minor
+"@effect/sql-sqlite-bun": minor
+---
+
+## Raw Database Client Access
+
+Both `@effect/sql-sqlite-node` and `@effect/sql-sqlite-bun` now expose the underlying database client through a new `rawClient` property. This provides direct access to the native database APIs when needed for advanced operations.
+
+**Example**
+
+```ts
+import { SqliteClient } from "@effect/sql-sqlite-bun"
+import { Effect } from "effect"
+
+const program = Effect.gen(function* () {
+  const client = yield* SqliteClient.make({ filename: "test.db" })
+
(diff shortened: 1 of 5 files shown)


fubhy:
Closing due to inactivity after maintainer feedback. The proposed public API still has unresolved design concerns around transaction safety and coupling to the underlying sqlite client. If this is still needed, please open a fresh PR that addresses the feedback, likely around accepting/providing a raw client without bypassing Effect SQL's transaction safeguards.

PR #5443, opened 2025-08-31
feat: `requires` field for dependencies in HttpApiMiddleware

Leka74:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds `requires` support to `HttpApiMiddleware.Tag` to allow middlewares to depend on other middleware outputs.

Details:

- New option: `requires?: Context.Tag | ReadonlyArray<Context.Tag>` when defining a middleware tag.

Example:

```
export class AdminUser extends HttpApiMiddleware.Tag<AdminUser>()("Http/Admin", {
  failure: HttpApiError.Forbidden,
  requires: CurrentUser // or: [CurrentUser, Session]
}) {}

// Inside the middleware implementation you can now safely use the required services
yield* CurrentUser
```

## Related

- Related Issue: N/A
- Closes: N/A
- Discord conversation: https://discord.com/channels/795981131316985866/1411695023128317973

### DIFF
--- a/.changeset/slimy-points-relate.md
+++ b/.changeset/slimy-points-relate.md
@@ -1,0 +1,21 @@
+---
+"@effect/platform": minor
+---
+
+Add `requires` support to `HttpApiMiddleware.Tag` to allow middlewares to depend on other middleware outputs.
+
+Details
+
+- New option: `requires?: Context.Tag | ReadonlyArray<Context.Tag>` when defining a middleware tag.
+
+Example
+
+```
+export class AdminUser extends HttpApiMiddleware.Tag<AdminUser>()("Http/Admin", {
+  failure: HttpApiError.Forbidden,
+  requires: CurrentUser // or: [CurrentUser, Session]
+}) {}
+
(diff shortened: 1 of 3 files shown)


tim-smart:
HttpApiMiddleware will have a type-level requires option: https://github.com/Effect-TS/effect-smol/blob/main/packages/effect/src/unstable/httpapi/HttpApiMiddleware.ts#L173

It could be back-ported at some stage.

patroza:
> It could be back-ported at some stage.

better soon than later. guess we can expand https://github.com/Effect-TS/effect/pull/5331 to http api

fubhy:
Closing this due to inactivity and because the current patch needs more work before it is reviewable. In particular, the middleware threading change appears to be able to run provider middleware effects multiple times, and the PR does not include tests covering dependency ordering. A fresh PR with tests and an implementation aligned with the current HttpApiMiddleware design would be welcome.

PR #5447, opened 2025-09-01
Simplify raceAll implementation

mikearnaldi:


packages/effect/src/internal/fiberRuntime.ts
```diff
@@ -2631,111 +2630,61 @@ export const raceAll: <Eff extends Effect.Effect<any, any, any>>(
...
+        for (const fiber of fibers) {
+          fiber.addObserver((exit) => {
+            if (exit._tag === "Success") {
+              winner = fiber
+              interruptAll()
+              resume(exit)
+            } else {
+              failed++
+              if (failed === fibers.length) {
+                resume(exit)
```
tim-smart, line 2664: Should this be a Cause.Parallel?

tim-smart: You can also remove the mask and do everything in a single observer: https://github.com/Effect-TS/effect/pull/5449

PR #5460, opened 2025-09-03
prepared‑statement cache to Bun sqlite client

subtleGradient:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Introduces a prepared‑statement cache to @effect/sql-sqlite-bun and invalidates it after loadExtension. This aligns behavior with the Node client and ensures that extension‑dependent queries re‑prepare with the updated function set and query planning after an extension is loaded.


tim-smart:
bun already prepares statements when you use db.query, so we don't need our own cache.

PR #5468, opened 2025-09-05
Add support for Drizzle relational queries

aduarterengifo:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds support for Drizzle relational queries by propagating the DrizzleConfig generic.

Before:
```ts
const schema = {
    account: pgTable("account", {
        id: integer()
    })
}

export const PgLive = PgClient.layerConfig({
	url: Config.redacted(""),
})

export const DrizzeLive = layerWithConfig({
	casing: "snake_case",
	schema,
}).pipe(Layer.provide(PgLive))
```

type error: 
```ts
The expected type comes from property 'schema' which is declared here on type 'DrizzleConfig<Record<string, never>>'
```

After:
```ts
const schema = {
    account: pgTable("account", {
        id: integer()
    })
}

export const PgLive = PgClient.layerConfig({
	url: Config.redacted(""),
})

export const DrizzeLive = layerWithConfigGeneric({
	casing: "snake_case",
	schema,
}).pipe(Layer.provide(PgLive))

export const DBLive = Layer.mergeAll(
	PgLive,
	DrizzeLive,
)

export const DrizzleInstance = PgDrizzleFactory<typeof schema>()

const eff = Effect.gen(function* () {
		const drizle = yield* DrizzleInstance
        // before account wouldn't be inferred.
		const result = yield* drizle.query.account.findFirst()
}).pipe(Effect.provide(DBLive))
```

### DIFF
--- a/.changeset/sixty-wasps-fold.md
+++ b/.changeset/sixty-wasps-fold.md
@@ -1,0 +1,5 @@
+---
+"@effect/sql-drizzle": minor
+---
+
+Adds a parallel API that preserves/propagates DrizzleConfig's generic through tag and service creation. This is necessary to be able to use Drizzle's relational queries.
--- a/packages/sql-drizzle/src/Pg.ts
+++ b/packages/sql-drizzle/src/Pg.ts
@@ -47,6 +47,72 @@ export class PgDrizzle extends Context.Tag("@effect/sql-drizzle/Pg")<
 >() {}
 
 /**
+ *
+ * @example
+ * ```ts
+ * const schema = {
+ *     account: pgTable("account", {
+ *         id: integer()
+ *     })
+ * }
+ * 
+ * export const PgLive = PgClient.layerConfig({
+ * 	url: Config.redacted(""),
+ * })
+ * 
+ * export const DrizzeLive = layerWithConfigGeneric({
+ * 	casing: "snake_case",
+ * 	schema,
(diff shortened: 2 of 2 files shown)


mikearnaldi:
This version is not really type safe as it requires manually typing the factory, also it would fail when multiple instances are provided. Generally the solution of having a generic layer has been discarded in the past. The pattern of supporting a typed schema is already supported as shown in the test:

```ts
import * as Pg from "@effect/sql-drizzle/Pg"
import { PgClient } from "@effect/sql-pg"
import * as D from "drizzle-orm/pg-core"
import { Config, Effect, Layer } from "effect"

export const users = D.pgTable("users", {
  id: D.serial("id").primaryKey(),
  name: D.text("name").notNull(),
  snakeCase: D.text("snake_case").notNull()
})

export class ORM extends Effect.Service<ORM>()("ORM", { effect: Pg.make({ schema: { users } }) }) {
  static Client = this.Default.pipe(
    Layer.provideMerge(
      PgClient.layerConfig({
        url: Config.redacted("")
      })
    )
  )
}

export const program = Effect.gen(function*() {
  const orm = yield* ORM

  yield* orm.query.users.findMany()
})
```

So unless there is some missing functionality I think we can close this PR

Issue #5465, opened 2025-09-04
Customizing HttpApiDecodeError responses


Avaq:
### What is the problem this feature would solve?

I have an _existing_ API that I want to create API specs for using the `HttpApi` module, before gradually implementing it using the `HttpApiBuilder`, phasing out its old implementation.

The API specs can be fully customised in what each endpoint responds with by using `addSuccess` and `addError`, except for one thing: The `HttpApiDecodeError`.

This means I'm unable to create an API spec to fully cover my existing API.

I [posted](https://discord.com/channels/795981131316985866/1399430657217921156/1399430657217921156) this on Discord a while ago, but never got a response.

### What is the feature you are proposing to solve the problem?

A `setDecodeError` method that takes a `Schema<HttpApiDecodeError, T>` so I can customize how those are encoded:

```ts
const HttpApiDecodeErrorFromMyWeirdValidationError =
  Schema.transform(
    MyWeirdValidationError,
    HttpApiDecodeError.HttpApiDecodeError,
    { encode: (yours) => ({...}), decode: (mine) => ({...}) }
  )
  
export default HttpApi.make('myApi')
.add(MyApiGroup)
.setDecodeError(HttpApiDecodeErrorFromMyWeirdValidationError) // <--- override
.addError(MyWeirdInternalServerError, { status: 500 })
```

### What alternatives have you considered?

Hacking the response with a middleware, and hacking the OpenAPI JSON object: https://github.com/Effect-TS/effect/issues/5367#issuecomment-3254982322

tim-smart:
Automated message: Effect v3 is now under a feature freeze and is now only considering bug fixes.

If this issue is still relevant for v4, then feel free to reopen :)

Issue #4751, opened 2025-04-16
Allow HttpApiEndpoint to override HttpApiDecodeError response


dmeehan1968:
### What version of Effect is running?

3.14.5

### What steps can reproduce the bug?

Create an endpoint and set success and error.  In this case, I just want 400 Bad Request to be the empty response, not the HttpApiDecodeError.  I don't see a method that allows me to remove the default error response, so defining my own just adds to the HttpApiDecodeError response (as is consistent with adding multiple error types).

As an aside, given that this endpoint doesn't defined a request payload or path/query params, it seems odd that the api should define a decode error response at all.

```ts
const api = HttpApi.make('api')
  .add(HttpApiGroup.make('devices')
    .add(HttpApiEndpoint.get('listDevices')`/devices`
      .addSuccess(DeviceListSchema)
      .addError(HttpApiError.BadRequest)
      .addError(HttpApiError.Unauthorized)
      .addError(HttpApiError.InternalServerError)
    )
  )
```

```yaml
paths:
  /devices:
    get:
      tags:
        - devices
      operationId: devices.listDevices
      parameters: []
      security: []
      responses:
        '200':
          description: A list of Device entities with pagination information
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/DeviceList'
        '400':
          description: The request did not match the expected schema
          content:
            application/json:
              schema:
                anyOf:
                  - $ref: '#/components/schemas/HttpApiDecodeError'
                  - $ref: '#/components/schemas/BadRequest'
        '401':
          description: Unauthorized
        '500':
          description: InternalServerError
```

Note that in the yaml (generated by using `OpenApi.fromApi` and the converting to YAML) under the 400 response, it now shows a union of response schema.  HttpApiDecodeError does not match the desired response format for this type of error, and in the implementation I would want to catch this error and produce my own (empty in this example, but could be another response schema that suits my application).

In my case, I am going to use the OpenApi spec to generate AWS Api Gateway infrastructure, and so I don't need an Effect implementation, as I will inject the necessary AWS extensions into the OpenApi spec to add my Lambda handlers.  I will intend to publish the OpenApi spec as output by @effect/platform as a way of documenting the API for consumers.

### What is the expected behavior?

```yaml
        '400':
          description: The request did not match the expected schema
          content:
            application/json:
              schema:
                  $ref: '#/components/schemas/BadRequest'
```

### What do you see instead?

```yaml
        '400':
          description: The request did not match the expected schema
          content:
[...]

PR #5479, opened 2025-09-08
Update addError to give precedence to newly added schemas

Avaq:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`HttpApiSchema.UnionUnify` creates a union whose members are listed in the order of `[...firstArgUnionMembers, ...secondArgUnionMembers]`. When encoding a union, Schema will try each member in the order that they appear.

So the `addError` method on each of the HttpApi* classes, by passing the user-provided schema as a second argument, causes older schemas that encompass newer schemas to take precedence over the newer ones.

By flipping the argument order, users are now able to use `addError` with a schema that (partly) encompasses an older one, to override the encoding behaviour for values conforming to those schemas.

## Related

- Related to https://github.com/Effect-TS/effect/issues/4751 (but doesn't fully fix it, as the default HttpApiDecodeError schema is still part of every HttpApi)
- Closes https://github.com/Effect-TS/effect/issues/5465

### DIFF
--- a/packages/platform/src/HttpApi.ts
+++ b/packages/platform/src/HttpApi.ts
@@ -184,10 +184,10 @@ const Proto = {
       identifier: this.identifier,
       groups: this.groups,
       errorSchema: HttpApiSchema.UnionUnify(
-        this.errorSchema,
         annotations?.status
           ? schema.annotations(HttpApiSchema.annotations({ status: annotations.status }))
-          : schema
+          : schema,
+        this.errorSchema
       ),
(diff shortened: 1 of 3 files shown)


gcanti:
What about a `replaceError(f: (prev: Schema | undefined) => Schema | undefined)` API instead of `addError`? This would let the user decide what comes first, while system-added schemas would always be placed at the end.

PR #5480, opened 2025-09-08
feat: add `it.effect.concurrent` and `it.for`, and fix `it.each`

coffeeispower:
## Type

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adds `it.effect.concurrent` and `it.for`, and fixes `it.each`.

`it.each` was broken because it's meant to have the same signature as in jest where if the case is an array, it spreads the arguments to the arguments of the test, so I changed it to match vitest, and created a `it.for` function which has the old `it.each` signature and provides the `TestContext` to have better feature parity with vitest.

### DIFF
--- a/packages/vitest/src/internal/internal.ts
+++ b/packages/vitest/src/internal/internal.ts
@@ -31,7 +31,7 @@ function assignDefaults<T extends object, U extends object>(target: T, source: U
   return target as Types.MergeLeft<T, U>
 }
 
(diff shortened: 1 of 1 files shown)


coffeeispower:
I realized there can be race conditions with concurrent because everything is working on the same `it` instance

PR #5482, opened 2025-09-09
feat: implement `Symbol.dispose` and `Symbol.asyncDisposable` for ManagedRuntime

patroza:


tim-smart:
> Would it be sufficient to conditionally add these methods

I'm not sure how this can be made type safe

PR #5496, opened 2025-09-12
Update `Array.match` to return `NonEmptyArray` when given `Array`

### LINKED ISSUE #5497: `Array.match` should return `NonEmptyArray` when given `Array`

3.16.12

```ts
import { Array, pipe } from 'effect';

const array: Array<number> = [1, 2, 3];

pipe(
  array,
  Array.match({
    onEmpty: () => 1,
    onNonEmpty: (xs) => {
      // ❌ `xs` has type `readonly [number, ...number[]]
      // Expected `[number, ...number[]]`
      return 1;
    },
  }),
);

```

See above.

### What do you see instead?

See above.

### Additional information

I tried to fix it but ran into an issue that I'm not sure how to fix: https://github.com/Effect-TS/effect/pull/5496

OliverJAsh:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes https://github.com/Effect-TS/effect/issues/5497

### DIFF
--- a/packages/effect/dtslint/Array.tst.ts
+++ b/packages/effect/dtslint/Array.tst.ts
@@ -1,6 +1,7 @@
 import { Array, Effect, Either, Option, Order, Predicate } from "effect"
 import { hole, identity, pipe } from "effect/Function"
 import { describe, expect, it, when } from "tstyche"
+import { NonEmptyReadonlyArray } from "effect/Array"
 
 declare const nonEmptyReadonlyStrings: Array.NonEmptyReadonlyArray<string>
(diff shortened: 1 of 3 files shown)


cyberixae:
I believe `Array.match` gives you a read-only array on purpose when matching a mutable array. This guarantees that the `onNonEmpty` function doesn't accidentally modify the input. Effect usually only returns a mutable array when it has created a copy, so the original input would not be affected by the mutation.

fubhy:
Closing this inactive draft/WIP PR. Please open a fresh PR when the change is ready for review.

PR #5503, opened 2025-09-15
Rethrow original cause from makeRemoteCallback

mikearnaldi:


packages/sql-drizzle/test/Pg.test.ts
```diff
@@ -63,4 +65,30 @@ describe.sequential("Pg", () => {
     ), {
     timeout: 60000
   })
+
+  it.effect.only("fail properly for conflict issues", () => {
```
LaureRC:
```suggestion
  it.effect("fail properly for conflict issues", () => {
```

PR #5542, opened 2025-09-21
Fix: JSONB handling in @effect/sql-drizzle with PostgreSQL 

### LINKED ISSUE #5539: From Discord: Issue with JSONB Column Handling in @effect/sql-drizzle with PostgreSQL
# Summary
When inserting JSONB columns using @effect/sql-drizzle with PostgreSQL, JSON values are being serialized to strings instead of stored as JSON, causing issues with JSONB database functionality. Using raw client or SqlClient directly correctly stores JSON, but the PgDrizzle wrapper serializes JSON values to strings. Using SQL query with explicit cast (`sql`${payload}::jsonb``) works as a workaround. Attempts to enable `transformJson` in sql-pg did not resolve the issue. The Drizzle client setup was shared, and it was noted that `yield*` can be used with rawClient due to global prototype patching by drizzle. Key takeaways: JSONB handling differs between raw client and PgDrizzle, explicit JSONB casting is a current workaround, and internal transformations or settings like `transformJson` may not affect this behavior.

# Discord thread

https://discord.com/channels/795981131316985866/1418585968746692860

Nithishvb:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Closes: #5539

This PR includes Automatic JSON handling for SQL parameters: object/array values are now auto-serialized to JSON for execution, enabling seamless insertion and querying of JSON/JSONB columns without manual conversion or API changes.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/sql-drizzle/src/internal/patch.ts
+++ b/packages/sql-drizzle/src/internal/patch.ts
@@ -47,7 +47,18 @@ export const makeRemoteCallback = Effect.gen(function*() {
   const constructionRuntime = yield* Effect.runtime<never>()
   return (sql: string, params: Array<any>, method: "all" | "execute" | "get" | "values" | "run") => {
     const runPromise = Runtime.runPromise(currentRuntime ? currentRuntime : constructionRuntime)
-    const statement = client.unsafe(sql, params)
+
+    const finalParams = params.map((e) => {
+      try {
+        if (e) {
+          return JSON.parse(e)
+        }
+      } catch (err) {
+        return e
+      }
+    })
+
+    const statement = client.unsafe(sql, finalParams)
     if (method === "execute") {
       return runPromise(Effect.either(Effect.map(statement.raw, (header) => ({ rows: [header] })))).then((res) => {
         if (res._tag === "Left") {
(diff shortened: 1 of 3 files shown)


packages/sql-drizzle/src/internal/patch.ts
```diff
@@ -47,7 +47,20 @@ export const makeRemoteCallback = Effect.gen(function*() {
   const constructionRuntime = yield* Effect.runtime<never>()
   return (sql: string, params: Array<any>, method: "all" | "execute" | "get" | "values" | "run") => {
     const runPromise = Runtime.runPromise(currentRuntime ? currentRuntime : constructionRuntime)
-    const statement = client.unsafe(sql, params)
+
+    const finalParams = params.map((e) => {
+      if (e === null || e === undefined) return e
+      if (typeof e !== "string") return e
+      const s = e.trim()
+      if (!(s.startsWith("{") || s.startsWith("["))) return e
```
tim-smart, line 55: Will this work if you have a non-json param that starts with these characters?

Nithishvb: Hey @tim-smart , yes if a non-JSON param happens to start with { or [, it will attempt JSON.parse. Since parsing will fail, it falls into the catch block and simply returns the original string unchanged.

tim-smart: Right, I don't think this is robust enough to merge sorry. I'm not very familiar with drizzle, but you would need to somehow identify params that are only for json columns.

fubhy:
Closing this because the current implementation was rejected as not robust enough and has not been updated. A new PR that can identify JSON/JSONB column params specifically would still be welcome.

PR #5595, opened 2025-10-02
Fix node stat

code-alexander:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
`blksize` has type `Option<Size>`, but the `Size` constructor fails if the argument is `undefined`.
I ran into this issue with [Electron ASAR archives](https://www.electronjs.org/docs/latest/tutorial/asar-archives#fake-stat-information-of-fsstat).

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,10 @@
       "workerd"
     ],
     "onlyBuiltDependencies": [
-      "better-sqlite3"
+      "@parcel/watcher",
+      "better-sqlite3",
+      "sharp",
+      "unrs-resolver"
     ]
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -91,7 +91,10 @@
       "workerd"
     ],
     "onlyBuiltDependencies": [
-      "better-sqlite3"
```
tim-smart, line 94: If you can revert this and then we can merge :)

tim-smart (approved):

PR #5601, opened 2025-10-02
fix consistent fallthrough to error case

janglad:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In `packages/ai/openai/src/OpenAiLanguageModel.ts` the logic around handling file parts currently always falls through to the error case for unsupported media type, this PR fixes that, bringing the logic in line with `packages/ai/anthropic/src/AnthropicLanguageModel.ts`

## Related

- Related Issue #
- Closes #


IMax153 (changes requested):
Thanks for the PR! This just needs a changeset which can be generated with `pnpm changeset`. You can just make it a patch release.

afonsomatos:
Can you guys run the CI to bump the version? I can't use this.

PR #5620, opened 2025-10-06
fix: incorrect prepareMessages for tool-call results in OpenAiLanguag…

bbohlender:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

as per definition in Prompt.ts the result of a `ToolResultPart` is unknown (directly the result of the tool call and not an Effect value)

prepareMessgaes in the OpenAILanguageModel currently treats the result of a ToolResultPart as an effect value causing function calling to not work as expected

### DIFF
--- a/packages/ai/openai/src/OpenAiLanguageModel.ts
+++ b/packages/ai/openai/src/OpenAiLanguageModel.ts
@@ -520,13 +520,10 @@ const prepareMessages: (
 
       case "tool": {
         for (const part of message.content) {
-          const result = part.result._tag === "Right"
-            ? part.result.right
-            : part.result.left
           messages.push({
             type: "function_call_output",
             call_id: part.id,
-            output: JSON.stringify(result)
+            output: JSON.stringify(part.result)
           })
         }
 


IMax153:
@bbohlender - thanks for the PR but I'm a bit confused.

Here we need to map the `EitherEncoded` we get from the SDK into a shape that OpenAI expects (i.e. something not wrapped in an `EitherEncoded` object).

The `Prompt` module's `ToolResultPart.result` property is now also an `EitherEncoded`.

Could you perhaps provide a small program reproducing the issue you're experiencing so I can investigate further?

IMax153:
See https://github.com/Effect-TS/effect/pull/5616 for the relevant changes which necessitated the code you're referencing.

bbohlender:
@IMax153 thanks for the response
Understood

Then I think the problem I am seeing originates from [here](https://github.com/Effect-TS/effect/blob/92a4f3f4f284b877f9ef72025aa2b4d8984fe4c1/packages/ai/ai/src/LanguageModel.ts#L1011)
```ts
return Effect.forEach(toolCalls, (toolCall) => {
    return toolkit.handle(toolCall.name, toolCall.params as any).pipe(
      Effect.map(({ encodedResult, result }) =>
        Response.makePart("tool-result", {
          id: toolCall.id,
          name: toolCall.name,
          result,
          encodedResult,
          providerExecuted: false,
          ...(toolCall.providerName !== undefined
            ? { providerName: toolCall.providerName }
            : {})
        })
      )
    )
  }, { concurrency })
```

Which might be fixed with something like this? (still new to effect-ts, so might not be optimal)

```ts
return Effect.forEach(toolCalls, (toolCall) => {
    return toolkit.handle(toolCall.name, toolCall.params as any).pipe(
      Effect.either,
      Effect.map((either) => Response.makePart("tool-result", {
          id: toolCall.id,
          name: toolCall.name,
          result: either,
          encodedResult: either._tag === "Left" ? either.left.encodedResult : undefined,
          providerExecuted: false,
          ...(toolCall.providerName !== undefined
            ? { providerName: toolCall.providerName }
            : {})
      }))
    )
  }, { concurrency })
```

IMax153:
No worries - I want to resolve the issue you are having, but I don't think there is an issue here with the current code. 

Toolkit.handle returns an object where the result and encodedResult should have the proper shape already. 

Is there any chance you could post a minimal program which reproduces the issue you're facing? Will make it much easier for me to identify the root cause.

IMax153:
@bbohlender - no worries! Thank you very much for your contribution / discussion anyway!

PR #5625, opened 2025-10-08
Fix `HttpLayerRouter` with `RpcServer` example

Ty3uK:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

By the definition of `RpcServer.layerHttpRouter` we need to explicitly provide `protocol` parameters or the code will treat route as Websocket and will register it with `GET` method:

https://github.com/Effect-TS/effect/blob/main/packages/rpc/src/RpcServer.ts#L758-L760

https://github.com/Effect-TS/effect/blob/main/packages/rpc/src/RpcServer.ts#L881

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/platform/README.md
+++ b/packages/platform/README.md
@@ -5088,7 +5088,8 @@ const UserHandlers = UserRpcs.toLayer({
 // Use `HttpLayerRouter` to register the rpc server
 const RpcRoute = RpcServer.layerHttpRouter({
   group: UserRpcs,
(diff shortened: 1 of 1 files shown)


tim-smart:
Websocket is the recommended way of using rpc, so I would prefer to keep it as is.

PR #5672, opened 2025-10-30
Fix: decoding of JsonRpc messages

### LINKED ISSUE #5665: JSON-RPC Serialization does not handle Array messages correctly

@effect/rpc: "0.71.1"

See [https://github.com/Effect-TS/effect/blob/main/packages/rpc/src/RpcSerialization.ts#L159](RpcSerialization), specifically the `decodeJsonRpcRaw` function.

When `decoded` is an Array, this function always returns an empty array (`messages`) and the code in the last line `Array.isArray(decoded) ? decoded.map(decodeJsonRpcMessage) : [decodeJsonRpcMessage(decoded)]` becomes unreachable.

Decoded messages should not be lost

### What do you see instead?

When decoding JSON-RPC batch requests (arrays), all decoded messages are lost

### Additional information

This can be tested by sending an Array of requests in a single HTTP Request to the RPC endpoint with `RpcSerialization.layerJsonRpc({})`

sharedRoutine:
Fixes returning decoded messages in `decodeJsonRpcRaw` when input is an Array.

Closes #5665

### DIFF
--- a/.changeset/violet-nails-jump.md
+++ b/.changeset/violet-nails-jump.md
@@ -1,0 +1,5 @@
+---
+"@effect/rpc": patch
+---
+
+Fix decoding JsonRpc messages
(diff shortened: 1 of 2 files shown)


fubhy:
Closing as stale due to lack of activity or active review path. The underlying issue still appears plausible, so please reopen or submit a fresh PR with an updated branch and tests if this is still needed.

PR #5673, opened 2025-10-30
Fix JsonRpc client side message handling

### LINKED ISSUE #5670: JSON-RPC Serialization: Requests are never handled on the client

@effect/rpc: "0.71.1"

On [RpcServers Response](https://github.com/Effect-TS/effect/blob/979d1c10cdb32817f64c1efb9daa41a1d8e8ff3d/packages/rpc/src/RpcServer.ts#L1013) we encode based on the RpcSerialization, which in the case of JsonRPC turns

```json
[{"_tag":"Exit","requestId":"18","exit":{"_tag":"Success","value":"some value"}}]
```

into

```json
[
    {
        "jsonrpc": "2.0",
        "id": 18,
        "result": "some value"
    }
]
```

The next thing that's executing is [run's on message handler on the client side](https://github.com/Effect-TS/effect/blob/979d1c10cdb32817f64c1efb9daa41a1d8e8ff3d/packages/rpc/src/RpcClient.ts#L703) which [defaults to Effect.void](https://github.com/Effect-TS/effect/blob/979d1c10cdb32817f64c1efb9daa41a1d8e8ff3d/packages/rpc/src/RpcClient.ts#L748) for an unknown tag (which the response does not have).

Therefore JsonRPC Serialization results in requests never finishing because `write` is never called.

JsonRPC Serialization should be properly treated on the client side

### What do you see instead?

Requests finish but client never knows about it

### Additional information

_No response_

sharedRoutine:
For non streamed responses the response is never decoded by the provided `RpcSerialization`. This is however required for JsonRpc to work. Once we are decoding however, json decoding no longer worked, which is why it needed to be adjusted.

Closes #5670

### DIFF
--- a/.changeset/beige-years-hug.md
+++ b/.changeset/beige-years-hug.md
@@ -1,0 +1,5 @@
+---
+"@effect/rpc": patch
+---
+
+Fix JsonRpc client message handling
--- a/packages/rpc/src/RpcClient.ts
+++ b/packages/rpc/src/RpcClient.ts
@@ -848,7 +848,8 @@ export const makeProtocolHttp = (client: HttpClient.HttpClient): Effect.Effect<
 
       if (isJson) {
         return client.post("", { body }).pipe(
-          Effect.flatMap((r) => r.json),
+          Effect.flatMap((r) => r.text),
+          Effect.map((text) => parser.decode(text)),
           Effect.mapError((cause) =>
             new RpcClientError({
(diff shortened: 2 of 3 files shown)


fubhy:
Closing this due to inactivity and no active review path. The issue/fix may still be relevant, but this PR has gone stale; please open a fresh PR against current main if you want to continue it.

Issue #5659, opened 2025-10-27
[@effect/cli] Passing --log-level flag hangs the program


nemmtor:
### What version of Effect is running?

3.18.4

### What steps can reproduce the bug?

```ts
import * as Command from "@effect/cli/Command";
import * as NodeContext from "@effect/platform-node/NodeContext";
import * as NodeRuntime from "@effect/platform-node/NodeRuntime";
import * as Effect from "effect/Effect";

const program = Effect.logInfo("Hello world");

const command = Command.make("foo");
export const cli = command.pipe(
  Command.withHandler(() => program),
  Command.run({
    name: "CLI issue reproduction",
    version: "v0.0.1",
  }),
);

cli(process.argv).pipe(Effect.provide(NodeContext.layer), NodeRuntime.runMain);
```

Running the program with `--log-level=debug` or any other value for log level makes the program to hang.

PR #5677, opened 2025-11-02
fix log level cli arg infinite loop

nemmtor:
## Type

- [x] Bug Fix

## Description
When passing `--log-level` argument with equal symbol syntax, for example `--log-level=debug` cli program ends up in infinite loop.

## Related

- Closes https://github.com/Effect-TS/effect/issues/5659


IMax153:
@nemmtor Thanks!

PR #5690, opened 2025-11-05
fix handling of multi-statement queries

lucas-barake:
Fixes handling of multi-statement queries in @effect/sql-pg. The pg library returns an array of Result objects for multi-statement queries, but the code was expecting a single Result object, causing result.rows to be undefined.

### DIFF
--- a/packages/sql-pg/test/Client.test.ts
+++ b/packages/sql-pg/test/Client.test.ts
@@ -258,17 +258,15 @@ it.layer(PgContainer.ClientTransformLive, { timeout: "30 seconds" })("PgClient t
     Effect.gen(function*() {
       const sql = yield* SqlClient.SqlClient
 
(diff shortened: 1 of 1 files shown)


packages/sql-pg/src/PgClient.ts
```diff
@@ -169,7 +169,11 @@ export const make = (
             if (err) {
               resume(Effect.fail(new SqlError({ cause: err, message: "Failed to execute statement" })))
             } else {
-              resume(Effect.succeed(result.rows))
+              // Multi-statement queries return an array of results
+              const rows = Array.isArray(result)
+                ? result.flatMap((r) => r.rows ?? [])
```
tim-smart:
```suggestion
                ? result.map((r) => r.rows ?? [])
```

I think there should be one set per query, the same as the mysql driver

PR #5708, opened 2025-11-07
feat(cli): add Command.withConditionalBehavior for predicate based  wrapping of commands

### LINKED ISSUE #5699: @effect/cli Programatically invoke wizard mode

A CLI author may want to enable Wizard mode in scenarios other than `mycli --wizard`.
In my case I prefer if Wizard mode is run whenever someone runs just the bare command itself with no options/args, so `mycli` invokes wizard mode. This is nice for lazy end-users (like me) or people completely new to the CLI.

Expose an API from the `@effect/cli` package which invokes wizard mode

_No response_

ryanbas21:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
 This PR introduces a composable combinator approach for adding conditional behaviors to commands based on CLI arguments.

 Rather than adding separate runWith* methods for each conditional behavior (which would lead to API explosion with runWithWizard, runWithHelp, runWithValidation, etc.), this PR provides a single, extensible combinator pattern that follows functional composition principles.

```ts
 Command.withConditionalBehavior(predicate, behavior)
```

A combinator that wraps a command to conditionally trigger a behavior based on the provided predicate function:

```ts
  const command = Command.make("greet", {
    name: Options.text("name")
  }, ({ name }) => Effect.log(`Hello, ${name}!`))
    .pipe(
      Command.withConditionalBehavior(
        (args) => args.length <= 1,  // Trigger when no args provided
        "wizard"                      // Run wizard mode
      )
    )

  const cli = Command.run(command, {
    name: "MyApp",
    version: "1.0.0"
  })
```

Parameters:
  - predicate: (args: ReadonlyArray<string>) => boolean - Function that determines when to trigger the behavior
  - behavior: "wizard" | CustomBehaviorFunction - Either the built-in "wizard" mode or a custom function

 
  Wizard mode on bare command:
  ```ts
  Command.make("deploy", config).pipe(
    Command.withConditionalBehavior(
      (args) => args.length <= 1,
      "wizard"
    )
  )
  ```

```ts
  Command.make("migrate", config).pipe(
    Command.withConditionalBehavior(
      (args) => args.includes("--interactive"),
      (cmd, args) => Effect.gen(function* () {
        // Custom prompting logic
        const confirmed = yield* promptForConfirmation()
        return confirmed ? args : Effect.fail(new UserCancelled())
      })
    )
  )
```

## Related

- Related Issue #5699
- Closes #5699

### DIFF
--- a/.changeset/soft-oranges-arrive.md
+++ b/.changeset/soft-oranges-arrive.md
@@ -1,0 +1,9 @@
+---
+"@effect/cli": minor
+---
+
+Add withConditionalBehavior combinator for composable conditional command behaviors
+
+Introduces a new `withConditionalBehavior` combinator that allows wrapping commands with conditional behaviors based on CLI arguments. This composable approach replaces the need for separate `runWith*` methods and can be used for wizard mode, help display, or any custom behavior.
+
+The combinator accepts Effect's `Predicate` type, enabling the use of predicate combinators like `Predicate.and`, `Predicate.or`, and `Predicate.not` for complex conditional logic.
--- a/packages/cli/src/Command.ts
+++ b/packages/cli/src/Command.ts
@@ -11,6 +11,7 @@ import type { HashSet } from "effect/HashSet"
 import type { Layer } from "effect/Layer"
 import type { Option } from "effect/Option"
 import { type Pipeable } from "effect/Pipeable"
+import type { Predicate } from "effect/Predicate"
 import type * as Types from "effect/Types"
 import type { Args } from "./Args.js"
 import type { CliApp } from "./CliApp.js"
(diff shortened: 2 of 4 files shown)


IMax153:
@ryanbas21 - thanks for the PR!

I'm not sure about having a separate `run` method with all the specifics around wizard mode. Because then you could make the argument that we should expose similar utilities for showing help documentation based on certain conditions, prompting on certain conditions, etc.

I think it would be nicer if a combinator could be exposed that would allow you to wrap a command and show the wizard mode based on a predicate.

ryanbas21:
> @ryanbas21 - thanks for the PR!
> 
> I'm not sure about having a separate `run` method with all the specifics around wizard mode. Because then you could make the argument that we should expose similar utilities for showing help documentation based on certain conditions, prompting on certain conditions, etc.
> 
> I think it would be nicer if a combinator could be exposed that would allow you to wrap a command and show the wizard mode based on a predicate.

@IMax153 good feedback. Re-implemented it in this way. Let me know what you think?

fubhy:
Closing as stale. This has not had activity for over 6 months and there is no active maintainer review path or ready-to-merge signal. The idea may still be worth revisiting, but the current patch would need a fresh proposal/update against current main and review of the command-combinator design.

PR #5721, opened 2025-11-10
Remove special handling of execute, fix #5714

### LINKED ISSUE #5714: db.execute types appear to be wrong in effect/drizzle

3.19.2

For the given code:

```ts
const rows = yield* db.execute<{
  id: number;
  created_at: number;
}>(sql`select id, created_at from example order by created_at desc limit 1`);
```

If you were to treat `db.execute` like it works with drizzle (outside of Effect)`, the return type of this would be

```ts
result: QueryResult<{
    id: number;
    created_at: number;
}>
```

And you could have code like

```ts
const rows = await db.execute<{
  id: number;
  created_at: number;
}>(sql`select id, created_at from example order by created_at desc limit 1`);

console.log(rows.rows[0].id);
```

But with drizzle, the type of the result is

```ts
const rows: {
    id: number;
    created_at: number;
}[]
```

But this is wrong: the actual returned type when you're using effect looks like:

```ts
[
  Result {
    command: 'SELECT',
    rowCount: 1,
    oid: null,
    rows: [ [Object] ],
    fields: [ [Field], [Field] ],
    _parsers: [ [Function: parseInteger], [Function: parseBigInteger] ],
    _types: TypeOverrides { _types: [Object], text: {}, binary: {} },
    RowCtor: null,
    rowAsArray: false,
    _prebuiltEmptyResultObject: { id: null, created_at: null }
  }
]
```

So really Effect's return value is more like

```ts
result: [QueryResult<{
    id: number;
    created_at: number;
}>]
```

A tuple of one with the QueryResult inside.

It looks like the problem might be here, in which the `execute` method gets some special treatment

https://github.com/Effect-TS/effect/blob/main/packages/sql-drizzle/src/internal/patch.ts#L52

The type parameter of db.execute should work in a way that reflects drizzle default behavior, or at least a predictable way.

### What do you see instead?

The type parameter does not reflect the actual data shape.

### Additional information

_No response_

mikearnaldi:
closes #5714

### DIFF
--- a/.changeset/wide-years-peel.md
+++ b/.changeset/wide-years-peel.md
@@ -1,0 +1,5 @@
+---
+"@effect/sql-drizzle": patch
+---
+
+Remove special handling of execute, fix #5714
--- a/packages/sql-drizzle/src/internal/patch.ts
+++ b/packages/sql-drizzle/src/internal/patch.ts
@@ -48,14 +48,6 @@ export const makeRemoteCallback = Effect.gen(function*() {
   return (sql: string, params: Array<any>, method: "all" | "execute" | "get" | "values" | "run") => {
     const runPromise = Runtime.runPromise(currentRuntime ? currentRuntime : constructionRuntime)
     const statement = client.unsafe(sql, params)
-    if (method === "execute") {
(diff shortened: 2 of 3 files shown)


fubhy:
Closing this because CI is failing and there has been no recent activity. Please reopen or open a fresh PR if this change is still relevant.

PR #5728, opened 2025-11-11
Fix UrlParams.setAll overwrite semantics

### LINKED ISSUE #5627: Confusing, possibly buggy `UrlParams.setAll`

effect@3.18.1, @effect/platform@0.92.1

```ts
import assert from 'node:assert/strict';
import { describe, it } from 'node:test';
import { UrlParams } from '@effect/platform';

describe('setAll', () => {
  it('concats, preferring elements from the second input with preservation', () => {
    const prev = UrlParams.fromInput({
      baz: 'c',
      foo: 'd',
    });

    const neww = UrlParams.fromInput({
      foo: 'a',
      bar: 'b',
    });

    assert.deepEqual(UrlParams.toRecord(UrlParams.setAll(prev, neww)), {
      foo: ['d', 'a'],
      bar: 'b',
    });
  });
});
```

The assertion result should instead be (order notwithstanding):

```ts
assert.deepEqual(UrlParams.toRecord(UrlParamsExtras.setAll(prev, neww)), {
  baz: 'c',
  foo: 'a',
  bar: 'b',
});
```

### What do you see instead?

The above repro assertion:

```ts
assert.deepEqual(UrlParams.toRecord(UrlParams.setAll(prev, neww)), {
  foo: ['d', 'a'],
  bar: 'b',
});
```

### Additional information

This behaviour is confusing. Take `UrlParams.set` for example, which overwrites a single key but leaves others alone:

```ts
assert.deepEqual(UrlParams.toRecord(UrlParams.set(prev, 'foo', 'a')), {
  baz: 'c',
  foo: 'a',
});
```

The test passes my expected assertion if the `keys.includes` predicate is negated/inverted here: https://github.com/Effect-TS/effect/blob/%40effect/platform%400.92.1/packages/platform/src/UrlParams.ts#L164

kitlangton:
- [x] Bug Fix

## Description

UrlParams.setAll was removing every key not present in the new input and also keeping the old values for keys that should have been overwritten. The fix filters existing entries using a set of the incoming keys, so only those keys are replaced and unrelated params stay intact. A regression test that mirrors the bug report now guards the behavior.

## Related

-  Closes #5627

### DIFF
--- a/packages/platform/src/UrlParams.ts
+++ b/packages/platform/src/UrlParams.ts
@@ -158,12 +158,16 @@ export const setAll: {
   (input: Input): (self: UrlParams) => UrlParams
   (self: UrlParams, input: Input): UrlParams
 } = dual(2, (self: UrlParams, input: Input): UrlParams => {
-  const toSet = fromInput(input)
-  const keys = new Set(toSet.map(([k]) => k))
-  return Arr.appendAll(
-    Arr.filter(self, ([k]) => !keys.has(k)),
-    toSet
-  )
+  const out = fromInput(input) as Array<readonly [string, string]>
+  const keys = new Set()
+  for (let i = 0; i < out.length; i++) {
+    keys.add(out[i][0])
+  }
+  for (let i = 0; i < self.length; i++) {
+    if (keys.has(self[i][0])) continue
+    out.push(self[i])
+  }
+  return out
 })
 
 /**


tim-smart:
I think you want to use appendAll instead

tim-smart (approved):

PR #5754, opened 2025-11-16
Fix Layer.updateService type signature

front-depiction:
## Summary

Fixes the type signature of `Layer.updateService` to correctly include the service `I` in both the input requirements (RIn) and output types (ROut).

## Problem

The `updateService` function:
- Reads an existing service `I` from the context to update it
- Outputs the updated service `I` along with the layer's outputs

The previous type signature was missing `I` in the output type (ROut position), which was incorrect.

## Solution

Updated the type signature:
- **ROut**: `I | A1` (outputs the updated service I plus the layer's original outputs)
- **E**: `E1` (same errors as input layer)
- **RIn**: `I | R1` (requires service I as input, plus the layer's original requirements)

## Changes

- Updated `Layer.updateService` type signature in both curried forms
- Added changeset documenting the fix

### DIFF
--- a/packages/effect/src/Layer.ts
+++ b/packages/effect/src/Layer.ts
@@ -1215,12 +1215,12 @@ export const updateService = dual<
   <I, A>(
     tag: Context.Tag<I, A>,
     f: (a: A) => A
-  ) => <A1, E1, R1>(layer: Layer<A1, E1, R1>) => Layer<A1, E1, I | R1>,
+  ) => <A1, E1, R1>(layer: Layer<A1, E1, R1>) => Layer<I | A1, E1, I | R1>,
   <A1, E1, R1, I, A>(
     layer: Layer<A1, E1, R1>,
     tag: Context.Tag<I, A>,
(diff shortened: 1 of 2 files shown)


packages/effect/src/Layer.ts
```diff
@@ -1215,12 +1215,12 @@ export const updateService = dual<
   <I, A>(
     tag: Context.Tag<I, A>,
     f: (a: A) => A
-  ) => <A1, E1, R1>(layer: Layer<A1, E1, R1>) => Layer<A1, E1, I | R1>,
+  ) => <A1, E1, R1>(layer: Layer<A1, E1, R1>) => Layer<I | A1, E1, I | R1>,
```
tim-smart, line 1218: This isn't correct as the update layer is using Layer.provide not Layer.provideMerge

front-depiction: Cheers, seem to have misunderstood the purpose of the function

PR #5786, opened 2025-11-21
Add Cron.prev reverse iteration support

### LINKED ISSUE #5666: Cron.prev method to get the last time that a cron ran

This would allow a Cron instance to produce the _last_ time that it would have triggered. This is useful if you're showing a UI that shows a progress bar of 'time till next run' and you want that to represent the time between cron invocations.

A method called Cron.prev, or internally, possibly a method that can calculate 'next' in both a forward and reverse time direction. This would also make computing a sequence of previous cron values possible.

This is admittedly a niche issue, but there is precedent for this method in libraries like [cron-parser](https://www.npmjs.com/package/cron-parser).

kitlangton:
## Summary
- add backward cron traversal with shared lookup tables and last-boundary resets
- align DST handling while walking backward and add reverse sequence iterator
- expand cron prev/next symmetry tests, DST overlap coverage, step/range backward cases, month/year wrap, and add changeset

## Context
Based on @tmcw's fork branch https://github.com/tmcw/effect/tree/cron-prev ; rebased on main and tightened tests/edge cases.
Closes #5666.

## Testing
- pnpm vitest packages/effect/test/Cron.test.ts

### DIFF
--- a/packages/effect/src/Cron.ts
+++ b/packages/effect/src/Cron.ts
@@ -525,11 +525,7 @@ const stepCron = (cron: Cron, startFrom: DateTime.DateTime.Input | undefined, di
         const currentMinute = current.getUTCMinutes()
         const nextMinute = table.minute[currentMinute]
         if (nextMinute === undefined) {
-          current.setUTCHours(
-            current.getUTCHours() + tick,
-            boundary.minute,
-            boundary.second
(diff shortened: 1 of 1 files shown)


fubhy (approved):
Welcome to team Cron

PR #5789, opened 2025-11-21
fix: local cluster requires message store

nwidynski:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Running the cluster socket in local mode currently causes a defect because persisted RPC messages require a message store. This fix uses the memory implementation over the noop.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/platform-node/src/NodeClusterSocket.ts
+++ b/packages/platform-node/src/NodeClusterSocket.ts
@@ -92,7 +92,7 @@ export const layer = <
     Layer.provide(runnerHealth),
     Layer.provideMerge(
       options?.storage === "local"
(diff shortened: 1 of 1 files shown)


tim-smart:
This is intentional, as if you need persistence you should use SQL storage.

I think we could add some separate layers for testing.

nwidynski:
Is there any way to disable persistence with the `NodeClusterSocket`? If not, then yeah, layers for in-memory execution would be amazing 👍 I can currently work around this by providing the entire testing stack of layers, but I would love to just use `NodeClusterSocket` with a different setting.

PR #5792, opened 2025-11-22
docs(platform): clarify FileSystem.watch recursive option behavior

Hoishin:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

Update API docs and previous change log for how `recursive` option for `FileSystem.watch` is supported, and how it's handled.

## Related

- Related PR #5174

It seems that it was AI slop that added "recursive option is only supported on macOS and Windows" to changelog and API docs. By the time recursive option was added, not only that Node.js v18 was already EOL, which meant all LTS at that time supported recursive watch for all OS, but also that `platform-node` was already using `@parcel/watch` by default which is always recursive.

### DIFF
--- a/packages/platform-node/CHANGELOG.md
+++ b/packages/platform-node/CHANGELOG.md
@@ -515,7 +515,7 @@
 
   Added a `recursive` option to `FileSystem.watch` that allows watching for changes in subdirectories. When set to `true`, the watcher will monitor changes in all nested directories.
 
-  Note: The `recursive` option behavior depends on the backend implementation. When using `@parcel/watcher` (default), watching is always recursive on all platforms and this option is ignored. When using the Node.js `fs.watch()` fallback, the `recursive` option is supported on all platforms (Node.js v20+).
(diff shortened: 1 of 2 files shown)


packages/platform-node/CHANGELOG.md
```diff
@@ -488,7 +488,7 @@
 
   Added a `recursive` option to `FileSystem.watch` that allows watching for changes in subdirectories. When set to `true`, the watcher will monitor changes in all nested directories.
 
-  Note: The recursive option is only supported on macOS and Windows. On other platforms, it will be ignored.
+  Note: The `recursive` option behavior depends on the backend implementation. When using `@parcel/watcher` (default), watching is always recursive on all platforms and this option is ignored. When using the Node.js `fs.watch()` fallback, the `recursive` option is supported on all platforms (Node.js v20+).
```
tim-smart: parcel watcher is not the default

Hoishin: I misunderstood the layer dependencies. Fixed now

PR #5827, opened 2025-11-28
add TestRunner & SingleRunner modules

tim-smart:


packages/workflow/src/Activity.ts
```diff
@@ -148,7 +149,15 @@ const retryOnInterrupt = (
  * @since 1.0.0
  * @category Error handling
  */
-export const retry: typeof Effect.retry = dual(
+export const retry: {
+  <E, O extends Types.NoExcessProperties<Omit<Effect.Retry.Options<E>, "schedule">, O>>(
```
vinassefranche, line 153: @tim-smart is there a reason why we can't set the `schedule` option on this anymore? We are using it in our current version of the code

adrian-gierakowski: same here

PR #5832, opened 2025-11-28
docs(platform): add protocol for rpc docs

k2on:
This example does not work without "protocol"


fubhy:
Closing as stale due to no activity or active review path for over 6 months. The docs issue may still be valid, so please open a fresh PR if you want to continue it against current main.

PR #5834, opened 2025-11-30
fix: Fixed "oneOf" toolChoice for google language model

timurrakhimzhan:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Fixed the following error coming from `GoogleAiLanguageModel` when trying to pass `oneOf` along with `mode: 'auto'`:
```{"error":{"code":400,"message":"Please set allowed_function_names only when function calling mode is ANY.","status":"INVALID_ARGUMENT"}}```

### DIFF
--- a/.changeset/eight-ears-sink.md
+++ b/.changeset/eight-ears-sink.md
@@ -1,0 +1,5 @@
+---
+"@effect/ai-google": patch
+---
+
+Fixed "oneOf" toolChoice for google language model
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: .changeset/eight-ears-sink.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/ai-google": minor
```
IMax153: Given this is a bug fix, we can make it a patch.

IMax153 (approved):

PR #5853, opened 2025-12-03
handle executable paths with spaces in CLI arguments, close #5845

### LINKED ISSUE #5845: From Discord: Built-in --log-level option causes argument parsing issues on Windows due to path splitting
# Summary
A user reported that using the built-in --log-level option on their CLI breaks access to parent options in subcommands, specifically when running on Windows. The issue arises because the executable path string containing spaces (e.g., 'C:\Program Files\nodejs\node.exe') is split incorrectly using a whitespace delimiter, which causes subsequent argument parsing to shift and fail. This manifests as the parser failing to find expected options such as '--config'. The maintainers acknowledge this as a Windows-specific bug related to path parsing and are considering solutions, with a temporary issue opened for tracking.

# Discord thread

https://discord.com/channels/795981131316985866/1443920112259432572

Masty88:
## Type

- [x] Bug Fix

## Description

Fixes executable path splitting when paths contain spaces (e.g., `C:\Program Files\nodejs\node.exe` on Windows or `/usr/local/My Programs/node` on Unix-like systems).

Instead of using `executable.split(/\s+/)` which breaks any path with spaces, we now extract runtime and script directly from the original args array using `Arr.take(originalArgs, 2)`.

**Changes:**
- Modified `handleBuiltInOption()` to accept `originalArgs` parameter
- Fixed `SetLogLevel` case to preserve paths with spaces when re-executing commands
- Fixed `ShowWizard` case to preserve paths with spaces when running wizard-constructed commands
- Added helper function `getExecutableArgs()` to extract runtime and script from original args
- Added test case for executable paths with spaces in `CliApp.test.ts`

**Test plan:**
- ✅ All existing 146 tests pass
- ✅ Added new test for executable paths with spaces
- ✅ Manually tested with minigit example: `npx tsx examples/minigit.ts --log-level info -c test=value clone https://github.com/test/repo`

## Related

- Closes #5845

### DIFF
--- a/packages/cli/src/internal/cliApp.ts
+++ b/packages/cli/src/internal/cliApp.ts
@@ -163,7 +163,7 @@ const handleBuiltInOption = <R, E, A>(
     case "SetLogLevel": {
       // Use first 2 elements from originalArgs (runtime + script) to preserve paths with spaces
       // Filter out --log-level from args before re-executing
-      const baseArgs = getExecutableArgs(originalArgs)
+      const baseArgs = Arr.take(originalArgs, 2)
       const filteredArgs: Array<string> = []
       for (let i = 0; i < args.length; i++) {
         if (isLogLevelArg(args[i]) || isLogLevelArg(args[i - 1])) {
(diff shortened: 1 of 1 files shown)


packages/cli/src/internal/cliApp.ts
```diff
@@ -339,6 +345,9 @@ const getWizardPrefix = (
   return Arr.appendAll(rootCommand.split(/\s+/), args)
 }
 
+const getExecutableArgs = (args: ReadonlyArray<string>): ReadonlyArray<string> =>
+  Arr.take(args, 2) as ReadonlyArray<string>
```
kitlangton: I don't think this `as` is necessary.

kitlangton: I'll port this over to smol :)

Masty88: Thanks for reviewing!

PR #5855, opened 2025-12-03
feat(opentelemetry): expose OtlpExporter as public API

jonastemplestein:
## Summary

Exposes the internal OTLP exporter as `OtlpExporter.make()` to allow building custom multi-endpoint tracers or composing multiple exporters without copying internal code.

## Motivation

When building applications that need to send traces to multiple backends (e.g., Honeycomb + Axiom + Sentry), the current options are:
1. Use `NodeSdk` with multiple `SpanProcessor`s (requires `@opentelemetry/sdk-trace-base`)
2. Copy the internal exporter code

Exposing the exporter allows Effect-native multi-endpoint tracing without any OpenTelemetry SDK dependencies:

```typescript
import { OtlpExporter, OtlpResource } from "@effect/opentelemetry"

const makeMultiExporter = (endpoints: Array<EndpointConfig>) =>
  Effect.gen(function*() {
    const exporters = yield* Effect.forEach(endpoints, e =>
      OtlpExporter.make({
        url: e.url,
        headers: e.headers,
        label: e.name,
        exportInterval: "5 seconds",
        maxBatchSize: 100,
        body: (spans) => ({ resourceSpans: [{ scopeSpans: [{ spans }] }] }),
        shutdownTimeout: "3 seconds"
      })
    )
    return {
      push: (data: unknown) => exporters.forEach(e => e.push(data))
    }
  })
```

## Changes

- Added `src/OtlpExporter.ts` - public module re-exporting the internal exporter with proper JSDoc documentation
- Updated `src/index.ts` - added `OtlpExporter` export
- Added changeset

## Test Plan

- [x] Existing tests pass (no behavioral changes)
- [ ] Would be good to add a test for the new public API

### DIFF
--- a/packages/opentelemetry/src/OtlpExporter.ts
+++ b/packages/opentelemetry/src/OtlpExporter.ts
@@ -1,0 +1,117 @@
+/**
+ * OTLP Exporter
+ *
+ * A generic OTLP exporter that can be used to build custom tracers, metrics exporters,
+ * or loggers that send data to OTLP-compatible backends.
+ *
+ * This exporter handles:
+ * - Batching of data with configurable batch size
+ * - Automatic retry with backoff on transient errors
+ * - Rate limiting with 429 Retry-After header support
+ * - Graceful shutdown with timeout
+ * - Automatic disabling for 60 seconds on persistent failures
+ *
+ * @since 1.0.0
+ * @example
+ * ```typescript
+ * import { OtlpExporter } from "@effect/opentelemetry"
+ * import { BunHttpClient } from "@effect/platform-bun"
+ * import { Effect, Layer } from "effect"
+ *
+ * const program = Effect.gen(function*() {
+ *   const exporter = yield* OtlpExporter.make({
+ *     url: "https://api.honeycomb.io/v1/traces",
+ *     headers: { "x-honeycomb-team": "your-api-key" },
+ *     label: "honeycomb",
+ *     exportInterval: "5 seconds",
+ *     maxBatchSize: 100,
+ *     body: (spans) => ({ resourceSpans: [{ scopeSpans: [{ spans }] }] }),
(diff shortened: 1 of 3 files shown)


packages/opentelemetry/src/OtlpExporter.ts
```diff
@@ -0,0 +1,117 @@
...
+ * - 429 rate limiting with Retry-After header support
+ * - Graceful shutdown ensuring pending data is flushed
+ * - Auto-disable for 60 seconds on persistent failures (logged via Effect.logDebug)
+ *
+ * @since 1.0.0
+ * @category Constructors
+ */
+export const make: (
+  options: MakeOptions
+) => Effect.Effect<Exporter, never, HttpClient.HttpClient | Scope.Scope> = internal.make as any
```
mikearnaldi, line 117: why do we need `as any`?

packages/opentelemetry/src/OtlpExporter.ts
```diff
@@ -0,0 +1,117 @@
...
+ *   Effect.runPromise
+ * )
+ * ```
+ */
+import type * as Headers from "@effect/platform/Headers"
+import type * as HttpClient from "@effect/platform/HttpClient"
+import type * as Duration from "effect/Duration"
+import type * as Effect from "effect/Effect"
+import type * as Scope from "effect/Scope"
+import * as internal from "./internal/otlpExporter.js"
```
tim-smart, line 48: I would just move the entire module up instead of re-exporting internal

fubhy:
Thanks for the PR. Maintainers left implementation feedback that still needs addressing: avoid the `as any`, and prefer moving the exporter module up rather than wrapping/re-exporting the internal module. Are you still interested in updating this?

tim-smart:

PR #5863, opened 2025-12-03
Widen otel logs peer dep range

mikearnaldi:


tmcw:
Thanks for doing this - can confirm that this fixes the peerDependencies situation with sdk-node!

tim-smart (approved):

PR #5866, opened 2025-12-04
Add Function.memo

jdharrisnz:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add pure memoization wrapper function to the Function module.

Questions:
- Do you think there is value in the memoThis alias? It's rare that `this` is used in type signatures, and the implementation uses it so it's valid, but it might just be unnecessary noise.
- Should the helpers be in some internal file or are they ok where they are with the `@internal` tag?
- Should this also be used internally, to replace `Hash.cached` for example? And if so, what are all the candidates for replacing manual caching? Would those changes belong in this PR or another?

### DIFF
--- a/.changeset/cute-chairs-kneel.md
+++ b/.changeset/cute-chairs-kneel.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Add Function.memo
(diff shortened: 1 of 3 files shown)


tim-smart:

PR #5871, opened 2025-12-05
feat: support `URL` in `HttpClientRequest.prependUrl`

f15u:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Add support `URL` in `HttpClientRequest.prependUrl`.

This is be useful working with full URLs, especially when added as base URLs to custom `HttpClient`s and/or working with `Schema` `Config`s using `Schema.URL` when the URL is get from env.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/light-comics-rest.md
+++ b/.changeset/light-comics-rest.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": minor
+---
+
+Support `URL` in `HttpClientRequest.prependUrl` parameter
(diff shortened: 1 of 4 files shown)


fubhy:
Thanks for the PR. Could you update the implementation to consistently use the normalized string value after accepting `string | URL`? In the current diff, the slash-handling branch still concatenates `url` instead of `clone`, which looks likely to fail type checking or at least leaves the implementation inconsistent. A test covering a base URL with a trailing slash would also help.

tim-smart:

PR #5899, opened 2025-12-12
fix HttpServerRequest.toWeb missing duplex option for streaming bodies

kitlangton:
## Type

- [x] Bug Fix

## Description

Fixes `HttpServerRequest.toWeb` throwing `TypeError: RequestInit: duplex option is required when sending a body` in Node.js when converting requests with streaming bodies (POST, PUT, PATCH).

### The Bug

When calling `HttpServerRequest.toWeb()` on a Node.js server request with a body, it throws:

```
TypeError: RequestInit: duplex option is required when sending a body.
    at new Request (node:internal/deps/undici/undici:10638:19)
```

### The Fix

Add `duplex: "half"` option when constructing a `Request` with a streaming body:

```typescript
return new Request(ourl.value, {
  method: self.method,
  body,
  headers: self.headers,
  // @ts-expect-error - duplex is required for streaming bodies in Node 18+
  duplex: body !== undefined ? "half" : undefined
})
```

### Why it only affects Node.js

The code has an early return for when `source` is already a `Request`:
```typescript
if (self.source instanceof Request) {
  return self.source
}
```
In Bun/Deno/web contexts, it returns the native Request directly. The bug only manifests in Node.js where we construct a new Request from an `IncomingMessage`.

### Test

Added test case that converts a POST request with JSON body via `toWeb()` and reads it back.

### DIFF
--- a/.changeset/tall-buses-sing.md
+++ b/.changeset/tall-buses-sing.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+fix HttpServerRequest.toWeb missing duplex option for streaming bodies in Node
--- a/packages/platform-node/test/HttpServer.test.ts
+++ b/packages/platform-node/test/HttpServer.test.ts
@@ -755,4 +755,30 @@ describe("HttpServer", () => {
       )
       expect(root).toEqual("root")
     }).pipe(Effect.provide(NodeHttpServer.layerTest)))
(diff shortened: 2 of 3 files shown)


kitlangton:
@copilot ... You opened a new pull request instead of reviewing this one. You have failed me today, brother.

PR #5905, opened 2025-12-14
Allow client middleware failure

mattiamanzati:
## Type

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Relax the RpcClientMiddleware allowing them to fail on client side raising the same error allowed in the server. This allows short circuit for errors such as "Unauthorized" when the client does not have credentials at all.

### DIFF
--- a/packages/rpc/src/RpcClient.ts
+++ b/packages/rpc/src/RpcClient.ts
@@ -118,7 +118,10 @@ export declare namespace RpcClient {
       infer _Error,
       infer _Middleware
     > ? [_Success] extends [RpcSchema.Stream<infer _A, infer _E>] ? AsMailbox extends true ? Effect.Effect<
-            Mailbox.ReadonlyMailbox<_A["Type"], _E["Type"] | _Error["Type"] | E | _Middleware["failure"]["Type"]>,
+            Mailbox.ReadonlyMailbox<
+              _A["Type"],
+              _E["Type"] | _Error["Type"] | E | _Middleware["failure"]["Type"] | _Middleware["~ClientError"]
+            >,
             never,
             | Scope.Scope
             | _Payload["Context"]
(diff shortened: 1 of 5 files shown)


packages/rpc/src/RpcClient.ts
```diff
@@ -352,12 +352,15 @@ export const makeNoSerialization: <Rpcs extends Rpc.Any, E, const Flatten extend
         headers: Headers.merge(parentFiber.getFiberRef(currentHeaders), headers)
       })
       if (discard) {
-        return Effect.flatMap(send, (message) =>
-          options.onFromClient({
-            message,
-            context,
-            discard
-          }))
+        return Effect.matchEffect(send, {
```
tim-smart, line 363: I would instead keep middleware errors as part of the discard type.

packages/rpc/src/RpcClient.ts
```diff
@@ -384,13 +387,15 @@ export const makeNoSerialization: <Rpcs extends Rpc.Any, E, const Flatten extend
...
           fiber = send.pipe(
-            Effect.flatMap((request) =>
-              options.onFromClient({
-                message: request,
-                context,
-                discard
-              })
-            ),
+            Effect.matchCauseEffect({
+              onFailure: (cause) => Effect.sync(() => entry.resume(Exit.failCause(cause))),
```
tim-smart: The failure is already resumed by the fiber observer.

tim-smart (changes requested):
In 4.0 we can have seperate client errors: https://github.com/Effect-TS/effect-smol/pull/818

packages/rpc/src/RpcClient.ts
mattiamanzati: Should maybe for this v3 PR add a Boolean flag to enable that? 

Because otherwise it would break existing clients by introducing a new error type; 
I was discarding it to emulate the same behaviour that would happen if the error was processed by the server.
Wdyt?

mattiamanzati: Or else we could add the separate client error type even in v3 like in v4, but would'nt that break semantics of knowing that everything that can fail is inside the schema and thous serializable?

tim-smart: We can make a minor bump, the rpc api is still unstable.

The error is happening on the client, so it makes the most sense to expose the error on the client.

mattiamanzati: I've updated and added support for the client error in the same v4 style, exception of automatically including failure types as well in discards.

This gets the minimum feature working without being too much breaking and old api still produce the same types for v3; lmk if we are ok with including the server middleware failure type as well, or if we keep that as a v4 change

fubhy:
Closing this because requested changes have been inactive for an extended period. Please open a fresh PR if you want to continue the change.

PR #5927, opened 2025-12-19
feat(opentelemetry): add protobuf protocol support for OTLP exporters

davidgoli:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add support for Protocol Buffers binary encoding as an alternative to JSON for OTLP trace, metrics, and log exports. This enables more efficient wire format when communicating with OpenTelemetry collectors. In particular, some OTLP collectors, such as Coralogix, only support protobuf and not json.

- Add `protocol` option ("json" | "protobuf") to Otlp.layer and individual OtlpTracer, OtlpMetrics, OtlpLogger layers
- Implement protobuf wire format encoding following opentelemetry-proto specs
- Set appropriate Content-Type header (application/x-protobuf vs application/json)
- No new dependencies - protobuf encoding implemented from scratch

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/opentelemetry/src/index.ts
+++ b/packages/opentelemetry/src/index.ts
@@ -34,6 +34,11 @@ export * as OtlpMetrics from "./OtlpMetrics.js"
 export * as OtlpResource from "./OtlpResource.js"
 
 /**
+ * OtlpSerialization service for tree-shakable protobuf support.
+ *
+ * This module provides the `OtlpSerialization` service that abstracts the
+ * encoding of OTLP telemetry data to HttpBody.
+ *
  * @since 1.0.0
  */
 export * as OtlpSerialization from "./OtlpSerialization.js"


packages/opentelemetry/src/Otlp.ts
```diff
@@ -32,6 +44,7 @@ export const layer = (options: {
   readonly metricsExportInterval?: Duration.DurationInput | undefined
   readonly tracerExportInterval?: Duration.DurationInput | undefined
   readonly shutdownTimeout?: Duration.DurationInput | undefined
+  readonly protocol?: OtlpProtocol | undefined
```
tim-smart: I think we should probably do seperate layers so the protobuf code doesn't end up in the bundle.

tim-smart: Or maybe even better, create a service that represents the chosen serialisation.

tim-smart (changes requested):

packages/opentelemetry/src/Otlp.ts
davidgoli: With a layer design, would json also be provided as a layer? I'm assuming we'd like to preserve backwards compatibility for existing Otlp code and not require an explicit json layer in userland, is that correct?

davidgoli:
@tim-smart done, please take another look

packages/opentelemetry/src/OtlpSerialization.ts
```diff
@@ -0,0 +1,86 @@
...
+import * as Layer from "effect/Layer"
+import * as OtlpProtobuf from "./internal/otlpProtobuf.js"
+
+/**
+ * Service interface for serializing OTLP telemetry data to HttpBody.
+ *
+ * @since 1.0.0
+ * @category Models
+ */
+export interface OtlpSerialization {
```
tim-smart: This will either need a type id, or use the class syntax

packages/opentelemetry/src/internal/otlpExporter.ts
```diff
@@ -9,6 +10,13 @@ import * as Num from "effect/Number"
 import * as Option from "effect/Option"
 import * as Schedule from "effect/Schedule"
 import * as Scope from "effect/Scope"
+import { OtlpSerialization } from "../OtlpSerialization.js"
+
+/**
+ * OTLP telemetry kind for encoding
+ * @internal
+ */
+export type OtlpKind = "traces" | "metrics" | "logs"
```
tim-smart: This can be removed. The exporter doesn't need to know what kind of data it is handling.

packages/opentelemetry/src/internal/otlpExporter.ts
```diff
@@ -53,6 +62,13 @@ export const make: (
     HttpClient.retryTransient({ schedule: policy, times: 3 })
   )
 
+  const serialization = yield* OtlpSerialization
+  const encode: (data: unknown) => HttpBody.HttpBody = options.kind === "traces"
```
tim-smart: This can be lifted into the corresponding service.

packages/opentelemetry/src/Otlp.ts
```diff
@@ -64,3 +71,52 @@ export const layer = (options: {
...
+ * ```typescript
+ * import { Otlp } from "@effect/opentelemetry"
+ *
+ * const layer = Otlp.layer({ baseUrl: "http://localhost:4318" })
+ * ```
+ *
+ * @since 1.0.0
+ * @category Layers
+ */
+export const layer = layerJson
```
tim-smart: It is ok for this to be a breaking change to align with effect 4. This package is still api unstable

davidgoli: Should we make the encoding format explicit then? or keep the json default?

tim-smart (approved):

PR #5938, opened 2025-12-30
Fix openai streaming decoding error when using provider webSearch tool

tensor2077:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
When sending a stream request, `OpenAiClient.streamRequest` decodes each SSE `event.data` into `ResponseStreamEvent`(`OpenAiClient.ts:1803`). 

However, in real streams, `ResponseOutputItemAddedEvent` can emit a
  `web_search_call` item before it’s fully populated: the `action` field is not present yet (it only shows up later in `response.output_item.done`). This can cause an early decode failure while
  `OpenAiLanguageModel.makeStreamResponse` is processing the tool-call stream parts.

https://github.com/Effect-TS/effect/blob/65e9e35157cbdfb40826ddad34555c4ebcf7c0b0/packages/ai/openai/src/OpenAiClient.ts#L402-L421

## Proposed Fix
I currently work around this by relaxing the schema to accept `web_search_call` without `action` in `ResponseOutputItemAddedEvent`, so it can accept an in-progress web_search_call item without action. (status="in_progress").

what do you think is the most appropriate way to model/handle this situations? @IMax153 

## Reproduce

You can reproduce this by enabling the `web_search` tool and asking the model to perform a search in streaming mode.

```ts
Toolkit.make(
  OpenAI.OpenAiTool.WebSearch({
    search_context_size: 'medium'
  })
)
```

### DIFF
--- a/.changeset/rude-bikes-tan.md
+++ b/.changeset/rude-bikes-tan.md
@@ -1,0 +1,5 @@
+---
+"@effect/ai-openai": patch
+---
+
+Fix streaming decode: response.output_item.added may emit web_search_call without action (when status="in_progress").
--- a/packages/ai/openai/src/OpenAiClient.ts
+++ b/packages/ai/openai/src/OpenAiClient.ts
@@ -393,6 +393,17 @@ export class ResponseFailedEvent extends Schema.Class<ResponseFailedEvent>(
   response: Generated.Response
 }) {}
 
+const WebSearchToolCallForAddEvent = Schema.asSchema(
+  Generated.WebSearchToolCall.pipe(
+    Schema.omit("action")
+  )
+)
+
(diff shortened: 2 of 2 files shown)


IMax153:
@tensor2077 - thank you for the detailed pull request. 

It doesn't feel quite right to implement this fix in the `OpenAiClient` - feels a little bit like a "hack", since if more tools are added in the future that require similar fixes this could become unmaintainable. But I also don't think there's a better short-term fix right now so I'm willing to move this forward.

The good thing is that in Effect v4, our OpenAPI generator will be **much** more robust, so we can actually generate the streaming output types, which should alleviate the need for these patches.

PR #5943, opened 2026-01-02
Ensure client is defined when catching error

sidpremkumar:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We have been seeing errors:
```
2025-12-31 04:02:27.243 | at <anonymous> (/app/node_modules/pg-pool/index.js:216:18)
2025-12-31 04:02:27.243 | at <anonymous> (/app/node_modules/@effect/sql-pg/dist/esm/PgClient.js:232:7)
2025-12-31 04:02:27.243 | TypeError: undefined is not an object (evaluating 'client.on')
2025-12-31 04:02:27.243 | ^
2025-12-31 04:02:27.243 | 232 | client.on("error", onError);
2025-12-31 04:02:27.243 | 231 | }
2025-12-31 04:02:27.243 | 230 | cause = cause_;
2025-12-31 04:02:27.243 | 229 | function onError(cause_) {
2025-12-31 04:02:27.243 | 228 | }
```

We have tried modifying and guarding better on our application code - but this issue persisted. We think it's coming from here, specifically the `client!` assertion that causes the entire app to crash and restart our application. 

Believe this can be fixed by making sure the client exists before calling on error. 

This seems to be known: 
Ref: https://github.com/brianc/node-postgres/issues/2492#issuecomment-797531997

> If the connection creation or acquisition encounters an error then the client arg is undefined and the done callback is a no-op function. 

## Related

n/a

### DIFF
--- a/.changeset/crazy-walls-wait.md
+++ b/.changeset/crazy-walls-wait.md
@@ -1,5 +1,5 @@
 ---
-"@effect/sql-pg": minor
+"@effect/sql-pg": patch
 ---
 
 Fix crash when pool.connect fails by ensuring client exists before attaching error handler


packages/sql-pg/src/PgClient.ts
```diff
@@ -315,25 +315,25 @@ export const make = (
...
-              })
-            ),
-            client!
-          ))
+          return
         }
-        function onError(cause_: Error) {
-          cause = cause_
-        }
-        client!.on("error", onError)
```
sidpremkumar, line 336: Ref: https://github.com/brianc/node-postgres/issues/2492#issuecomment-797531997

> If the connection creation or acquisition encounters an error then the client arg is undefined and the done callback is a no-op function.

tim-smart (approved):
👍

Just needs a lint fix and a changeset

.changeset/crazy-walls-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/sql-pg": minor
```
tim-smart:
```suggestion
"@effect/sql-pg": patch
```

sidpremkumar: sorry about that!

PR #5952, opened 2026-01-05
fix(opentelemetry): make sdk-trace-node and sdk-trace-web required peer deps

clayroach:
## Problem

`@effect/opentelemetry` marks `@opentelemetry/sdk-trace-node` and `@opentelemetry/sdk-trace-web` as optional peer dependencies. However, both are exported from the main `index.ts`, which creates static ESM imports. This causes module resolution errors when users don't have these packages installed:

```
Error: Cannot find package '@opentelemetry/sdk-trace-web' imported from 
.../@effect/opentelemetry/dist/esm/WebSdk.js
```

## Solution

Make `@opentelemetry/sdk-trace-node` and `@opentelemetry/sdk-trace-web` required peer dependencies by removing them from `peerDependenciesMeta`.

## Changes

- Removed `optional: true` for `@opentelemetry/sdk-trace-node`
- Removed `optional: true` for `@opentelemetry/sdk-trace-web`

## Testing

- Build passes
- All tests pass (15/15)

### DIFF
--- a/.changeset/move-sdk-to-subpath-exports.md
+++ b/.changeset/move-sdk-to-subpath-exports.md
@@ -1,5 +1,5 @@
 ---
-"@effect/opentelemetry": patch
+"@effect/opentelemetry": minor
 ---
 
 Make @opentelemetry/sdk-trace-node and @opentelemetry/sdk-trace-web required peer dependencies instead of optional. This fixes module resolution errors when importing from the main entry point.


clayroach:
Done

clayroach:
@tim-smart looks like this gets addressed in 4.0.0? https://github.com/Effect-TS/effect-smol/blob/main/packages/effect/src/unstable/observability/OtlpTracer.ts. Please confirm.

tim-smart:
> @tim-smart looks like this gets addressed in 4.0.0? https://github.com/Effect-TS/effect-smol/blob/main/packages/effect/src/unstable/observability/OtlpTracer.ts. Can you confirm?

No, that is the effect native tracer which doesn't depend on the otel libraries.

packages/opentelemetry/package.json
tim-smart: I think just get rid of all the optional deps

.changeset/move-sdk-to-subpath-exports.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/opentelemetry": patch
```
tim-smart:
```suggestion
"@effect/opentelemetry": minor
```

packages/opentelemetry/package.json
clayroach: All?  including `@opentelemetry/api`, `@opentelemetry/sdk-trace-base`, etc?

tim-smart: yes, they will all get used if the index barrel file is imported

.changeset/move-sdk-to-subpath-exports.md
tim-smart:
```suggestion
"@effect/opentelemetry": minor
```

PR #5961, opened 2026-01-07
doc: Clean up ZIO refs in Stream and Channel docs.

dpnova:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [X] Documentation Update

## Description

Just noticed in a hover state in the live docs for Stream there was a ZIO reference, so just removed the other one I spotted in that same file. Happy to to tidy up any other ones you'd like me to chase down. For now I just kept it to this file.

### DIFF
--- a/packages/effect/src/Channel.ts
+++ b/packages/effect/src/Channel.ts
@@ -1314,7 +1314,7 @@ export const mapOutEffect: {
 } = channel.mapOutEffect
 
 /**
- * Creates a channel that is like this channel but the given ZIO function gets
+ * Creates a channel that is like this channel but the given Effect function gets
  * applied to each emitted output element, taking `n` elements at once and
  * mapping them in parallel.
  *


packages/effect/src/Stream.ts
```diff
@@ -4432,7 +4432,7 @@ export const runIntoQueue: {
 } = internal.runIntoQueue
 
 /**
- * Like `Stream.runIntoQueue`, but provides the result as a scoped [[ZIO]]
+ * Like `Stream.runIntoQueue`, but provides the result as a scoped [[Effect]]
```
gcanti: I would remove the double parens here `[[Effect]]`

dpnova: thanks @gcanti - changed in ea85a3056

dpnova: or remove them all together? 

I've removed both - I don't see any convention with the squares in this file at least.

45ebfadc9

dpnova:
I removed the ZIO ref in the channel docs too - the rest of the ZIO refs in the codebase are either correct or referring to issues from the original ZIO tracker.

gcanti (approved):

PR #5981, opened 2026-01-14
Fix Array.flatten type inference loss (#5963)

### LINKED ISSUE #5963: Array.flatten losing type information

3.19.13

Not sure you can reproduce the bug with this but the situation is similar to the one I describe here, also I propose a solution at the end of this message.

```
declare const arg1: Effect.Effect<
  void,
  Error1 | Error2,
  Requirement1 | Requirement2
>;

declare const arg2: Effect.Effect<
  void,
  Error1 | Error2,
  Requirement1 | Requirement2 | Requirement3
>

const arg1BeforeArg2 = pipe(
    [
      arg1,
      arg2
    ],
    Array.flatten,
  );

const arg2BeforeArg1 = pipe(
    [
      arg2,
      arg1
    ],
    Array.flatten,
  );
```

Same signature for both `arg1BeforeArg2` and `arg2BeforeArg1`

```
arg1BeforeArg2;
// ^
// Effect.Effect<
//   void,
//   Error1 | Error2,
//   Requirement1 | Requirement2 | Requirement3
// >

arg2BeforeArg1;
// ^
// Effect.Effect<
//   void,
//   Error1 | Error2,
//   Requirement1 | Requirement2 | Requirement3
// >
```

### What do you see instead?

```
arg1BeforeArg2;
// ^
// Effect.Effect<
//   void,
//   Error1 | Error2,
//   Requirement1 | Requirement2
// >

arg1BeforeArg2;
// ^
// Effect.Effect<
//   void,
//   Error1 | Error2,
//   Requirement1 | Requirement2 | Requirement3
// >
```

The order of the items in the array shouldn't change the union type for `Requirements`

### Additional information

I was able to make it work by changing the type returned by the `flatten` function to be the one from this great library: https://github.com/inocan-group/inferred-types/blob/12da2b1282f2bab01f1e1d450aa87aa3bed85bcc/modules/types/src/lists/Flatten.ts

But it's really complex type 😅.

```ts
// packages/effect/src/Array.ts
import { Flatten } from 'inferred-types';

export const flatten: <const S extends ReadonlyArray<ReadonlyArray<any>>>(self: S) => Flatten<S> = flatMap(
  identity
) as any
```

Also by adding a `const` type parameter in `pipe`

```ts
// packages/effect/src/Pipeable.ts
export interface Pipeable {
  pipe<const A>(this: A): A
  pipe<const A, B = never>(this: A, ab: (_: A) => B): B
  pipe<const A, B = never, C = never>(this: A, ab: (_: A) => B, bc: (_: B) => C): C
  ...
}
```

```ts
// packages/effect/src/Function.ts
export function pipe<const A>(a: A): A
export function pipe<const A, B = never>(a: A, ab: (a: A) => B): B
export function pipe<const A, B = never, C = never>(
  a: A,
  ab: (a: A) => B,
  bc: (b: B) => C
): C
...
```

with both those changes, the type is inferred correctly.

bxff:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR resolves a type inference regression in `Array.flatten` where contravariant type parameters (like `Effect` requirements) were being prematurely unified, causing loss of precision in requirement unions.

### Problem Statement

When flattening nested arrays containing types with contravariant parameters, TypeScript's inference would collapse union types in contravariant positions. For example, flattening `[[Effect<R1 | R2>], [Effect<R1 | R2 | R3>]]` would incorrectly lose the `R3` requirement in the resulting union.

### Solution Overview

The fix implements two key improvements:

1. **Distributive Indexed Access**: Updated `ReadonlyArray.Flatten` to use `T[number][number]` instead of `infer A`. This correctly extracts and unions element types from nested arrays while preserving contravariant structure.

2. **Const Generic Modifier**: Added `const` modifier to the `S` generic parameter in the `flatten` function, enabling optional precision preservation for literal tuple structures.

### Core Changes

- **[packages/effect/src/Array.ts](file:///Users/dexdevlon/Developer/Github/effect/packages/effect/src/Array.ts)**:
  - `ReadonlyArray.Flatten`: Replaced `infer A` with distributive indexed access `T[number][number]`
  - `flatten`: Added `const` modifier to generic parameter `S`

- **[package.json](file:///Users/dexdevlon/Developer/Github/effect/package.json)**:
  - Restored `@effect/docgen` to stable version `^0.5.2` (fixes 404 from preview build)

- **Regression Tests**:
  - Added `Array.flatten` inference tests preserving `Effect` requirement unions in `dtslint/Array.tst.ts`
  - Added `pipe` literal tuple handling test in `dtslint/Function.tst.ts`

### Verification Results

**Type-level testing**: Full `dtslint` suite passes successfully:
- **55** test files, **772** tests, **3523** assertions

### Reproduction Example

```typescript
interface Effect<R> { 
  readonly _R: (_: R) => void 
}

interface R1 { _r1: unique symbol }
interface R2 { _r2: unique symbol }
interface R3 { _r3: unique symbol }

declare const arg1: Effect<R1 | R2>
declare const arg2: Effect<R1 | R2 | R3>

// Result element type is now correctly Eff<R1 | R2 | R3> | Eff<R1 | R2>
// which correctly preserves the union of requirements when simplified.
const result = Array.flatten([[arg1], [arg2]])
```

All contributing guidelines, tests, and documentation requirements have been satisfied.

## Related

- Related Issue #
- Closes #5963

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -53,8 +53,8 @@ importers:
         specifier: ^0.8.3
         version: 0.8.8
       '@effect/docgen':
-        specifier: ^0.5.2
-        version: 0.5.2(tsx@4.20.3)(typescript@5.8.3)
+        specifier: https://pkg.pr.new/Effect-TS/docgen/@effect/docgen@fd06738
+        version: https://pkg.pr.new/Effect-TS/docgen/@effect/docgen@fd06738(tsx@4.20.3)(typescript@5.8.3)
       '@effect/eslint-plugin':
         specifier: ^0.3.2
         version: 0.3.2
@@ -1704,13 +1704,14 @@ packages:
     engines: {node: '>=16.17.1'}
     hasBin: true
 
-  '@effect/docgen@0.5.2':
-    resolution: {integrity: sha512-gqBxAhp58R18vT5+ORobWRQ/2MaF5vH0k1zggSct54J41k8TKF5mYIW1qG5tkOVCejet+8K5MKsWK3gzIkaoMw==}
+  '@effect/docgen@https://pkg.pr.new/Effect-TS/docgen/@effect/docgen@fd06738':
+    resolution: {tarball: https://pkg.pr.new/Effect-TS/docgen/@effect/docgen@fd06738}
+    version: 0.5.2
     engines: {node: '>=18.0.0'}
     hasBin: true
     peerDependencies:
-      tsx: ^4.1.0
-      typescript: ^5.2.2
+      tsx: ^4.19.3
+      typescript: ^5.8.2
 
   '@effect/eslint-plugin@0.3.2':
     resolution: {integrity: sha512-c4Vs9t3r54A4Zpl+wo8+PGzZz3JWYsip41H+UrebRLjQ2Hk/ap63IeCgN/HWcYtxtyhRopjp7gW9nOQ2Snbl+g==}
@@ -5173,7 +5174,6 @@ packages:
 
   libsql@0.4.7:
     resolution: {integrity: sha512-T9eIRCs6b0J1SHKYIvD8+KCJMcWZ900iZyxdnSCdqxN12Z1ijzT+jY5nrk72Jw4B0HGzms2NgpryArlJqvc3Lw==}
-    cpu: [x64, arm64, wasm32]
     os: [darwin, linux, win32]
 
   lighthouse-logger@1.4.2:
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "@changesets/cli": "^2.29.4",
     "@edge-runtime/vm": "^5.0.0",
     "@effect/build-utils": "^0.8.3",
-    "@effect/docgen": "https://pkg.pr.new/Effect-TS/docgen/@effect/docgen@fd06738",
```
tim-smart, line 43: This needs to be reverted :)

bxff: Reverted.

tim-smart: Will need to update the lock file too

bxff: Apologies, forgot about that.

PR #5983, opened 2026-01-14
fix(cli): allow options after positional arguments

cevr:
## Summary

This PR fixes a usability issue in `@effect/cli` where options were required to appear before positional arguments.

**Before:**
```bash
myapp deploy --force staging  # worked
myapp deploy staging --force  # failed: "Received unknown argument: '--force'"
```

**After:**
```bash
myapp deploy --force staging  # works
myapp deploy staging --force  # works
```

## Problem

When using `@effect/cli`, options had to come **before** positional arguments. This is counter to how most CLI tools work (git, npm, docker, etc.) which allow options anywhere in the command.

The issue was in `parseCommandLine` for `Single` options - it only checked the first argument and if it wasn't a matching option, it failed with `MissingFlag`. Everything after the first positional arg was left as "leftover" and reported as "unknown argument".

## Solution

Updated the `parseCommandLine` function for `Single` options to scan through **all** arguments to find the option, regardless of position. Arguments before and after the option are collected as leftover for the positional argument parser.

The fix:
1. Scans through all arguments looking for the option name
2. If found, extracts the option and its value (for non-boolean options)
3. Returns all other arguments as leftover for positional arg parsing
4. Preserves existing auto-correct suggestions behavior

## Test Plan

Added tests in:
- `Options.test.ts` - 8 new tests for options after positional args
- `Command.test.ts` - 2 new end-to-end tests

Tests cover:
- Text options after positional args
- Text options with aliases after positional args  
- Boolean options after positional args
- Multiple options interspersed with positional args
- `--option=value` syntax after positional args
- Boolean options with explicit true/false values

### DIFF
--- a/packages/cli/src/internal/commandDescriptor.ts
+++ b/packages/cli/src/internal/commandDescriptor.ts
@@ -757,7 +757,6 @@ const parseInternal = (
                       if (!InternalCommandDirective.isUserDefined(childDirective)) {
                         return Effect.succeed(childDirective)
                       }
-                      // Check if child has leftover args that might be parent options
                       const childLeftover = childDirective.leftover
                       if (Arr.isEmptyReadonlyArray(childLeftover)) {
                         return Effect.succeed(InternalCommandDirective.userDefined(childLeftover, {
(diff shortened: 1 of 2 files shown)


cevr:
cant get it install locally because of @effect/docgen being custom. so i'm doing it the good ol' remote way

PR #5998, opened 2026-01-24
fix(@effect/sql-pg): readded stream to PgClientConfig

Brandon-Perry:
## Type

- [ ] Refactor
- [ ] Feature
- [ x ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In an older version of sql-pg, stream as an optional parameter was available, and has since been removed. I wanted to add it back since it doesn't seem like there's an alternative and sql-pg is just a wrapper around Pg.Pool. 

I'm assuming this was just accidentally removed during the previous refactor. If there's an Effect-ful way of doing this let me know!

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/sql-pg/src/PgClient.ts
+++ b/packages/sql-pg/src/PgClient.ts
@@ -123,7 +123,7 @@ export const make = (
       password: options.password ? Redacted.value(options.password) : undefined,
       ssl: options.ssl,
       port: options.port,
-      stream: options.stream ? options.stream : () => undefined,
+      stream: options.stream!,
       connectionTimeoutMillis: options.connectTimeout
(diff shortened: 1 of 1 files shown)


packages/sql-pg/src/PgClient.ts
```diff
@@ -77,6 +78,8 @@ export interface PgClientConfig {
   readonly username?: string | undefined
   readonly password?: Redacted.Redacted | undefined
 
+  readonly stream?: () => Duplex | undefined
```
tim-smart:
```suggestion
  readonly stream?: (() => Duplex) | undefined
```

PR #6000, opened 2026-01-26
test/docs(cli): document and test parent options after subcommand arguments

cevr:
## Summary

Follow-up to #5983. Adds tests and documentation for parent options appearing after subcommand arguments — a behavior #5983 already enabled via the `parseCommandLine` scan and child-leftover re-parse path.

- **Tests**: 30+ new tests covering parent options after subcommand args, shared option resolution (child wins), boolean options, `=` syntax, short aliases, and the `provideSync` service pattern
- **Docs**: Updated README to reflect flexible option placement with concrete `@effect/cli` code examples, and corrected the FAQ which previously stated options must precede subcommands/args
- **Cleanup**: Removed redundant `extractParentOptionsFromArgs` extraction layer (and all supporting helpers) that duplicated what #5983's re-parse path already handles — zero source diff on `commandDescriptor.ts`

### Key behaviors documented and tested

- `app push --force --verbose` — parent `--verbose` after subcommand option
- `app show main --verbose` — parent `--verbose` after subcommand positional arg
- `app --verbose push --force` — parent `--verbose` before subcommand (existing)
- `tool exec task --verbose` — shared option goes to child when after subcommand
- `tool --verbose exec task` — shared option goes to parent when before subcommand

## Test plan

- [x] `pnpm --filter @effect/cli test` — 182/182 pass
- [x] `pnpm --filter @effect/cli check` — clean
- [x] Verified new tests fail on pre-#5983 code (`Received unknown argument`)

### DIFF
--- a/.changeset/cli-parent-options-shared.md
+++ b/.changeset/cli-parent-options-shared.md
@@ -1,0 +1,9 @@
+---
+"@effect/cli": patch
+---
+
+Fix shared options and edge cases for parent options after subcommand arguments
+
+When a parent command and subcommand define the same option (e.g., both have `--verbose`), the subcommand now wins for options appearing after the subcommand. This matches CLI conventions (e.g., `git status --verbose` uses status's verbose, not git's).
+
+Also fixes boolean parent options incorrectly consuming the next token as a value, and respects `--` as a separator during parent option extraction.
--- a/.changeset/cli-parent-options-shared.md
+++ b/.changeset/cli-parent-options-shared.md
@@ -1,9 +1,0 @@
----
-"@effect/cli": patch
----
-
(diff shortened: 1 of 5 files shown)


IMax153:
@cevr - I don't think the PR can be accepted in it's current form. 

We would need to document the precise behavior of what occurs when a parent command and subcommand _share_ a common option. We would also need to document that that this behavior is _only_ valid for options (not arguments).

In addition, we would need several more tests to ensure the implemented behavior is what we expect.

fubhy:
Closing since Effect v3 is feature-frozen and the requested behavior was confirmed to work in Effect v4. If docs or tests are still needed for v4, please open the equivalent PR there.

PR #6011, opened 2026-01-29
Wrap primitive tool results in object for Gemini API compatibility

### LINKED ISSUE #6008: @effect/ai-google: Gemini rejects primitive tool results (requires object)
- `@effect/ai-google`: `0.12.1`
- `effect`: `3.19.15`

Define a tool with a primitive `success` or `failure` schema (e.g., `Schema.String`, `Schema.Number`):

```ts
const SearchTool = Tool.make("search", {
  description: "Search for items",
  parameters: { query: Schema.String },
  success: Schema.String, // or failure: Schema.String with failureMode: "return"
})
```

When the tool executes and returns a primitive value, the agentic loop fails on iteration 2 (when the tool result is sent back to Gemini) with:

```
Invalid value at 'contents[2].parts[0].function_response.response' (type.googleapis.com/google.protobuf.Struct), "Found 5 results"
```

Primitive tool results should work with Gemini.

### What do you see instead?

Gemini's `function_response.response` field requires a `Struct` (object), but primitive values are sent directly, causing a 400 error when the tool result is sent back on iteration 2.

lucas-barake:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Gemini's `function_response.response` field requires a `Struct` (i.e. a JSON object), but `@effect/ai-google` was passing `part.result` directly, which fails with a 400 error when the tool result is a primitive value (string, number, boolean, etc.).

This wraps primitive tool results in `{ result: value }` using `Predicate.isRecord` to detect when wrapping is needed. Plain objects pass through unchanged.

## Related

- Related Issue #6008
- Closes #6008

### DIFF
--- a/.changeset/fix-google-primitive-tool-results.md
+++ b/.changeset/fix-google-primitive-tool-results.md
@@ -1,0 +1,5 @@
+---
+"@effect/ai-google": patch
+---
+
+Wrap primitive tool results in an object before sending to Gemini
--- a/packages/ai/google/src/GoogleLanguageModel.ts
+++ b/packages/ai/google/src/GoogleLanguageModel.ts
@@ -386,11 +386,12 @@ const prepareMessages: (
           const parts: Array<typeof Generated.Part.Encoded> = []
 
           for (const part of message.content) {
+            const response = Predicate.isRecord(part.result) ? part.result : { result: part.result }
             parts.push({
               functionResponse: {
                 id: part.id,
                 name: part.name,
-                response: part.result as any
+                response: response as any
               }
             })
           }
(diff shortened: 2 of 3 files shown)


fubhy:
Thanks for the fix. This still looks relevant, but before maintainer review please add a regression test covering primitive Gemini tool results being wrapped into an object, and refresh the branch/checks if needed.

lucas-barake:
Rebased onto latest main and added a regression test. It captures the encoded request and asserts that string, number, boolean, null and array results all get wrapped as `{ result: ... }`, and that object results pass through unchanged.

Also re-checked the current API docs before refreshing this: `functionResponse.response` is still typed as a Struct, so it has to be a JSON object, and "result" is one of the wrapper keys the docs themselves suggest. So the fix is still valid as-is.

PR #6018, opened 2026-02-02
fix(Match): handle null/undefined in Match.tag and Match.tagStartsWith

### LINKED ISSUE #6017: Match.tag crashes on union type including null/undefined

3.19.15

When using `Match.tag` on union type that includes undefined/null, it will produce runtime error when the `Match.tag` is before `Match.when(undefined, ...)`. The code type checks correctly.

Reproduciton steps

```ts
import { Match } from "effect"

const value = undefined as ({_tag: "A"} | undefined)

const res = Match.value(value).pipe(
  Match.tag('A', () => 'mached A'), // <- crashes with `Cannot read properties of undefined (reading '_tag')`
  Match.when(undefined, () => 'matched undefined'),
  Match.exhaustive
);
console.log(res);
```

Or in playground: https://effect.website/play/#b2df3d05e29a

The code does not crash

### What do you see instead?

_No response_

### Additional information

_No response_

codewithkenzo:
## Summary

Fixes #6017

`Match.tag` and `Match.tagStartsWith` now correctly handle nullable union types instead of crashing with "Cannot read properties of undefined (reading '_tag')".

## Problem

When using `Match.tag` on a union type that includes `null` or `undefined`, the predicate would crash because it accessed `_[field]` without first checking if `_` is nullish:

```typescript
const value = undefined as ({_tag: "A"} | undefined)
const res = Match.value(value).pipe(
  Match.tag('A', () => 'matched A'),  // crashes!
  Match.when(undefined, () => 'matched undefined'),
  Match.exhaustive
)
```

## Solution

Add `_ != null &&` guard to the predicates in `discriminator` and `discriminatorStartsWith`, following the same pattern already used in `discriminators` (which powers `Match.tags`).

## Changes

- `packages/effect/src/internal/matcher.ts`: Add null checks to predicates
- `packages/effect/test/Match.test.ts`: Add test cases for nullable unions

## Testing

Added 3 test cases:
- `Match.tag` with `undefined` union
- `Match.tag` with `null` union  
- `Match.tagStartsWith` with `undefined` union

All 62 Match tests pass.

### DIFF
--- a/.changeset/fix-match-tag-nullable.md
+++ b/.changeset/fix-match-tag-nullable.md
@@ -1,0 +1,9 @@
+---
+"effect": patch
+---
+
+fix(Match): handle null/undefined in `Match.tag` and `Match.tagStartsWith`
+
+Added null checks to `discriminator` and `discriminatorStartsWith` predicates to prevent crashes when matching nullable union types.
+
+Fixes #6017
--- a/packages/effect/src/internal/matcher.ts
+++ b/packages/effect/src/internal/matcher.ts
@@ -351,8 +351,8 @@ export const discriminator = <D extends string>(field: D) =>
   const f = pattern[pattern.length - 1]
   const values: Array<P> = pattern.slice(0, -1) as any
   const pred = values.length === 1
-    ? (_: any) => _[field] === values[0]
-    : (_: any) => values.includes(_[field])
+    ? (_: any) => _ != null && _[field] === values[0]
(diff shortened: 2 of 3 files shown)


codewithkenzo:
This follows my comment on #6017. The fix adds the same `_ != null &&` guard that `discriminators` (used by `Match.tags`) already has.

Minimal change - just 3 lines modified in the predicates.

PR #6019, opened 2026-02-02
fix(HttpClient): backport retryTransient fixes from v4

### LINKED ISSUE #5995: `HttpClient.retryTransient` has inconsistent behavior

effect@3.19.14 @effect/platform@0.94.1

see:
* https://github.com/lewxdev/effect-repro-1
* https://github.com/lewxdev/effect-repro-1/actions/workflows/test.yaml

```ts
import { HttpClient, HttpClientResponse } from "@effect/platform";
import { describe, expect, it } from "@effect/vitest";
import { Effect, Fiber, Layer, Ref, Schedule, TestClock } from "effect";

const DURATION = 1000;
const RECUR = 3;

const httpClient = HttpClient.make((request, url) =>
  Effect.sync(() => {
    const status = Number(url.pathname.slice(1));
    const response = new Response(null, { status });
    return HttpClientResponse.fromWeb(request, response);
  })
);

const retryTransient = Layer.succeed(
  HttpClient.HttpClient,
  httpClient.pipe(
    HttpClient.retryTransient({
      schedule: Schedule.spaced(DURATION),
      times: RECUR,
    }),
  ),
);

const retryTransientResponseOnly = Layer.succeed(
  HttpClient.HttpClient,
  httpClient.pipe(
    HttpClient.retryTransient({
      mode: "response-only",
      schedule: Schedule.spaced(DURATION),
      times: RECUR,
      while: (response) => response.status === 408,
    }),
  ),
);

const testResponse = ({ status, isTransient }: { status: number; isTransient: boolean }) =>
  Effect.gen(function*() {
    const attemptsExpected = isTransient ? RECUR + 1 : 1;
    const attemptsRef = yield* Ref.make(0);
    const client = yield* Effect.map(
      HttpClient.HttpClient,
      HttpClient.tapRequest(() => Ref.update(attemptsRef, (n) => n + 1)),
    );

    const fiber = yield* Effect.fork(client.get(`http://_/${status}`));
    yield* TestClock.adjust(DURATION * RECUR);

    const response = yield* Fiber.join(fiber);
    expect(response.status).toBe(status);
    const attempts = yield* Ref.get(attemptsRef);
    expect(attempts).toBe(attemptsExpected);
  });

describe("HttpClient.HttpClient", () => {
  const responses = [
    { status: 200, isTransient: false },
    { status: 408, isTransient: true },
    { status: 500, isTransient: true },
    { status: 501, isTransient: false },
  ];
  const tests = [
    { name: "retryTransient", layer: retryTransient },
    { name: "retryTransient mode: 'response-only'", layer: retryTransientResponseOnly },
  ];

  for (const test of tests) {
    describe(test.name, () => {
      it.layer(test.layer)(({ effect }) => {
        for (const opts of responses) {
          effect(`GET /${opts.status}`, () => testResponse(opts));
        }
      });
    });
  }
});
```

Based on documentation, naming, and common HTTP semantics:
* HTTP 408 (Request Timeout) should be considered transient.
* HTTP 500 (Internal Server Error) should be considered transient.
* HTTP 501 (Not Implemented) should NOT be considered transient.

`HttpClient.retryTransient` should ALWAYS retry transient responses, and the `while` Predicate should add conditions to the schedule, not suppress expected ones (as the current JSDoc implies).

### What do you see instead?

* HTTP 408 is not retried, even when `"response-only"` mode is used and a `while` Predicate explicitly filters for responses with the 408 status.
* HTTP 500 is not retried in some configurations unless explicitly included in while.
* HTTP 501 is retried, even though it should not be considered transient.

The behavior differs significantly depending on whether `while` is provided, even for status codes not mentioned in the predicate.

### Additional information

_No response_

codewithkenzo:
## Summary

Fixes #5995

`HttpClient.retryTransient` now correctly identifies transient HTTP status codes.

## Problem

The previous logic `response.status >= 429` was incorrect:
- **408 (Request Timeout)** was NOT retried (408 < 429) — should be transient
- **501 (Not Implemented)** WAS retried (501 >= 429) — should NOT be transient (permanent)
- **505+ codes** were all retried — most are permanent errors

## Solution

Changed `isTransientResponse` from a simple comparison to an explicit allowlist matching industry standards (axios-retry, got, ky):

```typescript
// Before
const isTransientResponse = (response) => response.status >= 429

// After
const isTransientResponse = (response) => {
  const status = response.status
  return (
    status === 408 || // Request Timeout
    status === 429 || // Too Many Requests
    status === 500 || // Internal Server Error
    status === 502 || // Bad Gateway
    status === 503 || // Service Unavailable
    status === 504    // Gateway Timeout
  )
}
```

## Changes

- `packages/platform/src/internal/httpClient.ts`: Fix `isTransientResponse` logic
- `packages/platform/test/HttpClient.test.ts`: Add 9 tests for transient/non-transient status codes

## Testing

Added tests verifying:
- ✅ 408, 429, 500, 502, 503, 504 ARE retried (3 attempts with `times: 2`)
- ✅ 200, 501, 505 are NOT retried (1 attempt)

All 21 HttpClient tests pass.

## Note

I also noticed the `while` predicate uses `AND` for responses but `OR` for errors (see my comment on #5995). Happy to address that in a follow-up PR if desired.

### DIFF
--- a/packages/platform/src/HttpClient.ts
+++ b/packages/platform/src/HttpClient.ts
@@ -534,7 +534,7 @@ export const retryTransient: {
   >(
     options: {
       readonly mode?: Mode | undefined
-      readonly while?: Predicate.Predicate<NoInfer<Input>>
+      readonly while?: Predicate.Predicate<NoInfer<E>>
       readonly schedule?: Schedule.Schedule<B, NoInfer<Input>, R1>
       readonly times?: number
     } | Schedule.Schedule<B, NoInfer<Input>, R1>
@@ -552,7 +552,7 @@ export const retryTransient: {
     self: HttpClient.With<E, R>,
     options: {
       readonly mode?: Mode | undefined
-      readonly while?: Predicate.Predicate<NoInfer<Input>>
+      readonly while?: Predicate.Predicate<NoInfer<E>>
       readonly schedule?: Schedule.Schedule<B, NoInfer<Input>, R1>
       readonly times?: number
     } | Schedule.Schedule<B, NoInfer<Input>, R1>
--- a/.changeset/fix-retry-transient-status-codes.md
+++ b/.changeset/fix-retry-transient-status-codes.md
@@ -1,0 +1,11 @@
+---
+"@effect/platform": patch
+---
+
+Fix `retryTransient` to use correct transient status codes
+
+Changed `isTransientResponse` from `status >= 429` to an explicit allowlist (408, 429, 500, 502, 503, 504). This correctly excludes 501 (Not Implemented) and 505+ permanent errors, while including 408 (Request Timeout) which was previously missed.
+
+Also aligned response retry behavior with v4: the `while` predicate now only applies to error retries, not response retries. Response retries are determined solely by `isTransientResponse`. This matches the semantic intent since `while` is typed for errors, not responses.
+
+Fixes #5995
--- a/packages/platform/src/internal/httpClient.ts
+++ b/packages/platform/src/internal/httpClient.ts
@@ -755,7 +755,7 @@ export const retryTransient: {
   >(
     options: {
       readonly mode?: Mode | undefined
-      readonly while?: Predicate.Predicate<NoInfer<Input>>
+      readonly while?: Predicate.Predicate<NoInfer<E>>
       readonly schedule?: Schedule.Schedule<B, NoInfer<Input>, R1>
       readonly times?: number
     } | Schedule.Schedule<B, NoInfer<Input>, R1>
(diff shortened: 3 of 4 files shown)


codewithkenzo:
@tim-smart thank you for pointing out the other diffs!

Backported the remaining v4 fixes: ES type param, while predicate typing, and removed the Predicate.and from response retries.

Should match v4 exactly now... also noticed that you keep comments to a minimum, so I tried to stick to that approach as much as possible.

I kept v3's error structure and timeout check since those are structural changes, SQL artifacts also gone now, my bad :)

packages/platform/src/HttpClient.ts
```diff
@@ -526,6 +526,7 @@ export const retryTransient: {
   <
     B,
     E,
+    ES = never,
```
tim-smart: Schedules don't have errors in v3, so you don't need this.

tim-smart (changes requested):

PR #6021, opened 2026-02-02
fix(platform): properly join URL paths in HttpClientRequest.appendUrl

### LINKED ISSUE #5971: HttpClientRequest.appendUrl should use proper URL concatenation methods

`HttpClientRequest.appendUrl` just appends whatever string you give to it at the end of the current client base URL. I think it should instead follow the same mechanism as `new URL(path,base)`

Here you have a reproduction where you see how it behaves: https://effect.website/play#3b8b4bfe0a61

Use proper url concatenation mechanisms, the equivalent of doing `new URL(newPath,existingUrlInContext)`

for now, just being careful, or not using appendUrl

codewithkenzo:
## Summary

Fixes `HttpClientRequest.appendUrl` to properly join URL paths instead of using simple string concatenation.

Closes #5971

## Problem

The current implementation could produce invalid URLs:
```typescript
// Before (broken):
HttpClientRequest.get("https://api.example.com/v1").pipe(
  HttpClientRequest.appendUrl("users")
)
// Result: "https://api.example.com/v1users" (missing slash!)
```

## Solution

Ensure proper path joining by:
1. Adding a trailing slash to base URL if missing
2. Removing leading slash from path if present

```typescript
// After (fixed):
HttpClientRequest.get("https://api.example.com/v1").pipe(
  HttpClientRequest.appendUrl("users")
)
// Result: "https://api.example.com/v1/users"
```

This matches the behavior of other HTTP client libraries like axios, got, and ky.

## Test Cases

| Base URL | Path | Result |
|----------|------|--------|
| `https://api.com/v1` | `users` | `https://api.com/v1/users` |
| `https://api.com/v1/` | `users` | `https://api.com/v1/users` |
| `https://api.com/v1` | `/users` | `https://api.com/v1/users` |
| `https://api.com/v1/` | `/users` | `https://api.com/v1/users` |

## Verification

- [x] Type check passes (`pnpm check`)
- [x] All tests pass (18/18 in HttpClient.test.ts)
- [x] Changeset added

### DIFF
--- a/packages/platform/test/HttpClient.test.ts
+++ b/packages/platform/test/HttpClient.test.ts
@@ -202,6 +202,50 @@ describe("HttpClient", () => {
     )
   })
 
+  describe("appendUrl", () => {
+    it("joins path without trailing slash on base", () => {
+      const request = HttpClientRequest.get("https://api.example.com/v1").pipe(
+        HttpClientRequest.appendUrl("users")
+      )
+      strictEqual(request.url, "https://api.example.com/v1/users")
+    })
+
+    it("joins path with trailing slash on base", () => {
+      const request = HttpClientRequest.get("https://api.example.com/v1/").pipe(
+        HttpClientRequest.appendUrl("users")
+      )
+      strictEqual(request.url, "https://api.example.com/v1/users")
+    })
+
+    it("joins path with leading slash", () => {
+      const request = HttpClientRequest.get("https://api.example.com/v1").pipe(
+        HttpClientRequest.appendUrl("/users")
+      )
+      strictEqual(request.url, "https://api.example.com/v1/users")
+    })
+
+    it("joins path with both trailing and leading slashes", () => {
+      const request = HttpClientRequest.get("https://api.example.com/v1/").pipe(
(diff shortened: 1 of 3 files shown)


packages/platform/test/HttpClient.test.ts
```diff
@@ -202,6 +202,50 @@ describe("HttpClient", () => {
...
+        HttpClientRequest.appendUrl("users/123/posts")
+      )
+      strictEqual(request.url, "https://api.example.com/v1/users/123/posts")
+    })
+
+    it("handles empty path", () => {
+      const request = HttpClientRequest.get("https://api.example.com/v1").pipe(
+        HttpClientRequest.appendUrl("")
+      )
+      strictEqual(request.url, "https://api.example.com/v1/")
```
tim-smart:
```suggestion
      strictEqual(request.url, "https://api.example.com/v1")
```

codewithkenzo: Nice catch, updated so empty path returns the request unchanged, thank you!

PR #6022, opened 2026-02-02
fix(cli): fix Prompt.text clear when input wraps terminal lines

### LINKED ISSUE #5978: Prompt.select renders duplicate lines when choice descriptions cause terminal line wrapping
## Description

When using `Prompt.select` with choices that have long descriptions that wrap across terminal lines, navigating up/down causes duplicate prompt lines to appear instead of updating in place.

## Reproduction

```ts
import { Prompt } from "@effect/cli"
import { NodeContext, NodeRuntime } from "@effect/platform-node"

  const choice = yield* Prompt.select({
    message: "Select option:",
    choices: [
      { title: "opt-1", value: "a", description: "This is a very long description that will wrap to multiple lines in a narrow terminal window causing rendering issues" },
      { title: "opt-2", value: "b", description: "Another long description that exceeds typical terminal width and causes wrapping when displayed" },
      { title: "opt-3", value: "c", description: "Yet another verbose description to demonstrate the rendering bug in the select prompt" },
    ],
  })
  yield* Effect.log(\`Selected: \${choice}\`)
})

program.pipe(Effect.provide(NodeContext.layer), NodeRuntime.runMain)
```

Run with a narrow terminal (< 100 columns) and press up/down arrows:

```
? Select option: › 
? Select option: › 
? Select option: › 
? Select option: › 
```

Each navigation prints a new prompt line instead of updating in place.

## Root Cause

In `packages/cli/src/internal/prompt/select.ts`, `handleClear` calculates lines to erase incorrectly:

```ts
const text = "\n".repeat(Math.min(options.choices.length, options.maxPerPage)) + options.message
const clearOutput = InternalAnsiUtils.eraseText(text, columns)
```

**Problem 1**: Uses empty newlines to represent choices. `eraseText` counts 1 row per empty line, but actual rendered choices have content (`prefix + title + description`) that may wrap to multiple terminal rows.

**Problem 2**: The `clear` handler ignores state:

```ts
clear: () => handleClear(opts)  // state ignored
```

But descriptions are only rendered for the **selected** choice (in `renderChoiceDescription`):

```ts
if (!choice.disabled && choice.description && isSelected) { ... }
```

Without knowing `state`, `handleClear` can't determine which choice line includes a description and will wrap.

## Expected Behavior

The prompt should clear and redraw cleanly when navigating, regardless of description length or terminal width.

## Environment

- @effect/cli version: 0.72.1
- Platform: macOS/Linux
- Terminal: Any terminal with width < total line length

m9tdev:
## Summary

When the user's input in `Prompt.text` exceeds the terminal width and wraps to a new line, `renderClearScreen` miscalculates lines to erase because it only considers `options.message`, not the full rendered line including the input value.

This causes duplicate prompt lines to be printed on each keystroke after the text wraps:

```
? What would you like to say? › what do you propose to add to play around with? it can be
? What would you like to say? › what do you propose to add to play around with? it can be
? What would you like to say? › what do you propose to add to play around with? it can be
...
```

## Root Cause

In `renderClearScreen`, line erasure is calculated based only on `options.message`:

```typescript
const clearOutput = InternalAnsiUtils.eraseText(options.message, columns)
```

But the actual rendered line includes: `"? " + message + " › " + inputValue`

## Fix

Calculate the full rendered line length:

```typescript
const inputValue = state.value.length > 0 ? state.value : options.default
const fullLine = `? ${options.message} › ${inputValue}`
const clearOutput = InternalAnsiUtils.eraseText(fullLine, columns)
```

## Related

This is the same root cause as #5978 (for `Prompt.select`), and applies the equivalent fix from #5979 to `Prompt.text`.

Closes #5978

### DIFF
--- a/.changeset/fix-prompt-text-wrap.md
+++ b/.changeset/fix-prompt-text-wrap.md
@@ -1,0 +1,5 @@
+---
+"@effect/cli": patch
+---
+
+Fixed `Prompt.text` rendering duplicate lines when input text wraps to a new terminal line.
--- a/packages/cli/src/internal/prompt/text.ts
+++ b/packages/cli/src/internal/prompt/text.ts
@@ -49,7 +49,10 @@ function renderClearScreen(state: State, options: Options) {
         )
     })
     // Ensure that the prior prompt output is cleaned up
-    const clearOutput = InternalAnsiUtils.eraseText(options.message, columns)
+    // Calculate full rendered line: "? " + message + " › " + input
+    const inputValue = state.value.length > 0 ? state.value : options.default
+    const fullLine = `? ${options.message} \u203a ${inputValue}`
+    const clearOutput = InternalAnsiUtils.eraseText(fullLine, columns)
     // Concatenate and render all documents
     return clearError.pipe(
       Doc.cat(clearOutput),


IMax153:
> Thanks @m9tdev ! Would you be willing to test the `Prompt` impl in Effect v4 to see if a similar issue exists and open an issue there if so? If not, I can take a look.

Actually, nevermind. I think Tim's `lalph` bot is picking up backports we need to do. https://github.com/Effect-TS/effect-smol/pull/1159

PR #6035, opened 2026-02-05
EFF-412: Backport client interrupt error handling

tim-smart:
## Summary
- Backport effect v4 `causeResponse` interrupt scanning so a client abort interrupt (fiber id `-499`) yields HTTP 499 even when other interrupts appear earlier in the cause tree.
- Add regression test covering sequential interrupt causes.

## Notes
- v4 reference: `.repos/effect-v4/packages/effect/src/unstable/http/HttpServerError.ts`.

### DIFF
--- a/packages/platform/src/internal/httpServerError.ts
+++ b/packages/platform/src/internal/httpServerError.ts
@@ -28,53 +28,37 @@ export const clientAbortFiberId = globalValue(
 export const causeResponse = <E>(
   cause: Cause.Cause<E>
 ): Effect.Effect<readonly [HttpServerResponse, Cause.Cause<E>]> => {
-  const initial: readonly [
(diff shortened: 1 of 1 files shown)


packages/platform/src/internal/httpServerError.ts
```diff
@@ -28,17 +28,26 @@ export const clientAbortFiberId = globalValue(
 export const causeResponse = <E>(
   cause: Cause.Cause<E>
 ): Effect.Effect<readonly [HttpServerResponse, Cause.Cause<E>]> => {
-  const [effect, stripped] = Cause.reduce(
+  const initial: readonly [
```
tim-smart: You can just use mutable variables in the closure instead, to keep it simple.

PR #6047, opened 2026-02-09
fix: clean up previous runtime on Bun HMR reload

nounder:
When using `bun --hot`, each module re-evaluation creates a new program causing
memory leaks.

Here we ensure only one active runtime exists at any time by cleaning up
on each hot reload.

### DIFF
--- a/packages/platform-bun/src/BunRuntime.ts
+++ b/packages/platform-bun/src/BunRuntime.ts
@@ -1,11 +1,69 @@
 /**
  * @since 1.0.0
  */
-import * as NodeRuntime from "@effect/platform-node-shared/NodeRuntime"
-import type { RunMain } from "@effect/platform/Runtime"
+import { makeRunMain, type RunMain } from "@effect/platform/Runtime"
+import { constVoid } from "effect/Function"
+import { globalValue } from "effect/GlobalValue"
+
+// keep track of previous main fiber to clear it across hot reloads.
+const hmrState = globalValue(
+  Symbol.for("@effect/platform-bun/BunRuntime/hmrState"),
+  () => ({
+    fiber: undefined as any,
+    keepAlive: undefined as ReturnType<typeof setInterval> | undefined,
+    onSigint: undefined as (() => void) | undefined
+  })
+)
 
 /**
  * @since 1.0.0
  * @category runtime
  */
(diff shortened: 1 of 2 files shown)


packages/platform-bun/src/BunRuntime.ts
```diff
@@ -1,11 +1,69 @@
...
+  // Interrupt previous fiber from prior previous run
+  if (hmrState.fiber) {
+    if (hmrState.onSigint) {
+      process.removeListener("SIGINT", hmrState.onSigint)
+      process.removeListener("SIGTERM", hmrState.onSigint)
+    }
+    if (hmrState.keepAlive) {
+      clearInterval(hmrState.keepAlive)
+    }
+    hmrState.fiber.unsafeInterruptAsFork(hmrState.fiber.id())
```
tim-smart, line 35: Will probably need to wait for the previous fiber to exit before starting another one, in the case there are long running finalizers that could hold ports open.

tim-smart:
> I'm not sure why we need keepAlive timer here

For the case where nothing will keep the event loop busy.

In node this would exit the program immediately, not sure about bun 

```ts
runMain(Effect.async<void>(() => {}))
```

packages/platform-bun/src/BunRuntime.ts
nounder: When bun is `--hot`, Bun automatically updates server handles on subsequent runs so there's no port holding. My initial implementation waited for previous program to finalize but that caused intermissions in-between, similar to what happens when `--watch` is used.

Since `--hot` is explicit and opt-in, I think current behavior makes sense. I think even React hot reloading via react-fast-refresh is executing new tree and running cleanup after vdom compare.

nounder:
Just did a test on Bun and it does indeed exits without keep alive timer. Make sense since it aims for full node compat. In my codebase where I tested it without that timer I must have had some dangling timers.

Also added changeset and a comment about interruption behavior.

notkadez:
+1

fubhy:
Thanks for the PR. Could you update this to address the HMR state race? As written, an older interrupted fiber can finish after a newer run starts and clear the newer run's stored fiber/timer/signal handler state. Please also clarify or handle the earlier concern about long-running finalizers overlapping the next run.

PR #6056, opened 2026-02-12
fix(cluster): HttpLayerRouter variants for route registration in HttpRunner

0xh3x:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- `HttpRunner.layerHttpOptions` and `layerWebsocketOptions` registered routes on `HttpRouter.Default`, but `HttpLayerRouter.serve` (used by `layerHttp`/`layerWebsocket`)  serves from `HttpLayerRouter.HttpRouter` — a different router tag. This caused POST/GET requests to return 404.
- Switched to `RpcServer.layerProtocolHttpRouter` and `RpcServer.layerProtocolWebsocketRouter` so routes are registered on the correct router.
- Added a regression test verifying `layerHttpOptions` routes are reachable through `HttpLayerRouter`.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/some-icons-dream.md
+++ b/.changeset/some-icons-dream.md
@@ -1,0 +1,5 @@
+---
+"@effect/cluster": patch
+---
+
+use HttpLayerRouter variants for route registration in HttpRunner


0xh3x:
> cluster doesn't support HttpLayerRouter at the moment intentionally (as HttpLayerRouter is still experimental in effect v3). In effect v4 this will change :)

I see.
Does this mean that http transport is broken for BunCluster?
```ts
  const RunnerLive = PingLive.pipe(
    Layer.provideMerge(BunClusterHttp.layer({ transport: "http", storage: "local"}))
  )
  Layer.launch(RunnerLive).pipe(BunRuntime.runMain)
```
`POST` request on `http://localhost:34431/` results in 404
and as I understand the issue is with `HttpLayerRouter`

tim-smart:
Ah yep looks like when I backported some effect v4 changes it got switched to HttpLayerRouter.

tim-smart (approved):

PR #6065, opened 2026-02-15
feat(rpc): add defect schema option to Rpc.make

marbemac:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Right now `@effect/rpc` hardcodes `Schema.Defect` for serializing defects across RPC boundaries. This is fine for external APIs, but for internal service boundaries it strips stack traces and other diagnostic info which makes debugging painful.

This PR adds an optional `defect` parameter to `Rpc.make` that lets you override the defect schema per-RPC. It defaults to `Schema.Defect` so nothing changes for existing code.

```ts
Rpc.make("MyRpc", {
  success: Schema.String,
  defect: Schema.Unknown,
})
```

The custom schema flows through `Rpc.exitSchema()` and the server-side encoding in `RpcServer`. Protocol-level defects (fatal server crashes outside any specific RPC context) still use `Schema.Defect` — those represent infrastructure failures, not application errors.

### Changes

- **`Rpc.ts`**: Added `defectSchema` field to `Rpc` and `AnyWithProps` interfaces, threaded through `makeProto` and all Proto methods, added `defect` option to `Rpc.make`, updated `exitSchema` to use `rpc.defectSchema`
- **`RpcServer.ts`**: Added pre-computed `encodeDefect` to `Schemas` cache, threaded through `handleEncode` and `sendRequestDefect`
- **Tests**: Unit test for `exitSchema` round-trip with custom defect schema, plus full e2e test over HTTP with NDJSON serialization

### DIFF
--- a/.changeset/custom-defect-schema.md
+++ b/.changeset/custom-defect-schema.md
@@ -1,5 +1,5 @@
 ---
-"@effect/rpc": minor
+"@effect/rpc": patch
 ---
 
 Add optional `defect` parameter to `Rpc.make` for customizing defect serialization per-RPC. Defaults to `Schema.Defect`, preserving existing behavior.


packages/rpc/src/Rpc.ts
```diff
@@ -58,6 +58,7 @@ export interface Rpc<
   readonly payloadSchema: Payload
   readonly successSchema: Success
   readonly errorSchema: Error
+  readonly defectSchema: Schema.Schema.Any
```
tim-smart: The schema will have to have no requirements as well

packages/rpc/src/Rpc.ts
```diff
@@ -171,6 +172,7 @@ export interface AnyWithProps {
   readonly payloadSchema: AnySchema
   readonly successSchema: Schema.Schema.Any
   readonly errorSchema: Schema.Schema.All
+  readonly defectSchema: Schema.Schema.Any
```
tim-smart: The schema should be constrained to an input of unknown

tim-smart (changes requested):

packages/rpc/src/Rpc.ts
```diff
@@ -58,6 +58,7 @@ export interface Rpc<
   readonly payloadSchema: Payload
   readonly successSchema: Success
   readonly errorSchema: Error
+  readonly defectSchema: Schema.Schema<unknown, unknown>
```
tim-smart:
```suggestion
  readonly defectSchema: Schema.Schema<unknown, any>
```

The encoded type we don't care about, as long as we can give it an unknown value to encode.

tim-smart:
There were a few other places that need the encoded type updated to any. Also, does the client need updating? Or does it only use the exitSchema api.

.changeset/custom-defect-schema.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/rpc": minor
```
tim-smart:
```suggestion
"@effect/rpc": patch
```

PR #6073, opened 2026-02-19
feat(rpc): add supportsSpanPropagation option to RpcServer.toHttpApp

### LINKED ISSUE #6072: `supportsSpanPropagation` option for `RpcServer.toHttpApp`

When using `RpcServer.toHttpApp` as the transport, there's no way to opt into span propagation. The internal HTTP protocol hardcodes `supportsSpanPropagation: false`, so the server ignores `traceId`/`spanId`/`sampled` from incoming RPC messages — even though the client sends them.

All other protocols (WebSocket, Worker, Stdio) support span propagation. HTTP is the only one that doesn't, which makes sense as a default since a typical HTTP RPC server sits behind a reverse proxy that handles its own tracing. But for same-origin setups like Durable Object RPC, there's no proxy — the client and server share a trust boundary, and propagating spans across that boundary is the expected behavior.

Add an optional `supportsSpanPropagation` field to `toHttpApp`'s options:

```ts
const httpApp = RpcServer.toHttpApp(myGroup, {
  supportsSpanPropagation: true
})
```

The implementation is minimal — override the protocol's field before providing it to `make()`. No changes to `makeProtocolWithHttpApp`, no breaking changes.

_No response_

marbemac:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add optional `supportsSpanPropagation` option to `RpcServer.toHttpApp`.

The HTTP protocol hardcodes `supportsSpanPropagation: false`, which makes sense as a default for typical HTTP RPC servers behind reverse proxies. But for same-origin setups like Cloudflare Durable Object RPC, the client and server share a trust boundary and propagating spans is expected.

This adds a single optional field to `toHttpApp`'s options. When set to `true`, it overrides the protocol's `supportsSpanPropagation` before providing it to `make()`. No changes to `makeProtocolWithHttpApp` — no breaking changes.

```ts
const httpApp = RpcServer.toHttpApp(myGroup, {
  supportsSpanPropagation: true
})
```

## Related

- Closes #6072

### DIFF
--- a/.changeset/rpc-toHttpApp-span-propagation.md
+++ b/.changeset/rpc-toHttpApp-span-propagation.md
@@ -1,0 +1,5 @@
+---
+"@effect/rpc": minor
+---
+
+Add optional `supportsSpanPropagation` option to `RpcServer.toHttpApp` for enabling distributed tracing across HTTP RPC boundaries.
--- a/packages/rpc/src/RpcServer.ts
+++ b/packages/rpc/src/RpcServer.ts
@@ -1231,6 +1231,7 @@ export const toHttpApp: <Rpcs extends Rpc.Any>(
     readonly spanPrefix?: string | undefined
     readonly spanAttributes?: Record<string, unknown> | undefined
     readonly disableFatalDefects?: boolean | undefined
+    readonly supportsSpanPropagation?: boolean | undefined
   } | undefined
 ) => Effect.Effect<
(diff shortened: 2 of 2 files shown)


marbemac:
sigh, missed by a day! np, can patch locally for now. VERY excited to give v4 a go as soon as we can though - congrats on the milestone.

PR #6109, opened 2026-03-06
Print primitive default values in cli --help output

### LINKED ISSUE #6098: Display default value in --help documentation from @effect/cli

The current generated help documentation from --help does not display the default value you set with `Options.withDefault()`.

Example:

```ts
const uppercase = Options.choiceWithValue("uppercase", [["true", true], ["false", false]]).pipe(Options.withDefault(false));
```
The help doc entry for this is currently:

```
--uppercase true | false

  One of the following: true, false

  This setting is optional.
```

Display the default value.

Something like this:
```
--uppercase true | false (default: false)

    One of the following: true, false

    This setting is optional.
```

Currently you have to manually write it with `Options.withDescription()`

leemhenson:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Amends the behavior of `--help` to print out the default value when it's a primitive value not wrapped in an `Option`. For example:

```
This setting is optional. Defaults to: 42
```

## Related

- Related Issue #
- Closes #6098

### DIFF
--- a/.changeset/sad-areas-grin.md
+++ b/.changeset/sad-areas-grin.md
@@ -1,0 +1,5 @@
+---
+"@effect/cli": patch
+---
+
+print primitive default values in cli help output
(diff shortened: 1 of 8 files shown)


tim-smart:

PR #6150, opened 2026-03-30
Add Data.TaggedEnum example to taggedEnum JSDoc

danielo515:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

The taggedEnum constructor was missing an example showing the common pattern of using Data.TaggedEnum<{...}> as the type parameter. The existing examples only showed the verbose union syntax and the generics case.

## Related

Conversation about it: https://discord.com/channels/795981131316985866/1486327719217201252

### DIFF
--- a/packages/effect/src/Data.ts
+++ b/packages/effect/src/Data.ts
@@ -438,6 +438,20 @@ export declare namespace TaggedEnum {
  * ```
  *
  * @example
+ * ```ts
+ * import { Data } from "effect"
+ *
+ * type HttpError = Data.TaggedEnum<{
+ *   BadRequest: { readonly status: 400; readonly message: string }
+ *   NotFound: { readonly status: 404; readonly message: string }
+ * }>
+ *
+ * const { BadRequest, NotFound } = Data.taggedEnum<HttpError>()
+ *
+ * const notFound = NotFound({ status: 404, message: "Not Found" })
+ * ```
+ *
+ * @example
  * import { Data } from "effect"
  *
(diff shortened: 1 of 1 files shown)


packages/effect/src/Data.ts
```diff
@@ -438,6 +438,20 @@ export declare namespace TaggedEnum {
  * ```
  *
  * @example
+ * ```ts
```
tim-smart: It would be good to just have one example. Feel free to remove the other one.

danielo515: My example is indeed more concise and looks like a better practice. You suggest that the old example should be removed? How do you document such valid extra cases then?

danielo515: @tim-smart sorry for tagging you, but care to elaborate?

fubhy:
Thanks for the PR. Could you update this to keep just one non-generic `Data.taggedEnum` example, replacing the existing verbose union example with the `Data.TaggedEnum<{ ... }>` style? That matches the maintainer feedback and should make this easier to review.

danielo515:
> Thanks for the PR. Could you update this to keep just one non-generic `Data.taggedEnum` example, replacing the existing verbose union example with the `Data.TaggedEnum<{ ... }>` style? That matches the maintainer feedback and should make this easier to review.

@fubhy done!

tim-smart:

danielo515:
@fubhy can please the be considered?

PR #6154, opened 2026-04-03
fix(platform): preserve percent encoding in Url.setUrlParams and modifyUrlParams

### LINKED ISSUE #6153: `Url.modifyUrlParams` doesn't preserve percent encoding of spaces

@effect/platform@0.96.0

in Node

```javascript
const identity = x => x;
```

### What do you see instead?

### Additional information

_No response_

Gdhanush-13:
## What
Fixes #6153 - `Url.modifyUrlParams` converts `%20` to `+` in query parameters, even when using an identity function.

## Root Cause
`Url.setUrlParams` and `Url.modifyUrlParams` both use `UrlParams.toString` to serialize params back to a search string. `UrlParams.toString` delegates to `new URLSearchParams(...).toString()` which encodes spaces as `+` per the `application/x-www-form-urlencoded` spec. This is correct for form request bodies but wrong for URL query strings where `%20` should be preserved.

## Reproduction
```ts
const url = new URL('https://example.com?foo=bar%20baz');
url.href === url2.href; // false! %20 became +
```

## Fix
- Added `UrlParams.toPercentEncoded` which uses `encodeURIComponent` (producing `%20` for spaces)
- Updated `Url.setUrlParams` and `Url.modifyUrlParams` to use `toPercentEncoded` instead of `toString`
- Left `UrlParams.toString` unchanged for backward compatibility with `application/x-www-form-urlencoded` body encoding

## Files Changed
- `packages/platform/src/UrlParams.ts` - added `toPercentEncoded`  
- `packages/platform/src/Url.ts` - use `toPercentEncoded` in `setUrlParams` and `modifyUrlParams`

### DIFF
--- a/packages/platform/src/Url.ts
+++ b/packages/platform/src/Url.ts
@@ -257,7 +257,7 @@ export const setUrlParams: {
   (url: URL, urlParams: UrlParams.UrlParams): URL
 } = dual(2, (url: URL, searchParams: UrlParams.UrlParams) =>
   mutate(url, (url) => {
-    url.search = UrlParams.toString(searchParams)
+    url.search = UrlParams.toPercentEncoded(searchParams)
   }))
 
 /**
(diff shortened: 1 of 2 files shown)


fubhy:
Thanks for the PR. Could you add a changeset and tests covering preservation of percent-encoded query strings in `Url.setUrlParams` / `Url.modifyUrlParams`?

PR #6161, opened 2026-04-08
fix(rpc): make RpcSerialization.msgPack options configurable

### LINKED ISSUE #6170: Bug: RpcSerialization.msgPack silently swallows decode errors

`@effect/rpc` 0.75.0

The `decode` function in `RpcSerialization.msgPack` catches all errors from `unpackr.unpackMultiple()` and returns an empty array for any non-`incomplete` error:

```ts
decode: (bytes) => {
  // ...
  try {
    return unpackr.unpackMultiple(buf)
  } catch (error_) {
    const error = error_ as any
    if (error.incomplete) {
      incomplete = buf.subarray(error.lastPosition)
      return error.values ?? []
    }
    return []  // ← all other errors silently return empty array
  }
}
```

Any error that isn't an `incomplete` chunking error — corrupted data, encoding mismatches, runtime restrictions on `new Function()` (see #6169) — is silently discarded. The caller receives an empty array as if no messages were sent.

Non-`incomplete` decode errors should propagate (or at minimum be logged). These represent genuine failures — corrupted data, incompatible serialization settings, or environment restrictions — not expected chunking boundaries.

### What do you see instead?

Errors are silently swallowed and an empty array is returned. RPC calls fail with no indication of what went wrong, making debugging extremely difficult. In the case of #6169, this masked the root cause for weeks.

### Additional information

The `incomplete` error handling is correct and necessary — msgpackr uses it for chunked stream processing. But the catch-all `return []` fallback conflates expected chunking with unexpected failures.

bohdanbirdie:
Fixes https://github.com/Effect-TS/effect/issues/6170

## Summary

- Adds `makeMsgPack(options)` factory for creating MessagePack serialization with custom msgpackr options
- Existing `msgPack` and `layerMsgPack` exports are unchanged (no breaking change)
- Fixes silent error swallowing in msgPack decode — non-incomplete errors are now rethrown instead of returning `[]`

## Problem

On Cloudflare Workers with `allow_eval_during_startup` (default for `compatibility_date >= 2025-06-01`), `RpcSerialization.msgPack` silently fails when decoding messages containing 3+ objects with the same key structure.

**Root cause:** `new Packr()` with no options sets `this.structures = []` because `undefined != false` evaluates to `true`. This enables msgpackr's record/structure path. When the Unpackr's `readObject.count` exceeds the JIT threshold (2), msgpackr calls `new Function()` to compile a fast reader — which CF Workers blocks during request handling with `EvalError: Code generation from strings disallowed for this context`.

The existing `catch { return [] }` in `RpcSerialization.msgPack` silently swallows this error, returning an empty response. The caller never receives data.

## Reproduction

https://github.com/bohdanbirdie/repro-effect-rpc-msgpack-cf-workers

- **`main` branch** — "before" state: reproduces the bug using inline msgpackr code (same as `RpcSerialization.msgPack`)
- **`fixed` branch** — "after" state: uses `bun patch` on `@effect/rpc` to apply this PR's changes, worker uses `RpcSerialization.makeMsgPack({ useRecords: false })`

Deploy either branch to CF Workers with `npx wrangler deploy` and curl the endpoint.

## Fix

Use `makeMsgPack` to create a serialization with custom options:

```typescript
import { Layer } from "effect"
import { RpcSerialization } from "@effect/rpc"

Layer.succeed(
  RpcSerialization,
  RpcSerialization.makeMsgPack({ useRecords: false })
)
```

This skips `this.structures = []` in the Packr constructor, preventing the record/JIT path entirely.

## Test plan

- [x] New unit tests for `msgPack` and `makeMsgPack` with custom options
- [x] Existing 101 e2e RpcServer tests pass (http, ws, tcp, worker × ndjson, msgpack, jsonrpc)
- [x] CF Workers deployment verified with `{ useRecords: false }`

## References

- [livestore PR with full investigation](https://github.com/livestorejs/livestore/pull/1163)
- [CF Workers `new Function()` restriction docs](https://developers.cloudflare.com/workers/reference/security-model/)
- [msgpackr `useRecords` docs](https://github.com/kriszyp/msgpackr#records--structures)
- [CF `allow_eval_during_startup` compat flag](https://developers.cloudflare.com/workers/configuration/compatibility-flags/#dynamic-code-evaluation)

### DIFF
--- a/.changeset/long-buckets-love.md
+++ b/.changeset/long-buckets-love.md
@@ -1,0 +1,5 @@
+---
+"@effect/rpc": patch
+---
+
+add RpcSerialization.makeMsgPack
--- a/packages/platform-node/test/NodeClusterSocket.test.ts
+++ b/packages/platform-node/test/NodeClusterSocket.test.ts
@@ -1,0 +1,122 @@
+import {
+  ClusterSchema,
+  Entity,
+  MessageStorage,
+  RunnerAddress,
+  RunnerHealth,
+  RunnerStorage,
+  ShardingConfig,
+  SocketRunner
+} from "@effect/cluster"
+import { NodeClusterSocket } from "@effect/platform-node"
+import { Rpc, RpcSerialization } from "@effect/rpc"
+import { describe, it } from "@effect/vitest"
+import { BigDecimal, Effect, Layer, Logger, LogLevel, Option, PrimaryKey, Schema } from "effect"
+
+class TestPayload extends Schema.Class<TestPayload>("TestPayload")({
+  id: Schema.String,
+  amount: Schema.BigDecimal
+}) {
+  [PrimaryKey.symbol]() {
+    return this.id
+  }
+}
+
+const TestEntity = Entity
+  .make("TestEntity", [
+    Rpc.make("Process", {
+      payload: TestPayload,
+      success: Schema.Void
+    })
+  ])
+  .annotateRpcs(ClusterSchema.Persisted, true)
+  .annotateRpcs(ClusterSchema.Uninterruptible, true)
+
+const TestEntityLayer = TestEntity.toLayer(
+  Effect.succeed({
+    Process: () => Effect.void
+  })
+)
+
+const RUNNER_PORT = 50_123
+// Build shared storage instances once, so runner and client see the same state.
+// MessageStorage.layerMemory requires ShardingConfig, so we provide a minimal one.
+const SharedStorage = Layer.mergeAll(
+  RunnerStorage.layerMemory,
+  MessageStorage.layerMemory
+).pipe(
+  Layer.provide(ShardingConfig.layerDefaults)
+)
+
+const makeRunnerLayer = (port: number) =>
+  TestEntityLayer.pipe(
+    Layer.provideMerge(SocketRunner.layer),
+    Layer.provide(RunnerHealth.layerNoop),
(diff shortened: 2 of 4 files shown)


bohdanbirdie:
I'm not sure if the solution is good enough, but I think that the underlying issue is confirmed. I'm rather new to `effect` so PR changes might need some guidance, but I'm happy to carry on with it.

Reproduction repository should be easy to use and try.

Tim, let me know what you think (didn't want to ping and create noise).

IGassmann:
FYI, `msgpackr` has landed an [upstream fix](https://github.com/kriszyp/msgpackr/commit/c6e831d5c4af4b05824a72d351ef4de34d51b43c) that wraps the `new Function()` call in a try/catch, falling back to the interpreted path when it's blocked. Once that's released and `@effect/rpc` updates its `msgpackr` dependency, this would fix #6169 without needing `{ useRecords: false }`.

That said, the configurable `makeMsgPack(options)` API and the decode error-swallowing fix in this PR are still valuable independently.

IGassmann:
The upstream fix has been released in msgpackr 1.11.10 (see https://github.com/kriszyp/msgpackr/issues/179). I've opened #6191 to bump the minimum `msgpackr` version to `^1.11.10`, which resolves the Cloudflare Workers issue without needing configurable options or the `no-eval` entry point.

PR #6164, opened 2026-04-10
docs(Effect.fn): add example for explicit type annotations

### LINKED ISSUE #5860: Documentation for how to specify explicit types for Effect.fn
### What is the type of issue?

Documentation is missing

### What is the issue?

The `Effect.Effect` type is pretty well-documented and there are plenty of examples of explicit types for a returned Effect from a function. For example: https://effect.website/docs/getting-started/control-flow/#if-expression

The `Effect.fn` type though seems to be totally different and there isn't an obvious migration path from 'function that returns an effect' to 'effect.fn with explicit types'

It would be great if there was an example, because the Effect.fn method does seem like a nice improvement in dx but since it's relatively new people will be upgrading, and might not want to switch from explicit to inferred types.

### Where did you find it?

https://effect.website/docs/error-management/expected-errors/#effectfn

aymenhmaidiwastaken:
## Summary

- Add a "before/after" JSDoc example to `Effect.fn` showing how to migrate from a function with an explicit `Effect.Effect<A, E, R>` return type to `Effect.fn` using `Effect.fn.Return<A, E, R>`
- Uses `Data.TaggedError` for the error type to follow current Effect patterns

This addresses the documentation gap where users know how to type regular functions returning effects but don't know the equivalent pattern for `Effect.fn`.

Closes #5860

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -14622,6 +14622,32 @@ export namespace fn {
  * // timestamp=... level=ERROR fiber=#0 cause="Error: Boom! (<= after 1 second)
  * ```
  *
+ * **Example** (Explicit Type Annotations with Effect.fn)
+ *
+ * When migrating from a function with explicit return types, you can use
+ * `Effect.fn.Return` to annotate the generator return type:
+ *
+ * ```ts
+ * import { Effect, Data } from "effect"
+ *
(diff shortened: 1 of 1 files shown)


tim-smart (changes requested):
I would prefer if the existing example was updated instead of adding a new one.

fubhy:

tim-smart:

PR #6166, opened 2026-04-12
fix(platform): preserve percent-encoding in modifyUrlParams and setUrlParams

### LINKED ISSUE #6153: `Url.modifyUrlParams` doesn't preserve percent encoding of spaces

Yanhu007:
## Summary

Fixes #6153

`Url.modifyUrlParams` and `Url.setUrlParams` use `UrlParams.toString` which delegates to `URLSearchParams.toString()`. Per the WHATWG URL spec, `URLSearchParams` serializes spaces as `+` (`application/x-www-form-urlencoded`). This silently mutates URLs:

```typescript
const url = new URL('https://example.com?foo=bar%20baz')
const url2 = Url.modifyUrlParams(url, identity)
url.href === url2.href // false! %20 became +
```

## Approach

Add `UrlParams.toStringPercent` that uses `encodeURIComponent` (RFC 3986 percent-encoding, spaces → `%20`) and use it in the two URL-modifying functions.

The existing `UrlParams.toString` is **preserved unchanged** for HTTP form body encoding (`httpBody.ts`), where `application/x-www-form-urlencoded` format (`+` for spaces) is the correct standard.

| Function | Before | After |
|----------|--------|-------|
| `modifyUrlParams(identity)` | `foo=bar+baz` | `foo=bar%20baz` |
| `setUrlParams` | `foo=bar+baz` | `foo=bar%20baz` |
| `httpBody` form encoding | `foo=bar+baz` | `foo=bar+baz` (unchanged) |

### DIFF
--- a/.changeset/fix-url-params-percent-encoding.md
+++ b/.changeset/fix-url-params-percent-encoding.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+Preserve percent-encoding (`%20`) in `Url.modifyUrlParams` and `Url.setUrlParams`. Previously these functions used `URLSearchParams.toString()` which converts spaces to `+` per the `application/x-www-form-urlencoded` spec, silently mutating URLs. Added `UrlParams.toStringPercent` using RFC 3986 percent-encoding for URL context, while keeping the existing `toString` for form body encoding.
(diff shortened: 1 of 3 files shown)


tim-smart:
Duplicate PR

Issue #6126, opened 2026-03-16
detectAsyncLeaks false positive: MixedScheduler PROMISE leaks with @effect/vitest


gabriel-ecegi:
## Description

When using `@effect/vitest` with vitest's `detectAsyncLeaks: true`, every `it.effect()` / `it.scoped()` test reports false positive PROMISE leaks from Effect's `MixedScheduler`.

## Reproduction

```typescript
// vitest.config.ts
export default defineConfig({
  test: { detectAsyncLeaks: true }
});

// example.test.ts
import { it, describe, expect } from "@effect/vitest";
import { Effect } from "effect";

describe("leak demo", () => {
  it.effect("simple effect leaks PROMISE", () =>
    Effect.gen(function* () {
      expect(1).toBe(1);
    })
  );
});
```

**Output:**
```
⎯⎯⎯ Async Leaks 2 ⎯⎯⎯
PROMISE leaking in example.test.ts
```

**Stack trace:**
```
MixedScheduler.starveInternal Scheduler.js:68:17
Scheduler.js:84:47
```

## Root Cause

`MixedScheduler` uses `Promise.resolve(void 0).then(() => drain(depth + 1))` for fiber scheduling ([Scheduler.ts#L131](https://github.com/Effect-TS/effect/blob/main/packages/effect/src/Scheduler.ts#L131)). Vitest tracks every PROMISE async resource via `node:async_hooks` and only removes them on `promiseResolve`. The scheduler's intermediate promises haven't resolved by the time vitest collects leaks.

## What We Tried

| Approach | Result |
|----------|--------|
| `it.scoped()` | Still leaks — scheduler is outside scope |
| `Effect.scoped` pipe | Same |
| `afterAll` with 200ms delay | Doesn't help — vitest measures per-test |
| `SyncScheduler` via `Effect.locally` | Deadlocks on any async operation (fetch) |
| `MixedScheduler(0)` (force setTimeout) | Leaks `Timeout` instead of `PROMISE` |
| `Layer.locallyScoped(currentScheduler, SyncScheduler)` | Hangs with scoped effects |

## Expected Behavior

`@effect/vitest` test helpers should properly clean up MixedScheduler resources so that `detectAsyncLeaks` doesn't report false positives.

## Possible Solutions

1. `@effect/vitest` could flush the scheduler after each test before returning control to vitest
2. Use `queueMicrotask` instead of `Promise.resolve().then()` — microtasks don't create trackable async resources in `node:async_hooks`
[...]

PR #6167, opened 2026-04-12
fix(effect): replace hardcoded Date.now() in FiberId.unsafeMake with configurable clock source

copyleftdev:
## Problem

`FiberId.unsafeMake()` uses `Date.now()` for `startTimeMillis`, which feeds into `Hash.symbol()` and `Equal.symbol()`:

```typescript
// fiberId.ts:75
[Hash.symbol](): number {
  return Hash.cached(this, Hash.string(`${FiberIdSymbolKey}-${this._tag}-${this.id}-${this.startTimeMillis}`))
}
```

This means fiber identity hashing and equality are **non-deterministic across runs**. Two fibers created with the same counter ID but at different wall-clock times produce different hashes and are not equal.

This contributes to scheduler-related issues:
- #6124 — scheduler runner isolation (fiber-keyed `WeakMap` behaves inconsistently)
- #6126 — MixedScheduler PROMISE leaks in `@effect/vitest` (timing-dependent cache behavior)

It also makes deterministic testing of the fiber runtime impossible without monkey-patching `Date.now` globally, which breaks timers, logging, and concurrent tests.

## Fix

Replace the hardcoded `Date.now()` call with a `globalValue`-backed configurable clock source:

```typescript
const _clockSource = globalValue(
  Symbol.for("effect/Fiber/Id/_clockSource"),
  (): { now: () => number } => ({ now: () => Date.now() })
)
```

- **Default behavior is identical** — `Date.now()` is still called
- `setClockSource(fn)` / `resetClockSource()` allow swapping for deterministic testing
- No global mutation of `Date.now`
- 1 file changed, 22 insertions, 1 deletion

## Test plan

- [x] `Fiber.test.ts` — 16 tests pass
- [x] `Effect/concurrency.test.ts` — 24 tests pass
- [x] `Effect/racing.test.ts` — 4 tests pass
- [x] Zero regressions

### DIFF
--- a/.changeset/fix-fiberid-clock-source.md
+++ b/.changeset/fix-fiberid-clock-source.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Replace hardcoded `Date.now()` in `FiberId.unsafeMake()` with a configurable clock source. The `startTimeMillis` field feeds into `Hash.symbol()` and `Equal.symbol()`, making fiber identity non-deterministic across runs. The new `setClockSource()`/`resetClockSource()` API allows deterministic testing without monkey-patching `Date.now` globally.
--- a/packages/effect/src/internal/fiberId.ts
+++ b/packages/effect/src/internal/fiberId.ts
@@ -259,9 +259,30 @@ export const toSet = (self: FiberId.FiberId): HashSet.HashSet<FiberId.Runtime> =
   }
 }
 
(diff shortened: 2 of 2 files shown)


tim-smart:
Hello! Effect v3 is under a feature freeze now. Also in v4 fiber ids no longer contain a creation timestamp.

Issue #6014, opened 2026-01-30
@effect/workflow: Effect.all with concurrency > 1 wrapping Activity.make causes deadlock during replay


Necmttn:
### What version of Effect is running?

- `effect`: 3.19.15
- `@effect/workflow`: 0.16.0
- `@effect/platform`: 0.94.2
- `@effect/cluster`: 0.56.1
- Runtime: Bun 1.3.0

### What steps can reproduce the bug?

When a workflow uses `Effect.all(activities, { concurrency: N })` where `N > 1` and each element is an `Activity.make(...)`, the workflow completes successfully on first execution but **deadlocks during durable replay** (e.g. when the workflow engine re-executes the workflow to recover state).

Minimal reproduction pattern:

```ts
import { Activity, Workflow } from "@effect/workflow";
import { Effect, Schema, Schedule } from "effect";

const MyWorkflow = Workflow.make({
  name: "DeadlockRepro",
  payload: Schema.Struct({ items: Schema.Array(Schema.String) }),
  success: Schema.String,
  execute: ({ items }) =>
    Effect.gen(function* () {
      // Create an array of activities, one per item
      const activities = items.map((item) =>
        Activity.make({
          name: `process:${item}`,
          success: Schema.String,
          error: Schema.String,
          execute: Effect.succeed(`done:${item}`),
        })
      );

      // This works with concurrency: 1, deadlocks with concurrency > 1 during replay
      const results = yield* Effect.all(activities, { concurrency: 3 });
      return results.join(",");
    }),
});
```

Steps to reproduce:
1. Execute the workflow with `items: ["a", "b", "c", "d", "e"]`
2. All 5 activities complete successfully on first execution
3. Trigger a replay (e.g. restart the workflow engine, or let durable execution recovery kick in)
4. During replay, the workflow hangs indefinitely — activities never resume

The same `Effect.all({ concurrency: 3 })` pattern works perfectly outside of `@effect/workflow` (i.e. without `Activity.make`), confirming the issue is in the Activity replay mechanism interacting with concurrent suspension.

### What is the expected behavior?

`Effect.all(activities, { concurrency: N })` where `N > 1` should work correctly both during first execution and during replay. Activities should be replayed from the durable log and the concurrent combinator should resume all branches.

### What do you see instead?

The workflow hangs indefinitely during replay. No error is thrown, no timeout fires at the workflow level — the activities simply never complete. The workflow fiber appears to be suspended waiting for activity results that are never replayed.

[...]

PR #6168, opened 2026-04-12
feat(effect): add stepOne() to ControlledScheduler for fine-grained task execution

copyleftdev:
## Problem

`ControlledScheduler.step()` drains **all** pending tasks across **all** priority buckets in a single call:

```typescript
step() {
  const tasks = this.tasks.buckets
  this.tasks.buckets = []          // takes everything
  for (const [_, toRun] of tasks)  // runs everything
    for (let i = 0; i < toRun.length; i++)
      toRun[i]()
}
```

There is no way to execute exactly one task and observe the intermediate state. This makes it impossible to reproduce concurrency bugs in a controlled environment — you can't step through individual task interleavings to find the specific ordering that triggers the issue.

Relevant bugs that are hard to reproduce without fine-grained stepping:
- #6081 — semaphore race condition where permits could be leaked
- #6014 — workflow deadlock during durable replay with `concurrency > 1`
- #6139 — batched request resolver defects hanging consumer fibers

## Fix

Add `stepOne()` to `ControlledScheduler`:

```typescript
stepOne(): boolean {
  const buckets = this.tasks.buckets
  for (let i = 0; i < buckets.length; i++) {
    const [_, tasks] = buckets[i]!
    if (tasks.length > 0) {
      const task = tasks.shift()!
      if (tasks.length === 0) buckets.splice(i, 1)
      task()
      return true
    }
  }
  return false
}
```

- Executes exactly **one** task from the highest-priority bucket
- Returns `true` if a task was executed, `false` if empty
- `step()` behavior **unchanged**
- 1 file, 15 lines added, purely additive

## Test plan

- [x] `Schedule.test.ts` — 82 tests pass
- [x] `Effect/scheduling.test.ts` — 2 tests pass
- [x] `Fiber.test.ts` — 16 tests pass
- [x] Zero regressions

### DIFF
--- a/.changeset/add-controlled-scheduler-stepone.md
+++ b/.changeset/add-controlled-scheduler-stepone.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Add `stepOne()` to `ControlledScheduler` that executes exactly one pending task from the highest-priority bucket. Unlike `step()` which drains all tasks at once, `stepOne()` enables fine-grained control over task execution order for deterministic simulation testing and debugging of concurrency issues.
--- a/packages/effect/src/Scheduler.ts
+++ b/packages/effect/src/Scheduler.ts
@@ -264,6 +264,32 @@ export class ControlledScheduler implements Scheduler {
       }
     }
   }
+
+  /**
+   * Execute exactly one pending task from the highest-priority bucket.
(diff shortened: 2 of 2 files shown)


tim-smart:
Hello! Effect v3 is under a feature freeze now. Feel free to contribute this to the effect v4 repository.

PR #6194, opened 2026-04-21
fix: floor TestClock nanoseconds before BigInt conversion

mikearnaldi:
## Summary
- floor fractional millisecond instants before converting `TestClock` nanoseconds to `BigInt`
- add a regression test covering a fractional millisecond timestamp that previously threw `RangeError`
- add a patch changeset for the `effect` package

### DIFF
--- a/packages/effect/test/TestClock.test.ts
+++ b/packages/effect/test/TestClock.test.ts
@@ -26,9 +26,9 @@ describe("TestClock", () => {
 
     it.effect("should floor nanoseconds for fractional millisecond instants", () =>
       Effect.gen(function*() {
-        yield* TestClock.setTime(34199023438.000004)
+        yield* TestClock.setTime(199023438.0000004)
         const testClock = yield* TestClock.testClock()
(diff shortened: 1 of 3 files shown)


packages/effect/test/TestClock.test.ts
```diff
@@ -23,5 +23,12 @@ describe("TestClock", () => {
         const now = yield* DateTime.now
         deepStrictEqual(now, arbitraryDateTime)
       }))
+
+    it.effect("should floor nanoseconds for fractional millisecond instants", () =>
+      Effect.gen(function*() {
+        yield* TestClock.setTime(34199023438.000004)
+        const testClock = yield* TestClock.testClock()
+        strictEqual(testClock.unsafeCurrentTimeNanos(), 34199023438000004n)
```
tim-smart:
```suggestion
        yield* TestClock.setTime(199023438.0000004)
        const testClock = yield* TestClock.testClock()
        strictEqual(testClock.unsafeCurrentTimeNanos(), 199023438000000n)
```

PR #6201, opened 2026-04-28
Fix PCGRandom.integer() bias: use Lemire rejection sampling

### LINKED ISSUE #6184: PCG Random.integer() uses biased Math.round + modulo instead of rejection sampling

3.21.0

## Description

`PCGRandom.integer()` in `packages/effect/src/Utils.ts` uses `Math.round(this.number() * Number.MAX_SAFE_INTEGER) % max`, which introduces **two distinct sources of statistical bias** that deviate from the PCG reference implementation.

### Location

**File:** `packages/effect/src/Utils.ts`\
**Lines:** 612–615 (approximate)

```typescript
/**
 * Get a uniformly distributed 32 bit integer between [0, max).
 *
 * @category getter
 * @since 2.0.0
 */
integer(max: number) {
  return Math.round(this.number() * Number.MAX_SAFE_INTEGER) % max
}
```

**Consumed by:** `packages/effect/src/internal/random.ts`, line 50

```typescript
nextIntBetween(min: number, max: number): Effect.Effect<number> {
  return core.sync(() => this.PRNG.integer(max - min) + min)
}
```

### Issue 1: `Math.round()` bias

`Math.round()` creates systematic non-uniformity. For any continuous range mapped to discrete integers, the first and last values receive **half the probability mass** of interior values:

```
this.number() returns [0, 1)

Mapping to [0, 99] via Math.round(n * MAX_SAFE_INTEGER) % 99:

[0.0, 0.5)      → 0   (width: 0.5)
[0.5, 1.5)      → 1   (width: 1.0)
[1.5, 2.5)      → 2   (width: 1.0)
...
[98.5, 99.0)    → 99  (width: 0.5)  ← half probability!
```

The MDN documentation explicitly warns against using `Math.round()` for random integer generation:

> "Using `Math.round()` will give you a non-uniform distribution!"\
> — [MDN: Math.random()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random)

### Issue 2: Modulo bias

`Number.MAX_SAFE_INTEGER = 2^53 - 1 = 9,007,199,254,740,991` is **not evenly divisible** by most values of `max`.

For `max = 99`:

```
MAX_SAFE_INTEGER % 99 = 40
```

This means raw values `0` through `40` get **one extra hit** in the modulo cycle. Values `0`–`40` are overrepresented; values `41`–`98` are underrepresented.

> "Although this approach [modulo] is simple, it is also biased... For a 32-bit PRNG, a bounded range of less than 2^24 has a bias of less than 0.5% but above 2^31 the bias is 50%—some numbers will occur half as often as others."\
> — [PCG Blog: Bounded Rands](http://www.pcg-random.org/posts/bounded-rands.html)

### Issue 3: Double state consumption

`number()` calls `_next()` twice per value to construct a 53-bit float:

```typescript
number() {
  const hi = (this._next() & 0x03ffffff) * 1.0  // consumes 1 state advance
  const lo = (this._next() & 0x07ffffff) * 1.0  // consumes 1 state advance
  return (hi * BIT_27 + lo) / BIT_53
}
```

The PRNG state advances **2× per integer** when only 1× is needed. This is inefficient and advances the generator faster than necessary.

## Expected Behavior

`integer(max)` should produce a **uniformly distributed** integer in `[0, max)` with **no bias**.

## Reference Implementation

The official C++ PCG reference library (`pcg-cpp`) uses **rejection sampling** (OpenBSD `arc4random_uniform` style):

```cpp
// pcg-cpp/include/pcg_extras.hpp (lines 540–552)
template <typename RngType>
auto bounded_rand(RngType& rng, typename RngType::result_type upper_bound)
        -> typename RngType::result_type
{
    typedef typename RngType::result_type rtype;
    rtype threshold = (RngType::max() - RngType::min() + rtype(1) - upper_bound)
                    % upper_bound;
    for (;;) {
        rtype r = rng() - RngType::min();
        if (r >= threshold)
            return r % upper_bound;
    }
}
```

The original JavaScript `pcg-random` library (which Effect-TS adapted from) also intended to use rejection sampling:

```javascript
// Original pcg-random intention (Thom Chiovoloni)
var skew = ((-max >>> 0) % max) >>> 0
for (num = this.next32(); num < skew; num = this.next32()) {
  // rejection sampling
}
return num % max
```

## Proposed Fix

Replace the biased floating-point approach with **rejection sampling on raw 32-bit output**:

```typescript
integer(max: number): number {
  if (max <= 0) {
    throw new Error("max must be positive")
  }

  // For max that evenly divides 2^32, no rejection needed
  if ((0x100000000 % max) === 0) {
    return this._next() % max
  }

  // Rejection sampling to eliminate modulo bias
  const threshold = ((-max >>> 0) % max) >>> 0
  let result: number
  do {
    result = this._next() >>> 0
  } while (result < threshold)

  return result % max
}
```

This fix:

1. Eliminates `Math.round()` bias entirely
2. Eliminates modulo bias via rejection sampling
3. Consumes only **one** `_next()` call per integer (efficient)
4. Matches the official PCG C++ reference implementation

## Impact

Any code using `Random.nextIntBetween(min, max)` or `Random.nextInt` is affected. This includes:

- `Random.shuffle()` — shuffles may be biased
- `Random.choice()` — selection may favor certain elements
- User code calling `nextIntBetween` directly

The bias is small for large ranges but becomes significant for small ranges (e.g., `max = 2`, `max = 99`), where it can reach **several percent**.

## Environment

- **effect version:** All versions with current `PCGRandom.integer()` implementation
- **File:** `packages/effect/src/Utils.ts`
- **Related:** `packages/effect/src/internal/random.ts`

## Additional Context

- [PCG Blog: Bounded Rands](http://www.pcg-random.org/posts/bounded-rands.html) — detailed explanation of modulo bias and rejection sampling
- [MDN: Math.random()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random) — warning against `Math.round()`
- [OpenBSD arc4random_uniform](https://github.com/openbsd/src/blob/master/lib/libc/crypt/arc4random_uniform.c) — reference rejection sampling implementation
- [pcg-cpp reference implementation](https://github.com/imneme/pcg-cpp/blob/master/include/pcg_extras.hpp) — official C++ PCG bounded random

Zelys-DFKH:
## Summary

`PCGRandom.integer()` used `Math.round(this.number() * Number.MAX_SAFE_INTEGER) % max`, which baked two sources of statistical bias into one expression and consumed twice the PRNG state per call.

Closes #6184

ryanleecode's comment in the thread spotted and quantified the bias clearly. I wanted to make sure a fix landed.

## Problem

**Math.round boundary bias:** `Math.round` maps `[0, 1)` to integers non-uniformly. Buckets at each end of the range receive half the probability mass of interior values.

**Modulo bias:** `Number.MAX_SAFE_INTEGER` is not evenly divisible by most values of `max`, so some output values have one extra preimage in the modulo cycle and are overrepresented.

**Double state consumption:** `number()` calls `_next()` twice to build a 53-bit float. Integer generation only needs one 32-bit draw.

The practical bias magnitude is tiny (ryanleecode's analysis put it at ~10^-16), but a PRNG that is correct-by-construction is worth having.

## Fix

For `max <= 2^32`: Lemire (2018) rejection sampling on raw 32-bit output ([arXiv:1805.10941](https://arxiv.org/abs/1805.10941)). Same approach as OpenBSD's `arc4random_uniform` and the PCG reference implementation. Every value in `[0, max)` gets exactly `floor((2^32 - threshold) / max)` accepting preimages, provably uniform.

For `max > 2^32` (e.g. `nextInt` passing `Number.MAX_SAFE_INTEGER`): falls back to `Math.floor(this.number() * max)`. `Math.floor` removes the `Math.round` boundary bias; residual floating-point discretization bias at that scale is negligible.

## Tests

- Bounds test: `nextIntBetween` stays in `[min, max)` across several small ranges, where rejection sampling is most aggressively exercised.
- Deterministic test against a known seed to catch regressions in output sequence.

### DIFF
--- a/.changeset/pcg-random-integer-bias.md
+++ b/.changeset/pcg-random-integer-bias.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `PCGRandom.integer()` to use Lemire (2018) rejection sampling instead of `Math.round` + modulo, eliminating two sources of bias and halving state consumption per call.
--- a/packages/effect/src/Utils.ts
+++ b/packages/effect/src/Utils.ts
@@ -614,7 +614,20 @@ export class PCGRandom {
    * @since 2.0.0
    */
   integer(max: number) {
-    return Math.round(this.number() * Number.MAX_SAFE_INTEGER) % max
+    if (max <= 0x100000000) {
+      // Lemire (2018) rejection sampling on raw 32-bit output eliminates
+      // modulo bias and Math.round boundary bias, using one _next() call
+      // instead of two. https://arxiv.org/abs/1805.10941
+      const threshold = ((-max >>> 0) % max) >>> 0
+      let result: number
+      do {
+        result = this._next() >>> 0
+      } while (result < threshold)
+      return result % max
+    }
+    // max exceeds 32-bit range — use float-based generation.
+    // Math.floor avoids the boundary bias of the previous Math.round approach.
+    return Math.floor(this.number() * max)
   }
 
   /**
--- a/packages/effect/test/Random.test.ts
+++ b/packages/effect/test/Random.test.ts
@@ -1,6 +1,7 @@
 import { describe, it } from "@effect/vitest"
 import { assertTrue, deepStrictEqual, strictEqual } from "@effect/vitest/utils"
 import { Array, Cause, Chunk, Data, Effect, Random } from "effect"
+import { PCGRandom } from "effect/Utils"
 
 describe("Random", () => {
   it.effect("integer is correctly distributed", () =>
@@ -12,6 +13,28 @@ describe("Random", () => {
       }
       assertTrue(lastRandom >= tenYearsMillis / 2)
     }))
+
+  it.effect("nextIntBetween produces values within bounds for small ranges", () =>
+    Effect.gen(function*() {
+      const ranges: ReadonlyArray<readonly [number, number]> = [[0, 2], [0, 3], [5, 8], [0, 100]]
+      for (const [min, max] of ranges) {
+        const values = yield* Effect.all(Array.replicate(Random.nextIntBetween(min, max), 500))
+        for (const v of values) {
+          assertTrue(v >= min && v < max)
+        }
+      }
+    }))
+
+  it("PCGRandom.integer produces values in [0, max) and is deterministic", () => {
+    const prng = new PCGRandom(42)
+    const results = globalThis.Array.from({ length: 20 }, () => prng.integer(10))
+    assertTrue(results.every((v) => v >= 0 && v < 10))
+    deepStrictEqual(results, [6, 9, 5, 5, 7, 6, 0, 1, 4, 3, 5, 5, 8, 4, 3, 5, 8, 4, 1, 8])
+
+    const prng2 = new PCGRandom(42)
+    const ones = globalThis.Array.from({ length: 10 }, () => prng2.integer(1))
+    assertTrue(ones.every((v) => v === 0))
+  })
   it.effect("shuffle", () =>
     Effect.gen(function*() {
       const start = Array.range(0, 100)


Zelys-DFKH:

PR #6203, opened 2026-04-29
fix(platform): parse +json structured content types as JSON in HttpApiBuilder

### LINKED ISSUE #5908: ContentType defined in `HttpApiSchema.withEncoding` isn't returned in the endpoint response

3.19.9

When trying to change the contentType of an ApiEndpoint, the response contentType is always `application/json` instead of the changed contentType.

```typescript
const MyEndpoint = HttpApiEndpoint.get("foo")`/foo`
  .addSuccess(
    MyResultSchema.pipe(
      HttpApiSchema.withEncoding({
        kind: "Json",
        contentType: "application/scim+json", 
      }),
    ),
  )
```

Discussed this in the discord, i got pointed at changing the kind to `Text`. That returns the given contentType, but then the returned value isn't json anymore obviously.

returned contentType that is defined in this example "application/scim+json"

### What do you see instead?

returned contentType is always "application/json"

### Additional information

Also, when requesting an endpoint with contentType `application/scim+json` (given snippet above), the body isn't parsed as json and the schema (`MyResultSchema`) will fail. While its kind of expected based on the source code, shouldn't we be able to tell the endpoint how to parse the given contentType if its not one of the default ones?

Zelys-DFKH:
## Type
- [x] Bug Fix

## Description

#5908 has two complaints. The response `Content-Type` from `withEncoding` was already working correctly: `HttpServerResponse.json` already uses `encoding.contentType`. The request side is what's broken.

`requestPayload` in `HttpApiBuilder` only checked for `"application/json"`, so request bodies with `Content-Type: application/scim+json` (or any other RFC 6838 `+json` type like `application/vnd.api+json`) fell through to the `Uint8Array` branch instead of being decoded as JSON. The schema decode then fails because it receives raw bytes instead of a parsed object.

The fix is one line: `|| contentType.includes("+json")`. Consistent with how the other branches already use `includes`, and it handles content types with parameters like `application/scim+json; charset=utf-8`.

The regression was introduced in #4024 when explicit content-type dispatch replaced the catch-all JSON handling in `requestPayload`.

## Related

Closes #5908

### DIFF
--- a/.changeset/http-api-structured-json-content-type.md
+++ b/.changeset/http-api-structured-json-content-type.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+Fix `HttpApiBuilder` to parse request bodies with structured JSON content types (RFC 6838 `+json` suffix) as JSON instead of `Uint8Array`. Custom JSON content types like `application/scim+json` and `application/vnd.api+json` now decode correctly on the server when used with `HttpApiSchema.withEncoding({ kind: "Json", contentType: "..." })`.
--- a/packages/platform-node/test/HttpApi.test.ts
+++ b/packages/platform-node/test/HttpApi.test.ts
@@ -311,6 +311,122 @@ describe("HttpApi", () => {
     assert.deepStrictEqual(spec, OpenApiFixture as any)
   })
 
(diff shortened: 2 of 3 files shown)


Zelys-DFKH:

PR #6160, opened 2026-04-08
fix(ai-amazon-bedrock): emit finish part after metadata event to preserve token usage in streaming

jamesone:
## Type

- [x] Bug Fix

## Description

In the Bedrock ConverseStream API, events arrive in the order: `messageStop` → `metadata`. The `metadata` event carries all token usage data (`inputTokens`, `outputTokens`, `totalTokens`, `cacheReadInputTokens`, `cacheWriteInputTokens`) as well as trace information.

Previously, the streaming handler emitted the finish part during the `messageStop` event, before the `metadata` event had populated the usage object. Because the downstream `LanguageModel` layer decodes stream parts through a schema (creating new instances), the finish part was captured with uninitialized usage values — causing all token counts (including `cachedInputTokens`) to be lost in streaming responses.

### Fix

The fix introduces a `tryEmitFinish` guard that defers the finish part until both `messageStop` (which provides the stop reason) and `metadata` (which provides usage/trace) have been received. This:

- Ensures `cachedInputTokens`, `inputTokens`, `outputTokens`, `totalTokens` are all correctly populated in streaming responses
- Ensures `trace` and `cacheWriteInputTokens` metadata are present
- Handles event ordering defensively (emits on whichever arrives second)
- Preserves correct behavior when errors interrupt the stream between `messageStop` and `metadata` (no misleading finish part with empty usage)

The non-streaming (`Converse`) path was already correct since it reads directly from the decoded `ConverseResponse`.

### Tests

Adds comprehensive test suite for `AmazonBedrockLanguageModel` covering both streaming and non-streaming paths:

- Token usage population in the happy path
- Cached token counts (`cacheReadInputTokens` / `cacheWriteInputTokens`)
- Explicit zero cache counts preserved as `0` (not `undefined`)
- Trace metadata propagation
- Tool call content blocks with cached tokens
- Error after `messageStop` (no misleading finish part)
- Missing `messageStop` edge case (no finish part without a stop reason)
- Stop reason mapping
- Text part ordering relative to finish

## Related

- Closes #

PR #6206, opened 2026-05-02
fix(ai-amazon-bedrock): strip thinking from additionalModelRequestFields when using generateObject

### LINKED ISSUE #6185: @effect/ai-amazon-bedrock: withConfigOverride thinking config leaks into tool call handlers, breaking generateObject

- `@effect/ai`: ^0.33.0
- `@effect/ai-amazon-bedrock`: ^0.13.0

### Description

When using `AmazonBedrockLanguageModel.withConfigOverride` to enable extended thinking on a `generateText` call that includes a toolkit, the `Config` service propagates into tool call handlers resolved by `resolveToolCalls` in `LanguageModel.ts`. If any tool handler internally calls `generateObject` (which sets `toolChoice: { tool: { name: ... } }` in the Bedrock provider), the Bedrock API rejects the request with:

```
The model returned the following errors: Thinking may not be enabled when tool_choice forces tool use.
```

This is because Anthropic's API does not allow extended thinking when `toolChoice` forces a specific tool. The `@effect/ai-amazon-bedrock` provider correctly maps `generateObject` to a forced tool call (lines 214-229 of `AmazonBedrockLanguageModel.ts`), but doesn't account for the fact that `withConfigOverride` may have injected thinking configuration from an outer scope.

### Reproduction

```typescript
import { Chat, Toolkit } from "@effect/ai"
import { LanguageModel } from "@effect/ai/LanguageModel"
import { AmazonBedrockLanguageModel } from "@effect/ai-amazon-bedrock"
import { Effect, Schema } from "effect"

const ResultSchema = Schema.Struct({
  outcome: Schema.Literal("Pass", "Fail"),
  reasoning: Schema.String,
})

// Define a tool whose handler uses generateObject
const CheckTool = /* ... tool definition ... */

const CheckToolkit = Toolkit.make(CheckTool)

  const chat = yield* Chat.fromPrompt([{ role: "system", content: "You are a reviewer." }])

  // This call uses generateText with toolChoice: "auto" (compatible with thinking).
  // But the tool handler internally calls generateObject (forced toolChoice).
  // The thinking config leaks from the outer scope into the tool handler.
  const response = yield* chat.generateText({
    prompt: "Review this document",
    toolkit: /* resolved toolkit with handlers that call generateObject */,
  }).pipe(
    AmazonBedrockLanguageModel.withConfigOverride({
      additionalModelRequestFields: {
        thinking: { type: "enabled", budget_tokens: 4096 },
      },
    })
  )
})
```

**Error:** `400 Bad Request` from Bedrock Converse API:
```
"Thinking may not be enabled when tool_choice forces tool use."
```

### Root Cause

`withConfigOverride` provides the `Config` service via `Effect.provideService`:

```typescript
// AmazonBedrockLanguageModel.ts
(self, overrides) =>
  Effect.flatMap(
    Config.getOrUndefined,
    (config) => Effect.provideService(self, Config, { ...config, ...overrides })
  )
```

This makes the config (including `thinking`) available to the **entire** effect tree of `self`. When `resolveToolCalls` (in `LanguageModel.ts`) invokes tool handlers, those handlers run within this scope and inherit the thinking config. If a handler calls `generateObject`, the Bedrock provider's `makeRequest` reads the Config:

```typescript
const config = { modelId: options.model, ...options.config, ...context.unsafeMap.get(Config.key) }
```

It picks up `additionalModelRequestFields.thinking` from the outer scope, builds a request with both thinking enabled and forced `toolChoice`, and the Bedrock API rejects it.

### Current Workaround

Explicitly clear the thinking config on each tool handler:

```typescript
AmazonBedrockLanguageModel.withConfigOverride(handlerEffect, {
  additionalModelRequestFields: {},
})
```

### Expected Behavior

Enabling thinking on a `generateText` call should not affect nested `generateObject` calls made by tool handlers. The provider could:

1. Automatically strip `additionalModelRequestFields.thinking` when building requests where `responseFormat.type === "json"` (i.e. `generateObject`), or when `toolChoice` forces a specific tool — since these are known to be incompatible with thinking per Anthropic's API.
2. Alternatively, scope the `Config` service more narrowly so it only applies to the provider's own API call, not to tool handler resolution.

Zelys-DFKH:
## Type
- [x] Bug Fix

## Description
`generateObject` fails with "Thinking may not be enabled when tool_choice forces tool use" when extended thinking is configured via `withConfigOverride`. This affects both direct `generateObject` calls and the common case where a parent `generateText` propagates the `Config` service (via `Effect.provideService`) into a tool handler that internally calls `generateObject`.

### Root cause
`makeRequest` builds the Bedrock request by spreading `...config` first, which puts `additionalModelRequestFields.thinking` into the request object. The `additionalModelRequestFields` override at the end of the request object only fires when `additionalTools` is present. When `additionalTools` is `undefined` (the usual case), nothing overwrites the `thinking` key that `...config` already placed, and the request is rejected.

### Fix
Pre-compute `requestAdditionalFields` before the request object. In the json path (`generateObject`), merge `config.additionalModelRequestFields` with `additionalTools` and strip `thinking`. Return an explicit object even when empty so the final spread overwrites whatever `...config` put in the request. When no additional fields are configured at all, return `undefined` to avoid injecting an empty object.

## Tests
Adds `packages/ai/amazon-bedrock/test/AmazonBedrockLanguageModel.test.ts` with four cases:
- `thinking` is stripped for `generateObject` (forced toolChoice) and `toolChoice.tool` is set correctly
- `thinking` is preserved for `generateText` (no forced toolChoice)
- No `additionalModelRequestFields` is set when none was configured and `generateObject` is used
- Non-thinking fields survive alongside `thinking` being stripped from a mixed `additionalModelRequestFields`

Note: PR #6160 is also open and adds `AmazonBedrockLanguageModel.test.ts` (token usage in streaming). If that merges first, the two files can be merged with no conflicts - the describe blocks cover unrelated behavior.

## Related
- Closes #6185

### DIFF
--- a/packages/ai/amazon-bedrock/src/AmazonBedrockLanguageModel.ts
+++ b/packages/ai/amazon-bedrock/src/AmazonBedrockLanguageModel.ts
@@ -206,6 +206,26 @@ export const make = Effect.fnUntraced(function*(options: {
       const { messages, system } = yield* prepareMessages(providerOptions)
       const { additionalTools, betas, toolConfig } = yield* prepareTools(providerOptions, config)
       const responseFormat = providerOptions.responseFormat
+
+      // Anthropic rejects requests that combine extended thinking with forced tool use.
+      // generateObject always forces toolChoice, so strip "thinking" in the json path.
+      // Return an explicit object (even if empty) when fields are present so the spread
+      // overrides the thinking config already placed in the request by ...config above.
+      const requestAdditionalFields: Record<string, unknown> | undefined = responseFormat.type === "json"
+        ? (Predicate.isNotUndefined(config.additionalModelRequestFields) ||
+            Predicate.isNotUndefined(additionalTools))
+          ? (() => {
+            const { thinking: _thinking, ...rest } = {
+              ...config.additionalModelRequestFields,
+              ...additionalTools
+            }
+            return rest
+          })()
+          : undefined
+        : Predicate.isNotUndefined(additionalTools)
+        ? { ...config.additionalModelRequestFields, ...additionalTools }
+        : undefined
+
       const request: typeof ConverseRequest.Encoded = {
         ...config,
         system,
@@ -231,13 +251,8 @@ export const make = Effect.fnUntraced(function*(options: {
           ? { toolConfig }
           : {}),
         // Handle additional model request fields
-        ...(Predicate.isNotUndefined(additionalTools)
-          ? {
-            additionalModelRequestFields: {
-              ...config.additionalModelRequestFields,
-              ...additionalTools
-            }
-          }
+        ...(Predicate.isNotUndefined(requestAdditionalFields)
+          ? { additionalModelRequestFields: requestAdditionalFields }
           : {})
       }
       return { betas, request }
--- a/packages/ai/amazon-bedrock/src/AmazonBedrockLanguageModel.ts
+++ b/packages/ai/amazon-bedrock/src/AmazonBedrockLanguageModel.ts
@@ -208,23 +208,26 @@ export const make = Effect.fnUntraced(function*(options: {
       const responseFormat = providerOptions.responseFormat
 
       // Anthropic rejects requests that combine extended thinking with forced tool use.
-      // generateObject always forces toolChoice, so strip "thinking" in the json path.
-      // Return an explicit object (even if empty) when fields are present so the spread
-      // overrides the thinking config already placed in the request by ...config above.
-      const requestAdditionalFields: Record<string, unknown> | undefined = responseFormat.type === "json"
-        ? (Predicate.isNotUndefined(config.additionalModelRequestFields) ||
-            Predicate.isNotUndefined(additionalTools))
-          ? (() => {
-            const { thinking: _thinking, ...rest } = {
-              ...config.additionalModelRequestFields,
-              ...additionalTools
-            }
-            return rest
-          })()
-          : undefined
-        : Predicate.isNotUndefined(additionalTools)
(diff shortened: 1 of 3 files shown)


Zelys-DFKH:

packages/ai/amazon-bedrock/src/AmazonBedrockLanguageModel.ts
```diff
@@ -206,6 +206,26 @@ export const make = Effect.fnUntraced(function*(options: {
       const { messages, system } = yield* prepareMessages(providerOptions)
       const { additionalTools, betas, toolConfig } = yield* prepareTools(providerOptions, config)
       const responseFormat = providerOptions.responseFormat
+
+      // Anthropic rejects requests that combine extended thinking with forced tool use.
+      // generateObject always forces toolChoice, so strip "thinking" in the json path.
+      // Return an explicit object (even if empty) when fields are present so the spread
+      // overrides the thinking config already placed in the request by ...config above.
+      const requestAdditionalFields: Record<string, unknown> | undefined = responseFormat.type === "json"
```
IMax153: Let's make this block a bit more coherent by removing all the nested ternaries and using more imperative `if / else` logic.

IMax153 (changes requested):

Zelys-DFKH:
Done -- refactored to `if / else` as suggested. The nested version was harder to follow than I noticed writing it; this makes each branch's intent obvious. All 4 tests pass, tsc is clean.

IMax153 (approved):

PR #6207, opened 2026-05-02
fix(@effect/platform): flatten anyOf when accumulating same-content-type union payloads

### LINKED ISSUE #6052: @effect/platform: Nested anyOf emitted for flat union payloads in OpenAPI output

 @effect/platform 0.94, effect 3.19

1. Define a flat `Schema.Union(...)` for a request payload.
2. Use it as the payload for a single HTTP endpoint.
3. Generate OpenAPI (or JSON Schema) from the `HttpApi`.

```ts
import * as Schema from "effect/Schema";
import * as HttpApi from "@effect/platform/HttpApi";
import * as HttpApiEndpoint from "@effect/platform/HttpApiEndpoint";

const A = Schema.Struct({ _tag: Schema.Literal("A"), a: Schema.String });
const B = Schema.Struct({ _tag: Schema.Literal("B"), b: Schema.Number });
const C = Schema.Struct({ _tag: Schema.Literal("C"), c: Schema.Boolean });

export const API = HttpApi.make("Demo").add(
  HttpApiGroup.make("Foo").add(
    HttpApiEndpoint.post("Bar", "/bar").setPayload(Schema.Union(A, B, C)),
  ),
);

// Generate OpenAPI / JSON Schema from `api`
// Expected: anyOf: [A, B, C]
// Actual: anyOf: [ anyOf: [ anyOf: [A, B], C ] ] (nested)
```

The request body schema should have a flat `anyOf` array with the union members:

`anyOf: [A, B, C]`

### What do you see instead?

The request body schema is encoded as a recursive/left-associated `anyOf`, e.g.:

`anyOf: [ anyOf: [A, B ], C ]` (or deeper nesting with more members)

### Additional information

It looks like union members are merged incrementally (union of union), and the JSON Schema encoder preserves the AST shape. This is still valid JSON Schema but makes the output much harder to read and diff. There’s already `UnionUnifyAST` in `@effect/platform` which flattens unions

Zelys-DFKH:
## Type
- [x] Bug Fix

## Description
`OpenApi.fromApi` emits nested `anyOf` when `setPayload` receives a union where all members share the same content type. For example, `Schema.Union(A, B, C)` where all three are JSON structs produces `anyOf: [ anyOf: [A, B], C ]` instead of the expected flat `anyOf: [A, B, C]`.

### Root cause
`extractPayloads` accumulates same-content-type members with `AST.Union.make([current.ast, ast])` on each pass. That builds a left-nested union tree — `Union([Union([A, B]), C])` — which the JSON Schema encoder faithfully reflects as nested `anyOf`.

### Fix
One line: replace `AST.Union.make([current.ast, ast])` with `HttpApiSchema.UnionUnifyAST(current.ast, ast)`.

`UnionUnifyAST` is already in the package and is already used for the same reason by the response-map accumulator (`extractResponseMap`, line 382). It flattens both sides through `extractUnionTypes` before creating the Union node, so the result stays flat no matter how many members land on the same content type. The asymmetry between the two accumulators is what introduced the bug.

## Tests
Adds one test to `packages/platform/test/OpenApi.test.ts`: `Schema.Union(A, B, C)` (three JSON structs) on a POST endpoint produces a flat `anyOf: [A, B, C]` in the generated spec, not nested.

## Related
- Closes #6052

### DIFF
--- a/.changeset/fix-platform-flat-anyof-union-payload.md
+++ b/.changeset/fix-platform-flat-anyof-union-payload.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+`OpenApi.fromApi` now emits a flat `anyOf` array when multiple union members in `setPayload` share the same content type. Previously, each additional member was wrapped in a new `Union` node, producing deeply nested `anyOf: [ anyOf: [A, B], C ]` structures. The fix uses `HttpApiSchema.UnionUnifyAST` — already used by the response-map accumulator — to flatten unions as they are collected.
--- a/packages/platform/src/HttpApi.ts
+++ b/packages/platform/src/HttpApi.ts
@@ -419,7 +419,7 @@ const extractPayloads = (topAst: AST.AST): ReadonlyMap<string, {
         ast
       })
     } else {
-      current.ast = AST.Union.make([current.ast, ast])
+      current.ast = HttpApiSchema.UnionUnifyAST(current.ast, ast)
     }
   }
   if (topAst._tag === "Union") {
--- a/packages/platform/test/OpenApi.test.ts
+++ b/packages/platform/test/OpenApi.test.ts
@@ -1951,6 +1951,78 @@ describe("OpenApi", () => {
           expectSpecPaths(api, expected)
         })
       })
+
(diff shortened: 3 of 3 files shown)


Zelys-DFKH:

PR #6209, opened 2026-05-04
datner/acquire disposable

datner:
- feat(effect): add `Effect.acquireDisposable`
- test(effect): happy path and failure for `Effect.acquireDisposable`
- test(effect): change `assertTrue(equals` to `assertEquals`
- chore: add changeset

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -5461,6 +5461,29 @@ export const acquireRelease: {
 } = fiberRuntime.acquireRelease
 
 /**
+ * Constructs a scoped resource from an {@linkcode AsyncDisposable} or {@linkcode Disposable}
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/using}
+ *
+ * @example
+ * import sqlite from "node:sqlite";
+ * import { Effect } from "effect";
+ *
+ * // Define how the resource is acquired
+ * const acquire = Effect.sync(() => new sqlite.DatabaseSync(":memory:"))
+ * const resource = Effect.acquireDisposable(acquire)
+ *
+ * @see {@link acquireRelease} for more information about scopes.
+ *
+ * @since 3.0.0
+ * @category Scoping, Resources & Finalization
+ */
+export const acquireDisposable: {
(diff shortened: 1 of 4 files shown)


packages/effect/src/Effect.ts
```diff
@@ -5460,6 +5460,29 @@ export const acquireRelease: {
...
+ * import sqlite from "node:sqlite";
+ * import { Effect } from "effect";
+ *
+ * // Define how the resource is acquired
+ * const acquire = Effect.sync(() => new sqlite.DatabaseSync(":memory:"))
+ * const resource = Effect.acquireDisposable(acquire)
+ *
+ * @see {@link acquireRelease} for more information about scopes.
+ *
+ * @since 3.0.0
```
datner, line 5477: idk what to put here

packages/effect/src/internal/fiberRuntime.ts
```diff
@@ -1691,6 +1691,17 @@ export const acquireRelease: {
...
   ))
 
+/* @internal */
+export const acquireDisposable: {
+  <A extends AsyncDisposable | Disposable, E, R>(
+    acquire: Effect.Effect<A, E, R>
+  ): Effect.Effect<A, E, R | Scope.Scope>
+} = (acquire) =>
+  acquireRelease(acquire, (resource) =>
+    Predicate.hasProperty(resource, Symbol.asyncDispose)
```
KhraksMamtsov, line 1701: Hi! I just had a thought.
What if an object has both Dispose symbols—why does the asynchronous one take priority?
Maybe it would make sense to split this function into two, or allow configuring the priority?

mikearnaldi: maybe the assumption is that if there is an async one the sync doesn't wait?

datner: If it has both dispose symbols, are both invoked? In which order? 
The `mdn` suggests that [async always takes priority and is even valid for sync cases](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/await_using), but I can't find a definition of behavior for using both symbols... Open to suggestions 🤷🏻

tim-smart:

PR #6215, opened 2026-05-05
feat(@effect/platform): add tracer header filter combinators to HttpClient

### LINKED ISSUE #6198: HttpClient: allow filtering HTTP header span attributes (e.g. allowlist)
## Summary

The OTEL tracing integration in `@effect/platform`'s `HttpClient` unconditionally emits every request and response header as a span attribute (`http.request.header.<name>` / `http.response.header.<name>`), and there is no public API to restrict which headers are captured.

For chatty APIs this creates substantial trace bloat. For example, a typical Notion API response has ~30 headers and every one becomes a span attribute — `cf-ray`, `alt-svc`, `server`, `x-cache`, `strict-transport-security`, `set-cookie` flags, and other CDN-internal chatter — most of which has no debugging value.

## Source

[`packages/platform/src/internal/httpClient.ts`](https://github.com/Effect-TS/effect/blob/main/packages/platform/src/internal/httpClient.ts) — `make = f => makeWith(...)`, the two symmetric loops:

```ts
const redactedHeaders = Headers.redact(request.headers, redactedHeaderNames)
for (const name in redactedHeaders) {
  span.attribute(ATTR_HTTP_REQUEST_HEADER(name), String(redactedHeaders[name]))
}
// ...and the response-side equivalent with ATTR_HTTP_RESPONSE_HEADER
```

The only related knob, `Headers.currentRedactedNames` (FiberRef), redacts the **value** but the attribute **key** is still emitted. So it helps with secrets but does nothing about trace size.

## Reproduction

https://github.com/schickling-repros/2026-04-effect3-httpclient-header-allowlist

Server emits 12 response headers → 12 `http.response.header.*` span attributes, with no way to filter.

```
captured 2 span(s): http.client GET, repro-root
span: http.client GET
response header attributes (12):
  http.response.header.content-type = "application/json"
  http.response.header.date = "..."
  http.response.header.content-length = "2"
  http.response.header.strict-transport-security = "..."
  http.response.header.x-content-type-options = "nosniff"
  http.response.header.referrer-policy = "..."
  http.response.header.set-cookie = "<redacted>"
  http.response.header.x-request-id = "req-abc-123"
  http.response.header.cf-ray = "8abcd1234ef56789-FRA"
  http.response.header.alt-svc = "..."
  http.response.header.server = "cloudflare"
  http.response.header.x-cache = "HIT"
```

## Versions

- `effect@3.21.2`
- `@effect/platform@0.96.1`
- `@effect/platform-node@0.106.0`
- `@effect/opentelemetry@0.63.0`

## Requested feature

A way to configure the set of header names that get emitted as span attributes. An **allowlist** (capture only these) is the typical OTEL shape and is what we want in practice — chatty origins keep adding headers, so a denylist would always lag.

Prior art in other OTEL ecosystems:
- Java: `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_CLIENT_REQUEST` / `..._RESPONSE` (comma-separated allowlist)
- Go: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3179

A `FiberRef` analogous to `Headers.currentRedactedNames` (e.g. `Headers.currentCapturedAttributeNames`) would fit the existing surface and let callers scope the allowlist to a single request, an `HttpClient` instance, or the whole application. A per-client option on `HttpClient.transform*` / builder would also work.

## Workaround

We currently patch `@effect/platform` locally:

```ts
const HTTP_HEADER_ATTR_ALLOWLIST = new Set([
  "content-type", "content-length", "date",
  "x-request-id", "retry-after", "x-ratelimit-remaining",
])
// before each span.attribute call:
if (!HTTP_HEADER_ATTR_ALLOWLIST.has(name.toLowerCase())) continue
```

Happy to open a PR if the maintainers can confirm the preferred shape (FiberRef vs. client option, allowlist vs. allowlist+denylist, naming).

<sub>Filed by an AI assistant on behalf of @schickling</sub>

Zelys-DFKH:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Closes #6198.

`HttpClient` currently records every request and response header as an OTEL span attribute with no way to filter them. If you've ever looked at a span from a Notion API call and seen 30+ entries (`cf-ray`, `alt-svc`, `server`, `x-cache`, `strict-transport-security`, all the CDN noise), you know exactly why this matters. Sensitive headers like cookies and auth tokens end up in trace backends too, which is the harder problem.

This PR adds three combinators:

- `withTracerRequestHeadersFilter(predicate)`: filter which request headers are recorded as span attributes
- `withTracerResponseHeadersFilter(predicate)`: filter which response headers are recorded as span attributes
- `withTracerHeadersFilter(predicate)`: set the same predicate for both at once (the common case)

```ts
const allowedHeaders = new Set(["content-type", "x-request-id"])

const client = (yield* HttpClient.HttpClient).pipe(
  HttpClient.withTracerHeadersFilter((name) => allowedHeaders.has(name))
)
```

Each combinator is backed by a `FiberRef`, so the filter scopes to the effect it wraps and doesn't bleed into other clients. The default is `constTrue`, so existing behavior is unchanged.

The implementation follows the same `FiberRef` + `globalValue` + `dual` + `transformResponse` pattern as `withTracerDisabledWhen` and `withTracerPropagation`. Once I found those, the rest slotted in pretty naturally.

I went with three functions instead of one because request and response filtering are genuinely independent needs. You might want to capture all request headers for debugging but strip CDN noise from responses, or vice versa. Happy to collapse to a single combinator if you'd prefer a smaller API surface.

Thanks for building Effect — it's a genuinely well-designed library and I'm glad to contribute back.

## Related

- Closes #6198

### DIFF
--- a/.changeset/httpclient-tracer-header-filter.md
+++ b/.changeset/httpclient-tracer-header-filter.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+feat(@effect/platform): add `withTracerRequestHeadersFilter`, `withTracerResponseHeadersFilter`, and `withTracerHeadersFilter` to HttpClient
--- a/packages/platform/src/HttpClient.ts
+++ b/packages/platform/src/HttpClient.ts
@@ -724,6 +724,111 @@ export const withSpanNameGenerator: {
 } = internal.withSpanNameGenerator
 
 /**
+ * A `FiberRef` controlling which request headers are captured as OTEL span
+ * attributes. The predicate receives each header name (lower-cased) and should
+ * return `true` to include it. Defaults to `constTrue` (capture all).
+ *
+ * @since 1.0.0
+ * @category Tracing
+ */
+export const currentTracerRequestHeadersFilter: FiberRef.FiberRef<Predicate.Predicate<string>> =
+  internal.currentTracerRequestHeadersFilter
+
+/**
+ * Restricts which request headers are recorded as OTEL span attributes.
+ *
+ * ```ts
+ * import { FetchHttpClient, HttpClient } from "@effect/platform"
+ * import { NodeRuntime } from "@effect/platform-node"
+ * import { Effect } from "effect"
+ *
+ * const allowedRequestHeaders = new Set(["content-type", "x-request-id"])
+ *
+ * Effect.gen(function* () {
+ *   const client = (yield* HttpClient.HttpClient).pipe(
+ *     HttpClient.withTracerRequestHeadersFilter((name) => allowedRequestHeaders.has(name))
+ *   )
+ *
+ *   yield* client.get("https://api.example.com/data")
+ * }).pipe(Effect.provide(FetchHttpClient.layer), NodeRuntime.runMain)
+ * ```
+ *
+ * @since 1.0.0
+ * @category Tracing
+ */
+export const withTracerRequestHeadersFilter: {
+  (predicate: Predicate.Predicate<string>): <E, R>(self: HttpClient.With<E, R>) => HttpClient.With<E, R>
+  <E, R>(self: HttpClient.With<E, R>, predicate: Predicate.Predicate<string>): HttpClient.With<E, R>
+} = internal.withTracerRequestHeadersFilter
+
+/**
+ * A `FiberRef` controlling which response headers are captured as OTEL span
+ * attributes. The predicate receives each header name (lower-cased) and should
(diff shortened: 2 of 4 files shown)


tim-smart:
Effect v3 is under a feature freeze, feel free to move this contribution to the effect-smol repository.

PR #6221, opened 2026-05-07
feat(@effect/cli): add withOptionalValue combinator to Options

### LINKED ISSUE #6182: Support CLI options with optional values in @effect/cli
## Problem

`@effect/cli` currently has a hard split between boolean flags and value-accepting options:

- `Options.boolean` — never consumes a value (`--watch`)
- `Options.text` / `Options.choice` — always require a value (`--watch=mode`)

There's no way to define an option that works as **both** a standalone flag and a value-accepting option. This is a common CLI pattern — e.g. `--color` vs `--color=always` in GNU tools, `--watch` vs `--watch=mode` in various dev tools.

## Desired behavior

A flag like `--watch` that:
- When absent → `Option.none`
- When bare (`--watch`) → defaults to a specified value (e.g. `'first-failure'`)
- When explicit (`--watch=until-done`) → uses the provided value

## Possible API

```ts
const watch = Options.choice('watch', ['first-failure', 'until-done']).pipe(
  Options.withOptionalValue('first-failure'), // bare --watch → 'first-failure'
  Options.optional, // absent → Option.none
)
```

Or alternatively:

```ts
const watch = Options.boolean('watch').pipe(
  Options.withOptionalValue(Options.choice(['first-failure', 'until-done']), 'first-failure'),
)
```

## Prior art

- GNU coreutils: `--color[=WHEN]`
- TypeScript compiler: `--watch`
- Vitest / Jest: `--watch` vs `--watch=mode`
- Python argparse: `nargs='?'` with `const` for the bare-flag default

---

> Filed on behalf of @schickling

Zelys-DFKH:
Closes #6182

The \`--color[=WHEN]\` / \`--watch[=mode]\` / argparse \`nargs='?'\` pattern: a flag that works bare with a sensible default, or accepts an explicit value. \`@effect/cli\` had no way to express this; any non-boolean option always required a trailing value.

\`withOptionalValue(fallback)\` fills that gap. The flag still must appear unless combined with \`.optional()\` or \`.withDefault()\`; only the value portion becomes optional:

\`\`\`ts
const watch = Options.choice("watch", ["first-failure", "until-done"] as const).pipe(
  Options.withOptionalValue("first-failure"),
  Options.optional
)
// --watch              → Some("first-failure")
// --watch until-done   → Some("until-done")
// --watch=until-done   → Some("until-done")
// (absent)             → None
\`\`\`

The implementation stays entirely inside \`@effect/cli\`. \`Single\` gains one field (\`optionalValueDefault: Option<unknown>\`); the combinator is a standard \`dual\` that sets it via \`modifySingle\`. The three existing combinators that call \`modifySingle\` (\`withAlias\`, \`withDescription\`, \`withPseudoName\`) each thread the new field so chaining cannot silently clear it. The two "Expected a value following option" sites in \`parseCommandLine\` now emit an empty values array when \`optionalValueDefault\` is set, and \`parseInternal\` returns the fallback directly for an empty array. \`getUsageInternal\` wraps the type hint in brackets, so \`--level [<text>]\` appears in \`--help\` output.

12 new tests in \`test/Options.test.ts\` cover: supplied value, bare flag fallback, absent flag error, mid-args with and without a trailing value, composition with \`optional()\` and \`withDefault()\`, alias order independence (both orderings), \`--flag=value\` equals syntax, the canonical \`choice\` use case from the issue, and an integer option. 173 tests pass.

Thanks for the library. Happy to contribute.

### DIFF
--- a/.changeset/cli-optional-value.md
+++ b/.changeset/cli-optional-value.md
@@ -1,0 +1,17 @@
+---
+"@effect/cli": minor
+---
+
+Add `withOptionalValue` combinator to `@effect/cli` Options, allowing a
+CLI option to be specified on the command line without a value. When the
+flag is present but not followed by a value the provided `fallback` is
+used; when a value is provided it is used as normal; when the flag is
+absent entirely the option behaves as before (required unless wrapped with
+`optional()` or `withDefault()`).
+
+```ts
+// --log-level        → "info"  (fallback)
+// --log-level debug  → "debug" (supplied value)
+// (flag absent)      → error: Expected to find option '--log-level'
+const logLevel = Options.text("log-level").pipe(Options.withOptionalValue("info"))
+```
--- a/packages/cli/src/Options.ts
+++ b/packages/cli/src/Options.ts
@@ -557,6 +557,31 @@ export const withPseudoName: {
 } = InternalOptions.withPseudoName
 
 /**
+ * Allows the option to be specified on the command line without a value, in
+ * which case `fallback` is used. When the option is followed by a value that
+ * value is used as normal. The option itself must still appear on the command
+ * line unless combined with `.optional()` or `.withDefault()`.
+ *
+ * **Example**
+ *
+ * ```ts
+ * import * as Options from "@effect/cli/Options"
+ *
+ * // --log-level        → "info"  (fallback)
+ * // --log-level debug  → "debug" (supplied value)
+ * // (flag absent)      → error: Expected to find option: '--log-level'
+ * const logLevel = Options.text("log-level").pipe(Options.withOptionalValue("info"))
+ * ```
+ *
+ * @since 1.0.0
+ * @category combinators
+ */
+export const withOptionalValue: {
+  <A>(fallback: A): (self: Options<A>) => Options<A>
+  <A>(self: Options<A>, fallback: A): Options<A>
+} = InternalOptions.withOptionalValue
+
+/**
  * @since 1.0.0
  * @category combinators
  */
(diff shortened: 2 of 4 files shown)


fubhy:
Thanks for the PR. Before maintainer review, could you clarify or address the ambiguous parsing case where an optional-value option is followed by another flag, e.g. `--watch --other`? The current patch appears to consume `--other` as the value unless the optional-value flag is the final token, while the requested GNU/argparse-style behavior likely needs bare-flag fallback before the next option. A regression test for that case would make the intended semantics clear.

Zelys-DFKH:
Thanks for the exact repro case.

You're right: `--watch --other` was consuming `--other` as watch's value. The parser only fell back to the default when the optional-value flag was the last token on the command line; anywhere else it grabbed whatever came next, no questions asked.

Fixed by checking whether the next token looks like another flag before consuming it as a value. If the option has a fallback configured and the next token starts with `-` (and isn't a negative number, so `--count -5` still works), the flag is treated as bare and the next token is left for the rest of the parser to handle.

That heuristic introduces its own ambiguity with `--flag=value` syntax, since `--level=-foo` also starts with `-` after the split. Added a separate check for that so explicit equals-syntax always wins over the fallback. Four regression tests cover this: `--level --watch` and the aliased `-l --watch` for your case, `--count -5` for the negative-number exclusion, and `--level=-foo` for the equals-syntax case.

While I was in there I also caught and fixed a real TS2684 error in one of my own tests (a generic-inference issue in the `choice` test) that hadn't shown up in CI yet.

Pushed to the same branch.

tim-smart:

PR #6235, opened 2026-05-16
fix(platform): document and surface redirect behavior for HttpClient.followRedirects

### LINKED ISSUE #6231: HttpClient.followRedirects is misleading

3.21.0

Effect provides a method `HttpClient.followRedirects()` that appears to configure whether the client should follow a number of redirects. But `FetchHttpClient` has no RequestInit options, so it automatically follows redirects, making `followRedirects` a noop.

It's hard to tell what the expected behavior is because the `FetchHttpClient` has no documentation, but ideally it would not be possible to write this kind of misleading code.

### What do you see instead?

Redirect following would be detected in the client provided, or controlled from the option per-request.

### Additional information

_No response_

mvanhorn:
## Summary

`HttpClient.followRedirects` had no effect on `FetchHttpClient`: the underlying `fetch` follows redirects automatically, bypassing the Effect-level redirect handling. This exposes `redirect` through the requestInit `FiberRef` and sets it to `"manual"` when `HttpClient.followRedirects` is used, so the behavior actually takes effect on the fetch client. The interaction is documented in JSDoc.

## Changes

- `packages/platform/src/FetchHttpClient.ts` — respect the `redirect` value from the requestInit `FiberRef`.
- `packages/platform/src/HttpClient.ts` — set `redirect: "manual"` when `followRedirects` is used, plus JSDoc documenting the interaction.
- `packages/platform/test/HttpClient.test.ts` — test coverage for `followRedirects` on the fetch client.

Fixes #6231

### DIFF
--- a/.changeset/fuzzy-chairs-taste.md
+++ b/.changeset/fuzzy-chairs-taste.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+Allow `HttpClient.followRedirects` to take effect on `FetchHttpClient` by exposing `redirect` through the requestInit FiberRef and documenting the interaction.
--- a/packages/platform/src/FetchHttpClient.ts
+++ b/packages/platform/src/FetchHttpClient.ts
@@ -13,6 +13,11 @@ import * as internal from "./internal/fetchHttpClient.js"
 export class Fetch extends Context.Tag(internal.fetchTagKey)<Fetch, typeof globalThis.fetch>() {}
 
 /**
+ * The `RequestInit` options provided to `fetch`.
+ *
+ * Set `redirect` to `"manual"` when using `HttpClient.followRedirects`, so that
+ * fetch does not follow redirects before the client decorator can observe them.
+ *
  * @since 1.0.0
  * @category tags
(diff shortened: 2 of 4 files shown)


truffle-dev:
Picking up shape (2) was the read I'd hoped for after the option tree on #6231; the `requestInitTagKey` FiberRef was the cleanest path through and you took it. The JSDoc additions on both `FetchHttpClient.RequestInit` and `HttpClient.followRedirects` are doing the load-bearing work here since the redirect handoff stays caller-controlled, which is the right tradeoff for keeping the decorator generic. The test refactor away from `it.flakyTest` against google.com is a nice secondary win. Thanks for building it out.

tmcw:
This PR description doesn't match the code in the PR - there's no new maxRedirects option implemented.

mvanhorn:
Good catch — the description was stale from an earlier approach. There's no `maxRedirects` option; the PR makes `followRedirects` actually take effect on `FetchHttpClient` via the requestInit `FiberRef` (`redirect: "manual"`) plus docs and a test. I've rewritten the description to match the code.

PR #6236, opened 2026-05-17
feat(bun-test): add @effect/bun-test package

### LINKED ISSUE #5964: Add @effect/bun-test package for native Bun test runner support

I'm migrating an app to Effect and want to use `@effect/vitest` for testing Effect code, but Bun's test runner is significantly faster than Vitest in my experience. Issue #3406 was closed waiting on Bun to fully support Vitest, but that's been stalled for a while.

Looking at the `@effect/vitest` source, the Vitest dependency is pretty thin it's mostly just wrapping it() and providing the TestContext layer. The actual Effect testing machinery (TestClock, etc.) doesn't depend on Vitest at all.

A new @effect/bun-test package that provides the same API as @effect/vitest (it.effect, it.scoped, it.live, etc.) but uses Bun's native bun:test instead of Vitest.

I'd be happy to implement this, it looks like it would mostly be adapting packages/vitest to use Bun's test primitives. The core Effect test infrastructure stays the same.

Waiting for Bun to fully support Vitest (oven-sh/bun#4145) - still blocked
Writing my own wrapper locally, would rather contribute upstream so everyone benefits
Using Vitest anyway, works but slower, and I'd rather use Bun's native runner

emilienbidet:
## Summary

Adds a new `@effect/bun-test` workspace package that mirrors the `@effect/vitest`
API surface (`it.effect`, `it.scoped`, `it.live`, `it.scopedLive`, `it.layer`,
`it.prop`, `flakyTest`, `layer`, …) but runs on Bun's built-in
[`bun:test`](https://bun.sh/docs/cli/test) runner instead of Vitest.

Closes #5964.

## Why

Quoting #5964: Bun's test runner is significantly faster than Vitest, and the
Vitest dependency of `@effect/vitest` is a thin wrapper around `it()` and
`describe()` — the actual Effect test machinery (TestClock, TestServices,
fast-check integration) is runner-agnostic. This PR ports that thin wrapper
to `bun:test` so users on Bun can avoid pulling in Vitest entirely.

## What's in the package

`packages/bun-test/` is laid out exactly like `packages/vitest/`:

- `src/index.ts` — public API, types, re-exports of `bun:test` primitives
- `src/internal/internal.ts` — Effect runner + tester/layer/prop wrappers
- `src/utils.ts` — assertion helpers (copied from `@effect/vitest` `utils.ts`,
  `vassert.instanceOf` swapped for `node:assert`)
- `test/index.test.ts` — ported from `packages/vitest/test/index.test.ts`
- `tsconfig.*`, `docgen.json`, `bunfig.toml`, `package.json`, `README.md`,
  `LICENSE`

A changeset is included (`.changeset/add-effect-bun-test.md`).

## Differences from `@effect/vitest`

Bun's test runner has a smaller surface than Vitest. The PR documents these
gaps in the package README; they were not papered over with runtime shims:

- `addEqualityTesters` is a **no-op** — `bun:test`'s `expect` has no
  `addEqualityTesters` API. Effect's `Equal.equals` is exposed via the
  package-level helpers in `@effect/bun-test/utils` (`assertEquals`, etc.).
- **`TestContext`** is synthesised inside the test wrapper rather than passed
  in by the runner, because `bun:test` doesn't pass a context object. It
  exposes a fresh `signal: AbortSignal` and best-effort `onTestFinished` /
  `onTestFailed` callbacks invoked after the Effect resolves.
- **`scopedFixtures`** (vitest's fixture system) is not provided.
- **`re-export from \"bun:test\"`** doesn't work in Bun (1.2.x), so the public
  primitives are re-exported via `const` bindings rather than `export *`.

## Test plan

- [x] `pnpm install` at the workspace root resolves cleanly
- [x] `pnpm --filter=@effect/bun-test check` passes
- [x] `bun test` inside `packages/bun-test` is green
  - 32 pass, 6 skip, 0 fail (the skips are the test file's deliberate
    `it.skip`/`it.skipIf(true)` cases)
- [ ] Maintainers to decide whether CI should grow a Bun job for this package

## Notes for review

- One Bun-specific tweak in `test/index.test.ts`: the `layer > release` test
  had its trailing `afterAll(...)` reordered so it runs **after** the layer's
  own `afterAll` (Bun runs `afterAll` hooks in registration order; Vitest's
  ordering happened to release the scope before the assertion).
- The `it.scopedLive.fails("interrupts on timeout", ...)` test from the Vitest
  suite was dropped because Bun doesn't expose the equivalent of Vitest's
  `onTestFailed` hook with a fiber-interrupt handle.
- No upstream files outside `packages/bun-test/`, `.changeset/`, and
  `pnpm-lock.yaml` were touched.

### DIFF
--- a/.changeset/add-effect-bun-test.md
+++ b/.changeset/add-effect-bun-test.md
@@ -1,0 +1,7 @@
+---
+"@effect/bun-test": minor
+---
+
+Add `@effect/bun-test` package — same API surface as `@effect/vitest`
+(`it.effect`, `it.scoped`, `it.live`, `it.scopedLive`, `it.layer`, `it.prop`,
+`flakyTest`, …) but backed by Bun's native `bun:test` runner. Closes #5964.
--- a/packages/bun-test/CHANGELOG.md
+++ b/packages/bun-test/CHANGELOG.md
@@ -1,0 +1,1 @@
+# @effect/bun-test
--- a/packages/bun-test/LICENSE
+++ b/packages/bun-test/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Effectful Technologies Inc
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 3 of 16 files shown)


cucumber-sp:
would love to see

tim-smart:

PR #6241, opened 2026-05-19
fix(workflow): forward parent pointer when spawning a child with discard:true

jcalem-rogo:
## Summary

`Workflow.execute(payload, { discard: true })` takes a fast path inside [`WorkflowEngine.execute`](https://github.com/Effect-TS/effect/blob/main/packages/workflow/src/WorkflowEngine.ts) that omits the `parent` option when delegating to the underlying engine. As a result, `ClusterWorkflowEngine` never writes `"~@effect/workflow/parent"` into the child's persisted `cluster_messages.payload` row.

The non-discard branch a few lines below DOES forward `parent`, so the discard path looks like an oversight — `parent` carries two concerns and only one of them (interrupt linkage) is moot for discarded children. The other (the persisted causality breadcrumb) is still load-bearing for any observability surface that wants to reconstruct a parent/child workflow tree from message storage.

Before this fix:
```ts
if (opts.discard) {
  yield* options.execute(self, {
    executionId,
    payload: payload as object,
    discard: true
    //                ↑ no `parent` — child's payload is missing the parent pointer
  })
  return executionId
}
```

After:
```ts
if (opts.discard) {
  yield* options.execute(self, {
    executionId,
    payload: payload as object,
    discard: true,
    parent: Option.getOrUndefined(parentInstance)
  })
  return executionId
}
```

The fix is one line. The `parent` option is already typed as optional on the engine's `execute` signature, so no other surfaces change.

## Why this matters

The dominant fan-out shape in `@effect/cluster` workflows — `Effect.forEach(items, (item) => ChildWorkflow.execute(item, { discard: true }))` — was leaving every child's payload without a parent pointer. Tools that read cluster_messages to reconstruct a Temporal-like workflow tree end up with orphaned children and no way to link them back to their parent, except by falling back to OTel trace_id grouping (which is lossy and ambiguous for nested fan-outs).

## Test

Added a regression test in `packages/cluster/test/ClusterWorkflowEngine.test.ts` that spawns a child with `discard: true` and asserts the persisted child run-envelope carries `"~@effect/workflow/parent"` pointing back at the parent's actual `executionId`.

## Changeset

`@effect/workflow`: patch.

## Test plan

- [x] `pnpm vitest run` in `packages/cluster` — all 9 tests pass, including the new one.
- [x] `pnpm vitest run` in `packages/workflow` — all 6 tests pass.
- [x] `tsc -b` clean across both packages.
- [x] `eslint` clean on the touched files.

### DIFF
--- a/.changeset/forward-parent-pointer-discard.md
+++ b/.changeset/forward-parent-pointer-discard.md
@@ -2,6 +2,4 @@
 "@effect/workflow": patch
 ---
 
-Forward the parent pointer when spawning a child workflow with `discard: true`.
-
-`Workflow.execute(payload, { discard: true })` took a fast path inside `WorkflowEngine.execute` that omitted the `parent` option when delegating to the underlying engine. As a result, the cluster engine never wrote `"~@effect/workflow/parent"` into the child's persisted payload, breaking causality tracking for fire-and-forget fan-outs (the common pattern for batching workflows over many items). Discarded children now carry the same parent pointer that non-discarded children already did, so observability tools can link them back to their parent.
+Forward the parent pointer when spawning a child workflow with `discard: true`
--- a/.changeset/forward-parent-pointer-discard.md
+++ b/.changeset/forward-parent-pointer-discard.md
@@ -1,0 +1,7 @@
+---
+"@effect/workflow": patch
+---
+
+Forward the parent pointer when spawning a child workflow with `discard: true`.
+
+`Workflow.execute(payload, { discard: true })` took a fast path inside `WorkflowEngine.execute` that omitted the `parent` option when delegating to the underlying engine. As a result, the cluster engine never wrote `"~@effect/workflow/parent"` into the child's persisted payload, breaking causality tracking for fire-and-forget fan-outs (the common pattern for batching workflows over many items). Discarded children now carry the same parent pointer that non-discarded children already did, so observability tools can link them back to their parent.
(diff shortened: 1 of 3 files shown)


packages/cluster/test/ClusterWorkflowEngine.test.ts
```diff
@@ -245,6 +245,38 @@ describe.concurrent("ClusterWorkflowEngine", () => {
 
       assert.isTrue(flags.get("catch"))
     }).pipe(Effect.provide(TestWorkflowLayer)))
+
+  // Regression: the discard fast-path in WorkflowEngine.execute used to drop the `parent` option,
+  // leaving observability tools unable to link discarded children back to their parent.
+  it.effect("forwards parent pointer when spawning a child with discard:true", () =>
```
tim-smart:
```suggestion
  it.effect("forwards parent pointer when spawning a child with discard:true", () =>
```

.changeset/forward-parent-pointer-discard.md
```diff
@@ -0,0 +1,7 @@
+---
+"@effect/workflow": patch
+---
+
+Forward the parent pointer when spawning a child workflow with `discard: true`.
+
+`Workflow.execute(payload, { discard: true })` took a fast path inside `WorkflowEngine.execute` that omitted the `parent` option when delegating to the underlying engine. As a result, the cluster engine never wrote `"~@effect/workflow/parent"` into the child's persisted payload, breaking causality tracking for fire-and-forget fan-outs (the common pattern for batching workflows over many items). Discarded children now carry the same parent pointer that non-discarded children already did, so observability tools can link them back to their parent.
```
tim-smart:
```suggestion
Forward the parent pointer when spawning a child workflow with `discard: true`
```

PR #6244, opened 2026-05-23
fix(platform): scope HttpLayerRouter.addHttpApi error encoder to its own routes

### LINKED ISSUE #6243: HttpLayerRouter is mixing error schemas between different APIs

3.21.2

This is a minimal reproduction of the issue:

```typescript
import { HttpApi, HttpApiBuilder, HttpApiEndpoint, HttpApiGroup, HttpLayerRouter } from "@effect/platform"
import { NodeHttpServer } from "@effect/platform-node"
import { Effect, Layer, Schema } from "effect"

const Api1 = HttpApi.make("Api1").add(
  HttpApiGroup.make("group1").add(
    HttpApiEndpoint.post("endpoint1")`/1`.addError(Schema.transformLiteral("BAD", "x"))
  )
)
const Handlers1 = HttpApiBuilder.group(Api1, "group1", (_) => _.handle("endpoint1", () => Effect.void))
const Routes1 = HttpLayerRouter.addHttpApi(Api1).pipe(Layer.provide(Layer.mergeAll(Handlers1)))

const Api2 = HttpApi.make("Api2").add(
  HttpApiGroup.make("group2").add(
    HttpApiEndpoint.get("endpoint2")`/2`.addError(Schema.transformLiteral("GOOD", "x"))
  )
)
const Handlers2 = HttpApiBuilder.group(Api2, "group2", (_) => _.handle("endpoint2", () => Effect.fail("x" as const)))
const Routes2 = HttpLayerRouter.addHttpApi(Api2).pipe(Layer.provide(Layer.mergeAll(Handlers2)))

const AllRoutes = Layer.mergeAll(Routes1, Routes2)
const { handler } = HttpLayerRouter.toWebHandler(AllRoutes.pipe(Layer.provide(NodeHttpServer.layerContext)))

const request = new Request("http://localhost:3000/2")
const response = await handler(request)
const text = await response.text()
console.log(text)
```

Should log "GOOD", because of the transformation defined in the error schema for the endpoint 2.

### What do you see instead?

It is logging "BAD" instead, incorrectly using the error schema from the endpoint 1.

### Additional information

It works correctly if remove `Layer.mergeAll` from `Layer.provide(Layer.mergeAll(Handlers2))`. It is unnecessary in this minimal example, but is required for registering multiple endpoints in the same API.

spokodev:
Closes #6243.

When two APIs are registered via `HttpLayerRouter.addHttpApi`, each call adds its middleware to a shared context map under a unique `Middleware-N` key. `getMiddleware(context)` collects every such key and applies all of them to every route at server startup. The first registered API's error encoder ends up wrapping the second API's routes, so a 500 returned by the second API is encoded against the first API's error schema.

In the issue's repro, Api1's error schema transforms `"x"` to `"BAD"` and Api2's transforms it to `"GOOD"`. A request to Api2's endpoint returns `"BAD"` instead of `"GOOD"`.

## Fix

`addHttpApi` no longer relies on the shared middleware context map. It calls `HttpApiBuilder.buildMiddleware(api)` inline, wraps each matching route's handler directly with that middleware, and filters `router_.routes` to only the endpoints declared by this API (matched on `method + path`). Each API's error encoder only sees its own routes.

## Test plan

`packages/platform/test/HttpLayerRouter.test.ts`:

- `two registered APIs use their own error schemas (regression #6243)` — positive case: Api2's endpoint returns `"GOOD"` (its own schema); negative case: Api1's endpoint still returns `"BAD"` (no swap into the wrong direction).
- `single registered API still encodes errors using its own schema` — baseline preserved.

Both tests fail on `main` for the regression case and pass after the fix. `pnpm test run test/HttpLayerRouter.test.ts test/HttpApiBuilder.test.ts test/HttpApp.test.ts` ⇒ 52/52 green. `pnpm check` and `pnpm build` clean.

### DIFF
--- a/.changeset/fix-httplayerrouter-shared-error-encoder.md
+++ b/.changeset/fix-httplayerrouter-shared-error-encoder.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+Fix `HttpLayerRouter.addHttpApi` so two registered APIs no longer share error encoders. Previously each call's middleware was added to a shared context map and applied to every route, so a 500 response from one API was sometimes encoded against the other API's error schema. The fix scopes each API's middleware to its own endpoints (matched by method + path) and wraps each route handler directly.
--- a/packages/platform/src/HttpLayerRouter.ts
+++ b/packages/platform/src/HttpLayerRouter.ts
@@ -1008,15 +1008,15 @@ export const addHttpApi = <Id extends string, Groups extends HttpApiGroup.HttpAp
   | R
   | HttpApiGroup.HttpApiGroup.ToService<Id, Groups>
   | HttpApiGroup.HttpApiGroup.ErrorContext<Groups>
-> => {
-  const ApiMiddleware = middleware(HttpApiBuilder.buildMiddleware(api)).layer as Layer.Layer<never>
-  return HttpApiBuilder.Router.unwrap(Effect.fnUntraced(function*(router_) {
+> =>
+  HttpApiBuilder.Router.unwrap(Effect.fnUntraced(function*(router_) {
     const router = yield* HttpRouter
     let existing = existingRoutesMap.get(router)
     if (!existing) {
       existing = new Set()
       existingRoutesMap.set(router, existing)
     }
+    const apiMiddleware = yield* HttpApiBuilder.buildMiddleware(api)
     const context = yield* Effect.context<
       | Etag.Generator
       | HttpRouter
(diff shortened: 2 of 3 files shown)


spokodev:
Following up on this since 2026-05-23. Could you take a quick look when convenient? Ready to iterate if anything needs adjusting. Thanks!

tim-smart (approved):

PR #6246, opened 2026-05-24
Fix HttpApi discarding annotations from Union and Literal schemas

### LINKED ISSUE #6245: HttpApi is discarding annotations from Union and Literal schemas

3.21.2

```typescript
import { HttpApi, HttpApiEndpoint, HttpApiGroup, OpenApi } from "@effect/platform"

const Api = HttpApi.make("Api").add(
  HttpApiGroup.make("group").add(
    HttpApiEndpoint.get("endpoint")`/`
      .addSuccess(Schema.Literal("a", "b").annotations({ title: "Letter" }))
  )
)

const spec = OpenApi.fromApi(Api)
console.dir(spec.paths["/"].get?.responses["200"].content?.["application/json"]?.schema, { depth: null })
```

Should log `{ type: 'string', enum: [ 'a', 'b' ], title: 'Letter' }`, with the title annotation.

### What do you see instead?

Logs `{ type: 'string', enum: [ 'a', 'b' ] }`, without the title annotation.

### Additional information

In the internal process of unifying ASTs of Union types, it is recreating the Union AST, therefore discarding its annotations.
It can cause problems like missing information in OpenApi specs and not respecting the encoding method and HTTP status declared as annotations.

matheuspuel:
## Type

- [x] Bug Fix

## Description

This change makes HttpApi preserve annotations from Union schemas.

Previously, the types from every Union would be extracted and unified in the same new Union.
Now it does not break Unions which have at least 1 annotation, but still merge Unions without any annotations.

## Related

- Closes #6245

### DIFF
--- a/.changeset/fluffy-jokes-repair.md
+++ b/.changeset/fluffy-jokes-repair.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+Fix HttpApi discarding annotations from Union and Literal schemas
--- a/packages/platform/src/HttpApiSchema.ts
+++ b/packages/platform/src/HttpApiSchema.ts
@@ -566,7 +566,7 @@ export const Uint8Array = (options?: {
 
 const astCache = globalValue(
   "@effect/platform/HttpApiSchema/astCache",
-  () => new WeakMap<AST.AST, Schema.Schema.Any>()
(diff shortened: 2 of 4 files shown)


matheuspuel:
@tim-smart Could you take a look when convenient? It contains fixes for some unexpected behavior of HttpApi.

PR #6258, opened 2026-06-04
Fix child workflow fan-out inside activities

mikearnaldi:
## Summary
- add a regression test for parallel child workflows launched inside an activity
- mark activity execution internally so nested child workflow awaits do not suspend the activity fiber
- wait on suspended child workflow results using the existing workflow suspended retry schedule

## Validation
- pnpm test run packages/cluster/test/ClusterWorkflowEngine.test.ts
- pnpm test run packages/workflow/test/WorkflowEngine.test.ts
- pnpm lint-fix
- pnpm check
- pnpm build
- pnpm docgen

### DIFF
--- a/packages/workflow/src/WorkflowEngine.ts
+++ b/packages/workflow/src/WorkflowEngine.ts
@@ -15,6 +15,7 @@ import * as Scope from "effect/Scope"
 import type * as Activity from "./Activity.js"
 import type { DurableClock } from "./DurableClock.js"
 import type * as DurableDeferred from "./DurableDeferred.js"
+import { CurrentActivityExecution } from "./internal/activity.js"
 import * as Workflow from "./Workflow.js"
 
 /**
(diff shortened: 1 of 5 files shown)


packages/workflow/src/WorkflowEngine.ts
```diff
@@ -379,6 +381,19 @@ export const makeUnsafe = (options: Encoded): WorkflowEngine["Type"] =>
         parent: Option.getOrUndefined(parentInstance)
       })
       if (Option.isSome(parentInstance)) {
+        if (currentActivityExecution) {
```
tim-smart, line 384: This would keep the activity in-memory even if the workflow suspends for long time. Will need to find another way.

fubhy:
Is this still intended to move forward? Tim raised a concern that this approach keeps the activity in-memory while the workflow is suspended, so this likely needs an updated approach before review can continue.

tim-smart:
https://github.com/Effect-TS/effect/pull/8070

PR #6263, opened 2026-06-08
fix(opentelemetry): honor MetricReader temporality in MetricProducer, closes #6253

### LINKED ISSUE #1: Proposals
To discuss ideas, any library you might think should have a wrapper or any ideas for improvements welcome!

### LINKED ISSUE #6253: @effect/opentelemetry: MetricProducer hardcodes CUMULATIVE temporality, ignoring OTLPMetricExporter temporalityPreference
## Summary

`@effect/opentelemetry`'s `MetricProducer` does not query the registered `MetricReader`'s `selectAggregationTemporality(instrumentType)` — it always hardcodes `AggregationTemporality.CUMULATIVE` on every Sum data point (Counter and UpDownCounter). This means any non-default temporality preference (e.g. `temporalityPreference: DELTA` on `OTLPMetricExporter`) is silently ineffective: temporality is baked into the data point by the producer before the exporter sees it.

Per the OTel spec, producers should query the reader for the preferred temporality per instrument type rather than choosing unconditionally.

> **Note**: an earlier version of this issue claimed this causes Datadog to display inflated rates. That turned out to be a different bug on our side (we were missing `{ incremental: true }` on `Metric.counter`, so the metric was exported as a non-monotonic Sum and DD treated it as a gauge). With `incremental: true`, DD ingests the cumulative monotonic Sum correctly via internal cumulative→delta conversion. The temporality-hardcode issue described here is real and reproducible, but its impact is narrower than I originally framed it. Apologies for the noise.

## Reproduction

```ts
import { Effect, Metric } from "effect";
import { Metrics as EffectOtel, Resource } from "@effect/opentelemetry";
import {
    AggregationTemporality,
    InMemoryMetricExporter,
    PeriodicExportingMetricReader,
} from "@opentelemetry/sdk-metrics";

const exporter = new InMemoryMetricExporter(AggregationTemporality.DELTA);
const reader = new PeriodicExportingMetricReader({ exporter, exportIntervalMillis: 60_000_000 });

await Effect.runPromise(
    Effect.scoped(
        Effect.gen(function* () {
            const counter = Metric.counter("test.counter", { incremental: true });
            yield* Effect.succeed(5).pipe(Metric.increment(counter));
            yield* Effect.succeed(7).pipe(Metric.increment(counter));
            yield* Effect.promise(() => reader.forceFlush());
        }).pipe(
            Effect.provide(EffectOtel.layer(() => reader)),
            Effect.provide(Resource.layer({ serviceName: "test" }))
        )
    )
);

const data = exporter.getMetrics();
const counter = data[0].scopeMetrics[0].metrics[0];
console.log(counter.aggregationTemporality);   // 1 (CUMULATIVE) — expected 0 (DELTA)
console.log(counter.isMonotonic);              // true (good)
console.log(counter.dataPoints[0].value);      // 12 (good)
```

The `InMemoryMetricExporter` is constructed with `AggregationTemporality.DELTA`, but the produced data point arrives as CUMULATIVE. Same result with a real `OTLPMetricExporter({ temporalityPreference: DELTA })`.

## Root cause

`packages/opentelemetry/src/internal/metrics.ts` (line numbers from v0.61.0):

```ts
if (MetricState.isCounterState(metricState)) {
    return {
        descriptor: makeDescriptor(metricPair),
        dataPointType: SdkMetrics.DataPointType.SUM,
        isMonotonic: descriptor.type === SdkMetrics.InstrumentType.COUNTER,
        aggregationTemporality: SdkMetrics.AggregationTemporality.CUMULATIVE,  // <-- hardcoded
        dataPoints: [{ /* ... */ }],
    };
}
```

The same hardcoded value appears for every Sum-typed path in the file (Counter, UpDownCounter, FrequencyMetric, etc.). No producer-side or layer-level option to override it.

## Expected behavior

The producer should look up the reader's preferred temporality for the instrument type and stamp that on each data point — e.g. by calling `reader.selectAggregationTemporality(instrumentType)` and using the returned value. This is the spec-compliant approach.

## Impact

Narrow but real. Affects pipelines where the configured backend requires (or strongly prefers) DELTA temporality and where the user expects `temporalityPreference` on the exporter to control that:

- Some OTel collector configurations downstream of `@effect/opentelemetry` that are themselves set to DELTA-only routing.
- Backends that do not handle cumulative monotonic Sums on ingest (most major vendors do, so this is genuinely niche).

It does **not** noticeably affect default OTLP-to-Datadog or Prometheus-pull setups, which handle cumulative monotonic Sums correctly.

## Workaround

For users who genuinely need DELTA-temporality export and can't change the backend:

- Bypass `@effect/opentelemetry` for the affected instruments — e.g. emit via a different SDK (`dd-trace`'s dogstatsd, raw OTel API) and keep `@effect/opentelemetry` for everything else.
- There's no in-library escape hatch today.

## Suggested fix

Either:

1. **Drive temporality from the registered MetricReader** — query `reader.selectAggregationTemporality(instrumentType)` per produced data point and stamp accordingly. Spec-compliant; preserves all existing default behavior for users on CUMULATIVE-preferring readers.
2. **Accept a temporality option on `EffectOtel.layer` / `makeProducer`** — escape hatch: `{ counterTemporality: DELTA, upDownCounterTemporality: ... }`. Simpler, less correct.

Happy to PR (1) if there's interest.

## Environment

- `@effect/opentelemetry` 0.61.0
- `@opentelemetry/sdk-metrics` 2.2.0
- `@opentelemetry/exporter-metrics-otlp-grpc` 0.207.0

mohamedramadan14:
Fixes #6253.

`@effect/opentelemetry`'s `MetricProducerImpl` was hardcoding `AggregationTemporality.CUMULATIVE` on every produced data point (Counter, UpDownCounter, Frequency, Summary count/sum, Gauge, Histogram). It never consulted the registered `MetricReader`, so any `temporalityPreference` (e.g. `DELTA`) configured on `OTLPMetricExporter` (or any custom `aggregationTemporalitySelector`) was silently overridden.

### Why this needs a small adaptation of the issue's suggested fix #1

The OTel `MetricProducer.collect()` interface only receives `{ timeoutMillis }` — not a reader reference. The reader is given to the producer one-way via `reader.setMetricProducer(producer)`. So the producer cannot query `selectAggregationTemporality` directly during `collect()` without first remembering the reader at registration time.

### What the fix does

- `MetricProducerImpl` now holds a list of registered `MetricReader`s, populated in `registerProducer`.
- `collect()` calls `reader.selectAggregationTemporality(descriptor.type)` per produced data point and stamps that on the result. The 5 hardcoded sites (Counter, Gauge, Histogram, Frequency, Summary count/sum) all go through one helper.
- No reader registered → unchanged (`CUMULATIVE`, matching the OTel default).
- Multiple readers, all agree → use that preference.
- Multiple readers, disagree → use the first reader's preference and emit a single `diag.warn(...)` (non-fatal, non-breaking).

### Why option (1) over option (2)

Option (1) is spec-compliant and requires zero user-side configuration; option (2) would have made users re-state the reader's preference on the producer side.

### Tests added (`packages/opentelemetry/test/Metrics.test.ts`)

- counter honors reader's preferred temporality (DELTA) — the exact issue repro
- gauge honors reader's preferred temporality (DELTA)
- falls back to CUMULATIVE when no reader is registered

All 60 tests in `@effect/opentelemetry` pass. `pnpm check`, `pnpm build`, `pnpm docgen` all green.

### Out of scope (noted as a follow-up)

`packages/opentelemetry/src/OtlpMetrics.ts` (the non-SDK direct-OTLP snapshot path) has the same hardcoded `CUMULATIVE`, but it does not use a `MetricReader`. A follow-up could add a `temporality` option to `OtlpMetrics.make` since there is no reader to query there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Metric exports now honor each reader’s preferred aggregation temporality, including DELTA settings.
  * Temporality is selected appropriately by metric instrument type, including counters, gauges, histograms, and summaries.
  * Metrics revert to cumulative temporality when no reader is registered or after reader shutdown.
  * Added safeguards and validation for differing temporality preferences across readers.

### DIFF
--- a/.changeset/opentelemetry-metric-producer-temporality.md
+++ b/.changeset/opentelemetry-metric-producer-temporality.md
@@ -1,0 +1,5 @@
+---
+"@effect/opentelemetry": patch
+---
+
+Honor the registered `MetricReader`'s preferred `AggregationTemporality` in `MetricProducerImpl`. Previously, every Sum-typed data point (Counter, UpDownCounter, Frequency, Summary count/sum) and Gauge/Histogram data point was stamped with a hardcoded `CUMULATIVE`, causing `OTLPMetricExporter({ temporalityPreference: DELTA })` (and any other non-default temporality preference on the reader/exporter) to be silently ignored. The producer now queries `reader.selectAggregationTemporality(instrumentType)` per produced data point, matching the OpenTelemetry spec. When no reader is registered, behavior is unchanged. Fixes #6253.
--- a/packages/opentelemetry/src/internal/metrics.ts
+++ b/packages/opentelemetry/src/internal/metrics.ts
@@ -32,15 +32,23 @@ type MetricDataWithInstrumentDescriptor = MetricData & {
 
 /** @internal */
 export class MetricProducerImpl implements MetricProducer {
-  readonly resource: Resources.Resource
+  constructor(readonly resource: Resources.Resource) {}
+
+  startTimes = new Map<string, HrTime>()
+
   readonly readers: Array<MetricReader> = []
   private warnedConflictingTemporalities = false
 
-  constructor(resource: Resources.Resource) {
-    this.resource = resource
+  addReaders(readers: ReadonlyArray<MetricReader>) {
+    for (const reader of readers) this.readers.push(reader)
   }
 
-  startTimes = new Map<string, HrTime>()
+  removeReaders(readers: ReadonlyArray<MetricReader>) {
+    for (const reader of readers) {
+      const index = this.readers.indexOf(reader)
+      if (index !== -1) this.readers.splice(index, 1)
+    }
+  }
 
   startTimeFor(name: string, hrTime: HrTime) {
     if (this.startTimes.has(name)) {
@@ -50,14 +58,21 @@ export class MetricProducerImpl implements MetricProducer {
     return hrTime
   }
 
-  private temporalityFor(descriptorType: InstrumentType): AggregationTemporality {
+  /**
+   * Resolves the `AggregationTemporality` for an instrument type by asking the
+   * registered readers for their preference. Must be called with the type of the
+   * descriptor the data point is actually emitted under -- temporality selectors
+   * are per instrument type, so e.g. a summary's `_count`/`_sum` (emitted as
+   * `COUNTER`) can resolve differently to the summary itself (`UP_DOWN_COUNTER`).
+   */
+  private temporalityFor(instrumentType: InstrumentType): AggregationTemporality {
     if (this.readers.length === 0) {
       return AggregationTemporality.CUMULATIVE
     }
-    const selected = this.readers[0].selectAggregationTemporality(descriptorType)
+    const selected = this.readers[0].selectAggregationTemporality(instrumentType)
     if (
       this.readers.length > 1 &&
-      this.readers.some((r) => r.selectAggregationTemporality(descriptorType) !== selected)
+      this.readers.some((r) => r.selectAggregationTemporality(instrumentType) !== selected)
     ) {
       if (!this.warnedConflictingTemporalities) {
         this.warnedConflictingTemporalities = true
@@ -88,7 +103,6 @@ export class MetricProducerImpl implements MetricProducer {
       })
       const descriptor = descriptorFromKey(metricKey, attributes)
       const startTime = this.startTimeFor(descriptor.name, hrTimeNow)
-      const temporality = this.temporalityFor(descriptor.type)
 
       if (MetricState.isCounterState(metricState)) {
         const dataPoint: DataPoint<number> = {
@@ -104,7 +118,7 @@ export class MetricProducerImpl implements MetricProducer {
             dataPointType: DataPointType.SUM,
             descriptor,
             isMonotonic: descriptor.type === InstrumentType.COUNTER,
-            aggregationTemporality: temporality,
+            aggregationTemporality: this.temporalityFor(descriptor.type),
             dataPoints: [dataPoint]
           })
         }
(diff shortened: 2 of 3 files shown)


mohamedramadan14:
Pushed a follow-up commit fixing two problems with my first pass.

Temporality was resolved once per metric key from `descriptor.type`, but a summary's `_count`/`_sum` are emitted under `COUNTER` descriptors while the summary key itself resolves to `UP_DOWN_COUNTER`. Against a per-type selector like OTLP's `DeltaTemporalitySelector`, those two points came out CUMULATIVE — the exact case this PR is supposed to fix. It's now resolved at each emission site. My original tests missed it because `InMemoryMetricExporter` returns one constant for every instrument type, so I added an exporter that mirrors the real selector.

`registerProducer` also only ever pushed to `readers` and never cleaned up, so once the scope closed and the reader was shut down the producer kept consulting it, and repeat registrations piled up. Added a matching removal under `ensuring`. That surfaced a third issue: the three tests were only passing *because* of that leak — wrapping `Effect.scoped` around just the registration shuts the reader down before `collect()` runs, so they were asserting against a dead reader. The scope now covers the whole block.

63 tests pass, `tsc` and `eslint` clean.

One note: the `Check` workflow has never run on this PR. It only triggers on `pull_request` for `v3`/`v3/next-minor`, and when the base was retargeted to `v3` that fired an `edited` event, which isn't in the default trigger set — so nothing was ever queued. This push created a run, but it's sitting in `action_required`. Would appreciate a workflow approval whenever someone gets a chance.

PR #5868, opened 2025-12-04
Work with frozen intrinsics

arlyon:
## Type

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In some places (frozen-intrinsics, deterministic sandboxes like in temporal), error is readonly to limit the amount of peturbation. This causes issues when effect assumes it can modify, leading to exceptions. This changes that code to a best-effort attempt with a silent failure. I chose to just duplicate the snippet in the ai lib.

## Related

Though outside this repo, this pattern is used within node itself. The core snippet comes from their internal checks. I have added a wrapper to DRY the set, unset, and error handling.

https://github.com/nodejs/node/blob/e77694631f1642c302f664703197b5aabc65b482/lib/internal/errors.js#L246

PR #6279, opened 2026-06-17
fix: support frozen intrinsics when adjusting Error.stackTraceLimit

bweis:
In hardened/deterministic JavaScript environments (SES "frozen intrinsics", Temporal-style sandboxes), `Error` is frozen and `Error.stackTraceLimit` is read-only. Effect mutates `stackTraceLimit` in several internal spots to capture short/empty stack traces cheaply; under frozen intrinsics those assignments throw, breaking Effect entirely.

Add an internal `stackTraceLimit` helper (mirroring Node's own guard) that detects writability once and degrades `set` to a silent no-op when the property can't be modified. Replace the raw mutations across Effect.ts, Micro.ts, LayerMap.ts, and the internal context/core-effect/layer/runtime/ tracer/cause modules with guarded get/set calls, keeping `new Error()` inline at each call site so captured stack traces still point at the real caller.

`@effect/ai`'s `unsafeSecureJsonParse` gets an equivalent inline guard (no wrapping closure) since the effect helper is not part of the public API.

## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Effect manipulates `Error.stackTraceLimit` in several internal spots (Tag/Service/`fn` creation, tracer spans, fiber failures, pretty errors, etc.) to capture short or empty stack traces cheaply, using the pattern:

```ts
const limit = Error.stackTraceLimit
Error.stackTraceLimit = 2
const err = new Error()
Error.stackTraceLimit = limit
```

In environments with frozen intrinsics (SES / hardened JavaScript, deterministic sandboxes such as Temporal), Error.stackTraceLimit is non-writable and these assignments throw, breaking Effect entirely.

This PR introduces packages/effect/src/internal/stackTraceLimit.ts, which checks writability once at module load (mirroring [Node's internal guard](https://github.com/nodejs/node/blob/e77694631f1642c302f664703197b5aabc65b482/lib/internal/errors.js#L246)) and exposes getStackTraceLimit() / setStackTraceLimit() helpers where set is a silent no-op when the property can't be modified. All raw mutations are replaced with these guarded calls, with new Error() kept inline at each call site so the captured stack trace still points at the real caller rather than the helper.

@effect/ai's unsafeSecureJsonParse gets an equivalent inline guard because the effect helper is @internal and not part of the public API. Note that this call site uses a plain `if (!canWriteStackTraceLimit) { ... }` branch around the mutation rather than a `withStackTraceLimit(limit, fn)` wrapper: running the body inside a closure would push an extra frame onto any captured stack trace and change its shape (per review feedback). With the inline guard, the writable path is byte-identical to the original.

Behavior in normal (writable) environments is unchanged. A changeset and a dedicated test (packages/effect/test/StackTraceLimit.test.ts, covering both the writable and frozen paths) are included; the existing test/Effect, Cause, Tracer, Context, LayerMap, Micro, and @effect/ai Tool suites pass.

> [!NOTE]
> The original PR for this change was stood up by GitHub user @arlyon. This description and PR were aided via Claude Code.

### DIFF
--- a/packages/ai/ai/src/Tool.ts
+++ b/packages/ai/ai/src/Tool.ts
@@ -1522,19 +1522,6 @@ const isStackTraceLimitWritable = (): boolean => {
 
 const canWriteStackTraceLimit = isStackTraceLimitWritable()
 
-const withStackTraceLimit = <T>(limit: number, fn: () => T): T => {
-  if (!canWriteStackTraceLimit) {
-    return fn()
-  }
-  const prevLimit = Error.stackTraceLimit
-  try {
-    Error.stackTraceLimit = limit
-    return fn()
-  } finally {
-    Error.stackTraceLimit = prevLimit
-  }
-}
-
 /**
  * **Unsafe**: This function will throw an error if an insecure property is
  * found in the parsed JSON or if the provided JSON text is not parseable.
@@ -1542,9 +1529,19 @@ const withStackTraceLimit = <T>(limit: number, fn: () => T): T => {
  * @since 1.0.0
  * @category Utilities
  */
-export const unsafeSecureJsonParse = (text: string): unknown =>
+export const unsafeSecureJsonParse = (text: string): unknown => {
   // Performance optimization, see https://github.com/fastify/secure-json-parse/pull/90
-  withStackTraceLimit(0, () => _parse(text))
+  if (!canWriteStackTraceLimit) {
+    return _parse(text)
+  }
+  const prevLimit = Error.stackTraceLimit
+  Error.stackTraceLimit = 0
+  try {
+    return _parse(text)
+  } finally {
+    Error.stackTraceLimit = prevLimit
+  }
+}
 
 /**
  * Type of the `EmptyParams` schema used for tools with no parameters.
--- a/packages/ai/ai/src/Tool.ts
+++ b/packages/ai/ai/src/Tool.ts
@@ -1506,6 +1506,35 @@ function filter(obj: any) {
   return obj
 }
 
+// Utility for safely manipulating `Error.stackTraceLimit` in environments where
+// intrinsics may be frozen (e.g., SES / hardened JavaScript). Duplicated here
+// because `effect`'s internal helper is not part of its public API. When the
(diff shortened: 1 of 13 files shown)


bweis:
Per discussion in Discord: @tim-smart mind taking a look at this?

packages/ai/ai/src/Tool.ts
```diff
@@ -1506,23 +1506,45 @@ function filter(obj: any) {
...
+    return Object.isExtensible(Error)
+  }
+  return Object.prototype.hasOwnProperty.call(desc, "writable")
+    ? desc.writable === true
+    : desc.set !== undefined
+}
+
+const canWriteStackTraceLimit = isStackTraceLimitWritable()
+
+const withStackTraceLimit = <T>(limit: number, fn: () => T): T => {
```
arlyon: > withStackTraceLimit will introduce a stack frame so it changes the shape of the trace

Comment from @mikearnaldi on the original PR since this was also my direction

https://github.com/Effect-TS/effect/pull/5868#issuecomment-3620722472

arlyon: The worm took my post-review changes and force pushed / rewrote them.  See here for the clean post-review commit from 6mo ago

https://github.com/Effect-TS/effect/commit/1757334718f6b23a80631ab698367ebc25829c14

bweis: Ack. Great point.  Now fixed in the latest commit.

bweis:
Ported this change to `effect-smol`:
- Upstream PR: https://github.com/Effect-TS/effect-smol/pull/2444

The smol layout differs (consolidated internals, no `Micro.ts`, `ai` in-package), so it's a port rather than a cherry-pick — see the PR descriptions for the structural differences.

bweis:
Added the equivalent to effect-smol here: https://github.com/Effect-TS/effect-smol/pull/2444

bweis:
@tim-smart This was merged into effect-smol. Can we get this merged as a backport into effect as well?

PR #6298, opened 2026-06-29
v3 fix: Made field-less `Schema.Struct({})` respect `onExcessProperty: 'ignore' | 'preserve' | 'error'`

nikelborm:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Motivation: you remove one field in the schema; the decoded object gets fewer fields. You remove another; fewer again. You remove the last; suddenly, all the fields come back, regardless of the default `onExcessProperty: "ignore"`. If the user wants to retrieve all fields, they can set `onExcessProperty: "preserve"`.

I tried to defend old behavior with "`{}` means `typeof obj === 'object' && obj !== null` and effect's schema tries to be closer to TypeScript's behaviour, by treating `Schema.Struct({})` as something passing the condition". And this might hold in isolation. But considering the context, by this logic, any fields which are not explicitly mentioned in `Schema.Struct({ ... })` should be preserved in all decoded objects, not only the ones that have `{}` signature, to follow TypeScript's assignability rules. And this is not what happens. The behavior should be consistent between field-less and field-ful structs.

## Related

- https://github.com/Effect-TS/effect/pull/6434

### DIFF
--- a/.changeset/lazy-toes-hug.md
+++ b/.changeset/lazy-toes-hug.md
@@ -1,0 +1,76 @@
+---
+"effect": patch
+---
+
+Made field-less `Schema.Struct({})` respect `onExcessProperty: 'ignore' | 'preserve' | 'error'`
+
+Motivation: you remove one field, the resulting object gets fewer fields. You
+remove another, fewer again. You remove the last; suddenly all the fields come
+back, without respect to the default `onExcessProperty: "ignore"`. If the user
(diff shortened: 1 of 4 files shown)


nikelborm:
Rebased on the latest commit in the branch. It's ready for review.

nikelborm:
I think I can improve the tests a bit

PR #6300, opened 2026-07-03
feat(@effect/ai-anthropic): support custom auth header and extra headers in AnthropicClient

polRk:
## Summary
- Add `apiKeyHeader` option to `AnthropicClient.make`/`layer`/`layerConfig` — customize the header name used to send the API key (default `"x-api-key"`)
- Add `apiKeyScheme` option — prepend an auth scheme prefix (e.g. `Bearer`, `OAuth`) to the API key value
- Add `headers` option — merge arbitrary extra HTTP headers into every request

Useful when routing Anthropic requests through a gateway/proxy that expects different auth conventions (e.g. `authorization: Bearer <token>` instead of `x-api-key: <token>`).

## Test plan
- [x] `pnpm lint-fix`
- [x] `tsc -b` (package typecheck)
- [x] `pnpm docgen`
- [x] Manual verification with a mocked `HttpClient` confirming `authorization: Bearer <key>` and custom header appear in outgoing requests

### DIFF
--- a/.changeset/anthropic-client-custom-auth-header.md
+++ b/.changeset/anthropic-client-custom-auth-header.md
@@ -1,0 +1,5 @@
+---
+"@effect/ai-anthropic": minor
+---
+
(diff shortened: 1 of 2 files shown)


fubhy:
Thanks for the PR. Before maintainer review, could you add automated tests covering the custom auth header/scheme and extra headers behavior, and address the redaction story for auth-like values supplied through `headers`? Right now only `apiKeyHeader` is registered as redacted.

PR #6316, opened 2026-07-11
fix: prevent cachedWithTTL from suppressing interruption of the cached effect

### LINKED ISSUE #6304: `Effect.cachedWithTTL` and `Effect.cachedInvalidateWithTTL` suppress triggering of `Effect.timeout`

3.21.4

Run the following code:

```ts
import { Effect, pipe } from 'effect';

const causeTimeout = pipe(
  Effect.async(() => {}),
  Effect.timeoutFail({
    duration: '1 second',
    onTimeout: () => 'Timeout',
  })
);

await pipe(
  Effect.cachedWithTTL(causeTimeout, '1 minute'),
  Effect.andThen(run => run),
  Effect.runPromiseExit,
  promiseExit => Promise.race([
    promiseExit,
    new Promise((resolve) => setTimeout(() => resolve('Timeout waiting for Exit'), 5_000))
  ]).then(console.log)
);
```

A `Exit` object with failure `"Timeout"` is printed to the console

### What do you see instead?

 String `"Timeout waiting for Exit"` is printed to console

### Additional information

It seems like `Effect.cachedWithTTL` prevents the timeout from being triggered. Interestingly enough, `Effect.cached` works as expected.

`Effect.cachedInvalidateWithTTL` also suppresses to trigger the timeout

xianjianlf2:
Closes #6304

`Effect.cachedWithTTL` / `Effect.cachedInvalidateWithTTL` run the cached effect uninterruptibly: `getCachedValue` wraps populate-and-await in `core.uninterruptibleMask`, and `computeCachedValue` runs the user effect via `core.intoDeferred(self, deferred)` under that mask — so interruption-based combinators like `Effect.timeout` can never fire and the fiber hangs. (`Effect.cached` is unaffected because it maps to `effect.memoize`, a different path.)

The mask's `restore` was only applied to `deferredAwait`, not to running `self`. This threads `restore` into `computeCachedValue` and applies it to the cached effect:

```diff
-      core.intoDeferred(self, deferred)
+      core.intoDeferred(restore(self), deferred)
```

- Added `it.live` tests for both combinators asserting a timed-out cached effect yields `Exit.fail("Timeout")`; without the fix they hang ~15s and fail.
- `pnpm test run test/Effect/caching.test.ts` → 5 passed; `pnpm --filter=effect check` clean. Changeset included.

## Validation update (2026-09-22)
Merged the current v3 base, including its TSTyche 7.2.1 toolchain. The cached-effect regression remains unchanged. Validation passed: lint-fix; 5 caching tests; pnpm check; pnpm test-types --target '>=5.4' (7 compiler targets, 462 file runs, 5,600 tests); pnpm build; pnpm docgen. Head be17d310b.

AI assistance: OpenAI Codex.

### DIFF
--- a/.changeset/cached-with-ttl-interrupt.md
+++ b/.changeset/cached-with-ttl-interrupt.md
@@ -1,0 +1,7 @@
+---
+"effect": patch
+---
+
+Fix `Effect.cachedWithTTL` and `Effect.cachedInvalidateWithTTL` suppressing interruption (e.g. `Effect.timeout`) of the cached effect.
+
+The cached effect was executed under `uninterruptibleMask` while populating the cache, which prevented interruption-based combinators such as `Effect.timeout` from firing. The cached effect now runs with the caller's interruptibility restored, matching the behaviour of the non-cached effect.
--- a/packages/effect/src/internal/effect/circular.ts
+++ b/packages/effect/src/internal/effect/circular.ts
@@ -263,12 +263,13 @@ export const cachedInvalidateWithTTL: {
 const computeCachedValue = <A, E, R>(
   self: Effect.Effect<A, E, R>,
   timeToLive: Duration.DurationInput,
-  start: number
+  start: number,
+  restore: <AX, EX, RX>(effect: Effect.Effect<AX, EX, RX>) => Effect.Effect<AX, EX, RX>
 ): Effect.Effect<Option.Option<[number, Deferred.Deferred<A, E>]>, never, R> => {
   const timeToLiveMillis = Duration.toMillis(Duration.decode(timeToLive))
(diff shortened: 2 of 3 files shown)


xianjianlf2:
I rechecked the failing `Types` job. The test suites pass through TypeScript 5.9.3 and 6.0.3; the job then fails in TSTyche setup because `/home/runner/.local/share/TSTyche/typescript@7.0.2/lib/typescript.js` is missing (`ERR_MODULE_NOT_FOUND`). This is a runner/tool-cache provisioning failure rather than a type error from this PR. I attempted to rerun the failed job, but GitHub requires repository admin rights.

PR #6390, opened 2026-07-14
Retry failed Redis script loads

alvarosevilla95:
## Summary

Redis Lua scripts are loaded through `SCRIPT LOAD` and cached before being invoked with `EVALSHA`. The script cache currently stores failed loads indefinitely, so a transient Redis failure during `SCRIPT LOAD` can poison a running process until the Redis service is recreated.

This changes the script cache TTL so successful script loads continue to be cached forever, while failed loads are immediately evicted and retried on the next evaluation.

### DIFF
--- a/.changeset/retry-redis-script-load.md
+++ b/.changeset/retry-redis-script-load.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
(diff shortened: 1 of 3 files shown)


alvarosevilla95:
Original effect-smol PR: https://github.com/Effect-TS/effect-smol/pull/2614

PR #6394, opened 2026-07-14
feat: add graph set operators

lloydrichards:
## Type

- [x] Refactor
- [x] Feature

## Description

Have added 4 basic and 3 advanced set operators to the Graph module to be able to build and explore for interesting structures. These include:

- `Graph.compose` - $(G1 ∪ G2 = { V1 ∪ V2, E1 ∪ E2 })$
  merges two graphs by node identity, combining nodes and edges from both. Overlapping data from that wins.
- `Graph.intersection` - $(G1 ∩ G2 = { V1 ∩ V2, E1 ∩ E2 })$
  keeps only nodes and edges present in both graphs.
- `Graph.difference` - $(G1 \ G2 = { V1, E1 \ E2 })$
  keeps all nodes from self, but removes edges that also exist in that.
- `Graph.symmetricDifference` - $(G1 Δ G2 = { V1 ∪ V2, (E1 ∪ E2) \ (E1 ∩ E2) })$
  keeps edges that exist in exactly one graph, removing shared edges.
- `Graph.complement` - $(G' = { V, (V × V) \ E })$
  keeps the same nodes, but adds every missing non-self edge.
- `Graph.neighborhood` - $(Nᵣ(v) = G[{ u ∈ V | distance(v, u) ≤ r }])$
  builds an induced subgraph around a node within a radius.
- `Graph.sum` - $(G1 + G2 = { V1 ⊔ V2, E1 ⊔ E2 })$
  creates a disjoint union of two graphs without merging equal nodes.

Most of the implementation was done similar to `NetworkX` when it came to making decisions around when `self` or `that` would win on conflict and the main decision on node equality is left to the implementor where they need to pass a nodeId callback to be able to identify each node.

One additional refactor (separate commit) was the additional logic to `Graph.bfs` and `Graph.dfs` to allow for a `maxDepth`.  This was important to be able to reuse the BFS in the `neighbordhood` op and create a custom version like I did in my utils experiment.

Some fun and exciting things can be done with this functionality.

Screenshot 2026-06-23 at 22 26 49

## For Reviewing

To make things easier to review I've broken up the three main parts so they can be reviewed separately and left off the type tests till last. Any feedback or requests I will initial add as commits on top of the current tree and then rebase together to clean up once approved

## Related

Came up as a request in Slack with a bit of back and forth discussion with @fubhy.  I then spend a couple of days building [similar functions in my own repo](https://github.com/lloydrichards/edu_effect-okf/blob/main/apps/cli/src/lib/graph-utils.ts) to be sure of the api as well as test out their implementation with some real examples.

- Slack [thread](https://discord.com/channels/795981131316985866/1518570933864894485)
- effct-smol [PR #2482](https://github.com/Effect-TS/effect-smol/pull/2482)

### DIFF
--- a/packages/effect/src/Graph.ts
+++ b/packages/effect/src/Graph.ts
@@ -579,6 +579,444 @@ export const mutate: {
 })
 
 // =============================================================================
+// Set Operations
+// =============================================================================
+
+/** @internal */
+type NodeMaps<N> = {
+  readonly byId: Map<string, { readonly index: NodeIndex; readonly data: N }>
+  readonly byIndex: Map<NodeIndex, string>
+}
+
+/** @internal */
+type EdgeIdentity<E> = { readonly sourceId: string; readonly targetId: string; readonly data: E }
+
+/** @internal */
+const buildNodeMaps = <N, E, T extends Kind>(graph: Graph<N, E, T>, nodeId: (node: N) => string): NodeMaps<N> => {
+  const byId = new Map<string, { readonly index: NodeIndex; readonly data: N }>()
+  const byIndex = new Map<NodeIndex, string>()
+
+  for (const [index, data] of graph.nodes) {
+    const id = nodeId(data)
+    byId.set(id, { index, data })
+    byIndex.set(index, id)
+  }
+
+  return { byId, byIndex }
+}
+
+/** @internal */
+const edgeKey = (type: Kind, sourceId: string, targetId: string): string =>
+  type === "undirected" && sourceId > targetId ? `${targetId}\0${sourceId}` : `${sourceId}\0${targetId}`
+
+/** @internal */
+const emptyLike = <N, E, T extends Kind>(
+  graph: Graph<N, E, T>,
+  mutate: (mutable: MutableGraph<N, E, T>) => void
+): Graph<N, E, T> => {
+  const mutable = beginMutation(graph)
+
+  mutable.nodes.clear()
+  mutable.edges.clear()
+  mutable.adjacency.clear()
+  mutable.reverseAdjacency.clear()
+  mutable.nextNodeIndex = 0
+  mutable.nextEdgeIndex = 0
+  mutable.acyclic = Option.some(true)
+
+  mutate(mutable)
+
+  return endMutation(mutable)
+}
+
+/**
+ * Returns the union of two graphs, merging nodes by identity.
+ *
+ * **Details**
+ *
+ * Nodes and edges present in both graphs use data from `that`. The result has
+ * the same graph kind as `self`.
+ *
+ * `G1 ∪ G2 = {V1 ∪ V2, E1 ∪ E2}`
+ *
+ * **Example** (Composing graphs)
+ *
+ * ```ts
+ * import { Graph } from "effect"
+ *
+ * const left = Graph.directed<{ id: string }, string>((mutable) => {
+ *   const a = Graph.addNode(mutable, { id: "A" })
+ *   const b = Graph.addNode(mutable, { id: "B" })
+ *   Graph.addEdge(mutable, a, b, "A-B")
+ * })
+ *
+ * const right = Graph.directed<{ id: string }, string>((mutable) => {
+ *   const b = Graph.addNode(mutable, { id: "B" })
+ *   const c = Graph.addNode(mutable, { id: "C" })
+ *   Graph.addEdge(mutable, b, c, "B-C")
+ * })
+ *
+ * const result = Graph.compose(left, right, (node) => node.id)
+ *
+ * console.log(Graph.nodeCount(result)) // 3
+ * console.log(Graph.edgeCount(result)) // 2
+ * ```
+ *
+ * @category set operations
+ * @since 4.0.0
+ */
+export const compose: {
+  <N, E, T extends Kind = "directed">(
+    that: Graph<N, E, T>,
+    nodeId: (node: N) => string
+  ): (self: Graph<N, E, T>) => Graph<N, E, T>
+  <N, E, T extends Kind = "directed">(
+    self: Graph<N, E, T>,
+    that: Graph<N, E, T>,
+    nodeId: (node: N) => string
+  ): Graph<N, E, T>
+} = dual(
+  3,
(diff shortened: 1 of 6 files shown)


fubhy:
I closed the split PRs again (I just wanted them to review the seperate chunks independently).

packages/effect/src/Graph.ts
```diff
@@ -578,6 +578,650 @@ export const mutate: {
...
+  mutable.adjacency.clear()
+  mutable.reverseAdjacency.clear()
+  mutable.nextNodeIndex = 0
+  mutable.nextEdgeIndex = 0
+  mutable.acyclic = Option.some(true)
+
+  mutate(mutable)
+
+  return endMutation(mutable)
+}
```
fubhy: This is odd and adds unnecessary overhead and should not be needed.

fubhy: I'd probably do something like this:

```ts
const make =
  <T extends Kind>(type: T) =>
  <N, E>(
    mutate?: (mutable: MutableGraph<N, E, T>) => void
  ): Graph<N, E, T> => {
    const graph: Mutable<Graph<N, E, T>> = Object.create(ProtoGraph)

    graph.type = type
    graph.nodes = new Map()
    graph.edges = new Map()
    graph.adjacency = new Map()
    graph.reverseAdjacency = new Map()
    graph.nextNodeIndex = 0
    graph.nextEdgeIndex = 0
    graph.acyclic = Option.some(true)
    graph.mutable = false

    if (mutate !== undefined) {
      const mutable = beginMutation(graph)
      mutate(mutable)
      return endMutation(mutable)
    }

    return graph
  }

export const directed: <N, E>(
  mutate?: (mutable: MutableDirectedGraph<N, E>) => void
) => DirectedGraph<N, E> = make("directed")

export const undirected: <N, E>(
  mutate?: (mutable: MutableUndirectedGraph<N, E>) => void
) => UndirectedGraph<N, E> = make("undirected")
```

And then instead of using `emptyLike` you'd call `make(self.type)(mutate)`

packages/effect/src/Graph.ts
```diff
@@ -578,6 +578,650 @@ export const mutate: {
...
+    byId.set(id, { index, data })
+    byIndex.set(index, id)
+  }
+
+  return { byId, byIndex }
+}
+
+/** @internal */
+const edgeKey = (type: Kind, sourceId: string, targetId: string): string =>
+  type === "undirected" && sourceId > targetId ? `${targetId}\0${sourceId}` : `${sourceId}\0${targetId}`
```
fubhy: Still not sure about the identity handling tbh.

fubhy: I'll have to think about this for a bit.

fubhy: I still think there's something here with Equivalence & Hash

lloydrichards: Each time I've gone down the `Equivalence` rabbit hole I've found it much more complex to implement. `Equivalence` alone would make membership checks scans, so we also need a compatible hash to retain hash-based lookup. `Map` cannot use it, and Effect’s `HashSet` / `HashMap` expect identity to be implemented by their values. This is especially awkward for edge identity because it is composite: identified source node + target node + potentially edge data.

Maybe I am missing a simpler approach here, but I think a generic node key is a good smaller step:

```ts
type SetOperationIdentity<N, NodeId> = {
  readonly node: (node: N) => NodeId
}

const compose = N, E, NodeId, T extends Kind = "directed">(
    self: Graph<N, E, T>,
    that: Graph<N, E, T>,
    identity: SetOperationIdentity<N, NodeId>
  ): Graph<N, E, T>
  
Graph.compose(left, right, {
  node: (node) => node.id
})
```

This retains expected `O(V + E)` behavior, supports keys beyond strings, and keeps the existing simple-graph meaning where edges are identified by their endpoint node keys. Parallel edges therefore coalesce.

Existing conflict behavior remains unchanged where matching edgs continue to use the current left/right precedence rules. This only generalizes how nodes are identified.

It could later expand to include edge identity:

```ts
interface SetOperationIdentity<N, E, NodeId, EdgeId> {
  readonly node: (node: N) => NodeId
  readonly edge: (edge: E) => EdgeId
}

Graph.compose(left, right, {
  node: (node) => node.id,
  edge: (edge) => edge.relation
})
```

That would support multigraph semantics, where two edged can both connect the same two nodes. However the cost is not only another generic as internally we need to index by (sourceNodeId, targetNodeId, edgeId), typically with nested maps.

```ts
const left = Graph.directed<string, string>((mutable) => {
  const a = Graph.addNode(mutable, "A")
  const b = Graph.addNode(mutable, "B")
  const c = Graph.addNode(mutable, "C")

  Graph.addEdge(mutable, a, b, "teaches") // A -> B
  Graph.addEdge(mutable, a, b, "advises") // A -> B
  Graph.addEdge(mutable, b, c, "references")  // B -> C
})

const right = Graph.directed<string, string>((mutable) => {
  const a = Graph.addNode(mutable, "A")
  const b = Graph.addNode(mutable, "B")
  const c = Graph.addNode(mutable, "C")

  Graph.addEdge(mutable, a, b, "collaborates")  // A -> B
  Graph.addEdge(mutable, b, c, "cites")  // B -> C
})

const result = Graph.compose(left, right, {
  node: (node) => node
  // edge: (edge) => edge
})

console.log(Graph.nodeCount(result)) // 3
console.log(Graph.edgeCount(result)) // 2
```

The larger cost is semantic. Once edge identity is part of the API, we need to define what each operation means for multigraphs. This turns the operations from simple-graph set operations into explicitly multigraph-aware operations. I think generic node identity is a useful minimal API because it improves flexibility without changing the existing edge semantics. Explicit edge identity should be added only if supporting multigraph behavior is an intentional requirement.

packages/effect/src/Graph.ts
lloydrichards: implemented this with 8c85928 and is much cleaner 👍

fubhy:
@lloydrichards how about this?

lloydrichards:
> @lloydrichards how about this?

this looks great! i really like the API now with the optional identity, think this makes it cleaner and more powerful 💪

lloydrichards (commented):
Looks great! have tried out the local branch and linked to a cli project using some of the methods and everything seems to be working 👍

PR #6396, opened 2026-07-14
Add McpClient: Effect-native MCP client

polRk:
## Summary

Ports the Effect-native **`McpClient`** (Model Context Protocol client) from
[effect-smol#2534] into `Effect-TS/effect`, and bases it on the MCP
protocol-version-negotiation / spec-compliance work from [effect-smol#2601] so
the client relies on current, compliant schemas and a compliant server
counterpart.

Because GitHub cannot transfer PRs between repositories, this is the replacement
PR for the effect-smol → effect migration:

- **Client** — replaces [effect-smol#2534] (@polRk)
- **Base** — incorporates [effect-smol#2601] (@IMax153)

### Base — [effect-smol#2601] (@IMax153), 18 commits, brought in via `git cherry-pick`

The 18 commits of #2601 are included **verbatim with original authorship
preserved** because the client depends on the schema and negotiation work they
contain. The effect-smol base at #2601's branch point was byte-identical to
`effect/main` for these files, so the cherry-pick applied with **zero
conflicts**. #2601 provides:

- MCP protocol versions **`2025-11-25`** and `2025-06-18` with per-session
  negotiation and lifecycle enforcement.
- Fixes to the MCP schemas the client relies on — notably `CreateMessageResult`
  (adds required `role`/`content`), sampling `metadata` made optional,
  elicitation content constrained to primitives, base64 binary content, `_meta`.
- A Streamable-HTTP-compliant `McpServer.layerHttp` (session identity,
  `MCP-Protocol-Version` enforcement, 400/404/202, notification gating) — the
  compliant server this client is tested against.

### Client — [effect-smol#2534] (@polRk) + spec-compliance fixes on top

`McpClient`: automatic `initialize`/`initialized` handshake, stdio and Streamable
HTTP transports, typed methods for tools/resources/prompts/completion, and
host-side handlers for server-initiated sampling/roots/elicitation.

Adapted to #2601's stricter schemas, then fixed against the MCP `2025-06-18`
spec and the official `@modelcontextprotocol/sdk` (via a conformance audit):

- Send the required `Accept: application/json, text/event-stream` header on every
  Streamable HTTP request (compliant servers otherwise reply `406`).
- Derive supported/latest protocol versions from the shared
  `McpSchema.ProtocolVersion`; offer the latest and validate the negotiated one.
- Send `MCP-Protocol-Version` from the negotiated `initialize` **result body**
  (a compliant server is not required to echo it as a response header).
- Fork the `roots/list_changed` forwarder only after the handshake completes
  (MCP lifecycle ordering).
- Spawn the stdio server with `extendEnv: true` (a custom `env` no longer strips
  `PATH`/`HOME`) and drain the child's `stderr` (avoids deadlock on a full pipe).

## Known limitations / follow-ups (tracked separately)

Deferred to keep this PR reviewable; each needs its own tests:

- **SSE responses** — the HTTP transport decodes JSON responses only, not
  `text/event-stream`. Needed for servers (including the official SDK) that
  stream POST responses.
- **HTTP hang-safety** — in-flight requests aren't failed on child-process exit;
  `send()` lacks the terminal/empty-response guards from
  `RpcClient.makeProtocolHttp`.
- **Cancellation** — interrupting a request doesn't emit
  `notifications/cancelled`, and inbound `notifications/cancelled` is a no-op.
- **HTTP session lifecycle** — no `DELETE` termination, no 404 session-expiry
  handling.
- **Systemic (shared `effect-rpc` JSON-RPC serialization, affects `McpServer`
  too)** — JSON-RPC notifications are emitted carrying an `id`, which is not
  standard-JSON-RPC-compliant for third-party peers. Out of scope here.

## Test plan

- `pnpm check` — clean
- `pnpm test packages/effect/test/unstable/ai/` — **431 passed** (15 files),
  including #2601's schema/server/connection conformance suites and the
  McpClient stdio + Streamable HTTP suites
- `pnpm lint` and `pnpm docgen` — clean

[effect-smol#2534]: https://github.com/Effect-TS/effect-smol/pull/2534
[effect-smol#2601]: https://github.com/Effect-TS/effect-smol/pull/2601

### DIFF
--- a/.changeset/add-mcp-client-module.md
+++ b/.changeset/add-mcp-client-module.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Adds `McpClient`, an Effect-native Model Context Protocol client with automatic initialize/handshake, stdio and Streamable HTTP transports, typed methods for tools/resources/prompts/completion, and support for registering host-side handlers for server-initiated sampling, roots, and elicitation requests.
--- a/.changeset/fix-mcp-prompt-invalid-params.md
+++ b/.changeset/fix-mcp-prompt-invalid-params.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Preserve `InvalidParams` errors when MCP prompt arguments fail schema decoding.
--- a/.changeset/mcp-server-version-negotiation.md
+++ b/.changeset/mcp-server-version-negotiation.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Support MCP protocol versions `2025-11-25` and `2025-06-18`, including configurable server version negotiation, initialization lifecycle enforcement, current protocol schemas, and Streamable HTTP session and protocol-version validation.
--- a/.changeset/tender-spiders-cancel.md
+++ b/.changeset/tender-spiders-cancel.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Preserve numeric MCP request IDs when cancelling in-flight server requests.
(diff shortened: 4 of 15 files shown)


polRk:
@tim-smart I will fix conflicts and actualize over the next 7 days.

PR #6406, opened 2026-07-15
feat(http): add tracer header filter combinators to HttpClient

### LINKED ISSUE #6363: HttpClient: allow filtering HTTP header span attributes (allowlist / predicate)
### Problem

The HttpClient tracer in `effect/unstable/http` emits **every** request and response header as an OpenTelemetry span attribute (`http.request.header.<name>`, `http.response.header.<name>`). For chatty APIs (Notion, Cloudflare-fronted services, CloudFront, …) this means 15 – 30+ low-signal attributes per span (`cf-ray`, `alt-svc`, `cf-cache-status`, `server`, `x-amz-cf-id`, `via`, `nel`, `report-to`, `expect-ct`, …), which bloats trace storage and drowns the few attributes that matter.

The existing knob `Headers.CurrentRedactedNames` only redacts *values* of sensitive headers — the attributes are still emitted.

Source: [`packages/effect/src/unstable/http/HttpClient.ts`](https://github.com/Effect-TS/effect-smol/blob/main/packages/effect/src/unstable/http/HttpClient.ts), the two tight loops around the `http.request.header.` / `http.response.header.` `span.attribute(...)` calls.

### Reproduction

https://github.com/schickling-repros/2026-04-effect4-httpclient-header-allowlist

`bun install && bun repro.ts` prints 15 response-header span attributes from a single request, with no API to restrict them.

### Prior art in other OTEL ecosystems

- Java: `OTEL_INSTRUMENTATION_HTTP_CLIENT_CAPTURE_REQUEST_HEADERS` / `..._RESPONSE_HEADERS`
- Go: [open-telemetry/opentelemetry-go-contrib#3179](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3179)

### Requested feature

A user-configurable knob to restrict which headers are emitted as span attributes, independent of value redaction. Shape suggestion that fits the existing pattern in `HttpClient.ts`:

\`\`\`ts
export const TracerHeaderFilter = Context.Reference<
  Predicate.Predicate<string>
>(\"effect/http/HttpClient/TracerHeaderFilter\", {
  defaultValue: () => constTrue, // preserve current behavior
})
\`\`\`

and in the tracer:

\`\`\`ts
const keep = fiber.getRef(TracerHeaderFilter)
for (const name in redactedHeaders) {
  if (!keep(name)) continue
  span.attribute(\`http.response.header.\${name}\`, String(redactedHeaders[name]))
}
\`\`\`

A predicate is the most flexible primitive; an allowlist is trivially expressed as `(name) => allowed.has(name)`. Same treatment for request headers (either a shared predicate or two separate references).

Happy to open a PR if the direction looks right.

<sub>Filed by an AI assistant on behalf of @schickling</sub>

Zelys-DFKH:
Opening this at @IMax153's request to replace Effect-TS/effect-smol#2132. Now that V4 development has merged back into this repo's `main` branch, the linked issue lives here as #6363, so I'm bringing the same change over.

## What this adds

Three new `HttpClient` combinators for controlling which headers get captured as OTEL span attributes:

- `withTracerRequestHeadersFilter`: filters which request headers are captured
- `withTracerResponseHeadersFilter`: filters which response headers are captured
- `withTracerHeadersFilter`: applies the same predicate to both

Each takes a `Predicate.Predicate<string>` that receives the header name (already lower-cased) and returns `true` to capture it. The default is `constTrue`, so every header is captured just like today. This is fully backward compatible.

## Why

`HttpClient` currently captures every header value as a span attribute unconditionally. For chatty APIs (Notion, Cloudflare-fronted services, CloudFront, etc.) that means 15-30+ low-signal attributes per span (`cf-ray`, `alt-svc`, `via`, and the like), which bloats trace storage and buries the few attributes that actually matter for debugging. `Headers.CurrentRedactedNames` only redacts values; the attributes are still emitted regardless. This gives callers a way to opt individual headers out entirely, matching the header-capture allowlist knobs already common in other OTEL client instrumentations (Java's `OTEL_INSTRUMENTATION_HTTP_CLIENT_CAPTURE_REQUEST_HEADERS`, for example).

## Testing

Adds four tests to `HttpClient.test.ts` (a default-capture-all baseline plus one per combinator) and a `tstyche` typetest block that verifies data-first/data-last usage across all three combinators.

## Credit

Thanks to @schickling for the well-researched issue #6363, the reproduction repo made this easy to verify, and to @IMax153 for the review on the original Effect-TS/effect-smol#2132.

Fixes #6363

### DIFF
--- a/.changeset/httpclient-tracer-header-filter.md
+++ b/.changeset/httpclient-tracer-header-filter.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add `withTracerRequestHeadersFilter`, `withTracerResponseHeadersFilter`, and `withTracerHeadersFilter` combinators to `HttpClient` for controlling which headers are captured as OTEL span attributes.
--- a/packages/effect/src/unstable/http/HttpClient.ts
+++ b/packages/effect/src/unstable/http/HttpClient.ts
@@ -687,8 +687,11 @@ export const make = (
             }
             const redactedHeaderNames = fiber.getRef(Headers.CurrentRedactedNames)
             const redactedHeaders = Headers.redact(request.headers, redactedHeaderNames)
+            const requestHeaderFilter = fiber.getRef(TracerRequestHeadersFilter)
             for (const name in redactedHeaders) {
-              span.attribute(`http.request.header.${name}`, String(redactedHeaders[name]))
+              if (requestHeaderFilter(name)) {
+                span.attribute(`http.request.header.${name}`, String(redactedHeaders[name]))
+              }
             }
             request = fiber.getRef(TracerPropagationEnabled)
               ? HttpClientRequest.setHeaders(request, TraceContext.toHeaders(span))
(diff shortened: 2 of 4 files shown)


tim-smart:
Thanks for the PR. This was fixed in another PR

PR #6407, opened 2026-07-15
Fix Cache interrupting always when first consumer interrupts

mattiamanzati:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Previously if there were 2 awaiters, and the first one interrupts, the cache entry would be interrupted as well, now the cache interrupts upon no awaiters remaining (basically a refcount).
@tim-smart I tried to mimic the existing request resolver logic, maybe worth a second check
Also note that the publicly exposed Entry interface has to change, because awaiting a cache entry now is an effectful operation that keeps the lookup running.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache reliability when multiple requests share an in-progress lookup.
  * Interrupting one waiting request no longer cancels the lookup for other requests.
  * Shared lookups are cancelled when no requests remain, and later requests recompute the value as expected.

### DIFF
--- a/.changeset/fix-cache-shared-lookup-interruption.md
+++ b/.changeset/fix-cache-shared-lookup-interruption.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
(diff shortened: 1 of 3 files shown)


tim-smart:
Closing in favour of https://github.com/Effect-TS/effect/pull/6494/changes

PR #6410, opened 2026-07-15
Fix FiberImpl reentrancy and one-shot iterable bugs in Fiber interrup…

P0lip:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This addresses a chain of related bugs in `FiberImpl` interruption, all reachable when a fiber is interrupted while already mid-`runLoop` on the same call stack (e.g. `Fiber.runIn` interrupting a fiber synchronously against an already-closed `Scope`).

- **Reentrant `runLoop`** - `evaluate` could recurse into its own `runLoop`, corrupting the continuation stack and dropping a pending finalizer, leaking an acquired resource (stream reader, db handle, etc.). It now defers reentrant calls onto the scheduler instead of recursing synchronously.
- **Dropped interrupt** - that deferral could itself drop the interrupt: if the loop completed to a success exit before the scheduled task ran (still interruptible, never crossing an interruptibility boundary), the success was committed and the deferred interrupt discarded — so a fiber the caller asked to interrupt returned a value. `evaluate` now honors a pending interrupt before committing a success exit.
- **One-shot iterable** - `fiberInterruptAll`/`fiberInterruptAllAs` iterated their `fibers` argument twice, so a one-shot `Iterable` (e.g. a generator) was exhausted by the first pass and the await resolved without waiting for finalizers. Both now materialize once via `Arr.fromIterable` and reuse it.

Includes regression tests for all three.

### DIFF
--- a/packages/effect/test/Fiber.test.ts
+++ b/packages/effect/test/Fiber.test.ts
@@ -71,4 +71,173 @@ describe("Fiber", () => {
         assert.isTrue(Exit.hasInterrupts(exit))
       })
   )
+
+  describe("synchronous self-interrupt mid-runLoop", () => {
+    it.live("runs an async interrupt finalizer exactly once, in order", () =>
+      Effect.gen(function*() {
+        const events: Array<string> = []
+        const closedScope = yield* Scope.make()
+        yield* Scope.close(closedScope, Exit.void)
+
+        const child = yield* Effect.gen(function*() {
+          const self = Fiber.getCurrent()!
+          yield* Effect.sync(() => {
+            Fiber.runIn(self, closedScope)
+            events.push("acquired")
+          }).pipe(
+            Effect.onInterrupt(() =>
+              Effect.sync(() => {
+                events.push("finalizer-start")
+              }).pipe(
+                Effect.andThen(Effect.sleep(10)),
+                Effect.andThen(Effect.sync(() => {
+                  events.push("finalizer-end")
+                }))
+              )
+            )
+          )
+          events.push("unreachable")
+        }).pipe(Effect.forkChild({ startImmediately: true }))
+
+        const exit = yield* Fiber.await(child)
+        events.push("awaited")
+        assert.isTrue(Exit.hasInterrupts(exit))
+        assert.deepStrictEqual(events, ["acquired", "finalizer-start", "finalizer-end", "awaited"])
+      }))
+
+    it.effect("aborts an interruptible acquisition without running release", () =>
+      Effect.gen(function*() {
+        let released = false
+        const closedScope = yield* Scope.make()
+        yield* Scope.close(closedScope, Exit.void)
+
+        const child = yield* Effect.gen(function*() {
+          const self = Fiber.getCurrent()!
+          yield* Effect.acquireRelease(
+            Effect.sync(() => {
+              Fiber.runIn(self, closedScope)
+              return "resource"
+            }),
+            () =>
+              Effect.sync(() => {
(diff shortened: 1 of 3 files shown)


packages/effect/test/Fiber.test.ts
```diff
@@ -1,9 +1,108 @@
...
+            () =>
+              Effect.sync(() => {
+                released = true;
+              }),
+            { interruptible: true },
+          );
+        }).pipe(Effect.scoped, Effect.forkChild({ startImmediately: true }));
+
+        yield* Fiber.await(child);
+        assert.isTrue(released);
```
tim-smart: This should be false as the acquisition is interruptible with `{ interruptible: true }`.

packages/effect/test/Fiber.test.ts
```diff
@@ -1,9 +1,108 @@
...
+      }),
+  );
+
+  it.effect(
```
tim-smart: This test already passes on the main branch.

tim-smart (changes requested):
Thanks for the report. I'll fix the double iteration in another PR to prevent the mixing of concerns.

P0lip:
> I'll fix the double iteration in another PR to prevent the mixing of concerns.

yeah I considered that - they were semi-related (both surfaced in the same investigation) so I kinda did it all in one PR to avoid the double-review hassle, but works for me, I'll split going forward. saw you already landed the double iteration fix on main, thanks!

packages/effect/test/Fiber.test.ts
P0lip: You're right - I verified against main: with `interruptible: true` the interrupt aborts the acquisition and `released` stays `false`, which this PR broke by deferring interrupt delivery past the finalizer registration. The test was asserting behavior my own change introduced, so I'll rework this.

While verifying I did confirm the corruption I was originally chasing, with an async interrupt finalizer:

```ts
yield* Effect.sync(() => {
  Fiber.runIn(self, closedScope)
  events.push("acquired")
}).pipe(
  Effect.onInterrupt(() =>
    Effect.sync(() => events.push("fin-start")).pipe(
      Effect.andThen(Effect.sleep(10)),
      Effect.andThen(Effect.sync(() => events.push("finalized")))
    )
  )
)
```

On main this produces `["fin-start", "acquired"]` - the finalizer starts reentrantly inside the interrupting call, before the acquiring sync block has returned, and everything past the first async boundary is dropped while `Fiber.await` resolves. The deferral in this PR does not fix it either (the finalizer is then skipped entirely). I'll take a different approach - delivering the deferred self-interrupt at the next op boundary in the outer run loop rather than scheduling a task - and update the PR.

packages/effect/test/Fiber.test.ts
P0lip: Correct - it only fails with the deferral applied, the hazard was introduced by this PR itself. Will rework along with the above.

packages/effect/test/Fiber.test.ts
P0lip: 8ef3ceb7a

packages/effect/test/Fiber.test.ts
P0lip: 8ef3ceb7a

tim-smart:
I tried to push some changes but it seems I don't have permission. Fixed in #6484

P0lip:
Thanks, Tim!

PR #6411, opened 2026-07-15
Fix child process force-kill escalation

sbking:
## Summary

`forceKillAfter` currently times only delivery of the initial signal. Signal delivery completes immediately, so the `SIGKILL` fallback is never reached and termination can wait indefinitely when a child ignores the initial signal.

- include the child process exit wait in the `forceKillAfter` timeout
- escalate to `SIGKILL` when the initial signal does not stop the process
- cover both explicit `ChildProcessHandle.kill` calls and automatic scoped cleanup

## Testing

- `pnpm lint-fix`
- `pnpm check`
- `pnpm build`
- `pnpm test-types`
- `pnpm test packages/platform-node-shared/test/NodeChildProcessSpawner.test.ts -t "should force kill a process"`

### DIFF
--- a/packages/platform-node-shared/src/NodeChildProcessSpawner.ts
+++ b/packages/platform-node-shared/src/NodeChildProcessSpawner.ts
@@ -498,9 +498,8 @@ const make = Effect.gen(function*() {
               Effect.catch(
                 killProcessGroup(command, childProcess, signal),
                 () => killProcess(command, childProcess, signal)
-              )
+              ).pipe(Effect.andThen(Deferred.await(exitSignal)))
             ).pipe(
-              Effect.andThen(Deferred.await(exitSignal)),
               Effect.ignore
             )
           })
@@ -545,9 +544,8 @@ const make = Effect.gen(function*() {
             Effect.catch(
               killProcessGroup(command, childProcess, signal),
               () => killProcess(command, childProcess, signal)
-            )
+            ).pipe(Effect.andThen(Deferred.await(exitSignal)))
           ).pipe(
-            Effect.andThen(Deferred.await(exitSignal)),
             Effect.asVoid
           )
         }
--- a/packages/platform-node-shared/src/NodeChildProcessSpawner.ts
+++ b/packages/platform-node-shared/src/NodeChildProcessSpawner.ts
@@ -495,10 +495,10 @@ const make = Effect.gen(function*() {
             }
             // Process is still running, kill it
             return yield* killWithTimeout((command, childProcess, signal) =>
-              Effect.catch(
-                killProcessGroup(command, childProcess, signal),
-                () => killProcess(command, childProcess, signal)
-              ).pipe(Effect.andThen(Deferred.await(exitSignal)))
+              killProcessGroup(command, childProcess, signal).pipe(
+                Effect.catch(() => killProcess(command, childProcess, signal)),
+                Effect.andThen(Deferred.await(exitSignal))
+              )
             ).pipe(
               Effect.ignore
             )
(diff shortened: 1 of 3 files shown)


packages/platform-node-shared/test/NodeChildProcessSpawner.test.ts
```diff
@@ -456,6 +457,83 @@ describe("NodeChildProcessSpawner", () => {
...
+
+        it.effect("should force kill a process after the initial signal times out", () =>
+          Effect.gen(function*() {
+            const fs = yield* FileSystem.FileSystem
+            const path = yield* Path.Path
+            const directory = yield* fs.makeTempDirectoryScoped()
+            const ready = path.join(directory, "ready")
+            const handle = yield* ChildProcess.make("node", [
+              "-e",
+              "process.on('SIGTERM', () => {}); require('node:fs').writeFileSync(process.argv[1], ''); setInterval(() => {}, 1000)",
```
IMax153, line 470: Why is the `SIGTERM` listener required here?

packages/platform-node-shared/test/NodeChildProcessSpawner.test.ts
```diff
@@ -456,6 +457,83 @@ describe("NodeChildProcessSpawner", () => {
...
+        it.effect("should force kill a process when its scope closes", () =>
+          Effect.gen(function*() {
+            const fs = yield* FileSystem.FileSystem
+            const path = yield* Path.Path
+            const directory = yield* fs.makeTempDirectoryScoped()
+            const ready = path.join(directory, "ready")
+            const completed = yield* Effect.scoped(Effect.gen(function*() {
+              yield* ChildProcess.make("node", [
+                "-e",
+                "process.on('SIGTERM', () => {}); require('node:fs').writeFileSync(process.argv[1], ''); setTimeout(() => process.exit(0), 2000)",
```
IMax153, line 512: Why is the `SIGTERM` listener required here?

packages/platform-node-shared/src/NodeChildProcessSpawner.ts
```diff
@@ -545,9 +544,8 @@ const make = Effect.gen(function*() {
             Effect.catch(
               killProcessGroup(command, childProcess, signal),
               () => killProcess(command, childProcess, signal)
-            )
+            ).pipe(Effect.andThen(Deferred.await(exitSignal)))
```
IMax153: To make it a bit clearer what's going on in this pipeline, maybe we can refactor the pipeline to be more linear now:
```ts
killProcessGroup(command, childProcess, signal).pipe(
  Effect.catch(() => killProcess(command, childProcess, signal)),
  Effect.andThen(Deferred.await(exitSignal))
)
```

IMax153 (changes requested):

packages/platform-node-shared/test/NodeChildProcessSpawner.test.ts
sbking: Node exits on `SIGTERM` by default. Installing the listener suppresses that default exit so the child survives the initial signal and the test genuinely exercises `forceKillAfter` escalation to `SIGKILL`. The ready-file handshake ensures the listener is installed before termination begins; without it, the pre-fix implementation could pass. I added a comment here to make that explicit.

packages/platform-node-shared/test/NodeChildProcessSpawner.test.ts
sbking: Node exits on `SIGTERM` by default. Installing the listener suppresses that default exit so the child survives the initial signal and the test genuinely exercises `forceKillAfter` escalation to `SIGKILL`. The ready-file handshake ensures the listener is installed before termination begins; without it, the pre-fix implementation could pass. I added a comment here to make that explicit.

packages/platform-node-shared/src/NodeChildProcessSpawner.ts
sbking: Applied as suggested, and I made the same change in the symmetric scoped-cleanup path. The signal attempt, fallback, and exit wait now read as one linear pipeline while remaining inside the timed region.

sbking: (I used your verbatim pipeline; added you as co-authored-by to the commit)

IMax153 (approved):

PR #6427, opened 2026-07-16
Fix incompatible OpenAPI schema examples

### LINKED ISSUE #6356: @effect/openapi-generator can emit type-invalid annotations for invalid schema examples
`@effect/openapi-generator@4.0.0-beta.64` can emit generated code that fails typechecking when an OpenAPI spec contains invalid `example` values.

The generator converts OpenAPI `example` into Effect schema `examples`, but `examples` is type-parametric:

```ts
readonly examples?: ReadonlyArray<T>
```

So if the source spec has an example that does not match the schema type, the generated TypeScript can fail to typecheck.

Repro:

```sh
mkdir effect-openapi-repro
cd effect-openapi-repro
pnpm init
pnpm add effect@4.0.0-beta.64 @effect/platform-node@4.0.0-beta.64
pnpm add -D @effect/openapi-generator@4.0.0-beta.64 swagger2openapi typescript
curl -fsSL https://raw.githubusercontent.com/lxc/incus/main/doc/rest-api.yaml -o rest-api.yaml
pnpm exec openapigen --spec rest-api.yaml --name Incus --format httpclient > client.gen.ts
pnpm exec tsc --noEmit --moduleResolution Bundler --module ESNext --target ESNext --skipLibCheck --strict client.gen.ts
```

Example generated output:

```ts
Schema.Array(Schema.String).annotate({
  description: "When to trigger the template (create, copy or start)",
  examples: ["create"],
})
```

TypeScript reports:

```text
Type 'string' is not assignable to type 'readonly string[]'.
```

The source spec has an invalid example here: the schema is `type: array`, but the example is a string.

I think the generator should be robust to this class of invalid-but-real-world specs: if an `example`/`examples` value is obviously incompatible with the schema’s own `type`, omit it and emit a warning instead of producing TypeScript that does not typecheck.

Note: `swagger2openapi` is included in the repro only to work around Effect-TS/effect-smol#2155.

coyaSONG:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

OpenAPI 3.0 singular `example` values are normalized to Effect Schema `examples` annotations, but the generator previously copied them without checking their JSON type. Invalid real-world specifications could therefore generate annotations that fail TypeScript compilation.

This change filters schema examples against an explicitly declared JSON Schema `type`, including type arrays. It preserves compatible examples, leaves schemas without an explicit type unchanged, and emits an `invalid-schema-example-dropped` warning for each incompatible value. When every example is incompatible, the annotation is omitted.

Validation:

- `pnpm lint-fix`
- `pnpm --filter @effect/openapi-generator test --run` (70 tests)
- `pnpm check`
- end-to-end `openapigen` reproduction followed by strict `tsc --noEmit`

This contribution was implemented and validated with OpenAI Codex assistance.

## Related

- Related Issue #6356
- Closes #6356


coyaSONG:
Superseded by #7325, which was merged and closed #6356. Closing this draft to avoid duplicate review work. Thank you.

PR #6433, opened 2026-07-16
Fix published pre-response handler types

### LINKED ISSUE #6431: effect@4.0.0-beta.98 publishes empty preResponseHandler.d.ts used by HttpEffect
### What happened

`effect@4.0.0-beta.98` publishes `dist/unstable/http/HttpEffect.d.ts` importing and re-exporting `appendPreResponseHandlerUnsafe` from `./internal/preResponseHandler.ts`, but the published `dist/unstable/http/internal/preResponseHandler.d.ts` contains only `export {}`.

With strict TypeScript and `skipLibCheck: false`, importing the v4 HttpApi/HTTP server surface fails with:

```text
node_modules/effect/dist/unstable/http/HttpEffect.d.ts(9,10): error TS2305: Module '"./internal/preResponseHandler.ts"' has no exported member 'appendPreResponseHandlerUnsafe'.
```

### Reproduction

1. Install exact `effect@4.0.0-beta.98` and `@effect/platform-node@4.0.0-beta.98`.
2. Set `module`/`moduleResolution` to `NodeNext`, strict mode on, and `skipLibCheck: false`.
3. Import `HttpApiBuilder` from `effect/unstable/httpapi` or `NodeHttpServer` from `@effect/platform-node`.
4. Run `tsc --noEmit`.

### Package/source mismatch

The matching source file, `src/unstable/http/internal/preResponseHandler.ts`, exports both `requestPreResponseHandlers` and `appendPreResponseHandlerUnsafe`. The runtime JS also contains those exports. Only the generated/published declaration is empty.

### Expected

The declaration file should expose the same types-only surface as the source/runtime. No runtime behavior change appears necessary.

### Temporary downstream patch

We are temporarily restoring declarations for `PreResponseHandler`, `requestPreResponseHandlers`, and `appendPreResponseHandlerUnsafe` with `patch-package`. We will remove that patch when the published declaration is corrected.

coyaSONG:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Preserve `appendPreResponseHandlerUnsafe` in release declaration output. `HttpEffect` publicly re-exports this function, but its implementation declaration was tagged `@internal`. Release builds enable `stripInternal`, so the symbol disappeared from `internal/preResponseHandler.d.ts` while `HttpEffect.d.ts` still imported and re-exported it. Consumers using strict library checking then received TS2305.

This change removes the erroneous internal marker, extends the existing lint rule to flag public named re-exports of `@internal` symbols, adds regression coverage, and includes a patch changeset.

Validation:

- `pnpm lint-fix`
- `pnpm --dir packages/tools/oxc test --run` (51 tests)
- `pnpm vitest run packages/effect/test/unstable/http/HttpEffect.test.ts` (17 tests)
- `pnpm check`
- release-style declaration emit with `--stripInternal true` (the internal file now declares `appendPreResponseHandlerUnsafe` instead of emitting `export {}`)

Prepared with OpenAI Codex assistance; repository guidance, scope, diff, and validation results were checked before publication.

## Related

- Related Issue #6431
- Closes #6431

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected type declarations for the HTTP pre-response handler API.
  * Improved handling of pre-response handlers for both successful and failed HTTP responses.
  * Enhanced detection of internal exports unintentionally re-exported from public modules.

* **Developer Experience**
  * Added clearer diagnostics for invalid public re-exports of internal APIs.

* **Tests**
  * Added coverage for detecting internal exports re-exported by public modules.

### DIFF
--- a/packages/effect/src/unstable/http/internal/preResponseHandler.ts
+++ b/packages/effect/src/unstable/http/internal/preResponseHandler.ts
@@ -5,6 +5,7 @@ import type { HttpServerRequest } from "../HttpServerRequest.ts"
 /** @internal */
 export const requestPreResponseHandlers = new WeakMap<object, PreResponseHandler>()
 
+/** @internal */
 export const appendPreResponseHandlerUnsafe = (request: HttpServerRequest, handler: PreResponseHandler): void => {
   const prev = requestPreResponseHandlers.get(request.source)
   const next: PreResponseHandler = prev ?
--- a/packages/effect/src/unstable/http/internal/preResponseHandler.ts
+++ b/packages/effect/src/unstable/http/internal/preResponseHandler.ts
@@ -5,7 +5,6 @@ import type { HttpServerRequest } from "../HttpServerRequest.ts"
 /** @internal */
 export const requestPreResponseHandlers = new WeakMap<object, PreResponseHandler>()
 
-/** @internal */
 export const appendPreResponseHandlerUnsafe = (request: HttpServerRequest, handler: PreResponseHandler): void => {
   const prev = requestPreResponseHandlers.get(request.source)
   const next: PreResponseHandler = prev ?
--- a/.changeset/fix-http-pre-response-handler-types.md
+++ b/.changeset/fix-http-pre-response-handler-types.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix the published declaration for `HttpEffect.appendPreResponseHandlerUnsafe`.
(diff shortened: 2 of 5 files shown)


packages/effect/src/unstable/http/internal/preResponseHandler.ts
```diff
@@ -5,7 +5,6 @@ import type { HttpServerRequest } from "../HttpServerRequest.ts"
 /** @internal */
 export const requestPreResponseHandlers = new WeakMap<object, PreResponseHandler>()
 
-/** @internal */
```
mikearnaldi, line 8: we should not remove internal annotations from exports in the internal folder, we should instead annotate with the type the public re-export

PR #6436, opened 2026-07-16
Reject unknown properties in Match.when object patterns

### LINKED ISSUE #6230: Match.when object patterns accept unknown properties, hiding property renames
### What is the problem?

`Match.when` appears to accept object-pattern properties that do not exist on the matched value type. This can hide bugs after a property rename: the stale pattern still type checks, but the branch silently stops matching at runtime.

### Minimal repro

```ts
import { Match } from "effect";

declare const x: { _tag: "Transfer"; value: "foo" } | null;

Match.value(x).pipe(
  Match.when({ _tag: "Transfer", value: "foo" }, () => {}),
  Match.orElse(() => {}),
);
```

Now rename `value` on the input type:

```ts
import { Match } from "effect";

declare const x: { _tag: "Transfer"; renamedValue: "foo" } | null;

Match.value(x).pipe(
  // This still type checks, but can never match the Transfer value now.
  Match.when({ _tag: "Transfer", value: "foo" }, () => {}),
  Match.orElse(() => {}),
);
```

I would expect the second example to fail type checking because `value` is not a valid pattern key for the `Transfer` variant anymore.

### Why this is surprising

The stale property is treated like an excess property on the object pattern. At runtime, however, the matcher still checks that property, so the pattern becomes impossible to match for the intended variant.

For comparison, if the pattern is checked against a concrete object type with `satisfies`, TypeScript reports the stale property:

```ts
const pattern = { _tag: "Transfer", value: "foo" } satisfies Partial<{
  _tag: "Transfer";
  renamedValue: "foo";
}>;
// error: Object literal may only specify known properties, and 'value' does not exist ...
```

From looking at the published types, this may come from `Match.when` accepting a generic `P extends Types.PatternPrimitive<R> | Types.PatternBase<R>`, where `PatternBase<A>` is a recursive `Partial` over `keyof A`. Because the pattern is inferred as a generic parameter, TypeScript's usual fresh-object excess-property check does not seem to protect this call site.

### Expected behavior

Object patterns passed to `Match.when` should reject keys that are not present on the matched input/variant type, so property renames are caught by TypeScript.

### Actual behavior

The stale property still type checks, and the branch fails only at runtime by not matching.

### Environment

Observed with `effect@3.20.0` and TypeScript/native preview in a strict project.

coyaSONG:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Match.when` inferred object patterns through a generic parameter, which bypassed TypeScript's fresh-object excess-property check. A stale property could therefore keep compiling after the matched value type renamed that property, even though the branch could no longer match at runtime.

This adds an object-pattern key guard that accepts keys present on any object variant of the matched input while leaving predicates, refinements, and array patterns unchanged. The regression test covers the reported nullable tagged-union case.

## Related

- Closes #6230

## Validation

- `pnpm test-types Match.tst.ts` — 32 tests, 97 assertions, and 11 expected diagnostics passed
- `pnpm test packages/effect/test/Match.test.ts` — 62 tests passed
- `pnpm lint-fix`
- `pnpm lint`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`; the affected `packages/effect` docgen was rerun on the final source
- `git diff --check`

## AI assistance

OpenAI Codex assisted with issue screening, implementation, and validation. No human review is claimed.

### DIFF
--- a/.changeset/tidy-cats-match.md
+++ b/.changeset/tidy-cats-match.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Reject unknown properties in `Match.when` object patterns.
--- a/packages/effect/dtslint/Match.tst.ts
+++ b/packages/effect/dtslint/Match.tst.ts
@@ -84,6 +84,15 @@ describe("Match", () => {
   })
 
   describe("when", () => {
+    it("rejects unknown object pattern properties", () => {
+      pipe(
+        Match.value(hole<{ _tag: "Transfer"; renamedValue: "foo" } | null>()),
+        // @ts-expect-error: Type 'string' is not assignable to type 'never'
+        Match.when({ _tag: "Transfer", value: "foo" }, () => {}),
+        Match.orElse(() => {})
+      )
+    })
+
     it("schema exhaustive-literal", () => {
       expect(
         pipe(
--- a/packages/effect/src/Match.ts
+++ b/packages/effect/src/Match.ts
@@ -132,6 +132,16 @@ export interface ValueMatcher<in Input, out Filters, out Remaining, out Result, 
  */
 export type Case = When | Not
 
+type PatternKeys<A> = A extends object ? keyof A : never
+
+type PatternGuard<A, P> = P extends
+  | Predicate.Predicate<any>
+  | Predicate.Refinement<any, any>
+  | SafeRefinement<any>
+  | ReadonlyArray<any> ? unknown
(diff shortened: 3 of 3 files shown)


coyaSONG:
I am limiting my concurrent review load in this repository to two PRs. This draft is 34 upstream commits behind v3 and needs a refresh and revalidation before a credible review handoff, so I am closing it for now while preserving its branch and linked issue. It can be reopened after an active slot clears and the branch is refreshed.

PR #6448, opened 2026-07-17
Add HttpClientRequest header combinators

### LINKED ISSUE #6271: HttpClientRequest: add removeHeader / updateHeaders for removing & transforming headers
---

`HttpClientRequest` lets you *add/override* headers (`setHeader`, `setHeaders`) but provides **no way to remove a header or transform the header set** after a request has been built. The `Headers` module already has `Headers.remove` (and `set`/`setAll`/`merge`), but `HttpClientRequest` doesn't expose an equivalent, so there's no public, prototype-safe way to drop a single header from an existing request.

This isn't hypothetical — it bit us in production. `HttpClientRequest.bodyUnsafeJson` (→ `setBody`) writes an explicit `content-length` header derived from the body's byte length:

```js
// @effect/platform/internal/httpClientRequest.js — setBody
const contentLength = body.contentLength
if (contentLength) {
  headers = Headers.set(headers, "content-length", contentLength.toString())
}
```

When that request is executed through `FetchHttpClient`, the platform `fetch`/undici **also** derives and appends its own `content-length` from the body. On **Node 20**, undici then sees a duplicated `content-length: <n>, <n>` and rejects the request with:

```
RequestError (reason: Transport)
  └─ TypeError: fetch failed
       └─ InvalidArgumentError  UND_ERR_INVALID_ARG  'invalid content-length header'
```

(Node 22/24's undici is more lenient, so it only reproduces on Node 20 — which is exactly where many CI runners and Lambda runtimes sit.)

The natural fix is to **remove the redundant explicit `content-length` header** before execution and let fetch/undici compute it exactly once. But there's no API to do that. `HttpClientRequest.modify` only *applies* set-style operations (it routes `headers` through `setHeaders`), and there's no `removeHeader`/`updateHeaders`.

Add header-removal / header-transformation combinators to `HttpClientRequest`, mirroring the existing `setHeader`/`setHeaders` and delegating to the `Headers` module that already has the needed primitives:

```ts
// Remove one (or several) headers by key — mirrors Headers.remove
export const removeHeader: {
  (key: string): (self: HttpClientRequest) => HttpClientRequest
  (self: HttpClientRequest, key: string): HttpClientRequest
}

// General-purpose: transform the whole header set — the most flexible primitive
export const updateHeaders: {
  (f: (headers: Headers.Headers) => Headers.Headers): (self: HttpClientRequest) => HttpClientRequest
  (self: HttpClientRequest, f: (headers: Headers.Headers) => Headers.Headers): HttpClientRequest
}
```

`updateHeaders` is the more fundamental of the two — `removeHeader`, `setHeader`, etc. can all be expressed in terms of it — and it composes cleanly with the existing `Headers` combinators:

```ts
import { Headers, HttpClientRequest } from "@effect/platform"

request.pipe(HttpClientRequest.removeHeader("content-length"))
// or, equivalently:
request.pipe(HttpClientRequest.updateHeaders(Headers.remove("content-length")))
```

This also fixes the motivating case ergonomically, e.g. via `HttpClient.mapRequest`:

```ts
const client = HttpClient.mapRequest(
  baseClient,
  HttpClientRequest.removeHeader("content-length")
)
```

(Optionally, `HttpClientRequest.modify`'s options could also gain a `removeHeaders?: ReadonlyArray<string>` field for symmetry, but `removeHeader`/`updateHeaders` are the core asks.)

1. **Manual clone (current workaround).** Because `makeInternal` builds requests via `Object.create(Proto)` and there's no public constructor that takes pre-built headers without re-running `setBody` (which would re-add `content-length`), we reconstruct the request by hand while preserving the prototype:

   ```ts
   import { Headers, HttpClientRequest } from "@effect/platform"

   const removeHeader = (
     request: HttpClientRequest.HttpClientRequest,
     name: string
   ): HttpClientRequest.HttpClientRequest => {
     const headers = Headers.remove(request.headers, name)
     // Preserve the HttpClientRequest prototype (TypeId, pipe) while swapping headers.
     const next = Object.create(Object.getPrototypeOf(request))
     return Object.assign(next, request, { headers })
   }
   ```

   This works but reaches past the public API and is fragile to internal representation changes — exactly the kind of thing a first-class combinator should make unnecessary.

2. **Rebuild via `HttpClientRequest.make`/`modify`.** Not viable: passing `body` back through `modify`/`make` re-runs `setBody`, which re-adds the `content-length` header we're trying to remove. There's no public path to set headers *after* the body without the body re-injecting them.

3. **Fix `FetchHttpClient` to not forward a `content-length` it already set.** This is a reasonable orthogonal improvement (and arguably a separate bug report), but a general `removeHeader`/`updateHeaders` is broadly useful well beyond this one case — conditional auth/trace headers, redaction, normalizing proxy headers, stripping hop-by-hop headers, etc. — and brings `HttpClientRequest` to parity with the `Headers` module and with server-side response builders.

### Environment

- `@effect/platform` 0.96.1, `effect` 3.21.2
- Node.js 20 (where undici's stricter `content-length` validation surfaces the duplicate-header rejection)

coyaSONG:
## Summary

- add dual `HttpClientRequest.removeHeader` and `HttpClientRequest.updateHeaders` combinators
- preserve the original request while allowing the complete header set to be transformed
- add regression coverage and a patch changeset for `@effect/platform`

## Testing

- `pnpm lint-fix`
- `pnpm test packages/platform/test/HttpClient.test.ts` (31 tests)
- `pnpm check`

Closes #6271

### DIFF
--- a/.changeset/calm-headers-transform.md
+++ b/.changeset/calm-headers-transform.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+Add `HttpClientRequest.removeHeader` and `HttpClientRequest.updateHeaders`.
--- a/packages/platform/src/HttpClientRequest.ts
+++ b/packages/platform/src/HttpClientRequest.ts
@@ -168,6 +168,24 @@ export const setHeaders: {
  * @since 1.0.0
  * @category combinators
  */
+export const updateHeaders: {
+  (f: (headers: Headers.Headers) => Headers.Headers): (self: HttpClientRequest) => HttpClientRequest
+  (self: HttpClientRequest, f: (headers: Headers.Headers) => Headers.Headers): HttpClientRequest
+} = internal.updateHeaders
+
+/**
+ * @since 1.0.0
+ * @category combinators
+ */
+export const removeHeader: {
+  (key: string): (self: HttpClientRequest) => HttpClientRequest
+  (self: HttpClientRequest, key: string): HttpClientRequest
+} = internal.removeHeader
+
+/**
+ * @since 1.0.0
+ * @category combinators
+ */
 export const basicAuth: {
   (username: string | Redacted, password: string | Redacted): (self: HttpClientRequest) => HttpClientRequest
   (self: HttpClientRequest, username: string | Redacted, password: string | Redacted): HttpClientRequest
--- a/packages/platform/src/internal/httpClientRequest.ts
+++ b/packages/platform/src/internal/httpClientRequest.ts
@@ -167,6 +167,31 @@ export const setHeaders = dual<
     self.body
   ))
 
+/** @internal */
+export const updateHeaders = dual<
+  (
+    f: (headers: Headers.Headers) => Headers.Headers
+  ) => (self: ClientRequest.HttpClientRequest) => ClientRequest.HttpClientRequest,
+  (
+    self: ClientRequest.HttpClientRequest,
+    f: (headers: Headers.Headers) => Headers.Headers
+  ) => ClientRequest.HttpClientRequest
+>(2, (self, f) =>
+  makeInternal(
+    self.method,
+    self.url,
+    self.urlParams,
+    self.hash,
+    f(self.headers),
+    self.body
+  ))
+
+/** @internal */
+export const removeHeader = dual<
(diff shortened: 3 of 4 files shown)


tim-smart:
v3 is under a feature freeze. Feel free to reopen against v4 :)

PR #6450, opened 2026-07-17
Support Effects in vitest setup hooks

### LINKED ISSUE #6096: beforeEach/beforeAll should support effects

beforeEach/beforeAll exported by @effect/vitest do not support returning Effects (that are executed).  This makes it hard to reset state before each test is run.

beforeEach/beforeAll should extend vitest's native calls to support returning an effectual callback (e.g. () => Effect<...>)

I can work around it by using runSync in the before callback, or calling the reset within the test, both of which add noise.

coyaSONG:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`@effect/vitest` currently re-exports Vitest's `beforeAll` and `beforeEach` hooks unchanged. Vitest does not recognize a returned `Effect` as asynchronous work, so the Effect remains lazy and the hook body is skipped.

This change explicitly wraps both hooks. When a callback returns an Effect, it is executed and awaited through the package's existing Effect runner; synchronous and Promise-returning callbacks are passed through unchanged. Regression tests cover both suite-level and per-test Effect hooks, including the repository's concurrent test configuration.

Validation:

- `pnpm --filter @effect/vitest test --run` (47 passed, 6 skipped)
- `pnpm --filter @effect/vitest check`
- `pnpm lint-fix`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`
- `pnpm changeset status`

## Related

- Closes #6096

### DIFF
--- a/.changeset/tidy-hooks-run.md
+++ b/.changeset/tidy-hooks-run.md
@@ -1,0 +1,5 @@
+---
+"@effect/vitest": patch
+---
+
+Run Effects returned by `beforeAll` and `beforeEach` hooks.
--- a/packages/vitest/src/index.ts
+++ b/packages/vitest/src/index.ts
@@ -181,6 +181,22 @@ export namespace Vitest {
 export const addEqualityTesters: () => void = internal.addEqualityTesters
 
 /**
+ * Registers a callback before all tests in the current suite. If the callback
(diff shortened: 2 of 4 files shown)


tim-smart:

PR #6452, opened 2026-07-17
Deliver scheduled cluster messages at their deadline instead of the next poll

sbking:
## Summary

Messages saved with a future `deliverAt` are only discovered by the periodic storage poll, so a scheduled message can be delivered up to a full poll interval after its deadline.

- add a required `MessageStorage.nextDeliverAt` operation that returns the delay until the earliest future scheduled delivery for a set of shards, with memory and SQL driver implementations
- add `ShardingConfig.timelyScheduledMessageDelivery`, disabled by default
- when enabled, discover the next deadline in the background after each storage read and arm a replaceable wakeup for it
- when disabled, preserve the existing poll-cadence behavior and perform no `nextDeliverAt` queries
- if an enabled `nextDeliverAt` lookup fails, log at debug level and fall back to the existing poll
- use explicit null checks in the memory driver so due-message selection and `nextDeliverAt` agree when `deliverAt` is zero

PostgreSQL and SQLite answer the query with `MIN(deliver_at)` over a partial index on `(shard_id, deliver_at)` filtered to unprocessed scheduled rows. MySQL and SQL Server seek per shard over `(shard_id, processed, deliver_at, last_read)`. Migration `0004_deliver_at_index` follows the existing `0003_pg_messages_rowid_index` migration and creates the index idempotently where supported.

## Testing

- `pnpm lint`
- `pnpm check`
- `pnpm test packages/effect/test/cluster/MessageStorage.test.ts packages/effect/test/cluster/Sharding.test.ts` (73 passed)
- `EFFECT_INTEGRATION_TESTS=1 pnpm test packages/platform/node/test/cluster/SqlMessageStorage.integration.test.ts` (62 passed across PostgreSQL, MySQL, and SQLite)
- `pnpm test` reached 9,995 passed, 1 expected failure, and 55 skipped; the only failure was the unrelated timing-sensitive `packages/platform/browser/test/RpcWorker.test.ts` stream witness under full-suite load, which passed when rerun in isolation (8 passed)

The Sharding witnesses include both directions of the option: the default path performs zero `nextDeliverAt` calls and waits for poll cadence, while the enabled path delivers at the persisted deadline.

Benchmark protocol, reproduction harnesses, and raw results: https://gist.github.com/sbking/0a56b3f94ca518f4e50308f045e75a45

### DIFF
--- a/packages/effect/src/unstable/cluster/MessageStorage.ts
+++ b/packages/effect/src/unstable/cluster/MessageStorage.ts
@@ -14,6 +14,7 @@ import * as Arr from "../../Array.ts"
 import { Clock } from "../../Clock.ts"
 import * as Context from "../../Context.ts"
 import * as Data from "../../Data.ts"
+import * as Duration from "../../Duration.ts"
 import * as Effect from "../../Effect.ts"
 import * as Exit from "../../Exit.ts"
 import { constFalse, identity } from "../../Function.ts"
@@ -151,6 +152,17 @@ export class MessageStorage extends Context.Service<MessageStorage, {
   readonly unprocessedMessagesById: <R extends Rpc.Any>(
     messageIds: Iterable<Snowflake.Snowflake>
   ) => Effect.Effect<Array<Message.Incoming<R>>, PersistenceError>
+
+  /**
+   * Retrieves the delay until the earliest scheduled message for the
+   * specified shards becomes deliverable.
+   *
+   * The delay must measure when this implementation's `unprocessedMessages`
+   * predicate will consider the message due.
+   */
+  readonly nextDeliverAt: (
+    shardIds: Iterable<ShardId.ShardId>
+  ) => Effect.Effect<Option.Option<Duration.Duration>, PersistenceError>
 
   /**
    * Reset the mailbox state for the provided shards.
@@ -399,6 +411,18 @@ export type Encoded = {
     }>,
     PersistenceError
   >
+
+  /**
+   * Retrieves the earliest `deliverAt` of the scheduled messages for the
+   * given shards that is later than `now`.
+   *
+   * The returned time must use the same clock domain as this implementation's
+   * `unprocessedMessages` eligibility predicate.
+   */
+  readonly nextDeliverAt: (
+    shardIds: Arr.NonEmptyArray<string>,
+    now: number
+  ) => Effect.Effect<Option.Option<number>, PersistenceError>
 
   /**
    * Reset the mailbox state for the provided addresses.
@@ -673,6 +697,18 @@ export const makeEncoded: (encoded: Encoded) => Effect.Effect<
         (messages) => decodeMessages(storage, messages)
       )
     },
+    nextDeliverAt(shardIds) {
+      const shards = Array.from(shardIds, (id) => id.toString())
+      if (!Arr.isArrayNonEmpty(shards)) return Effect.succeedNone
+      return Effect.suspend(() => encoded.nextDeliverAt(shards, clock.currentTimeMillisUnsafe())).pipe(
+        Effect.map((next) =>
+          Option.map(
+            next,
+            (deliverAt) => Duration.max(Duration.millis(deliverAt - clock.currentTimeMillisUnsafe()), Duration.zero)
+          )
+        )
+      )
+    },
     resetAddress: (address) => encoded.resetAddresses([address]),
     resetAddresses: (addresses) => addresses.length === 0 ? Effect.void : encoded.resetAddresses(addresses),
     clearAddress: encoded.clearAddress,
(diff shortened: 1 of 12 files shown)


IMax153:
@coderabbitai review

packages/effect/src/unstable/cluster/SqlMessageStorage.ts
```diff
@@ -403,6 +403,47 @@ export const make: (options?: {
...
+      `,
+    orElse: () => (shardIds: ReadonlyArray<string>, now: number) =>
+      sql<{ readonly next_deliver_at: bigint | null }>`
+        SELECT MIN(deliver_at) AS next_deliver_at
+        FROM ${messagesTableSql}
+        WHERE shard_id IN (${sql.literal(shardIds.map(wrapString).join(","))})
+        AND processed = ${sqlFalse}
+        AND deliver_at > ${sql.literal(String(now))}
+      `
+  })
```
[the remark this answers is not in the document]
sbking: Addressed in bb0e8802b131000b3dfd5d9944afc21073eb06d5.

Shard values now reach the query as bound parameters in all three branches: the MySQL `UNION ALL` legs and the fallback `IN` list are built from bound fragments, and the MSSQL `VALUES` rows bind each value while keeping the `VARCHAR(50)` cast so the `shard_id` index seek stays sargable. A regression test drives a quote-bearing shard group (`quote's-group:1`) through the decoded `MessageStorage.nextDeliverAt` boundary — the same unprepared execution path this code runs in production — on all five storage drivers; with the previous interpolation restored, MySQL fails with `ER_PARSE_ERROR` 1064.

While validating the fix we also found and corrected a read-ordering gap in the storage read loop this PR adds: `unprocessedMessages` captured its delivery cutoff before `nextDeliverAt` computed the wake time, so a deadline landing between the two reads was neither delivered nor armed and slipped to the next poll interval. The loop now arms the wake before reading unprocessed messages, making the two predicates overlap instead of leaving a gap. A deterministic test forces a deadline across that boundary; with the previous ordering restored it fails with zero deliveries.

packages/effect/src/unstable/cluster/MessageStorage.ts
```diff
@@ -143,6 +143,14 @@ export class MessageStorage extends Context.Service<MessageStorage, {
     messageIds: Iterable<Snowflake.Snowflake>
   ) => Effect.Effect<Array<Message.Incoming<R>>, PersistenceError>
 
+  /**
+   * Retrieves the earliest `deliverAt` of the scheduled messages for the
+   * specified shards that are not yet deliverable.
+   */
+  readonly nextDeliverAt: (
+    shardIds: Iterable<ShardId.ShardId>
+  ) => Effect.Effect<Option.Option<number>, PersistenceError>
```
tim-smart: It should return a `Option<Duration>` as the database clock and runner clock could be out of sync.

packages/effect/src/unstable/cluster/Sharding.ts
```diff
@@ -573,6 +598,7 @@ const make = Effect.gen(function*() {
           pendingNotifications.forEach((entry) => removableNotifications.add(entry))
         }
 
+        yield* rediscoverStorageWake
```
tim-smart: This should probably be done in the background to keep it out of the read loop.

tim-smart (changes requested):
I'm also not sure about the idea in general. I'll think about it some more.

packages/effect/src/unstable/cluster/MessageStorage.ts
sbking: Good call. I changed this to return `Option<Duration>` so storage owns the clock math and Sharding only receives how long to wait. I kept the existing runner-time semantics to match the polling behavior

packages/effect/src/unstable/cluster/Sharding.ts
sbking: Yep, moved this into the background so the read loop doesn’t wait on the extra query. Discovery and the armed timer use separate handles so refreshing the deadline can’t cancel a timer we already know is valid. I added tests for a blocked lookup and for the deadline passing while the lookup is running.

sbking:
@tim-smart That's fair, let me know what you think! I also started [this discord thread](https://discord.com/channels/795981131316985866/1527611015641038900) the other day if you want to chat about the idea there

sbking:
@tim-smart I updated this PR and gated it behind a `ShardingConfig.timelyScheduledMessageDelivery` configuration option (defaults to `false` so timely delivery is opt-in)

Would really appreciate if you could take a second look! Without this I think I will have to ship my own `Sharding` and `MessageStorage` implementations to my users, which I'd really like to avoid

PR #6469, opened 2026-07-17
Fix Cause.Fail JSON error key

### LINKED ISSUE #4683: Inconsistency between Cause.Fail instance and toJSON

3.14.1

```ts
import { Cause } from "effect";

const cause = Cause.fail("Error");

console.log(cause);

if (Cause.isFailType(cause)) {
  // console.log(cause.failure); <-- does not type check
  console.log(cause.error);
}
```

The `console.log(cause)` prints the following

```
{ _id: 'Cause', _tag: 'Fail', failure: 'Error' }
```

Indicating that there should be a `failure` property on the `cause` object. But in reality there is none as can be seen in the if statement. There is a `error` property though.

I was expecting that what the `console.log` emitted should correspond to the structure of the `cause` object.

### What do you see instead?

The `console.log(cause)` indicates that there is a `failure` property on the `cause` object, when there in fact is no such property.

### Additional information

I believe this are due to the `toJSON` implementation for the `Cause.Fail` in `effect/src/internal/cause.ts`.

coyaSONG:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Cause.Fail` exposes its recoverable error through the `error` property, but its `toJSON` and Node inspection output serialized the same value under `failure`. This aligns the serialized key with the public property and updates the direct and nested regression expectations.

The change includes a patch changeset and was validated with:

- `pnpm test packages/effect/test/Cause.test.ts`
- `pnpm lint-fix`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`

Implementation prepared with OpenAI Codex. No human code review has been performed.

## Related

- Closes #4683

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated `Cause.Fail` JSON serialization to use the `error` field, aligning it with the corresponding error property.
  * Updated serialized output for nested sequential and parallel causes to reflect the same field name.

* **Documentation**
  * Added a patch release note documenting the JSON output alignment.


mikearnaldi:
@coderabbitai investigate the CI failures

mikearnaldi:
@coderabbitai fix the tests

mikearnaldi:
@coderabbitai create a commit fixing the test issues

PR #6470, opened 2026-07-17
Add DateTime.toEpochSeconds

### LINKED ISSUE #6148: @effect DateTime Get unix timestamp of a Date

hello, since we already have a `DateTime.toEpochMillis` that get us the timestamp of a date, could we have something to get the Unix timestamp of a date ? so in seconds ?

I surely have missed something but its painful to do thing like below whenever i need an unix timestamp :(
```
const reservationEventTtlUnix = Math.round(
   DateTime.toEpochMillis(yield* DateTime.now) / 1000,
)
```
i don't know maybe something like DateTime.toEpochSeconds

DateTime.toEpochSeconds(DateTime) maybe ?

_No response_

coyaSONG:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds `DateTime.toEpochSeconds` for obtaining whole seconds since the Unix epoch. The conversion floors the millisecond timestamp so instants before the epoch retain the conventional Unix timestamp semantics (for example, `-1` millisecond becomes `-1` second rather than `0`).

The change includes regression coverage for fractional positive and pre-epoch timestamps, along with the required patch changeset. It was validated with:

- `pnpm vitest packages/effect/test/DateTime.test.ts --run`
- `pnpm lint-fix`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`
- `git diff --check`

Implementation prepared with OpenAI Codex. No human code review has been performed.

## Related

- Closes #6148

### DIFF
--- a/.changeset/add-datetime-epoch-seconds.md
+++ b/.changeset/add-datetime-epoch-seconds.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add `DateTime.toEpochSeconds` for whole Unix timestamps.
--- a/packages/effect/src/DateTime.ts
+++ b/packages/effect/src/DateTime.ts
@@ -939,6 +939,14 @@ export const zonedOffsetIso: (self: Zoned) => string = Internal.zonedOffsetIso
 export const toEpochMillis: (self: DateTime) => number = Internal.toEpochMillis
 
 /**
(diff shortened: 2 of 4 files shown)


gcanti:
@coyaSONG Version 3 is now in feature freeze as we prepare the release of v4

PR #6487, opened 2026-07-18
fix(cli): remove doubled Expected prefix from InvalidValue messages

### LINKED ISSUE #6312: CliError.InvalidValue.message doubles the "Expected: Expected ..." prefix for choice/schema-backed primitives
# Title
`CliError.InvalidValue.message` doubles the "Expected: Expected ..." prefix for `choice`/schema-backed primitives

# Body

### Effect version

`4.0.0-beta.93` (`effect/unstable/cli`)

### Description

Any flag or argument backed by `Primitive.choice` (i.e. `Flag.choice`/`Flag.choiceWithValue`) or by the schema-backed primitives `integer`, `float`, `boolean`, `date` renders a doubled "Expected: Expected ..." prefix when given an invalid value, and also repeats the offending value twice in the same message.

### Reproduction

```ts
import { Primitive, CliError } from "effect/unstable/cli"

const sizeChoice = Primitive.choice([
  ["small", "small"],
  ["medium", "medium"],
  ["large", "large"],
])

  return yield* Effect.mapError(
    sizeChoice.parse("bogus"),
    (expected) =>
      new CliError.InvalidValue({ option: "size", value: "bogus", expected, kind: "flag" })
  )
})

Effect.runPromise(program.pipe(Effect.catch((error) => Effect.sync(() => {
  console.log("error.expected:", JSON.stringify(error.expected))
  console.log("error.message:", JSON.stringify(error.message))
  return error
}))))
```

This is exactly the path a real CLI hits through `Flag.choice("size", [...])` used in a `Command` — `Param.ts`'s `parseFlag`/`parseArgument` map a failed `primitiveType.parse(arg)` straight into `CliError.InvalidValue`'s `expected` field.

### Actual output

```
error.expected: "Expected \"small\" | \"medium\" | \"large\", got \"bogus\""
error.message: "Invalid value for flag --size: \"bogus\". Expected: Expected \"small\" | \"medium\" | \"large\", got \"bogus\""
```

Note both defects in `error.message`:
1. `"Expected: Expected ..."` — doubled prefix.
2. `"bogus"` appears twice — once as the flag's own value, once again via the primitive's `, got "bogus"` suffix.

### Expected output

Something like:

```
Invalid value for flag --size: "bogus". Expected: "small" | "medium" | "large"
```

(a single "Expected" prefix, and the invalid value shown once)

### Root cause

- [`Primitive.choice`](https://github.com/Effect-TS/effect/blob/9de58cc0bb4b45d8a30f77abc75071ca5f2e74ea/packages/effect/src/unstable/cli/Primitive.ts#L298-L308) fails with a raw string that already starts with `"Expected "` and already appends `, got ${value}`:

  ```ts
  return Effect.fail(`Expected ${validChoices}, got ${format(value)}`)
  ```

  The schema-backed primitives (`boolean`, `float`, `integer`, `date`, built via [`makeSchemaPrimitive`](https://github.com/Effect-TS/effect/blob/9de58cc0bb4b45d8a30f77abc75071ca5f2e74ea/packages/effect/src/unstable/cli/Primitive.ts#L108-L114)) fail the same way, surfacing `Schema`'s own decode-error message, which is also of the shape `"Expected <X>, got <Y>"`.

- [`Param.ts`'s `parseFlag`/`parseArgument`](https://github.com/Effect-TS/effect/blob/9de58cc0bb4b45d8a30f77abc75071ca5f2e74ea/packages/effect/src/unstable/cli/Param.ts#L1885-L1897) take that raw failure string verbatim and store it as `CliError.InvalidValue`'s `expected` field:

  ```ts
  const value = yield* Effect.mapError(
    primitiveType.parse(arg),
    (error) =>
      new CliError.InvalidValue({
        option: name,
        value: arg,
        expected: error,
        kind: "flag"
      })
  )
  ```

- [`CliError.InvalidValue`'s `message` getter](https://github.com/Effect-TS/effect/blob/9de58cc0bb4b45d8a30f77abc75071ca5f2e74ea/packages/effect/src/unstable/cli/CliError.ts#L359-L364) unconditionally prepends its own `"Expected: "` label on top of `expected`, regardless of whether `expected` already starts with the word "Expected":

  ```ts
  override get message() {
    if (this.kind === "argument") {
      return `Invalid value for argument <${this.option}>: "${this.value}". Expected: ${this.expected}`
    }
    return `Invalid value for flag --${this.option}: "${this.value}". Expected: ${this.expected}`
  }
  ```

So any primitive whose failure text is a full sentence beginning with "Expected" (rather than a bare description of the expected shape, e.g. `"small" | "medium" | "large"`) gets double-prefixed, and any primitive whose failure text also repeats the offending value (via a `, got <value>` suffix) shows that value twice.

### Suggested fix

Pick one canonical contract for what a `Primitive`'s failure message / `InvalidValue.expected` should contain, then make all primitives conform:

- **Option A** — `expected` should be a bare description of the accepted shape (e.g. `"small" | "medium" | "large"`, no `"Expected "` prefix, no `, got <value>` suffix), and `InvalidValue.message` stays as the single source of the "Expected:"/"got" wording. This would mean adjusting `Primitive.choice` (drop the `Expected ${...}, got ${...}` wrapper, just fail with `validChoices`) and stripping/reshaping the `Schema` decode error before it reaches `expected` in the schema-backed primitives.
- **Option B** — `InvalidValue.message` detects and strips a leading `"Expected "` (and/or a trailing `, got "<value>"`) from `this.expected` before composing the final sentence, so it degrades gracefully regardless of what a given `Primitive` reports.

Happy to open a PR for whichever direction maintainers prefer — we worked around this downstream in our own CLI (built on `effect/unstable/cli`) by post-processing the doubled substring in our error-formatting layer, but that's brittle against upstream wording changes and doesn't fix the duplicated-value wart.

rvaccone:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`CliError.InvalidValue.message` rendered a doubled `Expected: Expected ...` prefix for values backed by `Primitive.choice` or the schema backed primitives (`integer`, `float`, `boolean`, `date`), and for choices it also repeated the offending value twice.

Before:

```
Invalid value for flag --size: "bogus". Expected: Expected "small" | "medium" | "large", got "bogus"
```

After:

```
Invalid value for flag --size: "bogus". Expected: "small" | "medium" | "large"
Invalid value for argument <count>: "3.14". Expected an integer, got 3.14
```

This combines both directions suggested in the issue:

1. `Primitive.choice` now fails with a bare description of the accepted values. The offending value and the `Expected:` label are added by `InvalidValue.message`, so the value is no longer shown twice.
2. `InvalidValue.message` only prepends its `Expected:` label when `expected` does not already start with `Expected ` or `Expected:`. Failure texts that are already complete sentences (schema decode messages, user supplied strings like the `Param.filter` doc examples) are appended as their own sentence instead of being double labeled.

The schema backed primitives keep their full schema decode sentence. Extracting a bare expected description from a schema issue would mean duplicating the `SchemaIssue` formatter traversal inside the CLI package, because the `AnyOf` branch of the default formatter hardcodes the `Expected X, got Y` sentence and is not reachable through the `leafHook`/`checkHook` extension points. The decoded actual after `got` is also genuinely informative when it differs from the raw input (`got Invalid Date`, `got NaN`).

Tests: updated the `Primitive.choice` assertions, added unit coverage for both `InvalidValue.message` branches, added a test asserting the rendered stderr line through `Command.runWith`, and corrected a stale inline snapshot in `Arguments.test.ts` that encoded wording which no longer exists in the codebase.

## Related

Closes Effect-TS/effect#6312

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CLI invalid-value and missing-value error formatting to avoid duplicated “Expected” prefixes.
  * Improved invalid-choice error output to focus on the allowed options without repeating the invalid “got” framing.
  * Standardized CLI argument/flag validation wording for more consistent messaging.
* **Tests**
  * Expanded CLI test coverage for invalid value, invalid choice, and argument/primitive parsing error messages.
* **Documentation**
  * Updated the release notes entry describing the corrected “Expected” prefix behavior.

### DIFF
--- a/.changeset/fix-invalid-value-doubled-expected.md
+++ b/.changeset/fix-invalid-value-doubled-expected.md
@@ -1,0 +1,25 @@
+---
+"effect": patch
+---
+
+Fix doubled `Expected: Expected ...` prefixes in CLI `InvalidValue` error messages, closes #6312.
+
+`Primitive.choice` now fails with a bare description of the accepted values, and `CliError.InvalidValue` no longer prepends its own `Expected:` label when the underlying failure message (e.g. a schema decode message) already reads as a complete `Expected ...` sentence.
+
+Before:
+
+```
+Invalid value for flag --size: "bogus". Expected: Expected "small" | "medium" | "large", got "bogus"
+```
+
+After:
+
+```
+Invalid value for flag --size: "bogus". Expected: "small" | "medium" | "large"
+```
+
+Schema-backed primitives (`integer`, `float`, `boolean`, `date`) keep their full schema decode sentence, so the decoded actual is still shown after `got`:
+
+```
+Invalid value for flag --count: "3.14". Expected an integer, got 3.14
+```
--- a/.changeset/fix-invalid-value-doubled-expected.md
+++ b/.changeset/fix-invalid-value-doubled-expected.md
@@ -3,23 +3,3 @@
 ---
 
 Fix doubled `Expected: Expected ...` prefixes in CLI `InvalidValue` error messages, closes #6312.
-
-`Primitive.choice` now fails with a bare description of the accepted values, and `CliError.InvalidValue` no longer prepends its own `Expected:` label when the underlying failure message (e.g. a schema decode message) already reads as a complete `Expected ...` sentence.
-
-Before:
-
-```text
-Invalid value for flag --size: "bogus". Expected: Expected "small" | "medium" | "large", got "bogus"
-```
-
-After:
-
-```text
-Invalid value for flag --size: "bogus". Expected: "small" | "medium" | "large"
-```
-
-Schema-backed primitives (`integer`, `float`, `boolean`, `date`) keep their full schema decode sentence, so the decoded actual is still shown after `got`:
-
-```text
-Invalid value for flag --count: "3.14". Expected an integer, got 3.14
-```
--- a/.changeset/fix-invalid-value-doubled-expected.md
+++ b/.changeset/fix-invalid-value-doubled-expected.md
@@ -8,18 +8,18 @@ Fix doubled `Expected: Expected ...` prefixes in CLI `InvalidValue` error messag
 
 Before:
 
-```
+```text
 Invalid value for flag --size: "bogus". Expected: Expected "small" | "medium" | "large", got "bogus"
 ```
 
 After:
 
-```
+```text
 Invalid value for flag --size: "bogus". Expected: "small" | "medium" | "large"
 ```
 
 Schema-backed primitives (`integer`, `float`, `boolean`, `date`) keep their full schema decode sentence, so the decoded actual is still shown after `got`:
 
-```
+```text
 Invalid value for flag --count: "3.14". Expected an integer, got 3.14
 ```
--- a/packages/effect/src/unstable/cli/CliError.ts
+++ b/packages/effect/src/unstable/cli/CliError.ts
@@ -359,16 +359,25 @@ export class InvalidValue extends Schema.TaggedErrorClass<InvalidValue>(
   /**
    * Formats the invalid flag or argument value with the expected input.
    *
+   * **Details**
+   *
+   * When `expected` is already a complete `Expected ...` sentence (e.g. a
+   * schema decode message), it is appended as-is instead of being prefixed
+   * with another `Expected:` label.
+   *
    * @since 4.0.0
    */
   override get message() {
+    const expectation = this.expected.startsWith("Expected ")
+      ? this.expected
+      : `Expected: ${this.expected}`
     if (this.kind === "argument") {
-      return `Invalid value for argument <${this.option}>: "${this.value}". Expected: ${this.expected}`
+      return `Invalid value for argument <${this.option}>: "${this.value}". ${expectation}`
     }
     if (this.value.length === 0) {
-      return `Missing value for flag --${this.option}. Expected: ${this.expected}`
+      return `Missing value for flag --${this.option}. ${expectation}`
     }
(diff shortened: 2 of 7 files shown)


.changeset/fix-invalid-value-doubled-expected.md
```diff
@@ -0,0 +1,25 @@
...
+
+```text
+Invalid value for flag --size: "bogus". Expected: "small" | "medium" | "large"
+```
+
+Schema-backed primitives (`integer`, `float`, `boolean`, `date`) keep their full schema decode sentence, so the decoded actual is still shown after `got`:
+
+```text
+Invalid value for flag --count: "3.14". Expected an integer, got 3.14
+```
```
IMax153:
```suggestion
Fix doubled `Expected: Expected ...` prefixes in CLI `InvalidValue` error messages, closes #6312.
```

packages/effect/src/unstable/cli/CliError.ts
```diff
@@ -359,16 +359,25 @@ export class InvalidValue extends Schema.TaggedErrorClass<InvalidValue>(
   /**
    * Formats the invalid flag or argument value with the expected input.
    *
+   * **Details**
+   *
+   * When `expected` is already a complete `Expected ...` sentence (e.g. a
+   * schema decode message), it is appended as-is instead of being prefixed
+   * with another `Expected:` label.
```
IMax153:
```suggestion
```

IMax153 (changes requested):
Looks good, let's just trim down the verbosity in the changeset / jsdocs.

IMax153 (approved):

PR #6510, opened 2026-07-21
@effect/sql-mysql2: add disablePreparedStatements (text protocol for Hyperdrive-style proxies)

### LINKED ISSUE #6509: @effect/sql-mysql2: no way to disable prepared statements — unusable behind Cloudflare Hyperdrive (COM_STMT_PREPARE)

`@effect/sql-mysql2@4.0.0-beta.99` (effect 4.0.0-beta.99)

Use `@effect/sql-mysql2` from a Cloudflare Worker through a [Hyperdrive](https://developers.cloudflare.com/hyperdrive/) connection string (e.g. fronting PlanetScale MySQL):

```ts
const SqlLive = MysqlClient.layer({
  url: Redacted.make(env.HYPERDRIVE.connectionString)
})

const rows = yield* sql`SELECT * FROM users WHERE id = ${id}`
```

Queries execute.

### What do you see instead?

Every statement fails with:

```
Hyperdrive does not currently support MySQL COM_STMT_PREPARE messages
```

Hyperdrive multiplexes client connections onto pooled upstream connections, and MySQL prepared-statement handles are per-connection server state, so Hyperdrive rejects `COM_STMT_PREPARE` outright. `MysqlClient`'s connection implementation routes every parameterized statement through mysql2's `connection.execute()` (the prepared/binary protocol), so the client is unusable behind Hyperdrive — and behind other transaction-mode MySQL proxies with the same limitation.

### Additional information

Two adjacent gaps make the workaround impossible from config alone:

1. There is no way to opt out of prepared statements. mysql2's text path (`connection.query()`) escapes parameters client-side, so parameterization is preserved — this is the same escape hatch postgres.js exposes as `prepare: false` for transaction-mode poolers like pgbouncer. Notably `executeStream` already uses `connection.query(...).stream()`, so streaming works on the text protocol today; only the default `execute`/`executeValues`/`executeRaw` paths prepare.
2. When `url` is set, `poolConfig` is ignored (the `createPool` call in the `url` branch doesn't spread it), so `disableEval: true` — required on Cloudflare Workers because mysql2's JIT row parsers call `new Function(...)`, which workerd disallows — cannot be combined with a connection URI.

Proposal: add `disablePreparedStatements?: boolean` to `MysqlClientConfig` that flips the default method from `"execute"` to `"query"`, and spread `poolConfig` in the `url` branch as well. PR incoming from my fork.

sam-goodwin:
Fixes #6509

Adds `disablePreparedStatements` to `MysqlClientConfig`: statements run over the text protocol (`connection.query`) instead of `connection.execute`, for proxies that reject `COM_STMT_PREPARE` (e.g. Cloudflare Hyperdrive). Equivalent of postgres.js's `prepare: false`. Only the default of the existing `method` parameter changes; `executeStream` already used `query`.

Also spreads `poolConfig` into `createPool` when `url` is set (previously ignored on that path).

Includes a testcontainers case; suite passes locally against a real MySQL container.

## Summary by CodeRabbit

* **New Features**
  * Added a `disablePreparedStatements` option for MySQL connections to force text-protocol execution instead of prepared statements, improving compatibility with proxies that block binary prepared-statement traffic.
  * Parameterized queries remain safely parameterized when the option is enabled.
* **Bug Fixes**
  * Improved MySQL pool defaults so connection limits and idle timeouts correctly apply from `poolConfig` when explicit values aren’t provided.
* **Tests**
  * Added coverage to confirm that, with `disablePreparedStatements` enabled, query execution never uses the prepared-statement execution path.

### DIFF
--- a/.changeset/mysql2-disable-prepared-statements.md
+++ b/.changeset/mysql2-disable-prepared-statements.md
@@ -2,13 +2,4 @@
 "@effect/sql-mysql2": patch
 ---
 
-Add `disablePreparedStatements` to `MysqlClientConfig`, running every statement over the MySQL text protocol (`connection.query`) instead of prepared statements (`connection.execute` / `COM_STMT_PREPARE`).
-
-Some MySQL proxies do not support the binary protocol — Cloudflare Hyperdrive rejects every prepared statement with "Hyperdrive does not currently support MySQL COM_STMT_PREPARE messages", which previously made this client unusable behind it. mysql2 still escapes parameters client-side on the text path, so parameterization is preserved. Streaming already used the text protocol and is unchanged.
-
-```ts
-MysqlClient.layer({
-  url: Redacted.make(env.HYPERDRIVE.connectionString),
-  disablePreparedStatements: true
-})
-```
+Add `disablePreparedStatements` to `MysqlClientConfig`, to completely disable prepared statements
--- a/.changeset/mysql2-disable-prepared-statements.md
+++ b/.changeset/mysql2-disable-prepared-statements.md
@@ -9,10 +9,6 @@ Some MySQL proxies do not support the binary protocol — Cloudflare Hyperdrive 
 ```ts
 MysqlClient.layer({
   url: Redacted.make(env.HYPERDRIVE.connectionString),
-  disablePreparedStatements: true,
-  // mysql2's JIT row parsers use `new Function`, which workerd disallows
-  poolConfig: { disableEval: true }
+  disablePreparedStatements: true
 })
 ```
-
-`poolConfig` is now also applied when `url` is set (it was previously ignored on that path), so options like `disableEval` can be combined with a connection URI.
--- a/.changeset/mysql2-disable-prepared-statements.md
+++ b/.changeset/mysql2-disable-prepared-statements.md
@@ -1,0 +1,18 @@
+---
+"@effect/sql-mysql2": patch
+---
+
+Add `disablePreparedStatements` to `MysqlClientConfig`, running every statement over the MySQL text protocol (`connection.query`) instead of prepared statements (`connection.execute` / `COM_STMT_PREPARE`).
+
+Some MySQL proxies do not support the binary protocol — Cloudflare Hyperdrive rejects every prepared statement with "Hyperdrive does not currently support MySQL COM_STMT_PREPARE messages", which previously made this client unusable behind it. mysql2 still escapes parameters client-side on the text path, so parameterization is preserved. Streaming already used the text protocol and is unchanged.
+
+```ts
+MysqlClient.layer({
+  url: Redacted.make(env.HYPERDRIVE.connectionString),
+  disablePreparedStatements: true,
+  // mysql2's JIT row parsers use `new Function`, which workerd disallows
+  poolConfig: { disableEval: true }
+})
+```
+
+`poolConfig` is now also applied when `url` is set (it was previously ignored on that path), so options like `disableEval` can be combined with a connection URI.
--- a/packages/sql/mysql2/src/MysqlClient.ts
+++ b/packages/sql/mysql2/src/MysqlClient.ts
@@ -194,6 +194,19 @@ export interface MysqlClientConfig {
   readonly maxConnections?: number | undefined
   readonly connectionTTL?: Duration.Input | undefined
 
+  /**
+   * Extra options passed through to `mysql2`'s `createPool`.
+   *
+   * **Details**
+   *
+   * Explicit config fields (`maxConnections`, `connectionTTL`, …) take
+   * precedence over the equivalent `poolConfig` entries. When `url` is set,
+   * the connection-identity fields (`uri`, `host`, `port`, `user`,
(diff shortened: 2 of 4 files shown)


packages/sql/mysql2/src/MysqlClient.ts
```diff
@@ -301,14 +315,15 @@ export const make = (
...
+        ...options.poolConfig,
         uri: Redacted.value(options.url),
         multipleStatements: true,
         supportBigNumbers: true,
         connectionLimit: options.maxConnections!,
         idleTimeout: options.connectionTTL
           ? Duration.toMillis(Duration.fromInputUnsafe(options.connectionTTL))
           : undefined as any
-      })
+      } as Mysql.PoolOptions)
```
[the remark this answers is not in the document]
sam-goodwin: Fixed in 28e7e7a93 — both branches now fall back to `poolConfig.connectionLimit`/`idleTimeout` when `maxConnections`/`connectionTTL` are unset.

packages/sql/mysql2/test/MysqlClient.test.ts
```diff
@@ -32,4 +32,31 @@ describe("MysqlClient", () => {
...
+
+        yield* sql`INSERT INTO text_protocol_test ${sql.insert([{ value: "a" }, { value: "b" }])}`
+
+        const rows = yield* sql<{ value: string }>`SELECT value FROM text_protocol_test WHERE value = ${"a"}`
+        assert.deepStrictEqual(rows, [{ value: "a" }])
+
+        const streamed = yield* Stream.runCollect(
+          sql<{ value: string }>`SELECT value FROM text_protocol_test ORDER BY id`.stream
+        )
+        assert.deepStrictEqual(streamed, [{ value: "a" }, { value: "b" }])
```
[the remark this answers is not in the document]
sam-goodwin: In v4 `Stream.runCollect` returns `Effect<Array<A>>` (packages/effect/src/Stream.ts), not a `Chunk` — the assertion is correct as written and the test passes against a real MySQL container.

packages/sql/mysql2/src/MysqlClient.ts
```diff
@@ -301,14 +315,15 @@ export const make = (
 
     const pool = options.url
       ? Mysql.createPool({
+        ...options.poolConfig,
```
IMax153: The issue here is that `poolConfig` can contain connection options like `host`, `port`, `user`, `password`, etc. 

Spreading the `poolConfig` here when `options.url` is explicitly set would go against the documented behavior of `options.url`.

```ts
/**
 * Connection URI. Setting this will override the other connection options
 */
readonly url?: Redacted.Redacted
```

This is because `mysql2` treats any fields supplied alongside the connection URL as **overrides** - it only copies a value parsed from the URL when the corresponding explicit option is falsy.

packages/sql/mysql2/test/MysqlClient.test.ts
```diff
@@ -32,4 +32,31 @@ describe("MysqlClient", () => {
       Effect.provide(MysqlContainer.layerClient),
       Effect.catchTag("ContainerError", () => Effect.void)
     ), { timeout: 60_000 })
+
+  it.effect(
```
IMax153: This test doesn't really test that `disablePreparedStatements` does anything - it would pass even if `disablePreparedStatements` was `false`. 

A better test would have some sort of injected spy or mock that detects that query mode is indeed in use.

IMax153 (changes requested):

packages/sql/mysql2/src/MysqlClient.ts
sam-goodwin: Went further in 1d7579c: the `url` branch no longer touches `poolConfig` at all — it's byte-identical to what's on main, so `url` stays fully authoritative (including query-string params). `poolConfig` applies only with the discrete connection options, where it always has. The PR is now just `disablePreparedStatements` plus the `connectionLimit`/`idleTimeout` fallbacks on the non-url branch.

IMax153 (approved):
ccing @tim-smart to also give this a glance

.changeset/mysql2-disable-prepared-statements.md
```diff
@@ -0,0 +1,14 @@
...
+Add `disablePreparedStatements` to `MysqlClientConfig`, running every statement over the MySQL text protocol (`connection.query`) instead of prepared statements (`connection.execute` / `COM_STMT_PREPARE`).
+
+Some MySQL proxies do not support the binary protocol — Cloudflare Hyperdrive rejects every prepared statement with "Hyperdrive does not currently support MySQL COM_STMT_PREPARE messages", which previously made this client unusable behind it. mysql2 still escapes parameters client-side on the text path, so parameterization is preserved. Streaming already used the text protocol and is unchanged.
+
+```ts
+MysqlClient.layer({
+  url: Redacted.make(env.HYPERDRIVE.connectionString),
+  disablePreparedStatements: true
+})
+```
```
tim-smart:
```suggestion
Add `disablePreparedStatements` to `MysqlClientConfig`, to completely disable prepared statements
```

tim-smart (approved):

Issue #5594, opened 2025-10-01
Sql-Drizzle/Sqlite throwing STypeError: this.batchCLient is not a function


Sillyvan:
### What version of Effect is running?

"effect: ^3.18.1", "@effect/sql-d1": "^0.44.0", "@effect/sql-drizzle": "^0.45.0",

### What steps can reproduce the bug?

Theres no error in my Editor. Error only happens when it tries to run db.batch()
Im assuming this is related to Either sql-drizzle or d1 itself.

Failed to load analytics data: (FiberFailure) Error: Failed to get platform stats: TypeError: this.batchCLient is not a function
    at catch (/Users/silvan/dev/chaufi-mono/apps/web/src/routes/admin/analytics/analytics.remote.ts:66:32)
    at file:///Users/silvan/dev/chaufi-mono/node_modules/.pnpm/effect@3.18.1/node_modules/effect/dist/esm/internal/core-effect.js:579:51
    at file:///Users/silvan/dev/chaufi-mono/node_modules/.pnpm/effect@3.18.1/node_modules/effect/dist/esm/internal/fiberRuntime.js:931:41

Issue #3888, opened 2024-11-04
Ask for D1 batch() in sql-d1 to support a transaction around multiple statements


mw10013:
### What is the problem this feature would solve?

D1 seems to only support a transaction around multiple statements with batch() (https://developers.cloudflare.com/d1/build-with-d1/d1-client-api/#batch-statements).

More color in this comment by mikearnaldi: https://github.com/Effect-TS/effect/pull/3045#issuecomment-2183990029

### What is the feature you are proposing to solve the problem?

Add batch() to D1Client in sql-d1.

fubhy:
Closing this as a stale feature request. The requested D1 batch support may still be a valid design area, but there has not been recent maintainer ownership or a concrete next step on this issue.

PR #6524, opened 2026-07-22
Add D1 batch support

nr1brolyfan:
## Summary

- Add `D1Client.batch`, backed by native `D1Database.batch`
- Single D1 request for a fixed collection of statements
- Atomic rollback when any statement fails
- Typed heterogeneous tuple results, preserving statement order

## Details

- Reuses the prepared statement cache
- Supports parameter binding
- Applies query/result name transforms and `withoutTransforms()`
- Applies `Statement.CurrentTransformer`
- Adds `sql.execute` tracing with the `batch` operation
- Maps native failures to `SqlError`
- Empty batches return `[]` without calling D1
- Keeps `withTransaction` behavior unchanged
- D1-specific API rather than a generic `SqlClient.batchTx`

Enables a follow-up implementation of `.batch()` in `drizzle-orm/effect-d1`.

## Tests

- Ordered and typed results
- Name transforms and transformer replacement
- Empty batches
- Atomic rollback with Miniflare

## Related

Refs #3888 and #5594. Previous `batchTx` discussion: #3045.

## Summary by CodeRabbit

* **New Features**
  * Added support for executing multiple SQL statements as a single atomic D1 batch.
  * Batch results are returned in statement order, including support for empty batches.
  * Query and result transformations are applied consistently to batched statements.

* **Bug Fixes**
  * Batched execution failures now report structured SQL errors and roll back changes to preserve data integrity.

### DIFF
--- a/packages/sql/d1/src/D1Client.ts
+++ b/packages/sql/d1/src/D1Client.ts
@@ -133,61 +133,61 @@ const makeBatch = (options: {
   readonly spanAttributes: ReadonlyArray<readonly [string, unknown]>
   readonly getClient: () => D1Client
 }): D1Client["batch"] =>
-  Effect.fnUntraced(function*<const Statements extends ReadonlyArray<Statement.Statement<any>>>(
-    statements: Statements
-  ) {
-    if (statements.length === 0) {
-      return [] as unknown as BatchResults<Statements>
-    }
-    return yield* Effect.useSpan(
-      "sql.execute",
-      { kind: "client" },
-      (span) =>
-        Effect.withFiber(Effect.fnUntraced(function*(fiber) {
-          const transformer = yield* Statement.CurrentTransformer
-          const prepared: Array<D1PreparedStatement> = []
-          const transforms: Array<TransformRows | undefined> = []
-          const queryTexts: Array<string> = []
+<const Statements extends ReadonlyArray<Statement.Statement<any>>>(
+  statements: Statements
+) => {
+  if (statements.length === 0) {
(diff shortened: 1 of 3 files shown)


packages/sql/d1/src/D1Client.ts
```diff
@@ -182,7 +212,68 @@ export const make = (
...
+        },
+        SqlError
+      > => {
+        if (statements.length === 0) {
+          return Effect.succeed([]) as any
+        }
+        return Effect.useSpan(
+          "sql.execute",
+          { kind: "client" },
+          (span) =>
```
tim-smart: Use Effect.fnUntraced here

packages/sql/d1/src/D1Client.ts
```diff
@@ -182,7 +212,68 @@ export const make = (
...
+            Effect.gen(function*() {
+              const compiled = yield* Effect.forEach(statements, (statement) =>
```
tim-smart: We can use a normal for loop here, and extract the transformer using `yield* Statement.CurrentTransformer` above the loop.

We can also construct the `prepared` array in the same loop

packages/sql/d1/src/D1Client.ts
```diff
@@ -182,7 +212,68 @@ export const make = (
           catch: (cause) => new SqlError({ reason: classifyError(cause, "Failed to execute statement", "execute") })
         })
 
-      return identity<Connection>({
+      const makeBatch = (
```
tim-smart: Can this function be at the module level?

tim-smart (changes requested):

PR #6547, opened 2026-07-23
Add API diff tooling for v3-to-v4 migration

tim-smart:
## Summary

- Add a new API diff tool that discovers, snapshots, maps, and compares exported APIs across Effect versions.
- Generate JSON and Markdown v3-to-v4 migration reports covering moved, split, added, removed, and changed APIs.
- Add CLI documentation, package configuration, workspace integration, and focused tests for the diff pipeline.

## Testing

- Added unit tests for API discovery, snapshot generation, mapping, and diff behavior.
- Not run (test results were not provided).

## Summary by CodeRabbit

* **New Features**
  * Added an API comparison command that compares two repository revisions and produces JSON snapshots, a structured diff, and a Markdown report.
  * Reports identify added, removed, moved, renamed, and structurally changed APIs, with suggested replacements where applicable.
  * Added automatic discovery of public package entry points and deterministic snapshot generation.
* **Documentation**
  * Expanded the v3-to-v4 migration mapping and guide with updated module and API rename information.
  * Added usage documentation for the API comparison tool.


tim-smart (commented):
Reviewed with a focus on correctness and simplicity, plus one specific acceptance check: the tool must discover renamed/removed v3 APIs purely from the TypeScript code, with no hand-maintained lists. Ran the package tests (`pnpm vitest run`, 9/9 pass) and typecheck (`pnpm check`, clean) on the branch.

## Acceptance check: discovery is purely code-driven ✅

The pipeline is genuinely list-free: `Discovery.ts` enumerates entrypoints from each package's `exports` map plus built `.d.ts` files, `Snapshot.ts` extracts every exported symbol through the TypeScript compiler API, and `Diff.ts` matches base/head entities structurally (fingerprints, member sets, name similarity, JSDoc tokens). The earlier `Mapping.ts`, which consumed `migration/v3-to-v4.json`, was removed in the final commit. Good.

## Main issue: the PR still ships the hand-maintained map it just made obsolete

- `migration/v3-to-v4.json` (3,988 lines) is orphaned in the final tree — nothing reads it and nothing generates it.
- `migration/v3-to-v4.md` is rewritten to claim "This file is generated from `migration/v3-to-v4.json`. Do not edit it directly." — that was true mid-branch (via the now-deleted `Mapping.ts`) but is false in the final state.

These files are exactly the hand-maintained-list shape this tool is meant to replace, and the tool's actual outputs (`diff.json` / `report.md`) do not emit this format. Suggest either dropping both `migration/` changes from the PR, or keeping a generator in-tree so the map is reproducibly derived from a diff run. As-is, the next person to touch `migration/` has a "generated" file with no generator.

## Correctness (minor)

- `packages/tools/api-diff/src/Diff.ts:470` — `movementChange(match)` in the authoritative-match loop is dead code: authoritative matches come only from exact-id lookups (`unmatchedHead.get(baseEntity.id)`), and the id embeds `module#path#bucket`, so neither `api-renamed` nor `api-moved` can ever fire there. Net effect: every rename is reported only as a non-authoritative suggestion plus an `api-removed`/`api-added` pair. That matches the README's stated design, but the dead call reads as if authoritative renames exist. Either delete the call or (if authoritative renames are wanted) promote high-confidence exact-fingerprint suggestions to matches.
- `packages/tools/api-diff/src/ApiDiff.ts:95` — the `report.length === 0` guard is unreachable (`renderMarkdownReport` always emits the header) and runs after all four files are written anyway. Delete it.
- `packages/tools/api-diff/src/Model.ts:112` — `module-moved`, `module-split`, `module-consolidated`, and `unsupported` are never produced by `diffSnapshots`; they're leftovers from the mapping-based version and are still referenced in the `Report.ts` grouping tables. Pruning them keeps the classification set honest for downstream doc tooling.

## Simplicity

- `packages/tools/api-diff/src/Discovery.ts:193` — the entire `if (requested !== undefined)` pre-pass (~53 lines) is redundant: the general scan below it already filters both exact and wildcard export keys by `requested`, applies the same null-exclusion rules, and `missing` is computed from the output map either way. I verified this empirically: with the pre-pass deleted, all 9 tests still pass, including the requested-modules/missing assertions. Suggest removing it.
- `packages/tools/api-diff/src/Cli.ts:8` — all three flags are `Flag.optional`, and `ApiDiff.run` then manually errors when any is `None`. Making the flags required lets the CLI layer enforce and document this, turns `ApiDiffOptions` into plain strings, and deletes the manual check.
- `plan.md` — a 325-line design scratchpad at the repository root. Suggest dropping it from the PR (or moving it under `packages/tools/api-diff/` if it should live on).
- `.changeset/detect-api-removals.md`, `.changeset/migrate-api-diff-cli.md` — `@effect/api-diff` is private and the changeset config has `privatePackages: false`, so these are silently ignored no-ops (verified with `pnpm changeset status`). They can be dropped.

## Non-blocking nits

- `packages/tools/api-diff/src/Snapshot.ts:80` — `symbol !== undefined &&` is redundant immediately after the `undefined` early return.
- `packages/tools/api-diff/src/Discovery.ts:123` — the package walk also picks up `dist/package.json` files themselves, producing duplicate `PackageInfo` entries for built packages. Harmless today (the output map dedupes by module) but skipping paths already inside a discovered package root would avoid double work.

PR #6550, opened 2026-07-23
fix(openapi): return fresh cached specs

### LINKED ISSUE #6330: OpenApi.fromApi exposes shared mutable state
OpenApi.fromApi returns the same cached OpenAPISpec instance for repeated calls with the same API.

Because OpenAPISpec is mutable, any caller can modify the cached object and silently change the specification seen by every subsequent caller. The generated output therefore depends on prior external mutations.

xianjianlf2:
Fixes #6330

## Summary
- keep the internal OpenApi.fromApi cache but return fresh deep copies to callers
- preserve special OpenAPI keys like __proto__ as own enumerable properties while cloning
- add regression coverage for external mutation not affecting later cached calls

## Tests
- pnpm test packages/effect/test/unstable/httpapi/OpenApi.test.ts
- pnpm lint-fix
- pnpm check

## Summary by CodeRabbit

* **Bug Fixes**
  * OpenAPI specifications are now returned as fresh copies, preventing changes made to one result from affecting later results.
  * Cached OpenAPI data no longer leaks mutations across requests or calls.


xianjianlf2:
Merged latest `origin/main` into `fix/openapi-from-api-copy-6330` to refresh the stale failing CI run. The previous CodeRabbit request was already addressed in the current branch head.

Validation:
- `pnpm test packages/effect/test/unstable/httpapi/OpenApi.test.ts` (18 passed)

Current head: `35ceff4f6f0388fd698d2aa7800984faf76d546a`

xianjianlf2:
Follow-up after CI: the failing Node/Deno shard was `OpenApiRepresentation.test.ts`, which still asserted `OpenApi.fromApi(Api)` returned the same object reference. This PR intentionally returns fresh copies from the cached spec, so I updated that assertion to verify a distinct object with equal content.

Additional validation:
- `pnpm test packages/effect/test/unstable/httpapi/OpenApi.test.ts packages/effect/test/unstable/httpapi/OpenApiRepresentation.test.ts` (24 passed)
- `pnpm exec dprint check packages/effect/src/unstable/httpapi/OpenApi.ts packages/effect/test/unstable/httpapi/OpenApi.test.ts packages/effect/test/unstable/httpapi/OpenApiRepresentation.test.ts`
- `git diff --check -- packages/effect/src/unstable/httpapi/OpenApi.ts packages/effect/test/unstable/httpapi/OpenApi.test.ts packages/effect/test/unstable/httpapi/OpenApiRepresentation.test.ts`

Current head: `5ff5ddf5d9c83832a9cde0d5c8795e30ab1ec830`

gcanti:
I pushed two focused follow-up commits:

1. The cloning loop now uses `InternalRecord.assignProperty` instead of duplicating the `Object.defineProperty` logic. This preserves keys such as `__proto__` as own enumerable properties while reusing the repository’s existing safe record-assignment abstraction.
2. On a cache miss, `fromApi` now stores a deep-cloned private snapshot and returns the newly generated specification. Cache hits still return a fresh clone. This prevents references retained by `OpenApi.Override` or `OpenApi.Transform` inputs from mutating the cached value after generation. I also added a regression test covering mutation through an `Override` input.

PR #6412, opened 2026-07-16
feat: deno platform

lishaduck:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Init @effect/platform-deno.

This is slightly more minimal than the original PR, I split out the Deno cleanups that weren't directly in the new package to #6405 and removed the Deno KV layer (Deno KV isn't supported by the new Deno Deploy and I don't know of anybody who actually uses it locally).

## Related

Port of from https://github.com/Effect-TS/effect-smol/pull/859 to the main Effect repository.
Stacked on #6405.

- Related Issue Effect-TS/effect-smol#79
- Closes Effect-TS/effect#1806

## Summary by CodeRabbit

* **New Features**
  * Added Deno platform support for path handling, runtime execution, workers, worker runners, and Web Storage.
  * Added the new `@effect/platform-deno` package with public exports, documentation, licensing, and publishing configuration.
  * Added Deno-specific support for RPC worker applications and end-to-end workflows.

* **Tests**
  * Added coverage for storage, path operations, worker execution, and RPC scenarios.

* **Chores**
  * Updated project configuration to support Deno tooling and runtime-specific test execution.

PR #6555, opened 2026-07-23
test: add file system test suite

lloydrichards:
## Type
- [x] Optimization for v4

## Description

I wanted to explore adding a `MemoryFileSystem` but before jumping too deep into this realized that the `NodeFileSystem.test `was a little thin.  Ideally I would have some sort of suite that could be applied to both `NodeFileSystem` and `MemoryFileSystem` to make sure the implementation is aligned.  I notice a similar pattern with the `KeyValueStore` where you create a test suite that can accept an implementation and thus wanted to apply the same thing to FileSystem here before moving on to the implementation.

The interface is pretty simple:

```ts
const suite: (
  name: string,
  layer: Layer.Layer<FileSystem.FileSystem, unknown>
) => it.layer(layer, { timeout: { seconds: 30 } })(`FileSystem (${name})`, (it) => {...})

FileSystemTest.suite("node", NodeFileSystem.layer);
FileSystemTest.suite("memory", MemoryFileSystem.layer);
// FileSystemTest.suite("deno", DenoFileSystem.layer);
```

In the first commit I just ported what was already in `NodeFileSystem.test` but in the second I expanded and refactored the test suite to be a bit more behaviour driven, looking at what is currently implemented in Node but also using `sol` to help with assumptions and expectations.

In the end I ended up with 66 tests across 9 areas:

<details>
<summary>Full Test Suite</summary>

```txt
 ✓  @effect/platform-node-shared  test/NodeFileSystem.test.ts (66 tests) 78ms
   ✓ FileSystem (node) (63)
     ✓ path operations (17)
       ✓ should preserve binary bytes when reading and writing files 2ms
       ✓ should preserve UTF-8 text when reading and writing files 1ms
       ✓ should list directory entries and metadata when creating nested directories 1ms
       ✓ should include nested files when listing directories recursively 1ms
       ✓ should require parent directories when creation is not recursive 1ms
       ✓ should keep recursive directory creation idempotent when the path is a directory 1ms
       ✓ should report existence when a path is created and removed 1ms
       ✓ should access existing paths and reject missing paths when checking access 1ms
       ✓ should remove non-empty directories when removal is recursive 1ms
       ✓ should reject non-empty directories when removal is not recursive 1ms
       ✓ should ignore missing paths only when removal is forced 0ms
       ✓ should apply truncating, appending, and exclusive flags when writing files 1ms
       ✓ should copy independent file and directory-tree contents when copying paths 3ms
       ✓ should replace an existing copy destination when overwrite is requested 1ms
       ✓ should move files and directories when renaming paths 1ms
       ✓ should truncate and zero-extend files when changing their size 1ms
       ✓ should normalize stable filesystem errors when path operations fail 1ms
     ✓ temporary resources (1)
       ✓ should clean up scoped temporary resources when their scope closes 2ms
     ✓ streaming (3)
       ✓ should stream bounded file ranges when offsets and chunk sizes are configured 1ms
       ✓ should write stream chunks when using a file sink 1ms
       ✓ should append stream chunks when a file sink uses the append flag 1ms
     ✓ file handles and open modes (18)
       ✓ should keep read cursors independent when a file has multiple handles 1ms
       ✓ should advance and seek the write cursor when writing through a file handle 10ms
       ✓ should append primitive writes when a handle uses the append flag 1ms
       ✓ should truncate and extend files when using a file handle 1ms
       ✓ should return partial reads and EOF when reading beyond file contents 1ms
       ✓ should reject operations when a file-handle scope closes 1ms
       ✓ supports the 'r' open flag 1ms
       ✓ supports the 'r+' open flag 1ms
       ✓ supports the 'w' open flag 1ms
       ✓ supports the 'wx' open flag 1ms
       ✓ supports the 'w+' open flag 1ms
       ✓ supports the 'wx+' open flag 1ms
       ✓ supports the 'a' open flag 1ms
       ✓ supports the 'ax' open flag 1ms
       ✓ supports the 'a+' open flag 1ms
       ✓ supports the 'ax+' open flag 1ms
       ✓ should reject traversal through a file when a child path targets it 1ms
       ✓ should apply open flags consistently when high-level writes target existing and missing paths 4ms
     ✓ open file state (5)
       ✓ should zero-fill gaps when sparse writes extend a file 0ms
       ✓ should keep open handles attached when their file is renamed 1ms
       ✓ should preserve both open handles when renaming over an existing destination 1ms
       ✓ should keep unlinked contents accessible when a file handle remains open 0ms
       ✓ should preserve the read cursor when appending through a handle 1ms
     ✓ copy and rename (4)
       ✓ should preserve existing destinations when source-based mutations fail 1ms
       ✓ should replace existing file destinations when copying or renaming files 1ms
       ✓ should preserve existing destination links and handles when copying file contents 1ms
       ✓ should preserve an existing copy destination when overwrite is false 1ms
     ✓ temporary resource lifecycle (2)
       ✓ should create uniquely named temporary resources when prefixes and suffixes are requested 1ms
       ✓ should clean up scoped temporary resources when their scope fails 1ms
     ✓ stream lifecycle (3)
       ✓ should stream no chunks when files are empty or offsets are beyond EOF 1ms
       ✓ should preserve the original path when opening a file stream fails 0ms
       ✓ should finalize derived stream and sink handles when their operations succeed or fail 1ms
     ✓ adapter capabilities (10)
       ✓ should normalize missing-path errors when primitive path operations fail 1ms
       ✓ should check readable and writable access when no virtual identity is available 1ms
       ✓ should share contents and metadata when creating hard links 1ms
       ✓ should preserve contents through remaining links and handles when hard links are removed 1ms
       ✓ should resolve symbolic links when targets are relative or absolute 1ms
       ✓ should preserve error context when symbolic links form a loop 1ms
       ✓ should report updated access and modification timestamps when metadata is available 1ms
       ✓ should preserve the modification timestamp when copying with metadata preservation 1ms
       ✓ should exclude matching entries when globbing from a root 4ms
       ✓ should preserve error context when watching a missing path 0ms
   ✓ FileSystem (node-specific) (3)
     ✓ should read a pre-existing host file when the fixture exists 0ms
     ✓ should keep the file-handle cursor when truncation does not shorten before it 1ms
     ✓ should clamp the file-handle cursor when truncation shortens before it 1ms

 Test Files  1 passed (1)
      Tests  66 passed (66)
   Start at  21:09:44
   Duration  1.07s (transform 718ms, setup 516ms, import 394ms, tests 78ms, environment 0ms)
```
</details>

There are actually some decisions that could still be made about more tests to add as well as what is expected from the `FileSystem` which I've added a `// TODO` which I'll comment on in this PR to see if it makes sense to address now or raise as issues separately.

I wanted to get this in before working on the `MemoryFileSystem` mostly cause I think this would also be useful for other features like #6412 where they could reuse the same suite while also extracting deno-specific tests as needed. And also not to bloat the other PR with test cases.

## Related

- Discord [Thread](https://discord.com/channels/795981131316985866/1529516568567091294)

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive shared conformance test suite for filesystem adapter behavior, including path operations, scoped temp lifecycle cleanup, streaming, file-handle modes/cursors/truncation, copy/rename edge cases, links/symlinks, globbing, and consistent error reporting.
  * Refactored the Node.js filesystem tests to reuse the shared suite, reducing coverage to two targeted scenarios: fixture file reading and cursor behavior across truncation.


packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+          const filePath = path("eof.txt")
+          yield* fs.writeFileString(filePath, "abc")
+          const file = yield* fs.open(filePath)
+
+          const contents = yield* readAllocUpTo(file, 10)
+          assert.strictEqual(decode(contents), "abc")
+          assert.isTrue(Option.isNone(yield* file.readAlloc(1)))
+        }))
+
+      // TODO: Decide whether closed-handle operations normalize OS-specific EBADF as BadResource. Node reports Unknown.
```
lloydrichards: > Decide whether closed-handle operations normalize OS-specific EBADF as BadResource.

On Node, those operations fail at the OS level with `EBADF` (“bad file descriptor”). The Node error mapper currently has no EBADF case, so it falls through to `SystemErrorTag.Unknown`.

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+          yield* file.seek(0, "start")
+          yield* file.writeAll(encoder.encode("bar"))
+          assert.strictEqual(decode(yield* readAllocUpTo(file, 3)), "foo")
+
+          yield* file.writeAll(encoder.encode("baz"))
+          assert.strictEqual(decode(yield* readAllocUpTo(file, 6)), "barbaz")
+          assert.strictEqual(yield* fs.readFileString(filePath), "foobarbaz")
+        }))
+
+      // TODO: Decide whether File.seek returns the resulting cursor, then change its public return type from void to Size.
```
lloydrichards: > Decide whether File.seek returns the resulting cursor, then change its public return type from void to Size.

```ts
readonly seek: (offset: SizeInput, from: SeekMode) => Effect.Effect<void>
```

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+      // it.effect("should return the resulting cursor when seeking", () => {})
+
+      // TODO: Decide whether File.truncate clamps the calling handle's cursor.
```
lloydrichards: > Decide whether File.truncate clamps the calling handle's cursor.

Node clamps the cursor in this case, Node-specific test proves that. The reusable suite does not yet require it because the public `File.truncate` contract does not say what happens to the cursor.

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+          yield* fs.copyFile(copySource, copyDestination)
+          yield* fs.rename(renameSource, renameDestination)
+
+          assert.strictEqual(yield* fs.readFileString(copySource), "copied")
+          assert.strictEqual(yield* fs.readFileString(copyDestination), "copied")
+          assert.isFalse(yield* fs.exists(renameSource))
+          assert.strictEqual(yield* fs.readFileString(renameDestination), "renamed")
+        }))
+
+      // TODO: Define whether overwrite: false succeeds without changing an existing destination or fails with AlreadyExists.
```
lloydrichards: > Define whether overwrite: false succeeds without changing an existing destination or fails with AlreadyExists.

What remains undecided:
- Successful no-op <- copying to an existing destination with overwrite: false succeeds but leaves it untouched. This suits idempotent setup flows.
- AlreadyExists failure <- it reports that the requested copy did not happen. This suits callers that need to distinguish "copied" from "skipped"

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+        Effect.gen(function*() {
+          const { fs, path } = yield* makeTestContext
+          const missing = path("missing-stream.txt")
+
+          const error = yield* fs.stream(missing).pipe(Stream.runDrain, Effect.flip)
+
+          assertSystemError(error, { tag: "NotFound", method: "open", pathOrDescriptor: missing })
+        }))
+
+      // TODO: Add a deterministic synchronization point before covering stream and sink finalization after interruption.
```
lloydrichards: > Add a deterministic synchronization point before covering stream and sink finalization after interruption.

Add a test-only synchronization hook that blocks after the derived handle is acquired? Then interrupt the stream or sink and assert that its handle finalizer runs.

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+          assert.isTrue(Result.isFailure(failedResult))
+          if (Result.isFailure(failedResult)) {
+            assert.strictEqual(failedResult.failure, "expected failure")
+          }
+          assert.deepStrictEqual(yield* Ref.get(finalized), [streamed, sunk, failed])
+        }))
+    })
+
+    describe("adapter capabilities", () => {
+      // TODO: Define the public working-directory contract. Decide separately whether the memory adapter defaults to virtual `/`.
```
lloydrichards: > Define the public working-directory contract. Decide separately whether the memory adapter defaults to virtual `/`

In node, an absolute path is normally relative to the host process working directory. But for Memory filesystem there no natural process directory, so it needs a virtual default or explicit configuration.

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+            [undefined, fs.utimes(missing, 0, 0)]
+          ] as const
+
+          for (const [method, operation] of operations) {
+            const error = yield* Effect.flip(operation)
+            assertSystemError(error, { tag: "NotFound", method, pathOrDescriptor: missing })
+          }
+        }))
+
+      // TODO: Define what ok, readable, and writable mean without user identity or permission state.
```
lloydrichards: > Define what ok, readable, and writable mean without user identity or permission state.

In Node, those checks can use the OS’s effective user and file permission bits, but for a memory filesystem has no natural equivalent unless it explicitly models.

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+          const second = path("second-link")
+          yield* fs.symlink("second-link", first)
+          yield* fs.symlink("first-link", second)
+
+          const error = yield* Effect.flip(fs.realPath(first))
+
+          assertSystemError(error, { method: "realPath", pathOrDescriptor: first })
+        }))
+
+      // TODO: Add positive chmod and chown conformance in a POSIX-only metadata capability suite.
```
lloydrichards: > Add positive chmod and chown conformance in a POSIX-only metadata capability suite.

should an in memory adapter intentionally omit identity/permission modeling?

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+
+          if (Option.isSome(info.atime)) {
+            assert.strictEqual(info.atime.value.getTime(), atime.getTime())
+          }
+          if (Option.isSome(info.mtime)) {
+            assert.strictEqual(info.mtime.value.getTime(), mtime.getTime())
+          }
+        }))
+
+      // TODO: Define whether preserveTimestamps guarantees only mtime or both atime and mtime. Node preserves mtime but resets atime.
```
lloydrichards: > Define whether preserveTimestamps guarantees only mtime or both atime and mtime.

Node preserves mtime but resets atime, so baseline conformance currently checks only mtime

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+          const matches = (yield* fs.glob("**/*.ts", {
+            root,
+            exclude: ["**/*.test.ts"]
+          })).map((entry) => entry.replaceAll("\\", "/")).sort()
+
+          assert.strictEqual(matches.length, 1)
+          assert.isTrue(matches[0].endsWith("src/index.ts"))
+        }))
+
+      // TODO: Add deterministic watcher-registration and subscription-control hooks before specifying event ordering,
```
lloydrichards: > Add deterministic watcher-registration and subscription-control hooks before specifying event ordering, normalized paths, or watcher cleanup

packages/effect/test/FileSystemTest.ts
```diff
@@ -0,0 +1,1084 @@
...
+      // normalized paths, or watcher cleanup.
+      // it.effect("should emit normalized events when watching registered paths", () => {})
+
+      // TODO: Decide whether a missing watch target reports `watch` or the host preflight operation. Node exposes `stat`.
```
lloydrichards: > TODO: Decide whether a missing watch target reports `watch` or the host preflight operation. Node exposes `stat`.

Node currently performs that verification with an internal stat call. If the path is missing, the exposed PlatformError therefore says `method: "stat", tag: "NotFound"` even though the method was `watch`

lloydrichards:
Since Deno was added in the mean time I've added this too with the suite and even managed to find a bug with the open flag and error normalization

packages/effect/test/FileSystemTest.ts
lloydrichards: Both Node and Deno implementations return the resulting logical cursor, while the public interface erases it as void. Change the return type to `Effect<Size>` and enable the portable test. This is an exported API change and needs type validation plus a changeset.

packages/effect/test/FileSystemTest.ts
lloydrichards: Both Node and Deno clamp the non-append logical cursor when truncation shortens beneath it. Node already has two Node-specific tests proving this. Move/generalize those into the portable suite so Deno is covered too.

packages/effect/test/FileSystemTest.ts
lloydrichards: Both implementations now perform stat(path) before selecting the watcher backend, so a missing target consistently reports method: "stat". Enable that assertion and remove the TODO.

lloydrichards:
- [test: add suite to DenoFileSystem tests](https://github.com/Effect-TS/effect/pull/6555/commits/9a47c7ebc56e875422f8f12a57e768b521ad1428) [9a47c7e](https://github.com/Effect-TS/effect/pull/6555/commits/9a47c7ebc56e875422f8f12a57e768b521ad1428)
- [fix: deno writeFile honors open flags](https://github.com/Effect-TS/effect/pull/6555/commits/293fb8f623bb6ac7ba8355c65cbc67f96db47c9a) [293fb8f](https://github.com/Effect-TS/effect/pull/6555/commits/293fb8f623bb6ac7ba8355c65cbc67f96db47c9a)

Looks like this is already successful as both issues flagged have been fixed in separate PRs: #7076, #7080 and #7086

lloydrichards:
I noticed that with #6682 there was a `testLayer` added that had a bit of overlap with the `FileSystemTest` suite that I'm adding here. I've ported what was missing and removed it as I feel that it was still a bit thin compared to what I was testing the `MemoryFileSystem` against (#6573)

lloydrichards:
@tim-smart since its been a minute on this PR I just wanted to give you a gentle judge to see if you still want to review it.  I understand reviewing tests isn't much fun but there are some open questions i've added comments to which could be points for discussion or further development.  

That said, if you think the `testLayer` from the Deno PR is all you need and don't want to add this test suite then you are also welcome to let me know and close the PR.  I can still work with this as a local branch for the work on the `MemoryFileSystem` as needed.

packages/effect/test/FileSystemTest.ts
lloydrichards: This is now implemented by merged #7076: the public `File.seek` return type is `Size`. The [shared suite asserts the returned cursor](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L798). The old TODO is gone; this thread can remain resolved.

packages/effect/test/FileSystemTest.ts
lloydrichards: This is now covered by the [shared truncate-cursor tests](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L807), including preservation when the cursor remains within the file, clamping when truncation shortens beneath it, and reading appended bytes after clamping. Node and Deno use the same logical-cursor behavior. The old TODO is gone; this thread can remain resolved.

packages/effect/test/FileSystemTest.ts
lloydrichards: The [shared missing-watch test](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L1215) now asserts `method: "stat"`. Both Node and Deno perform that preflight before selecting the watch backend. This records the existing adapter behavior; the old TODO is gone and this thread can remain resolved.

packages/effect/test/FileSystemTest.ts
lloydrichards: Resolving this as a scope decision for this PR. The [shared fixture](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L9) uses paths under a scoped temporary directory, so it does not need a common working-directory default. Relative-path behavior stays adapter-specific. For the MemoryFileSystem proposal in #6573, the chosen default is virtual `/`; configurable working-directory behavior can be considered separately. I removed the broad TODO and empty test placeholder. Also, my original comment should have said "relative path", not "absolute path".

packages/effect/test/FileSystemTest.ts
lloydrichards: Resolving this as a scope decision for the shared suite. The [existing access test](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L1053) covers an accessible file and a missing path; it does not require a virtual user model. For #6573, the MemoryFileSystem policy is to store permission metadata without enforcing a virtual identity. Permission-denial behavior stays adapter-specific. The TODO is now a note stating that boundary.

packages/effect/test/FileSystemTest.ts
lloydrichards: Follow-up from checking the current adapters: this thread was already resolved, but the error-tag discrepancy is not fixed. Node still has no `EBADF` case and falls back to `Unknown`; Deno recognizes `BadResource`. I restored a [concrete suite TODO](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L506) separating shared closed-handle failure coverage from adapter error-tag normalization and its regression test. Keeping the historical thread resolved does not mean that follow-up is complete.

packages/effect/test/FileSystemTest.ts
lloydrichards: This remains a public-contract question. Node skips an existing destination with `overwrite: false`; Deno reports `AlreadyExists`, with its error mapping addressed in #7086. The [suite still accepts either outcome and checks that the destination contents are preserved](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L896). I updated the TODO with that concrete difference. Leaving this open for a decision on whether success/failure must agree across adapters.

packages/effect/test/FileSystemTest.ts
lloydrichards: The remaining work is a test addition, not a public synchronization API. The existing `trackedFs.open` wrapper can signal acquisition through a `Deferred`; then the test can interrupt a blocked stream or sink and assert finalization. I made the [TODO specific](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L989). Leaving this open because interruption coverage has not been implemented in this comment-only cleanup.

packages/effect/test/FileSystemTest.ts
lloydrichards: Partially answered by #7943: Deno now uses successful `chmod` to test that writing an existing file preserves its mode. The MemoryFileSystem proposal in #6573 stores permissions and ownership without enforcing a virtual identity. The remaining [TODO](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L1151) is positive chmod/chown conformance gated by adapter/host support and ownership privileges. I removed the empty test placeholder. Leaving the coverage work open.

packages/effect/test/FileSystemTest.ts
lloydrichards: The [shared test continues to assert mtime preservation](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L1173). I narrowed the TODO to the unanswered part: whether `preserveTimestamps` must also preserve `atime`. The previous blanket statement about Node resetting atime is removed; this cleanup does not establish a stronger timestamp guarantee. Leaving the contract question open.

packages/effect/test/FileSystemTest.ts
lloydrichards: Partially addressed: the [Node-specific startWatch helper](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/platform/node-shared/test/NodeFileSystem.test.ts#L27) already waits for a sentinel event before testing changes. I split the [TODO](https://github.com/Effect-TS/effect/blob/47ffc73e26598d29256a6dbd66f7bd9262985f38/packages/effect/test/FileSystemTest.ts#L1211) into adapter readiness/cleanup coverage and a separate portable event-path decision. Node emits filenames while Deno forwards native event paths; exact native event ordering should not be a prerequisite for cleanup tests. Leaving the remaining work open.

packages/effect/test/FileSystemTest.ts
lloydrichards: Correction to my previous reply: keeping the working-directory default adapter-specific does not prevent shared relative-path coverage. I have now implemented the placeholder in [b45bf328](https://github.com/Effect-TS/effect/commit/b45bf3288ef1decbf243eba3d03f89bd70f75b11).

The test obtains the adapter's working directory with `realPath(".")`, creates a scoped directory beneath it, writes through a relative path and reads through the absolute path, checks that both resolve to the same file, then updates through the absolute path and reads through the relative path. It does not change the host process working directory.

Validation: Node filesystem tests 79/79, Deno 73/73, and the new case against MemoryFileSystem through a temporary suite entry point. Lint and root type-check passed. I also temporarily changed Node's relative writes to resolve from `/`; this test failed with `NotFound`, then passed after restoring the adapter. The thread now has implemented coverage behind its resolution.

lloydrichards:
I revisited the open questions against the recent Deno work and updated the inline threads. Seek, truncate-cursor behavior, and missing-watch errors are covered now. Working-directory and permission semantics stay adapter-specific.

The two contract questions I’d appreciate input on are:

1. `overwrite: false`: Node skips an existing destination successfully, while Deno reports `AlreadyExists`. Should the shared suite allow both, provided the destination stays unchanged, or require the same outcome?
2. `preserveTimestamps`: Is preserving `mtime` enough for shared conformance, or should this also guarantee `atime`?

My preference is to keep the current minimum guarantees for this PR and handle stronger guarantees separately. I’ve made the remaining test and error-normalization TODOs concrete so those can progress independently.

PR #6561, opened 2026-07-23
Fix CLI positional argument overflow

hsyntax:
A positional variadic with a `max` stopped consuming values once it reached the limit, but any remaining operands were ignored after parsing finished.

For example:

```ts
const upload = Command.make("upload", {
  files: Argument.string("file").pipe(
    Argument.variadic({ max: 2 })
  )
})
```

Running:

```sh
upload a.txt b.txt c.txt
```

previously succeeded with `["a.txt", "b.txt"]` and silently discarded `c.txt`. It now fails with:

```text
Unexpected positional argument: "c.txt"
```

The fix rejects positional operands left over after the complete command has been parsed.

Variadics can still pass operands to positional arguments declared after them, so this remains valid:

```ts
const copy = Command.make("copy", {
  sources: Argument.string("source").pipe(
    Argument.variadic({ max: 2 })
  ),
  destination: Argument.string("destination")
})
```

```sh
copy a.txt b.txt output/
```

Partial parsing for shared flags and parent command context also continues to preserve operands, since those may belong to the active command or subcommand.

*This PR was raised with the help of Codex ( GPT-5.6-Sol-high )*

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated CLI error for unexpected leftover positional arguments, with clear messaging for one or multiple extra operands.
* **Bug Fixes**
  * CLI parsing now rejects unconsumed positional arguments after command/context parsing, including cases beyond configured variadic limits.
  * Refined shared and inherited option parsing so leftover tokens are handled consistently when combining parsers.
* **Tests**
  * Expanded coverage for unexpected-argument scenarios and verified error message formatting.

### DIFF
--- a/.changeset/fix-cli-unexpected-arguments.md
+++ b/.changeset/fix-cli-unexpected-arguments.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Reject unexpected positional arguments left after command parsing, including values exceeding `Argument.variadic` maximum bounds.
--- a/packages/effect/src/unstable/cli/CliError.ts
+++ b/packages/effect/src/unstable/cli/CliError.ts
@@ -2,11 +2,11 @@
  * Defines structured errors for the unstable CLI parser and runner.
  *
  * CLI errors describe problems such as unknown or duplicate flags, missing
- * flags or arguments, invalid values, unknown subcommands, user handler
- * failures, and requests to show command help. This module includes the
- * `CliError` union, the `isCliError` guard, schema-backed error classes with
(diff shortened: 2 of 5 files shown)


IMax153:
I think this is the behavior that we want / expect, no?

hsyntax:
> I think this is the behavior that we want / expect, no?

Just to clarify, are you asking if the original behaviour was correct? 

The core problem here is that the original implementation was silently dropping user input - especially when a `max` num was defined. 

Codex also states that the v3 @effect/cli implementation did fail this case with a `Received unknown argument: '<value>'` error, but I didn't verify that myself.

PR #6564, opened 2026-07-23
Run SQL persistence cleanup in bounded batches

AVtheking:
## Summary

- add expiry indexes for the shared SQL persistence table across PostgreSQL, MySQL, SQL Server, and SQLite
- replace the unbounded startup deletion with a scoped worker that deletes expired entries in batches of 1,000 until drained
- use dialect-appropriate atomic deletion queries and retain a short delay between batches
- add integration coverage for bounded cleanup while preserving live and future entries

## Motivation

Previously, expiration cleanup was coupled to `BackingPersistence.make`. There was no independent cleanup process, so expired entries for a store were only considered when that store was opened again. If `make` was never called again, those rows could remain in `effect_persistence` indefinitely.

When cleanup did run, it issued one unbounded `DELETE` for every expired entry in that store. The `expires` column also had no supporting index, so the database could scan a large part of the table and delete an arbitrary number of rows in a single statement. That combination could cause significant database load, longer-held locks, and unpredictable startup latency.

Each app instance runs its own cleanup worker. The bounded statements remain safe when workers overlap; PostgreSQL and SQL Server coordinate through row locking, while a MySQL deadlock aborts one cleanup attempt and is logged before the next periodic retry.

## Validation

- `pnpm lint-fix`
- `pnpm check`
- `pnpm test packages/sql/pg/test/Persistence.test.ts` (14 tests)
- `pnpm test packages/sql/mysql2/test/Persistence.test.ts` (12 tests)
- `pnpm test packages/sql/sqlite-node/test/Persistence.test.ts` (11 tests)
- SQL Server 2022 container verification for the actual 1,000-row and trailing-batch delete query

Closes EFF-256

## Summary by CodeRabbit

* **New Features**
  * SQL-backed shared-table persistence now automatically cleans up expired entries in indexed, bounded background batches.
  * Cleanup runs periodically, deleting expired rows until none remain while preserving active/non-expiring entries.
  * The required expiration index is created if missing to support efficient cleanup across SQL dialects.
* **Bug Fixes**
  * Expired-row cleanup is no longer limited to startup for a single store; it now applies consistently to shared SQL persistence.
* **Tests**
  * Added SQL persistence cleanup coverage for MySQL, PostgreSQL, and SQLite, including batch deletion behavior and expiration index verification.

*This PR was raised with the help of Codex ( GPT-5.6-Sol-high )*


packages/effect/src/unstable/persistence/Persistence.ts
```diff
@@ -538,7 +539,7 @@ export const layerBackingSql: Layer.Layer<
       `,
     mssql: () =>
       sql`
-        IF NOT EXISTS (SELECT * FROM sysobjects WHERE name=${table} AND xtype='U')
+        IF NOT EXISTS (SELECT * FROM sysobjects WHERE name=${"effect_persistence"} AND xtype='U')
```
hsyntax:
```suggestion
        IF NOT EXISTS (SELECT * FROM sysobjects WHERE name=${table} AND xtype='U')
```

PR #6570, opened 2026-07-24
Guard import.meta.env in ConfigProvider.fromEnv

### LINKED ISSUE #6358: ConfigProvider.fromEnv references import.meta.env in unsupported runtimes
### What happened?

Using Effect v4 (`effect@4.0.0-beta.64`) inside Convex fails during Convex module analysis when `Config` uses the default `ConfigProvider`.

Convex reports:

```txt
Failed to analyze http.js: Uncaught TypeError: import.meta unsupported
    at fromEnv (...)
```

The relevant default in `ConfigProvider.fromEnv()` appears to merge both `process.env` and `import.meta.env`:

```ts
const env = options?.env ?? {
  ...globalThis?.process?.env,
  ...import.meta?.env
}
```

Convex does not support `import.meta` in pushed modules, so the module fails while being analyzed/bundled before the app can start.

### Minimal shape

```ts
import { Config, Effect } from "effect"

Effect.runSync(
  Effect.gen(function* () {
    return yield* Config.url("SITE_URL")
  })
)
```

In Convex, this code is imported from the `convex/http.ts` graph, so Convex analyzes it during push.

### Workaround

Providing an explicit env-backed provider avoids the default `import.meta.env` lookup:

```ts
import { ConfigProvider, Effect } from "effect"

program.pipe(
  Effect.provideService(
    ConfigProvider.ConfigProvider,
    ConfigProvider.fromEnv({ env: process.env as Record<string, string> })
  )
)
```

### Question

Is there a recommended way for runtimes that do not support `import.meta` to opt out globally, or should the default provider avoid touching `import.meta` unless it is known to be supported?

xianjianlf2:
Fixes #6358

`ConfigProvider.fromEnv()` currently reads `import.meta.env` while building the default environment. Some runtimes can throw when `import.meta` is accessed, so this guards that lookup and falls back to `process.env` / an empty Vite env instead.

Tested:
- `pnpm --filter effect check`
- `pnpm --filter effect exec vitest run test/ConfigProvider.test.ts -t "fromEnv" --config vitest.config.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated `ConfigProvider.fromEnv()` to resolve its default environment from `process.env` (when available) instead of using any `import.meta.env` fallback.
* **Documentation**
  * Refreshed `fromEnv` guidance to match the new default behavior and clarify how to pass `env` in custom runtimes.
* **Tests**
  * Added tests covering default environment loading, restoration of `process.env`, and that an explicit `env` option overrides the default.
  * Added a check ensuring the implementation does not reference `import.meta`.

### DIFF
--- a/.changeset/configprovider-import-meta-env.md
+++ b/.changeset/configprovider-import-meta-env.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Remove the default `import.meta.env` lookup from `ConfigProvider.fromEnv`, fixing module analysis failures in runtimes that do not support `import.meta`, closes #6358.
--- a/packages/effect/CONFIG.md
+++ b/packages/effect/CONFIG.md
@@ -347,7 +347,7 @@ Effect.runSync(host) // "localhost"
 
 **How `_` splitting works**: env var names are split on `_` to build a tree. This means `DATABASE_HOST=localhost` is accessible at both `["DATABASE_HOST"]` (flat) and `["DATABASE", "HOST"]` (nested). Querying `["DATABASE"]` returns a Record node with child key `"HOST"`.
 
-Pass `{ env: { ... } }` for testing. Omit to use `process.env` (merged with `import.meta.env` when available).
+Pass `{ env: { ... } }` for testing. Omit to use `process.env` when available.
 
 ### `ConfigProvider.fromUnknown` — Plain JS Objects
 
--- a/packages/effect/src/ConfigProvider.ts
+++ b/packages/effect/src/ConfigProvider.ts
@@ -850,8 +850,8 @@ function emptyStringAsMissing(value: string | undefined, preserveEmptyStrings: b
  * purely numeric names, the node is reported as an `Array`; otherwise as a
  * `Record`.
  *
- * The default environment merges `process.env` and `import.meta.env` (when
- * available). Override by passing `{ env: { ... } }`.
+ * The default environment reads `process.env` when available. For runtimes that
+ * expose environment variables elsewhere, pass `{ env: { ... } }`.
  *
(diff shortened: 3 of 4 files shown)


tim-smart (changes requested):
This won't fix the issue as it fails during a module analysis step, not at runtime.

xianjianlf2:
Thanks, that makes sense. I updated the branch so `ConfigProvider` no longer touches `import.meta` at all, including during module analysis. The default path is now just `process.env` when it exists, and I also cleaned up the docs wording to match.

Local check: `pnpm --dir packages/effect test ConfigProvider.test.ts --run` passes.

xianjianlf2:
已推送 `b097ad8e`，同步最新 `main` 并保留不访问 `import.meta.env` 的修复；ConfigProvider 测试 99/99 通过，dprint 检查通过。@tim-smart 请复审。

xianjianlf2:
@codex review

xianjianlf2:
已同步最新 `main` 并解决 `ConfigProvider.ts` 的 rebase 冲突，保留 `process.env`-only 实现，避免 common module 访问 `import.meta.env`。新 head：`56e01fb7c`。

验证结果：
- `pnpm exec vitest run packages/effect/test/ConfigProvider.test.ts`：99/99 通过
- `pnpm --dir=packages/effect check`：通过
- `pnpm exec oxlint -f unix packages/effect/src/ConfigProvider.ts packages/effect/test/ConfigProvider.test.ts`：通过
- `pnpm exec dprint check packages/effect/src/ConfigProvider.ts packages/effect/test/ConfigProvider.test.ts packages/effect/CONFIG.md`：通过

请重新复审。

tim-smart:
I don't think we want to do this as it then breaks support for vite.

PR #6579, opened 2026-07-24
Scope cluster reply serialization failures and peer-delivered defects to their own request

marbemac:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

If a volatile entity reply can't be serialized, the failure currently takes down the whole runner-to-runner connection. `RunnerServer` dies with the `Reply.serialize` error, `RpcServer` broadcasts it as a connection-level `Defect` frame (killing every in-flight request on that connection), and `Runners.makeRpc` treats the delivered defect as `RunnerUnavailable`. `Sharding.sendOutgoing` then re-sends the same request id every `sendRetryInterval`, and since the entity already processed it, the caller gets `AlreadyProcessingMessage` for a request that actually succeeded — only the reply couldn't be encoded.

Each step is reasonable on its own; it's the combination that turns one unencodable reply into a connection-wide failure and a retry loop. This PR changes all three steps, so no two of them can recombine this way:

- `RunnerServer` now responds to the affected request with a defect reply (`ReplyWithContext.fromDefect`) when reply serialization fails, the same way request serialization failures are already handled in `Runners`. The `Stream` handler also ends its queue, since the fallback reply is terminal.
- `Runners.makeRpc` only converts transport failures (`RpcClientError`) to `RunnerUnavailable`. Defects delivered by the peer resolve the request with the real error instead of being re-sent. Persisted requests still map to `RunnerUnavailable`, because their reply can be recovered from storage.
- The runner protocol servers (`RunnerServer.layer`, `HttpRunner.toHttpEffect{,Websocket}`) set `disableFatalDefects: true`, so other handler defects also stay scoped to a single request. From what I can find nothing depends on connection-fatal defects on this channel (a `Defect` frame doesn't close the socket, and reconnection is driven by socket errors), and the entity level already works this way — but if connection-fatal defects are intentional here, this change can be dropped without affecting the other two.

Tests: `test/cluster/Runners.test.ts` covers both fixes at the unit level (also the first tests for the `RunnerUnavailable` / `AlreadyProcessingMessage` paths), `SocketRunner.test.ts` adds a two-runner repro over real sockets checking that a sibling in-flight request survives, and `RpcServer.test.ts` gets a `disableFatalDefects` isolation test.

## Related

Same direction as #2457, which did this for unknown request tags.

## Summary by CodeRabbit

* **Bug Fixes**
  * Isolated reply serialization failures to only the affected request, preventing shared runner/connection shutdowns.
  * Improved concurrent request resilience: sibling in-flight calls continue successfully even when one request fails.
  * Enhanced fault handling and recovery mappings for peer-delivered and transport-layer errors, especially for persisted requests.
  * Kept active streams running when unrelated requests encounter fatal defects or routing misses.
* **Tests**
  * Added/expanded coverage for reply serialization failure isolation and fatal-defect isolation using shared harnesses and new runner scenarios.

### DIFF
--- a/packages/effect/src/unstable/cluster/HttpRunner.ts
+++ b/packages/effect/src/unstable/cluster/HttpRunner.ts
@@ -150,7 +150,8 @@ export const toHttpEffect: Effect.Effect<
   const handlers = yield* Layer.build(RunnerServer.layerHandlers)
   return yield* RpcServer.toHttpEffect(Runners.Rpcs, {
     spanPrefix: "RunnerServer",
-    disableTracing: true
+    disableTracing: true,
+    disableFatalDefects: true
   }).pipe(Effect.provideContext(handlers))
 })
 
@@ -173,7 +174,8 @@ export const toHttpEffectWebsocket: Effect.Effect<
   const handlers = yield* Layer.build(RunnerServer.layerHandlers)
   return yield* RpcServer.toHttpEffectWebsocket(Runners.Rpcs, {
     spanPrefix: "RunnerServer",
-    disableTracing: true
+    disableTracing: true,
+    disableFatalDefects: true
   }).pipe(Effect.provideContext(handlers))
 })
 
--- a/packages/effect/src/unstable/cluster/RunnerServer.ts
+++ b/packages/effect/src/unstable/cluster/RunnerServer.ts
@@ -10,6 +10,7 @@
  *
  * @since 4.0.0
  */
+import type * as Cause from "../../Cause.ts"
 import * as Effect from "../../Effect.ts"
 import type * as Exit from "../../Exit.ts"
 import * as Fiber from "../../Fiber.ts"
@@ -17,6 +18,7 @@ import { constant } from "../../Function.ts"
 import * as Layer from "../../Layer.ts"
 import * as Option from "../../Option.ts"
 import * as Queue from "../../Queue.ts"
+import type * as Rpc from "../rpc/Rpc.ts"
 import * as RpcServer from "../rpc/RpcServer.ts"
 import type * as ClusterError from "./ClusterError.ts"
 import * as Message from "./Message.ts"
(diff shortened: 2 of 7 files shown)


packages/effect/src/unstable/cluster/HttpRunner.ts
```diff
@@ -150,7 +150,8 @@ export const toHttpEffect: Effect.Effect<
   const handlers = yield* Layer.build(RunnerServer.layerHandlers)
   return yield* RpcServer.toHttpEffect(Runners.Rpcs, {
     spanPrefix: "RunnerServer",
-    disableTracing: true
+    disableTracing: true,
+    disableFatalDefects: true
```
marbemac, line 154: Least sure about this one — the other two changes fix the issue without it, this just stops other handler defect classes from failing every in-flight request on the shared connection. Happy to drop it if connection-fatal defects are intentional on this channel (same with the other 2 flips of `disableFatalDefects` to true).

packages/effect/src/unstable/cluster/RunnerServer.ts
```diff
@@ -30,6 +32,30 @@ import { ShardingConfig } from "./ShardingConfig.ts"
 
 const constVoid = constant(Effect.void)
 
+// The original reply never left this runner, so its id can be reused for the
```
tim-smart: Remove these comments as they aren't really helping

packages/effect/src/unstable/cluster/RunnerServer.ts
```diff
@@ -168,7 +205,10 @@ export const layer: Layer.Layer<
   RpcServer.Protocol | Sharding.Sharding | MessageStorage.MessageStorage
 > = RpcServer.layer(Runners.Rpcs, {
   spanPrefix: "RunnerServer",
-  disableTracing: true
+  disableTracing: true,
+  // a handler defect concerns a single request, so it must not tear down
+  // every request multiplexed on the runner connection
+  disableFatalDefects: true
```
tim-smart, line 204:
```suggestion
  disableTracing: true,
  disableFatalDefects: true
```

tim-smart (changes requested):

packages/effect/src/unstable/cluster/RunnerServer.ts
marbemac: sure thing, done across the board

tim-smart (approved):

PR #6593, opened 2026-07-24
Fix mergeAll outer failure handling

tim-smart:
## Summary

- preserve permit-before-pull ordering for normal `Channel.mergeAll` operation
- at a saturated concurrency window, run the outer pull in a child fiber and race permit acquisition only against an observer of that fiber; the pull itself is never aborted
- propagate real outer failures immediately through the output queue, while preserving wait-on-normal-`Done` behavior; normal driver teardown interrupts and awaits tracked children
- cover concurrent `Stream.flatMap` at the exact concurrency limit, successful saturated pulls, permit arrival during an in-flight pull, slow finite inners, and the corresponding `Stream.switchMap` failure cases
- include an `effect` patch changeset

## Root cause

The merge driver gated every outer termination cause on `doneLatch`, which is correct only for normal completion. It also could not observe a pending outer failure while every permit was held by never-ending inner channels. Together these behaviors left the stream hung indefinitely.

## Implementation

Normal outer pulls remain guarded by a semaphore permit. When the concurrency window is already full, the driver forks the next pull as a child and races permit acquisition against a `Fiber.await` observer. An outer failure therefore preempts the permit wait. If the permit wins while the pull is still running, only the observer is interrupted; the pull fiber continues and the driver joins its cached exit afterward. This prevents both successful-probe deadlocks and lost elements or failures from an aborted stateful pull.

On a real outer failure, the driver fails the output queue immediately. The queue's first terminal cause is retained while the driver's interrupt-children middleware interrupts and awaits the tracked child fibers during normal teardown, so inner finalizers complete without manual fiber-set mutation.

## Validation

- `pnpm --dir packages/effect test --run` (7,072 passed, 4 skipped)
- the six focused regressions pass
- the saturated outer-failure/cause/finalizer regression passed in 25 repeated runs
- `pnpm lint-fix`

Closes EFF-25

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent stream/Channel merging so outer failures propagate promptly.
  * When an outer failure occurs, active inner streams are interrupted and non-“done” failures fail the result immediately.
  * Reduced upstream pull latency under saturated concurrency by prefetching pulls when possible.
* **Tests**
  * Added coverage to verify prompt failure propagation, correct finalization behavior, permit release, and interruption timing for `flatMap` and `switchMap` under concurrency edge cases.

### DIFF
--- a/packages/effect/src/Channel.ts
+++ b/packages/effect/src/Channel.ts
@@ -6030,32 +6030,27 @@ export const mergeAll: {
         yield* Scope.addFinalizer(forkedScope, Queue.shutdown(queue))
 
         const pull = yield* toTransform(channels)(upstream, scope)
-        let pendingChannel: Pull.Success<typeof pull> | undefined
 
         yield* Effect.gen(function*() {
           while (true) {
+            let pullFiber: Fiber.Fiber<Pull.Success<typeof pull>, any> | undefined
             if (semaphore) {
-              const takePermit = semaphore.take(1)
               if (fibers.size < concurrencyN) {
-                yield* takePermit
+                yield* semaphore.take(1)
               } else {
+                pullFiber = yield* Effect.forkChild(pull)
                 yield* Effect.raceFirst(
-                  takePermit,
-                  Effect.uninterruptibleMask((restore) =>
-                    restore(pull).pipe(
-                      Effect.flatMap((channel) =>
-                        Effect.sync(() => {
-                          pendingChannel = channel
-                        })
-                      ),
-                      Effect.andThen(Effect.never)
-                    )
+                  semaphore.take(1),
+                  Effect.flatMap(
+                    Fiber.await(pullFiber),
+                    (exit) => Exit.isSuccess(exit) ? Effect.never : exit
                   )
                 )
               }
             }
-            const channel = pendingChannel ?? (yield* pull)
-            pendingChannel = undefined
+            const channel = pullFiber === undefined
+              ? yield* pull
+              : yield* Effect.flatten(Fiber.await(pullFiber))
             const childScope = Scope.forkUnsafe(forkedScope)
             const childPull = yield* toTransform(channel)(upstream, childScope)
 
--- a/packages/effect/src/Channel.ts
+++ b/packages/effect/src/Channel.ts
@@ -6041,16 +6041,13 @@ export const mergeAll: {
                 pullFiber = yield* Effect.forkChild(pull)
                 yield* Effect.raceFirst(
                   semaphore.take(1),
-                  Effect.flatMap(
-                    Fiber.await(pullFiber),
-                    (exit) => Exit.isSuccess(exit) ? Effect.never : exit
-                  )
+                  Effect.andThen(Fiber.join(pullFiber), Effect.never)
                 )
               }
             }
             const channel = pullFiber === undefined
               ? yield* pull
-              : yield* Effect.flatten(Fiber.await(pullFiber))
+              : yield* Fiber.join(pullFiber)
             const childScope = Scope.forkUnsafe(forkedScope)
             const childPull = yield* toTransform(channel)(upstream, childScope)
 
@@ -6092,7 +6089,6 @@ export const mergeAll: {
             }
             const inFlight = Arr.fromIterable(fibers)
             fibers.clear()
-            doneLatch.openUnsafe()
             return Effect.uninterruptible(Effect.withFiber((parent) => {
               // Signal every child before publishing the failure. The driver's
               // interrupt-children middleware awaits them when the driver exits.
--- a/packages/effect/src/Channel.ts
+++ b/packages/effect/src/Channel.ts
@@ -6087,16 +6087,7 @@ export const mergeAll: {
             if (Result.isSuccess(halt)) {
               return doneLatch.whenOpen(Queue.failCause(queue, cause))
             }
-            const inFlight = Arr.fromIterable(fibers)
-            fibers.clear()
-            return Effect.uninterruptible(Effect.withFiber((parent) => {
-              // Signal every child before publishing the failure. The driver's
-              // interrupt-children middleware awaits them when the driver exits.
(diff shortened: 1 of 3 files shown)


packages/effect/src/Channel.ts
```diff
@@ -6033,8 +6033,8 @@ export const mergeAll: {
 
         yield* Effect.gen(function*() {
           while (true) {
-            if (semaphore) yield* semaphore.take(1)
             const channel = yield* pull
+            if (semaphore) yield* semaphore.take(1)
```
tim-smart: The semaphore is supposed to guard the pull, so it should take a permit before pulling.

tim-smart: Addressed in f2ea73717. The normal path now takes the semaphore permit before pulling. At a saturated window, permit acquisition is raced with an interruptible outer pull so a failure/completion can preempt the wait; if that pull produces a channel first, it is retained atomically and used once the permit arrives, so no element is dropped. This keeps speculative evaluation confined to the otherwise-deadlocked saturated case. The exact two-never-ending-inner repro still fails promptly with both finalizers run (`finalized === 2`), and the full Effect suite passes (7,070 passed, 4 skipped).

tim-smart: Revised in 2390d19a9 after the blocking re-review. The saturated pull now runs in a child fiber that is never a race loser: `raceFirst` observes it through `Fiber.await`, so a permit win interrupts only the observer, and the driver then joins the original pull fiber's cached exit. This removes the uninterruptible-park deadlock and ensures an in-flight stateful pull cannot lose either an element or a failure. I also added both deterministic regressions from re-review (successful saturated pull followed by permit release, and permit arrival mid-pull); both pass in 5 repeated runs. The full Effect suite passes with 7,072 tests and 4 skips, plus package typecheck and lint-fix.

packages/effect/src/Channel.ts
```diff
@@ -6033,8 +6033,21 @@ export const mergeAll: {
 
         yield* Effect.gen(function*() {
           while (true) {
-            if (semaphore) yield* semaphore.take(1)
-            const channel = yield* pull
+            let pullFiber: Fiber.Fiber<Pull.Success<typeof pull>, any> | undefined
```
tim-smart, line 6036: I think the `pullFiber` code can be reverted now?

tim-smart: Tested this empirically by locally reverting `pullFiber` to the original permit-before-pull shape and running `interrupts all inner streams when the outer fails at the concurrency limit`. It timed out at 5,000 ms. With both permits held by never-ending inners, the driver remains parked in `semaphore.take(1)`, so it never pulls the pending outer failure and the teardown-only failure handler cannot run. I restored `pullFiber`; the saturated pull observer is still required.

packages/effect/test/Stream.test.ts
```diff
@@ -1031,6 +1031,133 @@ describe("Stream", () => {
...
+          (n) => n === 3 ? Stream.make(3) : Stream.fromEffect(Deferred.await(innerGate).pipe(Effect.as(n))),
+          { concurrency: 2 }
+        ).pipe(
+          Stream.runCollect,
+          Effect.forkChild
+        )
+        yield* Deferred.await(probing)
+        yield* Deferred.succeed(innerGate, void 0)
+        for (let i = 0; i < 10; i++) yield* Effect.yieldNow
+        yield* Deferred.succeed(gate, void 0)
```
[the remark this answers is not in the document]
tim-smart: Fixed in 489d7a354. The test now uses an `innerCompleted` Deferred resolved by the gated inners’ `ensuring` finalizer and awaits that observed completion before opening the blocked outer pull. The revised regression passed 20/20 repeated runs, all six focused regressions, and the full effect suite.

tim-smart:
**Final review pass at `851dfa6c6` (teardown-only failure path): Approved.**

(Posted here because the reviewer agent's Multica session expired mid-run; this is the review-of-record for EFF-25.)

**Cause-clobbering: the queue-state gate holds, by construction.** `Queue.failCauseUnsafe` is a strict no-op returning `false` when the queue state is not `Open` (`packages/effect/src/Queue.ts:924-927`) — first failure wins, nothing overwrites. The ordering makes the outer cause always first: the `catchCause` handler's `Queue.failCause(queue, cause)` completes inside the driver's run loop, the driver then exits, and only at exit does the interrupt-children middleware interrupt the inners. So every interruption-driven trip through the child `onError` (`packages/effect/src/Channel.ts:6076`) hits an already-failed queue and is a guaranteed no-op. The `fibers.has` early-return now serves only its original purpose (the switch-mode eviction path, which still deletes before interrupting); middleware-interrupted children instead run their full cleanup — scope close, semaphore release, latch open — all harmless post-mortem. The only interleaving that can still race the outer cause is a child failing *naturally* with a real error at the same moment, where first-wins between two genuine causes is the same pre-existing, acceptable behavior as every earlier shape.

**Promptness: unchanged.** The queue is failed in the handler itself, before any child interruption or finalizer runs — downstream unblocks exactly as early as the manual-interrupt version did. Finalizer ordering is preserved: downstream teardown awaits the driver via the `forkIn` scope finalizer, and the driver only settles after the middleware has interrupted *and awaited* the children — which is why the saturated repro still deterministically observes `finalized === 2` before the stream exit settles.

Empirical verification at this head: Stream suite 296 passed; all six focused regressions passed 10/10 (outer-failure set) and 3/3 (saturated-pull set) stress runs; full `packages/effect` suite 7,066 passed with only the 6 pre-existing environment failures baselined throughout this review (`EffectKeepAlive`, `unstable/cli/Command` snapshots, `OtlpMetrics`); package `tsc` clean. `8dcd25f35` applied the three simplifications from the simplicity pass exactly as verified. CI snapshot at review time: 15/16 checks green, only CodeRabbit pending.

The failure path is now a single `Queue.failCause` — the simplest correct form this fix has had. Review complete.

PR #6618, opened 2026-07-25
Hash SQL message deduplication keys to prevent message_id overflow

### LINKED ISSUE #6317: @effect/cluster: cluster_messages.message_id VARCHAR(255) overflows for DurableQueue deferred envelopes — DurableDeferred.done fails to persist and the workflow never resumes
## Summary

`SqlMessageStorage` creates `cluster_messages.message_id` as `VARCHAR(255)`, but the dedupe key stored there is `Envelope.primaryKeyByAddress` — `${entityType}/${entityId}/${tag}/${primaryKey}`. For the `deferred` envelope sent by `DurableDeferred.done`, that composes to:

```
Workflow/<workflowName>/<executionId>/deferred/DurableQueue/<queueName>/<sha256-hex(64)>
```

i.e. ~100+ characters of fixed overhead around the executionId. Once the executionId passes roughly 110–130 characters, every attempt to persist the deferred-done envelope fails with:

```
PersistenceError: value too long for type character varying(255)
```

`DurableDeferred.done` dies after its retries and the awaiting workflow stays `Suspended` forever.

Long executionIds are easy to hit legitimately: a workflow that runs child workflows must give them **deterministic** executionIds derived from its own (e.g. `` `${parentExecutionId}-item-${i}` ``), otherwise a replay after a durable suspend spawns fresh children on every pass. Two or three levels of that (import → per-item population → per-item queue job) blows past the budget.

## Environment

- `@effect/cluster` 0.59.0
- `@effect/workflow` 0.18.2
- `effect` 3.21.4, `@effect/sql-pg` 0.52.1 (PostgreSQL dialect)
- `SingleRunner.layer({ runnerStorage: "sql" })`

## Root cause

`SqlMessageStorage` (pg dialect) — `dist/esm/SqlMessageStorage.js`:

```sql
CREATE TABLE IF NOT EXISTS cluster_messages (
  id BIGINT PRIMARY KEY,
  rowid BIGSERIAL,
  message_id VARCHAR(255),
  ...
  UNIQUE (message_id)
)
```

`Envelope.js` already anticipates the problem:

```ts
export const primaryKeyByAddress = (options) =>
  // hash the entity address to save space?
  `${options.address.entityType}/${options.address.entityId}/${options.tag}/${options.id}`
```

For a DurableQueue job the deferred's primaryKey (`DeferredRpc`'s `primaryKey: ({ name }) => name` in `ClusterWorkflowEngine.js`) is the deferred name built in `DurableQueue.process`:

```ts
const queueName = `DurableQueue/${self.name}`
const id = yield* Activity.idempotencyKey(`${queueName}/${self.idempotencyKey(payload)}`) // sha-256 hex, 64 chars
const deferred = DurableDeferred.make(`${self.deferred.name}/${id}`, ...) // self.deferred.name === `DurableQueue/${self.name}`
```

So the full `message_id` is:

```
Workflow/<workflowName>/<executionId>/deferred/DurableQueue/<queueName>/<64 hex chars>
```

With e.g. a 20-char workflow name and a 15-char queue name that is ~133 chars before the executionId — an executionId over ~122 chars overflows the column, and the `UNIQUE (message_id)` insert fails on every retry.

## Reproduction sketch

```ts
const JobQueue = DurableQueue.make({
  name: "job-queue",
  payload: Schema.Struct({ id: Schema.String }),
  idempotencyKey: ({ id }) => id,
  success: Schema.Void,
})

// slow enough that the workflow durably suspends before the job finishes
const JobWorker = DurableQueue.worker(JobQueue, () => Effect.sleep("1 second"))

const AwaitJob = Workflow.make({
  name: "AwaitJob",
  payload: Schema.Struct({ id: Schema.String }),
  idempotencyKey: ({ id }) => id,
})
const AwaitJobLayer = AwaitJob.toLayer(({ id }) => DurableQueue.process(JobQueue, { id }))

// execute with a long deterministic executionId, as a parent workflow
// deriving child ids would:
const engine = yield* WorkflowEngine
const result = yield* engine.execute(AwaitJob, {
  executionId: `${"parent-execution-id-".repeat(7)}item-0`, // > ~130 chars
  payload: { id: "1" },
})
```

The workflow suspends, the worker completes, `DurableDeferred.done` → `client(executionId).deferred(...)` → the persist of the deferred envelope fails with `value too long for type character varying(255)`, and `engine.poll` returns `Suspended` forever.

(Note `entity_id` is also `VARCHAR(255)`, so executionIds have a hard cap regardless — but `message_id` overflows far earlier because of the composed prefix/suffix.)

## Suggested fix

Either of:

- hash the composed key (as the `Envelope.ts` comment suggests) — it is only used for dedupe via the `UNIQUE` constraint, so a digest works; or
- create `message_id` as `TEXT` on PostgreSQL.

## Workaround we're using

Widening the column right after cluster's migration runs:

```ts
const widenClusterMessageIdColumn = Effect.flatMap(
  PgClient.PgClient,
  (sql) => sql`ALTER TABLE cluster_messages ALTER COLUMN message_id TYPE text`,
).pipe(Effect.orDie, Effect.asVoid)

export const WorkflowEngineLive = ClusterWorkflowEngine.layer.pipe(
  Layer.provide(
    SingleRunner.layer({ runnerStorage: "sql" }).pipe(
      Layer.tap(() => widenClusterMessageIdColumn),
    ),
  ),
)
```

which fully resolves the failures in our e2e suite.

tim-smart:
## Summary

The composed request deduplication key (`entityType/entityId/tag/primaryKey`) can legally exceed the 255-character `message_id` column: `entity_type VARCHAR(150)` + `entity_id VARCHAR(255)` + `tag VARCHAR(50)` alone allow 458 characters of address before the RPC primary key is appended. Any sufficiently long combination overflows the column (or silently violates its purpose via `INSERT IGNORE`), surfacing as `PersistenceError`s from deep inside SQL storage — the failure reported in #6317 via long workflow deferred keys.

`SqlMessageStorage` now conditionally hashes the composed key at the SQL boundary:

- Keys that fit in 255 characters are stored as plaintext, **byte-compatible with rows written by previous versions** — existing deployments keep deduplicating with no transition handling at all.
- Longer keys are stored as a SHA-256 digest (256 bits, 64 hex chars via `Encoding.encodeHex`) — fixed-size, fits every dialect's indexable column width, collision probability negligible (birthday bound 2^128). Digests never contain `/` while composed keys always do, so the two encodings cannot collide.
- `Envelope.primaryKeyByAddress` stays synchronous and plaintext; memory storage and the no-op storage are unchanged.
- **No schema change**: no new columns, no migration.

### Legacy rows with over-length keys (SQLite only)

On the width-enforcing dialects (pg/mysql/mssql), keys longer than 255 characters could never be stored, so no legacy handling is needed. SQLite's `message_id` is `TEXT`, so pre-upgrade SQLite deployments can hold plaintext keys longer than 255 characters; for exactly that case (long key, non-width-enforcing dialect), save-time deduplication and `requestIdForPrimaryKey` fall back to a plaintext read when the digest misses.

### Layer requirements

- `SqlMessageStorage.make`/`layer`/`layerWith` now require `Crypto.Crypto`; digest failures are mapped into `PersistenceError` like every other storage failure.
- `SingleRunner.layer` propagates the requirement (`SqlClient | Crypto.Crypto`).
- `NodeClusterHttp`/`NodeClusterSocket`/`BunClusterHttp`/`BunClusterSocket` provide `NodeCrypto.layer`/`BunCrypto.layer` internally, so their public requirements are unchanged.
- `ClusterWorkflowEngine.layer` is untouched.

### Tests

Added to the pg/mysql/sqlite matrix in `SqlMessageStorage.test.ts`:

- long composed keys (450-char primary key) save, deduplicate, and resolve via `requestIdForPrimaryKey`; `message_id` is verified to be a 64-char hex digest
- keys within the column width are verified to be stored as byte-identical plaintext (backwards compatibility), and deduplicate/resolve
- SQLite-specific regression: legacy plaintext rows with keys > 255 chars (possible only on SQLite's `TEXT` column) still deduplicate via the fallback

All 36 tests pass across the three dialects; memory storage tests are unchanged and pass.

Closes #6317
Closes EFF-48

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL message deduplication for long keys by using compact hashed identifiers when necessary.
  * Preserved compatibility with existing plaintext deduplication keys and legacy stored records.
  * Ensured duplicate requests continue to resolve to their original request IDs.

* **Developer Experience**
  * SQL-backed cluster storage now automatically provides the required cryptographic support in Node.js and Bun environments.

* **Tests**
  * Added coverage for long, Unicode, compatible, and legacy deduplication keys.

### DIFF
--- a/packages/effect/src/unstable/cluster/SqlMessageStorage.ts
+++ b/packages/effect/src/unstable/cluster/SqlMessageStorage.ts
@@ -113,9 +113,21 @@ export const make: (options?: {
     })
 
   // Rows written before `message_id` was hashed store the plaintext composed
-  // key. Keys longer than the column's 255-character limit can never exist as
-  // legacy rows, so the fallback read is skipped for them.
-  const mayHaveLegacyRow = (primaryKey: string): boolean => primaryKey.length <= 255
+  // key; the fallback reads below exist only for that transition and can be
+  // removed in a future release. On the width-enforcing dialects keys longer
+  // than the column's 255-character limit can never exist as legacy rows, so
+  // the fallback read is skipped for them — but sqlite declares `message_id`
+  // as TEXT and does not enforce VARCHAR widths, so legacy rows there can
+  // hold keys of any length. (MySQL's `INSERT IGNORE` truncated over-long
+  // legacy keys to a 255-character prefix; those rows were already corrupt
+  // for deduplication purposes pre-digest and are deliberately not matched.)
+  const messageIdEnforcesWidth = sql.onDialectOrElse({
+    mssql: () => true,
+    mysql: () => true,
+    pg: () => true,
+    orElse: () => false
+  })
+  const mayHaveLegacyRow = (primaryKey: string): boolean => !messageIdEnforcesWidth || primaryKey.length <= 255
 
   const envelopeToRow = (
     envelope: Envelope.Encoded,
--- a/packages/effect/src/unstable/cluster/SqlMessageStorage.ts
+++ b/packages/effect/src/unstable/cluster/SqlMessageStorage.ts
@@ -10,8 +10,8 @@
  *
  * Request deduplication keys are hashed with the `Crypto` service before they
  * are written to the unique `message_id` column, so composed keys of any
- * length are supported; the plaintext key is stored in the non-unique
- * `primary_key` column for diagnostics.
+ * length are supported. The key's components remain readable via the
+ * `entity_type`, `entity_id`, `tag`, and `payload` columns.
  *
  * @since 4.0.0
  */
@@ -101,8 +101,8 @@ export const make: (options?: {
   // and the 64-character hex encoding fits every dialect's indexable column
   // width. The digest never contains "/" while legacy plaintext keys always
   // do, so the two encodings cannot collide and legacy rows can be matched
-  // with a dual-read fallback. The plaintext is kept in the non-unique
-  // `primary_key` column for diagnostics.
+  // with a dual-read fallback. The key's components remain readable via the
+  // entity_type, entity_id, tag, and payload columns.
   const encoder = new TextEncoder()
   const hashPrimaryKey = (primaryKey: string): Effect.Effect<string, PlatformError.PlatformError> =>
     Effect.map(crypto.digest("SHA-256", encoder.encode(primaryKey)), Encoding.encodeHex)
@@ -127,7 +127,6 @@ export const make: (options?: {
   const envelopeToRow = (
     envelope: Envelope.Encoded,
     message_id: string | null,
-    primary_key: string | null,
     deliver_at: number | null
   ): MessageRow => {
     switch (envelope._tag) {
@@ -135,7 +134,6 @@ export const make: (options?: {
         return {
           id: envelope.requestId,
           message_id,
-          primary_key,
           shard_id: ShardId.toString(envelope.address.shardId),
           entity_type: envelope.address.entityType,
           entity_id: envelope.address.entityId,
@@ -160,7 +158,6 @@ export const make: (options?: {
         return {
           id: envelope.id,
           message_id,
-          primary_key,
           shard_id: ShardId.toString(envelope.address.shardId),
           entity_type: envelope.address.entityType,
           entity_id: envelope.address.entityId,
@@ -179,7 +176,6 @@ export const make: (options?: {
         return {
           id: envelope.id,
           message_id,
-          primary_key,
           shard_id: ShardId.toString(envelope.address.shardId),
           entity_type: envelope.address.entityType,
           entity_id: envelope.address.entityId,
@@ -446,7 +442,7 @@ export const make: (options?: {
         let insert: Effect.Effect<ReadonlyArray<Row>, SqlError | PlatformError.PlatformError>
         if (primaryKey !== null) {
           insert = Effect.flatMap(hashPrimaryKey(primaryKey), (messageId) => {
-            const row = envelopeToRow(envelope, messageId, primaryKey, deliverAt)
+            const row = envelopeToRow(envelope, messageId, deliverAt)
             if (!mayHaveLegacyRow(primaryKey)) {
               return insertEnvelope(row, messageId)
             }
@@ -456,7 +452,7 @@ export const make: (options?: {
             )
           })
         } else {
-          const row = envelopeToRow(envelope, null, null, deliverAt)
+          const row = envelopeToRow(envelope, null, deliverAt)
           insert = Effect.as(sql`INSERT INTO ${messagesTableSql} ${sql.insert(row)}`.unprepared, [])
           if (envelope._tag === "AckChunk") {
             insert = sql`UPDATE ${repliesTableSql} SET acked = ${sqlTrue} WHERE id = ${envelope.replyId}`.pipe(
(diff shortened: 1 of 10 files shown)


packages/effect/src/unstable/cluster/SqlMessageStorage.ts
```diff
@@ -84,16 +92,43 @@ export const make: (options?: {
...
+  // `message_id` therefore stores a SHA-256 digest of the composed key: 256
+  // bits keeps the collision probability negligible (birthday bound 2^128),
+  // and the 64-character hex encoding fits every dialect's indexable column
+  // width. The digest never contains "/" while legacy plaintext keys always
+  // do, so the two encodings cannot collide and legacy rows can be matched
+  // with a dual-read fallback. The plaintext is kept in the non-unique
+  // `primary_key` column for diagnostics.
+  const encoder = new TextEncoder()
+  const hashPrimaryKey = (primaryKey: string): Effect.Effect<string, PlatformError.PlatformError> =>
+    Effect.map(crypto.digest("SHA-256", encoder.encode(primaryKey)), (bytes) => {
```
tim-smart: You could use the Encoding module I think here

tim-smart: Done in 9ad793bdd — switched to `Encoding.encodeHex`.

packages/effect/src/unstable/cluster/SqlMessageStorage.ts
```diff
@@ -84,6 +93,37 @@ export const make: (options?: {
...
+  // and the 64-character hex encoding fits every dialect's indexable column
+  // width. The digest never contains "/" while legacy plaintext keys always
+  // do, so the two encodings cannot collide and legacy rows can be matched
+  // with a dual-read fallback. The key's components remain readable via the
+  // entity_type, entity_id, tag, and payload columns.
+  const encoder = new TextEncoder()
+  const hashPrimaryKey = (primaryKey: string): Effect.Effect<string, PlatformError.PlatformError> =>
+    Effect.map(crypto.digest("SHA-256", encoder.encode(primaryKey)), Encoding.encodeHex)
+
+  // Rows written before `message_id` was hashed store the plaintext composed
```
tim-smart: Remove this comment

packages/effect/src/unstable/cluster/SqlMessageStorage.ts
```diff
@@ -84,6 +93,37 @@ export const make: (options?: {
...
+  // tag(50) alone total 458 characters before the RPC primary key is appended.
+  // `message_id` therefore stores a SHA-256 digest of the composed key: 256
+  // bits keeps the collision probability negligible (birthday bound 2^128),
+  // and the 64-character hex encoding fits every dialect's indexable column
+  // width. The digest never contains "/" while legacy plaintext keys always
+  // do, so the two encodings cannot collide and legacy rows can be matched
+  // with a dual-read fallback. The key's components remain readable via the
+  // entity_type, entity_id, tag, and payload columns.
+  const encoder = new TextEncoder()
+  const hashPrimaryKey = (primaryKey: string): Effect.Effect<string, PlatformError.PlatformError> =>
```
tim-smart: Maybe we could only hash the primary key if it exceeds 255 chars, to keep compatibility with previous versions

tim-smart: Done in a1f96bca — `message_id` now stays the plaintext composed key when it fits in 255 chars (byte-compatible with previous versions, no fallback reads needed), and only over-length keys get the SHA-256 digest. The plaintext fallback is kept for one narrow case: sqlite's TEXT column stored over-length plaintext keys pre-upgrade, so long-key lookups there still dual-read.

packages/effect/src/unstable/cluster/SqlMessageStorage.ts
tim-smart: Removed in a1f96bca (the block shrank to a one-liner explaining the remaining sqlite-only fallback).

PR #6619, opened 2026-07-25
fix(cli): fail when excess positional arguments are provided

### LINKED ISSUE #6594: unstable/cli: excess positional arguments are silently discarded instead of failing
**Version:** effect 4.0.0-beta.101

## What happens

The parser silently drops positional arguments that no declared param consumes, and no combinator can make it fail:

```ts
import { Command, Argument } from "effect/unstable/cli"

// Command with no arguments:
const init = Command.make("init", {})
// `mycli init extra` succeeds — "extra" is silently ignored

// Command with a single positional:
const run = Command.make("run", { prompt: Argument.string("prompt") })
// `mycli run "hi" oops` succeeds with prompt = "hi" — "oops" is silently ignored

// Cardinality combinators don't help:
Argument.string("prompt").pipe(Argument.between(1, 1)) // or atMost(1)
// max only stops consuming; the extra positional is still silently dropped
```

## Cause

- `internal/command.js` `parseParams` threads `currentArguments` through each declared param and discards whatever remains after the loop — there is no "unexpected argument" check.
- `Param.js` `parsePositionalVariadic` enforces `min` with an error but treats `max` purely as a stop-consuming bound.

## Why this looks like a bug

- Asymmetric with flags: unknown flags raise `UnrecognizedOption`, and `parseOptionVariadic` fails with `InvalidValue` when occurrences exceed `max` — positionals do neither.
- Regression from v3: stable `@effect/cli` documented a `ValidationError` for excess arguments.

## Suggested fix

Fail the parse when positionals remain unconsumed after `parseParams`, and/or make positional `max` reject extras the way flag `max` does.

wasim-builds:
Fixes #6594

This throws a ValidationError when there are unconsumed positional arguments after parsing all parameters. Previously, excess positionals were silently discarded.

## Summary by CodeRabbit

* **New Features**
  * Added an `UnrecognizedArgument` CLI error type for invalid command-line input.
* **Bug Fixes**
  * Commands now error when extra positional arguments remain after parsing.
  * Error messages include the unexpected tokens (e.g., `Excess arguments: ...`).
* **Tests**
  * Added coverage for fixed vs. variadic positional arguments, ensuring excess operands are handled correctly.

### DIFF
--- a/packages/effect/src/unstable/cli/CliError.ts
+++ b/packages/effect/src/unstable/cli/CliError.ts
@@ -93,7 +93,7 @@ export type CliError =
   | UnknownSubcommand
   | ShowHelp
   | UserError
-  | ValidationError
+  | UnrecognizedArgument
 
 /**
  * Error thrown when an unrecognized option is encountered.
@@ -491,14 +491,14 @@ export class UserError extends Schema.TaggedErrorClass<UserError>(
 }
 
 /**
- * Error thrown when a validation fails (e.g. excess positional arguments).
+ * Error thrown when an unrecognized argument is encountered.
  *
  * @category models
  * @since 4.0.0
  */
-export class ValidationError extends Schema.TaggedErrorClass<ValidationError>(
-  `${TypeId}/ValidationError`
-)("ValidationError", {
+export class UnrecognizedArgument extends Schema.TaggedErrorClass<UnrecognizedArgument>(
+  `${TypeId}/UnrecognizedArgument`
+)("UnrecognizedArgument", {
   error: Schema.String
 }) {
   /**
@@ -537,6 +537,7 @@ export const NonShowHelpErrors: Schema.Union<
     typeof MissingArgument,
     typeof InvalidValue,
     typeof UnknownSubcommand,
+    typeof UnrecognizedArgument,
     typeof UserError
   ]
 > = Schema.Union([
@@ -546,6 +547,7 @@ export const NonShowHelpErrors: Schema.Union<
   MissingArgument,
   InvalidValue,
   UnknownSubcommand,
+  UnrecognizedArgument,
   UserError
 ])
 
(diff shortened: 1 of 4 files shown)


packages/effect/src/unstable/cli/internal/command.ts
```diff
@@ -291,6 +291,12 @@ const parseParams: (parsedArgs: Param.ParsedArgs, params: ReadonlyArray<Param.An
     currentArguments = remainingArguments
   }
 
+  if (currentArguments.length > 0) {
+    return yield* new CliError.ValidationError({
```
IMax153: I think instead we should introduce an `UnrecognizedArguments` error to mirror `UnrecognizedOption`.

packages/effect/src/unstable/cli/CliError.ts
IMax153, line 532: You'll also need to add the error to this schema once renamed.

IMax153 (commented):
This PR will also need a changeset - you can generate one with `pnpm changeset`. It can just be a `patch`.

IMax153 (changes requested):

IMax153:
@wasim-builds thank you! We also need a series of regression tests as described by the CodeRabbit review.

IMax153:
@wasim-builds - I apologize, I realized there was a pre-existing PR here: https://github.com/Effect-TS/effect/pull/6561 to fix the issue.

I'm going to give precedence to the PR that was opened earlier - plus it also addresses a few more edge cases. 

Please let me know if it addresses your issues, and if not feel free to re-open a PR / issue.

IMax153:
> Is there a process or pathway for becoming a member of the Effect-TS GitHub organization? I'd love to continue helping out where I can!

That's wonderful! I'm glad to hear the Effect codebase has been a source of inspiration :) We always welcome contributions from the community! The organization access is intentionally quite limited for security reasons, but always happy to review PRs and collaborate, either here or in Discord!

PR #6625, opened 2026-07-25
feat: add MCP protocol adapters

### LINKED ISSUE #6617: Multi-protocol McpServer for supporting upcoming 2026-07-28 spec
This stems from a [discord thread](https://discord.com/channels/795981131316985866/1530221469056110674) about the state of the `v2026-07-28` spec for `McpServer`

For context, currently the `McpServer` is one spec behind on `v2025-06-18` and there is another issue for adding the latest `v2025-11-25` (#6157).  However the current RC for `v2026-07-28` throws a bit of a spanner in this as it will no longer be a progressive enhancement on top of the existing features but instead a bit of architectural shift toward a stateless protocol:

- the initialization handshake and protocol-level session are removed;
- protocol version and client capabilities are provided per request;
- `server/discover` advertises supported versions and server capabilities;
- some existing features are deprecated or moved into extensions;
- server-to-client interactions and HTTP behavior change.

This makes updating one shared set of MCP schemas increasingly difficult. Supporting one version by widening or replacing the current schemas risks changing the behavior expected by clients using another version.

Related work:

- #6157 — support MCP protocol version `2025-11-25`
- #6396 — add `McpClient`, including `2025-11-25` server groundwork
- #6610 — Streamable HTTP transport compliance
- #6353 — MCP Apps / interactive UI suppor

## Goal/Scope

Allow one `McpServer` endpoint to support multiple MCP protocol versions without combining incompatible versions into one permissive schema.

The server should declare the protocols it supports, then when a request arrives, the protocol version used by that client or request should determine:

- how the request is interpreted
- which lifecycle and capabilities apply
- which functionality is available
- the shape of the response returned to the client

Clients using different protocol versions should therefore be able to communicate with the same endpoint while each observes the schema and behavior defined by its selected version.

The protocol version must resolve the server to a specific wire shape and set of behaviors, rather than being metadata attached after the request has already been interpreted.

Initial built-in support would start at `2025-06-18`. Older revisions can be considered separately if they are still needed and we can leave out the extensions for the time being until we have a clear direction.

## Proposal

Allow an `McpServer` to declare the protocol versions it supports:

```ts
const Server = McpServer.layerHttp({
  name: "Acme MCP",
  version: "1.0.0",
  path: "/mcp",
  protocols: [
    McpProtocol.v2026_07_28,
    McpProtocol.v2025_11_25,
    McpProtocol.v2025_06_18
  ]
})
```

At a high level, each supported version would provide the protocol-specific interpretation of the server:

```mermaid
sequenceDiagram
  participant Client as McpClient
  participant Ingress as McpServer ingress
  participant Protocol as Selected protocol behavior
  participant Core as Shared MCP functionality

  alt 2025-06-18 or 2025-11-25
    Client->>Ingress: initialize with protocol version
    Ingress->>Protocol: Select stateful protocol behavior
    Protocol-->>Ingress: Version-specific initialize result
    Ingress-->>Client: Negotiated version and session
    Client->>Ingress: MCP request with session
  else 2026-07-28
    Client->>Ingress: MCP request with protocol version
    Ingress->>Protocol: Select stateless protocol behavior
  end

  Protocol->>Protocol: Validate and decode request using selected schema
  Protocol->>Core: Canonical MCP operation
  Core-->>Protocol: Canonical result
  Protocol->>Protocol: Encode result using the same selected schema
  Protocol-->>Ingress: Version-specific response
  Ingress-->>Client: Response matching the client's protocol version
```

Only one version-specific path is selected for a request. The same selection governs both request decoding and response encoding, so a request using `2025-06-18`, for example, cannot be decoded or returned using the shape of `2025-11-25`.

The shared MCP functionality represents what the server can do: tools, prompts, resources, and their handlers.

The version-specific behavior represents how a particular MCP specification exposes that functionality, including:

- request and response shapes
- capability representation
- initialization, discovery, and session behavior
- notifications and server-to-client interactions
- protocol-specific transport requirements

The exact implementation is open for discussion. One possible direction is a protocol adapter or similar boundary between the wire protocol and the shared server functionality. The important behavioral requirement is that protocol selection happens early enough that incompatible versions are interpreted independently, rather than being combined into one permissive schema.

For legacy protocols, selection may happen during initialization and remain associated with the session. For `2026-07-28`, selection would instead happen per request using the version information carried by that request.

The questions I would like feedback on are:

1. Does this describe the right behavioral boundary for supporting both stateful and stateless MCP revisions on one endpoint?
1. Are there capabilities or lifecycle behaviors that cannot reasonably be projected from shared tools, prompts, and resources?
1. Are there transport, authentication, or conformance requirements that should be considered before settling on an implementation?

lloydrichards:
## Type

- [x] Refactor
- [x] Feature

## Goal/Scope

Prepare `McpServer` to support multiple incompatible MCP protocol versions through one endpoint without merging their schemas into a permissive shared protocol.

Adds the version-isolation seam and the `2025-06-18` adapter, but does not implement the `2025-11-25` or `2026-07-28` protocols.

## Description

Servers now declare an ordered, non-empty list of implemented adapters:

```text
protocols: [McpProtocol.v2025_06_18]
```

An adapter contains its version and the four RPC groups used for client requests, client notifications, server requests, and server notifications. This prevents a server from advertising support using only a version string.

`ProtocolRegistry` snapshots the declaration, rejects duplicate versions, and selects an exact match or falls back to the first declared adapter. Allowing incoming methods to be routed to versioned internal keys before payload decoding:

```text
tools/call
→ @effect/mcp/2025-06-18/tools/call
```

The internal key selects exactly one adapter's schema and handler. The existing `RpcServer` still owns decoding, middleware, handler execution, and response encoding.

Successful initialization pins the selected adapter to the session. Later requests, reverse client requests, and outbound notifications all reuse that adapter.

## How to Review

1. Start with `McpProtocol.ts` and `internal/mcpProtocol.ts`. Confirm an adapter represents executable support rather than only a version identifier.

2. Review `internal/mcpProtocolRegistry.ts`. Check the declaration snapshot, duplicate rejection, first-adapter fallback, and versioned RPC-group merge.

3. Follow one request through `runWithRegistry` in `McpServer.ts`:

   ```text
   raw initialize
   → inspect protocolVersion
   → select adapter
   → route to versioned key
   → decode and execute through RpcServer
   → store adapter in session
   ```

   The important invariant is that adapter selection happens before schema decoding and is not renegotiated after initialization.

4. Review `layerHandlers` and `addProtocolHandlers`. Shared MCP domain handlers are prepared with `RpcGroup.toHandlers` and registered under each adapter's versioned key.

5. Review the protocol and server tests. They cover registry validation and fallback, incompatible schema isolation, session pinning, the HTTP protocol header, missing sessions, and preservation of the June stdio transcript.

## Comments

Only `McpProtocol.v2025_06_18` is public in this phase. Later versions can introduce distinct schemas and projections without changing the selection boundary.

Server-originated MCP notifications expose an existing mismatch with `RpcServer.Protocol.send`, which only models RPC responses. The narrow cast is retained with a TODO: changing the RPC transport contract is outside this PR.

## Related

- closes #6617

## Summary by CodeRabbit

* **New Features**
  * Added adapter-based MCP protocol support, with built-in support starting at `2025-06-18`.
  * MCP servers now negotiate, pin, and use the correct protocol per session for routing requests and notifications.
  * MCP server client context now exposes the negotiated `protocolVersion`.
* **Bug Fixes**
  * Improved MCP protocol header validation and protocol selection, including rejecting unsupported offered protocol versions.
  * Enhanced protocol-aware decoding; `notifications/cancelled` now behaves as an interrupt.
* **Documentation**
  * Updated MCP server examples to configure `protocols` explicitly and reflect the new support floor.
* **Tests**
  * Expanded MCP protocol/server tests and added type-level validation for the updated server API.

### DIFF
--- a/.changeset/quiet-mice-negotiate.md
+++ b/.changeset/quiet-mice-negotiate.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Add adapter-valued MCP server protocol declarations, route requests through the selected protocol before schema decoding, and add built-in support for MCP `2025-06-18`.
--- a/.changeset/quiet-mice-negotiate.md
+++ b/.changeset/quiet-mice-negotiate.md
@@ -1,5 +1,5 @@
 ---
-"effect": minor
+"effect": patch
 ---
 
 Add adapter-valued MCP server protocol declarations, route requests through the selected protocol before schema decoding, and add built-in support for MCP `2025-06-18`.
--- a/packages/effect/src/unstable/ai/McpProtocol.ts
+++ b/packages/effect/src/unstable/ai/McpProtocol.ts
@@ -3,6 +3,7 @@
  *
  * @since 4.0.0
  */
+import type * as RpcGroup from "../rpc/RpcGroup.ts"
 import * as Internal from "./internal/mcpProtocol.ts"
 import * as McpSchema from "./McpSchema.ts"
 
@@ -12,7 +13,7 @@ import * as McpSchema from "./McpSchema.ts"
  * @category protocols
  * @since 4.0.0
  */
-export const v2025_06_18 = Internal.make({
+export const v2025_06_18: ProtocolAdapter = Internal.make({
   protocolVersion: "2025-06-18",
   clientRpcs: McpSchema.ClientRpcs,
   clientNotificationRpcs: McpSchema.ClientNotificationRpcs,
@@ -26,7 +27,13 @@ export const v2025_06_18 = Internal.make({
  * @category models
  * @since 4.0.0
  */
-export type ProtocolAdapter = typeof v2025_06_18
+export type ProtocolAdapter = Internal.ProtocolAdapter<
+  "2025-06-18",
+  RpcGroup.Rpcs<typeof McpSchema.ClientRpcs>,
+  RpcGroup.Rpcs<typeof McpSchema.ClientNotificationRpcs>,
+  RpcGroup.Rpcs<typeof McpSchema.ServerRequestRpcs>,
+  RpcGroup.Rpcs<typeof McpSchema.ServerNotificationRpcs>
+>
 
 /**
  * The MCP protocol versions implemented by this release.
--- a/packages/effect/src/unstable/ai/McpProtocol.ts
+++ b/packages/effect/src/unstable/ai/McpProtocol.ts
@@ -1,0 +1,37 @@
+/**
+ * Defines the MCP protocol implementations that an `McpServer` can support.
+ *
+ * @since 4.0.0
+ */
+import * as Internal from "./internal/mcpProtocol.ts"
+import * as McpSchema from "./McpSchema.ts"
+
+/**
+ * The MCP 2025-06-18 protocol implementation.
+ *
+ * @category protocols
+ * @since 4.0.0
+ */
+export const v2025_06_18 = Internal.make({
+  protocolVersion: "2025-06-18",
+  clientRpcs: McpSchema.ClientRpcs,
+  clientNotificationRpcs: McpSchema.ClientNotificationRpcs,
+  serverRequestRpcs: McpSchema.ServerRequestRpcs,
+  serverNotificationRpcs: McpSchema.ServerNotificationRpcs
+})
+
+/**
+ * An implemented MCP protocol that can be supplied to `McpServer`.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export type ProtocolAdapter = typeof v2025_06_18
+
+/**
+ * The MCP protocol versions implemented by this release.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export type ProtocolVersion = ProtocolAdapter["protocolVersion"]
--- a/packages/effect/src/unstable/ai/McpServer.ts
+++ b/packages/effect/src/unstable/ai/McpServer.ts
@@ -40,9 +40,10 @@ import * as RpcMessage from "../rpc/RpcMessage.ts"
 import * as RpcSerialization from "../rpc/RpcSerialization.ts"
 import * as RpcServer from "../rpc/RpcServer.ts"
 import * as AiError from "./AiError.ts"
+import * as McpProtocolRegistry from "./internal/mcpProtocolRegistry.ts"
+import type * as McpProtocol from "./McpProtocol.ts"
 import {
   CallToolResult,
-  ClientNotificationRpcs,
   ClientRpcs,
   CompleteResult,
   Elicit,
@@ -62,7 +63,6 @@ import {
   Resource,
   ResourceTemplate,
   ServerNotificationRpcs,
-  ServerRequestRpcs,
   TextContent,
   Tool as McpTool
 } from "./McpSchema.ts"
@@ -334,29 +334,32 @@ export class McpServer extends Context.Service<McpServer, {
   static readonly layer: Layer.Layer<McpServer | McpServerClient> = Layer.effect(McpServer)(McpServer.make) as any
 }
 
-const LATEST_PROTOCOL_VERSION = "2025-06-18"
-const SUPPORTED_PROTOCOL_VERSIONS = [
-  LATEST_PROTOCOL_VERSION,
-  "2025-03-26",
-  "2024-11-05",
-  "2024-10-07"
-]
-const MCP_SESSION_ID_HEADER = "mcp-session-id"
-const MCP_PROTOCOL_VERSION_HEADER = "mcp-protocol-version"
+const mcpSessionIdHeader = "mcp-session-id"
+const mcpProtocolVersionHeader = "mcp-protocol-version"
+
+interface Session {
+  readonly initializePayload: typeof Initialize.payloadSchema.Type
+  readonly protocol: McpProtocol.ProtocolAdapter
+}
 
 class McpProtocolState extends Context.Service<McpProtocolState, {
-  readonly clientSessions: Map<string, typeof Initialize.payloadSchema.Type>
-  readonly latestProtocolVersion: string
-  readonly supportedProtocolVersions: ReadonlySet<string>
+  readonly clientSessions: Map<string, Session>
+  readonly protocolRegistry: McpProtocolRegistry.ProtocolRegistry<McpProtocol.ProtocolAdapter>
 }>()("effect/ai/McpServer/McpProtocolState") {}
 
-const makeMcpProtocolState = (): McpProtocolState["Service"] => ({
-  clientSessions: new Map(),
-  latestProtocolVersion: LATEST_PROTOCOL_VERSION,
-  supportedProtocolVersions: new Set(SUPPORTED_PROTOCOL_VERSIONS)
+const makeMcpProtocolState = Effect.fnUntraced(function*(
+  protocols: Arr.NonEmptyReadonlyArray<McpProtocol.ProtocolAdapter>
+) {
+  return McpProtocolState.of({
+    clientSessions: new Map(),
+    protocolRegistry: yield* McpProtocolRegistry.make(protocols)
+  })
 })
 
-const layerMcpProtocolState = Layer.sync(McpProtocolState, makeMcpProtocolState)
+const layerMcpProtocolState = (
+  protocols: Arr.NonEmptyReadonlyArray<McpProtocol.ProtocolAdapter>
+): Layer.Layer<McpProtocolState, Cause.IllegalArgumentError> =>
+  Layer.effect(McpProtocolState)(makeMcpProtocolState(protocols))
 
 /**
  * Runs an MCP server over the current `RpcServer.Protocol`.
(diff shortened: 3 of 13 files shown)


packages/effect/src/unstable/ai/McpServer.ts
```diff
@@ -532,31 +569,52 @@ export const run: (options: {
           case "Chunk":
           case "ClientProtocolError":
           case "Defect":
-            return RcMap.get(clients, clientId).pipe(
+            return RcMap.get(
+              clients,
+              `${clientId}\0${getProtocolForClient(clientProtocols, clientId, protocolRegistry).protocolVersion}`
```
IMax153: Why the null byte here?

packages/effect/test/unstable/ai/McpServer.test.ts
```diff
@@ -129,8 +137,81 @@ const toolResultText = (result: McpSchema.CallToolResult): string => {
...
+  return InternalMcpProtocol.make({
+    protocolVersion,
+    clientRpcs: McpSchema.ClientRpcs.omit("ping").add(Ping),
+    clientNotificationRpcs: McpSchema.ClientNotificationRpcs,
+    serverRequestRpcs: McpSchema.ServerRequestRpcs,
+    serverNotificationRpcs: McpSchema.ServerNotificationRpcs
+  })
+}
+
+const makeRawHttpServer = (
```
IMax153, line 156: I think we should consider creating a small module of utilities for the McpServer tests in `./test/unstable/ai/McpServer/utils.ts`. I had already started doing this but I wasn't very principled about it.

Then we can re-use these utilities as we expand the test suite out to cover the full specification.

packages/effect/src/unstable/ai/McpProtocol.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @category protocols
+ * @since 4.0.0
+ */
+export const v2025_06_18 = Internal.make({
+  protocolVersion: "2025-06-18",
+  clientRpcs: McpSchema.ClientRpcs,
+  clientNotificationRpcs: McpSchema.ClientNotificationRpcs,
+  serverRequestRpcs: McpSchema.ServerRequestRpcs,
+  serverNotificationRpcs: McpSchema.ServerNotificationRpcs
+})
```
IMax153: I like the idea of having an abstracted `McpProtocol` - but we will need to come up with a solution for `McpSchema` as we introduce support for other protocol versions.

Also please make sure all public exports have their type annotated - this does not.

packages/effect/src/unstable/ai/McpServer.ts
lloydrichards: cause in the lookup i use it to split the key and get the `selectedProtocol`.  Not the prettiest, but I was worried the same client might try to connect with different protocol versions (for gods know what reason) and then the RcMap wouldn't be able to tell:

```ts
 lookup: Effect.fnUntraced(function*(key: string) {
      const separator = key.indexOf("\0")
      const clientId = Number(key.slice(0, separator))
      const selectedProtocol = protocolRegistry.select(key.slice(separator + 1))
```

packages/effect/test/unstable/ai/McpServer.test.ts
lloydrichards: I could extract this out into #6630 if you don't mind the cleanup later

packages/effect/src/unstable/ai/McpProtocol.ts
lloydrichards: gave it a more explicit type then just `typeof`:

```ts
export type ProtocolAdapter = Internal.ProtocolAdapter<
  "2025-06-18",
  RpcGroup.Rpcs<typeof McpSchema.ClientRpcs>,
  RpcGroup.Rpcs<typeof McpSchema.ClientNotificationRpcs>,
  RpcGroup.Rpcs<typeof McpSchema.ServerRequestRpcs>,
  RpcGroup.Rpcs<typeof McpSchema.ServerNotificationRpcs>
>
```

I was waiting to adding `v2025-11-28` before having a hard think about the best type for this, but if you have ideas now i'm all ears

packages/effect/src/unstable/ai/McpServer.ts
IMax153: `RcMap` supports hashed keys in its lookup, so you can pass in a `Data.Class` or similar that implements Hash / Equals to avoid this kind of hacky logic.

packages/effect/test/unstable/ai/McpServer.test.ts
IMax153: Not at all - I commented there too, so I'll resolve this convo for now.

packages/effect/src/unstable/ai/McpProtocol.ts
IMax153: I'm honestly not sure what the best strategy is going to be. The schemas themselves will likely be quite similar between versions, with some minor changes, but I don't know for sure.

Maybe we implement sets of Rpcs for each protocol in `McpSchema`? Not sure how laborious that will be, but if we share individual Rpcs that don't change between versions, it might not be too bad?

packages/effect/src/unstable/ai/McpServer.ts
lloydrichards: much cleaner, thanks! 👨‍🍳

```ts
class McpClientKey extends Data.Class<{
  readonly clientId: number
  readonly protocolVersion: string
}> {}
```

see 968d042

packages/effect/src/unstable/ai/McpProtocol.ts
lloydrichards: ideally there is some "core" `McpSchema` that just needs to be extended with version specific shapes.  I already codified it in the McpServer that if an rpc is missing from something like the `.serverNotificationRpcs` then the broadcast is skipped since this might be something that the client on that particular version is missing:

```ts
      for (const clientId of server.initializedClients.keys()) {
        if (!clientIds.has(clientId)) {
          server.initializedClients.delete(clientId)
          continue
        }
        const selectedProtocol = clientProtocols.get(clientId)
        if (!selectedProtocol) {
          continue
        }
        const rpc = selectedProtocol.serverNotificationRpcs.requests.get(request.tag)
        if (!rpc) {
          continue
        }
        const encoded = yield* Schema.encodeUnknownEffect(
          Schema.toCodecJson(rpc.payloadSchema)
        )(request.payload)
        ...
      }
```

This kind of assumes that the Rpc shape could end up being quite different between say `v2025-06-18` and `v2026-07-28`

IMax153: Yea - I think we will need a robust solution for this as we introduce further protocol versions.

lloydrichards: I'm going to track this in #6639 so its easier to collect information and ideas.  If you want this to be a blocker, then I can build another PR that stacks into this one so its easier to review

IMax153: I don't _think_ one has to block the other. If this PR is ready for a final review, let me know and I can do a deep dive.

lloydrichards: i'll go over it again tonight to be sure but i think its just about ready.  see what i can bring from the McpCore exploration

lloydrichards: I'm happy with it.  Feel free to deep dive into this when you have time.  I'll try figure out how the other protocols will fit together on top of this

.changeset/quiet-mice-negotiate.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
```
IMax153:
```suggestion
"effect": patch
```

This can be a `patch` since we're still in beta.

packages/effect/src/unstable/ai/McpServer.ts
```diff
@@ -373,32 +382,46 @@ const layerMcpProtocolState = Layer.sync(McpProtocolState, makeMcpProtocolState)
 export const run: (options: {
   readonly name: string
   readonly version: string
+  readonly protocols: Arr.NonEmptyReadonlyArray<McpProtocol.ProtocolAdapter>
```
IMax153, line 393: Are we going to force the user to pass in versions, or should this be optional and default to the latest version we support currently?

packages/effect/src/unstable/ai/McpServer.ts
```diff
@@ -532,31 +579,55 @@ export const run: (options: {
...
         }
+        const selectedProtocol = clientProtocols.get(clientId)
+        if (!selectedProtocol) {
+          continue
+        }
+        const rpc = selectedProtocol.serverNotificationRpcs.requests.get(request.tag)
+        if (!rpc) {
+          continue
+        }
+        const encoded = yield* Schema.encodeUnknownEffect(
```
IMax153: To avoid re-computing these encoders / decoders every time we run through here, I'm wondering if our protocol abstraction should cache them somehow.

For example, in our `Toolkit` module, we cache the encoders / decoders we need on first creation.

packages/effect/src/unstable/ai/McpServer.ts
```diff
@@ -1386,145 +1466,180 @@ const layerHandlers = (serverInfo: {
...
+            }
+            return Effect.withFiber((fiber) => {
+              const httpRequest = Context.getOrUndefined(fiber.context, HttpServerRequest.HttpServerRequest)
+              const session: Session = {
+                initializePayload: params,
+                protocol: selectedProtocol
+              }
+              if (httpRequest) {
+                const sessionId = crypto.randomUUID()
+                options.clientSessions.set(`session:${sessionId}`, session)
```
IMax153: Not a huge fan of the string prefixed keys here - feels brittle.

IMax153 (changes requested):

packages/effect/src/unstable/ai/McpServer.ts
lloydrichards: I think making this required makes sense cause there needs to be a decision on what can be supported, especially with the upcoming `v2026-06-28` stateless shift.  Since each item in the array multiplies the number of RpcGroups it would be good to not hide it behind defaults.

In the `McpCore` version i'm working on now I've broken `v2024-11-05` and `v2025-03-26` from the `McpSchema` to test how the fallbacks work and this seems like an important decisions for the `McpServer` on what happens when an unexpected version'd client sends a request:

```ts
 const server = McpServer.layerHttp({
    name: "Acme Mcp",
    version: "1.0.0",
    path: "/mcp",
    protocols: [
      McpProtocol.v2025_06_18, // <- fallback
      McpProtocol.v2025_03_26,
      McpProtocol.v2024_11_05
    ],
  })
```

If you think this is too breaking of a change then we could have a default array, but it would need to be communicated well that this increases the size of the server by just updating versions

packages/effect/src/unstable/ai/McpServer.ts
lloydrichards: I split up the session map into a `bySessionId` and `byClientId`:

```ts
interface Sessions {
  readonly bySessionId: Map<string, Session>
  readonly byClientId: Map<number, Session>
}
```
So the http and stdio versions can be tracked separately and makes it a cleaner distinction.  

see a70aa6d

thought: I did notice while doing this though that there is no TTL for http sessions so there is a bit of resource leak here as sessions are stored indefinitely.  This is a separate issue as the current `McpServer` does the same, but want to flag it as I saw it.

packages/effect/src/unstable/ai/McpServer.ts
lloydrichards: see a70aa6d

.changeset/quiet-mice-negotiate.md
lloydrichards: see a70aa6d

packages/effect/src/unstable/ai/McpServer.ts
IMax153: No I think this argument is well thought out - we can keep the protocols array required for now.

IMax153 (approved):
I think this looks good to me in its current form - we can merge and test it out 👍

PR #6635, opened 2026-07-25
Fix empty struct union candidate selection

gcanti:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected schema handling for empty object structures in unions.
  * Empty object schemas now correctly accept all non-nullish values while continuing to reject `undefined`.

* **Tests**
  * Expanded coverage across primitive values, objects, arrays, functions, symbols, bigints, and `null`.


nikelborm:
Would I need to update https://github.com/Effect-TS/effect/pull/6434 after this lands?

PR #6637, opened 2026-07-25
Fix Latch stale flush resuming waiters registered after close

tim-smart:
## Problem

`Latch.open` / `Latch.release` schedule `flushWaiters` on the fiber dispatcher, which runs on a later macrotask. The flush drained whatever was in the `waiters` array at flush time, with no snapshot and no state check. If the latch was closed again before the flush ran, a waiter that registered **after** the close — never covered by any `open`/`release` — was incorrectly resumed:

```ts
const latch = Latch.makeUnsafe(false)
const w1 = yield* Effect.forkChild(latch.await) // waiters exist, so open() schedules a flush
yield* latch.open   // isOpen = true, flush queued for a later macrotask
yield* latch.close  // isOpen = false before the flush runs
yield* latch.await  // suspends on a CLOSED latch — but the stale flush resumed it
```

This violates the documented contract that `await` on a closed latch suspends until a subsequent `open`/`release`, and gating logic built on `Latch` could observe work proceeding while the gate is closed.

## Fix

Snapshot the waiters at schedule time: `open`/`release` move the current waiters into a pending `scheduled` array, and the scheduled flush resumes only that snapshot. Waiters that register afterwards stay in `waiters` and are untouched by a stale flush.

- A `release` while a flush is already pending extends the snapshot with the newly covered waiters (no extra dispatcher task).
- `await` interruption cleanup also removes the waiter from the pending snapshot.
- `openUnsafe` still flushes everything synchronously (pending snapshot + current waiters).

## Tests

Added two regression tests in `packages/effect/test/Latch.test.ts`:

- `open then close does not resume waiters registered after close` — fails on `main`, passes with the fix.
- `release while a flush is pending covers the new waiters` — exercises the snapshot-extension path with a manual scheduler.

`pnpm check`, oxlint/dprint, and the Latch/Effect/Channel/PubSub/Fiber/SubscriptionRef suites (320 tests) pass.

Closes EFF-52

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed latch wake-up behavior to avoid resuming waiters added after a close from a previously scheduled flush.
  * Wake-ups are now limited to the waiters captured for the related open/release, including reentrant close scenarios.
  * Improved cancellation so interrupted waiters aren’t resumed by queued wake-ups.
* **Tests**
  * Added concurrency and edge-case tests for stale flush prevention, pending flush behavior, repeated releases, reentrant close, and cancellation during queued wake-ups.
* **Documentation**
  * Updated documented latch behavior to reflect the corrected waiter flushing semantics.

### DIFF
--- a/packages/effect/src/internal/effect.ts
+++ b/packages/effect/src/internal/effect.ts
@@ -5516,9 +5516,6 @@ const succeedFalse = succeed(false)
 
 class Latch implements _Latch.Latch {
   waiters: Array<(_: Effect.Effect<void>) => void> = []
-  // waiters snapshotted at open/release time, awaiting the scheduled flush.
-  // Waiters that register after a subsequent close stay in `waiters` and are
-  // not resumed by a stale flush.
   scheduled: Array<(_: Effect.Effect<void>) => void> | undefined = undefined
   private _isOpen: boolean
 
--- a/packages/effect/src/internal/effect.ts
+++ b/packages/effect/src/internal/effect.ts
@@ -5550,9 +5550,12 @@ class Latch implements _Latch.Latch {
     }
   }
   private flushWaiters() {
-    this.flushScheduled()
+    // swap both arrays out before any resume runs: a resumed waiter can
+    // reentrantly close the latch and register new waiters, which must not
+    // be drained by this flush
     const waiters = this.waiters
     this.waiters = []
+    this.flushScheduled()
     for (let i = 0; i < waiters.length; i++) {
       waiters[i](exitVoid)
     }
--- a/packages/effect/src/internal/effect.ts
+++ b/packages/effect/src/internal/effect.ts
@@ -5516,7 +5516,10 @@ const succeedFalse = succeed(false)
 
 class Latch implements _Latch.Latch {
   waiters: Array<(_: Effect.Effect<void>) => void> = []
-  scheduled = false
+  // waiters snapshotted at open/release time, awaiting the scheduled flush.
+  // Waiters that register after a subsequent close stay in `waiters` and are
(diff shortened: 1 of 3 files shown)


packages/effect/src/internal/effect.ts
```diff
@@ -5516,25 +5516,46 @@ const succeedFalse = succeed(false)
 
 class Latch implements _Latch.Latch {
   waiters: Array<(_: Effect.Effect<void>) => void> = []
-  scheduled = false
+  // waiters snapshotted at open/release time, awaiting the scheduled flush.
```
tim-smart: Remove this comment

tim-smart: Removed in dfa988020.

packages/effect/test/Latch.test.ts
```diff
@@ -32,6 +32,139 @@ describe("Latch", () => {
...
+          latch.closeUnsafe()
+          second = yield* Effect.forkChild(Latch.await(latch), { startImmediately: true })
+          return yield* Effect.never
+        }),
+        { startImmediately: true }
+      )
+
+      yield* latch.release.pipe(Effect.provideService(Scheduler.Scheduler, scheduler))
+      latch.openUnsafe()
+      assert.isUndefined(second!.pollUnsafe())
```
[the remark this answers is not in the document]
tim-smart: Fixed in 8433e2b0a. The test now verifies one task is queued, drains it, then re-asserts the reentrant waiter remains suspended. I also ran the strengthened test against the pre-fix implementation; it fails at the post-drain assertion because the stale task resumes the waiter.

PR #6641, opened 2026-07-26
Add manual flush registry to OtlpExporter

tim-smart:
Adds a manual flush mechanism to the OTLP exporters so buffered telemetry can be drained on demand, for example with `ctx.waitUntil` in short-lived serverless invocations where neither the export interval, the batch threshold, nor scope shutdown fires in time.

## Flush registry

- Adds a public `Flusher` service and a module-level `layerFlusher`. Reusing one layer instance lets `Layer.MemoMap` give merged signal layers one shared registry.
- Makes registration unconditional and scoped in `OtlpExporter.make`, preventing missing wiring and removing exporters when their scopes close.
- Runs registered exporters concurrently. `flush` cannot fail, respects the existing 60-second `disabledUntil` circuit breaker, and has no built-in timeout; callers can bound it with `Effect.timeoutOption`.
- Keeps the existing overlap behavior: a flush waits for exports it starts but cannot await an interval export already in flight.

## Signal layers

- Updates `OtlpTracer`, `OtlpLogger`, and `OtlpMetrics` layers to provide and expose the shared `Flusher` with `Layer.provideMerge`.
- Returns `layerFlusher` from disabled `layerFromConfig` branches, so flushing remains an available no-op when telemetry is disabled.
- Changes signal layer output types additively to include `Flusher`; `OtlpExporter.make` now requires `Flusher` for custom exporters.

## Tests and documentation

- Covers immediate flush, registry sharing across traces/logs/metrics, the disabled circuit breaker, scope deregistration, empty and SDK-disabled registries, and collector failures.
- Documents the flush contract and Cloudflare Workers `ctx.waitUntil` call site.
- Documents that metrics flushing takes a snapshot and therefore narrows delta-temporality windows when called frequently.
- Adds a changeset for the public API and type changes.

Validation: `pnpm lint`, `pnpm --dir packages/effect docgen`, and `pnpm check`.

Closes EFF-56

## Summary by CodeRabbit

* **New Features**
  * Added coordinated manual flushing for OTLP batch exporters using a shared flusher service.
  * OTLP logs, metrics, and traces now expose this shared flushing support, including when OTLP export is disabled or unavailable.
  * Flushing drains all registered exporters and exporters are automatically deregistered when their scope ends.
* **Tests**
  * Expanded OTLP exporter test coverage with dedicated flush scenarios (registry sharing, no-op when disabled, scope deregistration, empty-registries, and resilient handling of collector failures).

### DIFF
--- a/packages/effect/src/unstable/observability/OtlpExporter.ts
+++ b/packages/effect/src/unstable/observability/OtlpExporter.ts
@@ -58,7 +58,7 @@ const policy = Schedule.forever.pipe(
  * @category flushing
  * @since 4.0.0
  */
-export interface Flusher {
+export class Flusher extends Context.Service<Flusher, {
   /**
    * Drains all registered exporters concurrently and cannot fail.
    *
@@ -86,17 +86,9 @@ export interface Flusher {
    */
   readonly flush: Effect.Effect<void>
   readonly register: (run: Effect.Effect<void>) => Effect.Effect<void, never, Scope.Scope>
-}
-
-/**
- * Context key for the shared exporter flush registry.
- *
- * @category flushing
- * @since 4.0.0
- */
-export const Flusher: Context.Key<Flusher, Flusher> = Context.Service<Flusher>(
+}>()(
   "effect/observability/OtlpExporter/Flusher"
-)
+) {}
 
 /**
  * Provides a `Flusher` backed by a fresh registry.
@@ -110,7 +102,7 @@ export const Flusher: Context.Key<Flusher, Flusher> = Context.Service<Flusher>(
  * layer per call would silently create one registry per signal.
  *
  * Registration is scoped — an exporter is removed from the registry when its
- * own scope closes. Flushing with an empty registry is a debug-logged no-op.
+ * own scope closes. Flushing with an empty registry is a no-op.
  *
  * Note that `flush` cannot await an export that was already in flight when it
  * was called (for example one started by the export interval); it only waits
@@ -124,7 +116,7 @@ export const layerFlusher: Layer.Layer<Flusher> = Layer.sync(Flusher, () => {
   return {
     flush: Effect.suspend(() => {
       if (registry.size === 0) {
-        return Effect.logDebug("Flush requested but no OTLP exporters are registered")
+        return Effect.void
       }
       return Effect.forEach(registry, identity, {
         concurrency: "unbounded",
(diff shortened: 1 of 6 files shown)


packages/effect/src/unstable/observability/OtlpExporter.ts
```diff
@@ -41,6 +43,80 @@ const policy = Schedule.forever.pipe(
...
+  readonly register: (run: Effect.Effect<void>) => Effect.Effect<void, never, Scope.Scope>
+}
+
+/**
+ * Context key for the shared exporter flush registry.
+ *
+ * @category flushing
+ * @since 4.0.0
+ */
+export const Flusher: Context.Key<Flusher, Flusher> = Context.Service<Flusher>(
```
tim-smart: This needs to use the class Context.Service syntax to ensure the type id is unique.

packages/effect/src/unstable/observability/OtlpExporter.ts
```diff
@@ -41,6 +43,80 @@ const policy = Schedule.forever.pipe(
...
+ *
+ * @category flushing
+ * @since 4.0.0
+ */
+export const layerFlusher: Layer.Layer<Flusher> = Layer.sync(Flusher, () => {
+  const registry = new Set<Effect.Effect<void>>()
+  return {
+    flush: Effect.suspend(() => {
+      if (registry.size === 0) {
+        return Effect.logDebug("Flush requested but no OTLP exporters are registered")
```
tim-smart:
```suggestion
        return Effect.void
```

We don't need to log here

packages/effect/src/unstable/observability/OtlpExporter.ts
```diff
@@ -122,6 +198,9 @@ export const make: (
     })
   )
 
+  const flusher = yield* Flusher
+  yield* Effect.provideService(flusher.register(runExport), Scope.Scope, scope)
```
tim-smart:
```suggestion
  yield* flusher.register(runExport)
```

`register` will use the correct scope without us providing it.

packages/effect/src/unstable/observability/OtlpExporter.ts
tim-smart: Updated `Flusher` to use the class `Context.Service` syntax in 7ef2bc188. Existing service call sites and the JSDoc example remain valid with the class tag.

packages/effect/src/unstable/observability/OtlpExporter.ts
tim-smart: Updated registration to `yield* flusher.register(runExport)` in 7ef2bc188.

packages/effect/src/unstable/observability/OtlpExporter.ts
tim-smart: Updated the empty-registry branch to return `Effect.void` and adjusted the JSDoc in 7ef2bc188.

PR #6679, opened 2026-07-28
Preserve buffered NodeTerminal readLine input

tim-smart:
## Summary

- keep one readline interface for the terminal scope
- buffer line events so sequential `readLine` calls do not drop piped input
- preserve `QuitError` after buffered lines are drained at EOF

## Testing

- `pnpm --filter @effect/platform-node-shared test --run test/NodeTerminal.test.ts`
- `pnpm lint-fix`
- `pnpm check`
- verified the two-line and EOF fixtures with Bun 1.3.13

Closes EFF-135

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed terminal input handling so buffered lines are preserved across sequential reads.
  * Improved cleanup when terminal input closes or raw mode is disabled.

* **Tests**
  * Added coverage for reading multiple lines in sequence and returning them in the correct order.

* **Chores**
  * Added a patch release note for the terminal input fix.

### DIFF
--- a/packages/platform-node-shared/src/NodeTerminal.ts
+++ b/packages/platform-node-shared/src/NodeTerminal.ts
@@ -16,7 +16,8 @@ import * as Option from "effect/Option"
 import { badArgument, type PlatformError } from "effect/PlatformError"
 import * as Predicate from "effect/Predicate"
 import * as Queue from "effect/Queue"
-import * as Scope from "effect/Scope"
+import * as RcRef from "effect/RcRef"
+import type * as Scope from "effect/Scope"
 import * as Terminal from "effect/Terminal"
 import * as readline from "node:readline"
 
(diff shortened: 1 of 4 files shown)


packages/platform-node-shared/src/NodeTerminal.ts
```diff
@@ -45,33 +44,34 @@ export const make: (
...
-    // Acquire readline interface with TTY setup/cleanup inside the scope
-    const rlRef = yield* RcRef.make({
-      acquire: Effect.acquireRelease(
-        Effect.sync(() => {
-          const rl = readline.createInterface({ input: stdin, escapeCodeTimeout: 50 })
-          readline.emitKeypressEvents(stdin, rl)
+    const lines = yield* Queue.make<string, Cause.Done>()
+
+    // Keep one readline interface so lines consumed together remain available.
+    yield* Effect.acquireRelease(
```
tim-smart: The layer construction should be side effect free - the readline interface should only be installed when the service is interacted with.

packages/platform-node-shared/src/NodeTerminal.ts
```diff
@@ -45,33 +44,34 @@ export const make: (
...
+      Effect.sync(() => {
+        const rl = readline.createInterface({ input: stdin, escapeCodeTimeout: 50 })
+        readline.emitKeypressEvents(stdin, rl)
+        rl.on("line", (line) => Queue.offerUnsafe(lines, line))
+        rl.once("close", () => Queue.endUnsafe(lines))
 
+        if (stdin.isTTY) {
+          stdin.setRawMode(true)
+        }
+        return rl
```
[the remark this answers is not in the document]
tim-smart: Addressed in 1f02a1a09. Lazy initialization checks the tracked EOF state before creating readline and immediately ends the line queue when stdin already ended, while retaining the readline close handler for later EOF. Added regression coverage.

packages/platform-node-shared/src/NodeTerminal.ts
tim-smart: Addressed in 1f02a1a09. The readline interface and line queue are now memoized and initialized only when input APIs are used; layer construction leaves stdin without a readline data listener. Added regression coverage.

PR #6684, opened 2026-07-28
Add Deno Crypto implementation

tim-smart:
## Summary

- add a Deno Web Crypto-backed `Crypto` layer and overridable `WebCrypto` reference
- chunk random byte requests at the Web Crypto 65,536-byte limit
- export the module from `@effect/platform-deno`

## Testing

- `pnpm lint-fix`
- `pnpm --filter @effect/platform-deno test --run test/DenoCrypto.test.ts`
- `deno task test --run packages/platform-deno/test/DenoCrypto.test.ts`
- `pnpm check`
- `deno check .`

The Deno tests intentionally differ from `BrowserCrypto.test.ts`: they use the real runtime Web Crypto implementation for SHA-256, a 70,000-byte random request, and UUID shape checks. They do not port the browser suite's fake-injection tests because those fakes replace the adapter's primitive operations and primarily retest `Crypto.make`, which is already covered in `packages/effect`.

Closes EFF-140

### DIFF
--- a/packages/platform-deno/src/DenoCrypto.ts
+++ b/packages/platform-deno/src/DenoCrypto.ts
@@ -1,12 +1,11 @@
 /**
  * Deno-backed implementation of Effect's Crypto service.
  *
- * This module uses Deno's global Web Crypto API. Unlike the browser
- * implementation, random byte requests are split into chunks of at most 65,536
- * bytes so they support the full size range accepted by `Crypto.make`. Deno's
- * supported runtime versions always provide `globalThis.crypto` and
- * `crypto.subtle`, so this implementation does not include browser availability
- * guards.
+ * This module uses Deno's global Web Crypto API. Random byte requests are split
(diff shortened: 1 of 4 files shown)


packages/platform-deno/src/DenoCrypto.ts
```diff
@@ -0,0 +1,67 @@
+/**
+ * Deno-backed implementation of Effect's Crypto service.
+ *
+ * This module uses Deno's global Web Crypto API. Random byte requests are split
+ * into chunks of at most 65,536 bytes so they support the full size range
+ * accepted by `Crypto.make`. Unlike browser runtimes, Deno's supported runtime
+ * versions always provide `globalThis.crypto` and `crypto.subtle`, so this
+ * implementation does not include availability guards.
```
tim-smart:
```suggestion
 * This module uses Deno's global Web Crypto API.
```

PR #6686, opened 2026-07-28
Add Deno Redis integration

tim-smart:
## Summary

- add a scoped Deno Redis integration backed by `@db/redis`
- support URL-based and explicit connection options
- mirror Redis persistence and queue integration coverage under Deno

## Testing

- `pnpm lint-fix`
- `pnpm check`
- `pnpm --filter @effect/platform-deno check`
- `deno task test --run packages/platform-deno/test/DenoRedis.test.ts`

Closes EFF-151

## Summary by CodeRabbit

* **New Features**
  * Added native Redis integration for Deno applications via new `DenoRedis` support.
  * Configure Redis using a URL or explicit connection details (explicit credentials override URL) via `layer` and `layerConfig`.
  * Enables Redis-backed cache and queue persistence with improved error handling.
* **Bug Fixes**
  * Improved queue failure handling by moving exhausted items to the failed queue.
* **Tests**
  * Added Deno Redis integration coverage for cache/queue persistence, connection lifecycle, and correct `AUTH` credential behavior.

### DIFF
--- a/packages/platform-deno/src/DenoRedis.ts
+++ b/packages/platform-deno/src/DenoRedis.ts
@@ -15,6 +15,8 @@ import * as Context from "effect/Context"
 import * as Effect from "effect/Effect"
 import * as Fn from "effect/Function"
 import * as Layer from "effect/Layer"
+import * as Predicate from "effect/Predicate"
+import * as Record from "effect/Record"
 import * as Redis from "effect/unstable/persistence/Redis"
 
 /**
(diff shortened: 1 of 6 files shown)


packages/platform-deno/src/DenoRedis.ts
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export type RedisOptions = Omit<RedisConnectOptions, "hostname"> & {
+  readonly hostname?: string
+  readonly url?: string
+}
+
+const withoutUndefined = <A extends object>(options: A): Partial<A> =>
```
tim-smart: Get rid of this helper

packages/platform-deno/src/DenoRedis.ts
```diff
@@ -0,0 +1,121 @@
...
+      try: () => {
+        const { url, ...connectOptions } = options
+        if (url === undefined) {
+          return connect({ hostname: "localhost", ...withoutUndefined(connectOptions) })
+        }
+        const { name: username, ...parsed } = parseURL(url)
+        return connect({
+          ...withoutUndefined({ ...parsed, username }),
+          ...withoutUndefined(connectOptions)
+        } as RedisConnectOptions)
```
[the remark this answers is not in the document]
tim-smart: Verified, but this does not apply in the current structure. RedisConnectOptions.signal is a caller callback (() => AbortSignal), not an AbortSignal. More importantly, Effect.acquireRelease protects acquisition with an uninterruptible region by default, so the tryPromise signal cannot abort during the connect handshake; interruption is delivered after acquisition and the registered release closes the client atomically. Passing that signal would therefore be a no-op for this case and could incorrectly replace the caller-provided timeout signal, which remains supported. The interrupt-during-connect regression covers this behavior.

packages/platform-deno/src/DenoRedis.ts
tim-smart: Removed in f61108af2. Connection options are now merged explicitly field by field, skipping undefined overrides. The URL-password and explicit-credential precedence cases are both covered by wire-level tests.

PR #6691, opened 2026-07-28
Allow enabling websocket compression in NodeHttpServer and BunHttpServer

t3dotgg:
tl;dr - I want to add compression to websockets in T3 Code. It will [help cut data transferred on T3 Code by as much as 80%](https://github.com/pingdotgg/t3code/pull/4705)

Happy to talk more about this. Figured the patch was simple enough to file a PR quick - no hard feelings if this gets closed :)

---- FROM HERE DOWN IS AI SLOP ----

Browsers offer `permessage-deflate` on every WebSocket connection, but an Effect HTTP server has no way to accept it. `NodeHttpServer` hardcodes `new WebSocketServer({ noServer: true })`, and `BunHttpServer` overwrites any `websocket` key in the serve options with its own handlers — so there is no way to enable compression or tune payload limits without patching the packages.

For apps streaming JSON over websockets this leaves a lot on the table: measured on real production traffic (small ~500B-1KB frames with repeated keys and UUIDs, raw TCP byte counts), accepting `permessage-deflate` with context takeover cuts wire bytes by **4x**.

## Solution

Both servers accept an optional `websocket` field in their existing options, forwarded to the underlying implementation. The wiring/lifecycle options the server manages (`noServer`, `open`/`message`/`close`, etc.) are excluded from the type so they cannot be overridden.

```ts
// Node: forwarded to the ws WebSocketServer
NodeHttpServer.layer(() => createServer(), {
  port: 3000,
  websocket: { perMessageDeflate: true }
})

// Bun: merged into Bun.serve's websocket handler
BunHttpServer.layer({
  port: 3000,
  websocket: { perMessageDeflate: true }
})
```

Omitting the option keeps today's behavior exactly; clients that don't offer the extension keep uncompressed frames on their connection.

Includes a test asserting the server accepts `permessage-deflate` when configured and offered, and that clients not offering it still connect uncompressed. `tsc -b` on both packages, oxlint, dprint, and jsdocs are clean; a changeset is included.

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Fable 5), reviewed and filed by a human.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable `websocket` option to both Node.js and Bun HTTP servers.
  * Enables forwarding WebSocket configuration (for example, per-message deflate settings) while keeping connection lifecycle behavior managed by the servers.
* **Bug Fixes**
  * Preserved and applied user-provided WebSocket settings instead of overriding them with fixed defaults.
* **Tests**
  * Added coverage to verify WebSocket option forwarding, including per-message compression behavior.  
* **Documentation**
  * Updated release documentation with usage examples for Node.js and Bun.

### DIFF
--- a/.changeset/http-server-websocket-options.md
+++ b/.changeset/http-server-websocket-options.md
@@ -1,0 +1,24 @@
+---
+"@effect/platform-node": patch
+"@effect/platform-bun": patch
+---
+
+Allow configuring the WebSocket server in `NodeHttpServer` and `BunHttpServer`.
+
+The HTTP server layers previously hardcoded their WebSocket wiring — `NodeHttpServer` always constructed `new WebSocketServer({ noServer: true })`, and `BunHttpServer` overwrote any `websocket` key in the serve options with its own handlers. There was no way to enable `permessage-deflate` compression (which browsers offer on every connection) or tune payload limits.
+
+Both servers now accept a `websocket` option that is forwarded to the underlying implementation, with the wiring/lifecycle options the server manages excluded from the type:
+
+```ts
+// Node: forwarded to the `ws` WebSocketServer
+NodeHttpServer.layer(() => createServer(), {
+  port: 3000,
+  websocket: { perMessageDeflate: true }
+})
+
+// Bun: merged into Bun.serve's websocket handler
+BunHttpServer.layer({
+  port: 3000,
+  websocket: { perMessageDeflate: true }
+})
+```
--- a/.changeset/http-server-websocket-options.md
+++ b/.changeset/http-server-websocket-options.md
@@ -5,8 +5,6 @@
 
 Allow configuring the WebSocket server in `NodeHttpServer` and `BunHttpServer`.
 
-The HTTP server layers previously hardcoded their WebSocket wiring — `NodeHttpServer` always constructed `new WebSocketServer({ noServer: true })`, and `BunHttpServer` overwrote any `websocket` key in the serve options with its own handlers. There was no way to enable `permessage-deflate` compression (which browsers offer on every connection) or tune payload limits.
-
 Both servers now accept a `websocket` option that is forwarded to the underlying implementation, with the wiring/lifecycle options the server manages excluded from the type:
 
 ```ts
--- a/packages/platform-bun/src/BunHttpServer.ts
+++ b/packages/platform-bun/src/BunHttpServer.ts
@@ -67,6 +67,25 @@ export type ServeOptions<R extends string> =
   & { readonly routes?: Bun.Serve.Routes<WebSocketContext, R> }
 
 /**
+ * WebSocket tuning options forwarded to `Bun.serve`'s `websocket` handler.
+ *
+ * **Details**
+ *
+ * The lifecycle handlers (`open`, `message`, `close`, ...) are managed by the
+ * server and cannot be overridden; everything else — such as
+ * `perMessageDeflate` compression, payload limits, and idle timeouts — passes
+ * through, e.g.
+ * `BunHttpServer.layer({ port: 3000, websocket: { perMessageDeflate: true } })`.
+ *
+ * @category options
+ * @since 4.0.0
+ */
+export type WebSocketOptions = Omit<
+  Bun.WebSocketHandler<WebSocketContext>,
+  "open" | "message" | "close" | "drain" | "ping" | "pong" | "data"
+>
+
+/**
  * Creates a scoped Bun `HttpServer` from `Bun.serve` options, stopping the server on scope finalization with optional graceful shutdown settings.
  *
  * @category constructors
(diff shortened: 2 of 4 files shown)


.changeset/http-server-websocket-options.md
```diff
@@ -0,0 +1,24 @@
+---
+"@effect/platform-node": patch
+"@effect/platform-bun": patch
+---
+
+Allow configuring the WebSocket server in `NodeHttpServer` and `BunHttpServer`.
+
+The HTTP server layers previously hardcoded their WebSocket wiring — `NodeHttpServer` always constructed `new WebSocketServer({ noServer: true })`, and `BunHttpServer` overwrote any `websocket` key in the serve options with its own handlers. There was no way to enable `permessage-deflate` compression (which browsers offer on every connection) or tune payload limits.
+
+Both servers now accept a `websocket` option that is forwarded to the underlying implementation, with the wiring/lifecycle options the server manages excluded from the type:
```
tim-smart, line 8:
```suggestion
Allow configuring the WebSocket server in `NodeHttpServer` and `BunHttpServer`.

Both servers now accept a `websocket` option that is forwarded to the underlying implementation, with the wiring/lifecycle options the server manages excluded from the type:
```

packages/platform-bun/src/BunHttpServer.ts
```diff
@@ -66,6 +66,25 @@ export type ServeOptions<R extends string> =
...
+ * `perMessageDeflate` compression, payload limits, and idle timeouts — passes
+ * through, e.g.
+ * `BunHttpServer.layer({ port: 3000, websocket: { perMessageDeflate: true } })`.
+ *
+ * @category options
+ * @since 4.0.0
+ */
+export type WebSocketOptions = Omit<
+  Bun.WebSocketHandler<WebSocketContext>,
+  "open" | "message" | "close" | "drain" | "ping" | "pong" | "data"
```
tim-smart: I think we should also drop "binaryType" here too, so it can't be set to arraybuffer.

packages/platform-node/src/NodeHttpServer.ts
```diff
@@ -62,6 +62,27 @@ import * as NodeMultipart from "./NodeMultipart.ts"
...
+ * @since 4.0.0
+ */
+export interface Options extends Net.ListenOptions {
+  readonly disablePreemptiveShutdown?: boolean | undefined
+  readonly gracefulShutdownTimeout?: Duration.Input | undefined
+  readonly websocket?:
+    | Omit<NodeWS.ServerOptions, "noServer" | "server" | "host" | "port" | "path">
+    | undefined
+}
+
```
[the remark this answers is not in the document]
tim-smart: Keeping it a type-only omit without runtime checks is fine.

tim-smart (changes requested):

packages/platform-node/src/NodeHttpServer.ts
```diff
@@ -62,6 +62,27 @@ import * as NodeMultipart from "./NodeMultipart.ts"
...
 
+/**
+ * Options accepted by the Node `HttpServer` constructors and layers.
+ *
+ * **Details**
+ *
+ * `websocket` is forwarded to the underlying `ws` `WebSocketServer`, minus the
+ * wiring options the server manages itself. Use it to enable
+ * `permessage-deflate` compression or tune payload limits, e.g.
+ * `NodeHttpServer.layer(() => createServer(), { port: 3000, websocket: { perMessageDeflate: true } })`.
```
moishinetzer: Have this comment inline with the property or comment on all the property keys.

moishinetzer (commented):
Small change, otherwise LGTM. I'm griping with the fact that shovelling down a config to the constructor feels like its not taking advantage of effect by creating a service with a default layer for the config but this is small enough that I'll leave it to you or @tim-smart whether that makes sense here.

.changeset/http-server-websocket-options.md
t3dotgg: Applied in 51c31fb — changeset trimmed to the suggested copy.

packages/platform-bun/src/BunHttpServer.ts
t3dotgg: Done in 51c31fb — `binaryType` added to the omit. Good call: the server's `message` handler feeds `Socket.fromWebSocket`, which expects the default `nodebuffer` framing, so letting callers flip it to `arraybuffer` would break message decoding.

packages/platform-node/src/NodeHttpServer.ts
t3dotgg: Moved inline onto the `websocket` property in 51c31fb. Left the neighboring shutdown options as they were since they predate this PR.

tim-smart (approved):

PR #6703, opened 2026-07-28
Fix docgen source links for modules in nested directories

leonitousconforti:
## Summary

`[Source](...)` links in docgen's generated markdown are built from only the file's base name:

```ts
const name = source.sourceFile.getBaseName()
return `\n\n[Source](${config.srcLink}${name}#L${position.line})`
```

For modules in subdirectories of `srcDir` (e.g. `src/frida/FridaRpcClient.ts`), the link drops the subdirectory and points one level too high (`<srcLink>/FridaRpcClient.ts` instead of `<srcLink>/frida/FridaRpcClient.ts`), producing 404s.

This changes the printer to use the module's path relative to `srcDir` (`source.path.slice(1).join("/")`), which is the same convention `getModuleMarkdownOutputPath` already uses for the generated markdown file layout. Links for modules directly under `srcDir` are unchanged.

## Notes

- The `makeSource` test helper previously modeled `Source.path` as `[basename]`, which doesn't match production (`file.path.split(path.sep)` including the `srcDir` prefix). It now derives the path from the source file's actual file path with a `src` prefix, matching what `parseFile` produces.
- Added a regression test covering a module in a nested directory.

Noticed while adopting `@effect/docgen@4.0.0-beta.102` in a project with nested modules: https://github.com/leonitousconforti/efffrida/pull/158

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed generated documentation **Source** links for modules in nested directories.
  - Source links now display and resolve the correct relative folder path and line reference instead of targeting the wrong directory level.

- **Tests**
  - Updated and expanded parser/docgen test coverage to ensure nested in-memory source files produce correctly formed **Source** links.

### DIFF
--- a/.changeset/fix-docgen-nested-source-links.md
+++ b/.changeset/fix-docgen-nested-source-links.md
@@ -1,0 +1,7 @@
+---
+"@effect/docgen": patch
+---
+
+Fix source links for modules in nested directories
+
+`[Source]` links in the generated markdown were built from only the file's base name, so modules in subdirectories of `srcDir` (e.g. `src/nested/File.ts`) linked one directory too high. Links now use the module's path relative to `srcDir`, matching how the markdown output paths are computed.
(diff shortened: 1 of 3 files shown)


fubhy:
Thanks. I am currently rewriting `docgen` to produce the semantic structure as needed for our new api reference docs on the new website. This issue is already resolved in the rewrite.

leonitousconforti:
Sweet!

PR #6705, opened 2026-07-28
Restore FileSystem.watch recursive control

### LINKED ISSUE #6698: FileSystem.watch: recursive control removed, Node/Bun watches forced recursive
In Effect 3, `FileSystem.watch(path, { recursive: false })` observes only direct children, and `{ recursive: true }` observes nested paths. In `4.0.0-beta.102` the options parameter is gone from `FileSystem.watch` and `WatchBackend.register`, and `@effect/platform-node-shared`'s `NodeFileSystem` calls `node:fs.watch` with `{ recursive: true }` unconditionally. `@effect/platform-bun` delegates to that layer.

The result is that every Node/Bun watch is now recursive, with no typed way to request the previous behaviour.

## Reproduction

Create `root/nested/`, watch `root`, then create `nested/child.txt` followed by `root/direct.txt`.

- **Effect 3** with `{ recursive: false }` — reports only `direct.txt`
- **beta.102** — reports both `nested/child.txt` and `direct.txt`

## Why this matters

Silently widening a watch is not a neutral default change. Build and codegen watchers that deliberately watch one directory level now receive events for the entire subtree, which can cause spurious rebuilds or feedback loops where a watcher's own output re-triggers it. Because the failure presents as flakiness rather than an error, it is unusually hard to attribute back to the upgrade.

## Prior art

`recursive` was not incidental — it was requested in #2986 and implemented for exactly this purpose, to opt *into* recursive watching because the default observed only the top level. beta.102 removes that control and makes the opposite behaviour mandatory.

(#5913 is a separate v3 issue about accidentally falling back to *non-recursive* watching through layer order; it does not cover removal plus forced recursion.)

## Request

Restore recursive control through the core interface, `WatchBackend`, and the Node implementation, with `false` preserving the Effect 3 non-recursive behaviour.

## Versions

- effect: 4.0.0-beta.102
- @effect/platform-node / platform-node-shared: 4.0.0-beta.102
- Compared against: effect 3.21.4

<details>
<summary>Posted on behalf of @schickling</summary>

| field | value |
| --- | --- |
| `agent_name` | cl1-heron |
| `agent_session_id` | 54711470-ae7e-4322-a3e2-129a2689a097 |
| `agent_tool` | Claude Code |
| `agent_tool_version` | 2.1.220 |
| `agent_runtime` | Claude Code 2.1.220 |
| `agent_model` | claude-opus-5 |
| `runtime_profile` | /nix/store/xg6r3nkr7spjyg2gbl3zprqbsag3fsjc-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/2vkpfkkbc9hjmcwknzkv62v2i44an8mz-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | effect-utils/schickling/2026-07-28-effect-4 |
| `machine` | dev3 |
| `tooling_profile` | dotfiles@unknown-dirty |
</details>

tylergibbs1:
## Summary

- restore the v3 `WatchOptions` API on `FileSystem.watch`
- forward the option through `WatchBackend` and the Node/Deno native implementations
- default omitted or `false` options to non-recursive watching
- add Node regression coverage for both direct-child and recursive behavior
- add a patch changeset for the affected packages

## Why

The v4 migration removed the watch options and hard-coded recursive watching in the Node-compatible implementation used by Node and Bun. That silently widened existing watchers to entire directory trees and could trigger extra rebuilds or feedback loops.

## Root cause

The established option from #5174 was dropped while the filesystem API moved into `effect`, and native watchers were changed to `recursive: true` instead of forwarding caller intent.

## Impact

`fs.watch(path)` and `fs.watch(path, { recursive: false })` again observe only direct children. Passing `{ recursive: true }` opts into subtree events.

## Verification

- focused regression tests added for `recursive: false` and `recursive: true`
- branch diff audited against current `main`
- GitHub Check and Snapshot workflows are awaiting the repository's required approval for forked workflows; no jobs have run yet

Closes #6698

## Summary by CodeRabbit

* **New Features**
  * Extended `FileSystem.watch` to accept optional `WatchOptions` (including `recursive?: boolean`) and threaded options through the Node and Deno implementations.
* **Bug Fixes**
  * Restored `recursive` handling for filesystem watching; non-recursive watching is now the default when `recursive` isn’t specified.
* **Tests**
  * Added/expanded Node filesystem watch tests to confirm first detected changes for `recursive: true` vs non-recursive (including the default).
* **Chores**
  * Added a changeset for patch-level release publishing.

### DIFF
--- a/.changeset/wise-files-watch.md
+++ b/.changeset/wise-files-watch.md
@@ -1,0 +1,7 @@
+---
+"@effect/platform-deno": patch
+"@effect/platform-node-shared": patch
+"effect": patch
+---
+
+Restore the `recursive` option for `FileSystem.watch`, with non-recursive watching as the default.
--- a/packages/effect/src/FileSystem.ts
+++ b/packages/effect/src/FileSystem.ts
@@ -341,9 +341,13 @@ export interface FileSystem {
     mtime: Date | number
   ) => Effect.Effect<void, PlatformError>
   /**
-   * Watch a directory or file for changes
+   * Watch a directory or file for changes.
+   *
+   * By default, only changes to the direct children of the directory are
+   * reported. Set the `recursive` option to `true` to watch for changes in
+   * subdirectories as well.
    */
-  readonly watch: (path: string) => Stream.Stream<WatchEvent, PlatformError>
+  readonly watch: (path: string, options?: WatchOptions) => Stream.Stream<WatchEvent, PlatformError>
   /**
    * Write data to a file at `path`.
    */
@@ -1248,6 +1252,19 @@ export declare namespace File {
 export type SeekMode = "start" | "current"
 
 /**
+ * Options for watching files or directories.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export interface WatchOptions {
+  /**
+   * When `true`, changes in subdirectories are also reported.
+   */
+  readonly recursive?: boolean | undefined
+}
+
+/**
  * Represents file system events emitted when watching files or directories.
  *
  * **When to use**
@@ -1363,5 +1380,9 @@ export declare namespace WatchEvent {
  * @since 4.0.0
  */
 export class WatchBackend extends Context.Service<WatchBackend, {
-  readonly register: (path: string, stat: File.Info) => Option.Option<Stream.Stream<WatchEvent, PlatformError>>
+  readonly register: (
+    path: string,
+    stat: File.Info,
+    options?: WatchOptions
+  ) => Option.Option<Stream.Stream<WatchEvent, PlatformError>>
 }>()("effect/platform/FileSystem/WatchBackend") {}
--- a/packages/effect/src/FileSystem.ts
+++ b/packages/effect/src/FileSystem.ts
@@ -343,6 +343,8 @@ export interface FileSystem {
   /**
    * Watch a directory or file for changes.
    *
+   * **Details**
+   *
    * By default, only changes to the direct children of the directory are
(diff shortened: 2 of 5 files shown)


tylergibbs1:
Fixed the actionable Check failures in 977d23e: updated the new tests to the v4 `Effect.forkChild` API and moved the extended watch documentation under a `Details` section for the JSDoc linter. A fresh CI run should start from this push.

tylergibbs1:
Addressed the latest review in 5724b7c: the watch tests now establish a real readiness handshake by repeatedly touching a sentinel until the stream observes it, and the sentinel is filtered from assertions. I also added explicit coverage for `fs.watch(root)` with omitted options defaulting to non-recursive behavior.

tylergibbs1:
Addressed the watcher-lifecycle review in d67acc0. The readiness wait now races the watcher fiber’s completion, so startup errors or early closure surface immediately, and an `ensuring` finalizer always interrupts the sentinel writer.

tim-smart (approved):

tylergibbs1:
Fixed the Node CI timeout in a15fc7165. The readiness signal used the test clock, so after its first pre-watcher write it never retried; it now delays on the live clock before writing, giving the watcher fiber time to acquire. Verified the focused NodeFileSystem suite 10 consecutive times on Linux/Node 26 (18/18 each), plus oxlint, dprint, and the package TypeScript build.

PR #6722, opened 2026-07-28
Fix OpenAI stable web search action parameters

mrtdurdenthe2:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Stable OpenAI `web_search` responses currently become Effect tool calls with `params: {}`, while `OpenAiTool.WebSearch` requires `params.action`. A successful provider response therefore fails Effect response decoding with `InvalidOutputError: Missing key at ["params"]["action"]`.

This change preserves the existing public tool schema and maps the completed provider action into stable web search tool-call parameters. For streaming responses, the initial `response.output_item.added` event may not contain `action`, so the stable tool call is emitted when `response.output_item.done` provides the completed action. `WebSearchPreview` keeps its existing empty-parameter behavior.

## OpenAI API Reference

OpenAI documents that Responses API web search output includes “a `web_search_call` output item with the ID of the search call, along with the action taken in `web_search_call.action`.”

https://platform.openai.com/docs/guides/tools-web-search#output-and-citations

The documented response shape includes:

```json
{
  "type": "web_search_call",
  "id": "ws_...",
  "status": "completed",
  "action": {
    "type": "search",
    "query": "latest news about AI"
  }
}
```

## Testing

- Added deterministic non-streaming coverage with `gpt-4.1` and `gpt-5.6` model configurations.
- Added streaming coverage where the partial item omits `action` and the completed item supplies it.
- Verified the provider action remains available in both tool-call parameters and the tool result.
- `pnpm test --run`: 96 tests passed.
- `pnpm check`: passed.
- `pnpm build`: passed.
- dprint, oxlint, and `git diff --check`: passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OpenAI web search responses so search actions are correctly preserved in tool calls and results.
  * Improved streamed web search handling by waiting for complete action details before emitting tool-call events.
  * Ensured stable web search actions include the expected search type and query.

### DIFF
--- a/packages/ai/openai/test/OpenAiLanguageModel.test.ts
+++ b/packages/ai/openai/test/OpenAiLanguageModel.test.ts
@@ -752,6 +752,39 @@ describe("OpenAiLanguageModel", () => {
           }
         }))))
 
+      for (const model of ["gpt-4.1", "gpt-5.6"] as const) {
+        it.effect(`maps stable web search action to tool call parameters with ${model}`, () =>
+          Effect.gen(function*() {
+            const toolkit = Toolkit.make(OpenAiTool.WebSearch({}))
+            const result = yield* LanguageModel.generateText({
+              prompt: "Search the web",
+              toolkit
+            }).pipe(Effect.provide(OpenAiLanguageModel.model(model)))
+
+            const toolCall = result.content.find((part) => part.type === "tool-call")
+            assert.isDefined(toolCall)
+            if (toolCall?.type === "tool-call") {
+              deepStrictEqual(toolCall.params, {
+                action: { type: "search", query: "Effect TypeScript" }
+              })
+            }
+
+            const toolResult = result.content.find((part) => part.type === "tool-result")
+            assert.isDefined(toolResult)
+            if (toolResult?.type === "tool-result") {
+              deepStrictEqual(toolResult.result, {
+                action: { type: "search", query: "Effect TypeScript" },
+                status: "completed"
+              })
+            }
+          }).pipe(Effect.provide(makeTestLayer({
+            body: {
+              model,
+              output: [makeWebSearchCall()]
+            }
+          }))))
+      }
+
       it.effect("uses canonical OpenAiMcp name for mcp_approval_request", () =>
         Effect.gen(function*() {
           const result = yield* LanguageModel.generateText({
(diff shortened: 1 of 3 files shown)


packages/ai/openai/test/OpenAiLanguageModel.test.ts
```diff
@@ -752,6 +752,39 @@ describe("OpenAiLanguageModel", () => {
           }
         }))))
 
+      for (const model of ["gpt-4.1", "gpt-5.6"] as const) {
```
IMax153: Let's use `it.each` here and run the effect tests using `Effect.runPromise`

packages/ai/openai/test/OpenAiLanguageModel.test.ts
```diff
@@ -752,6 +752,39 @@ describe("OpenAiLanguageModel", () => {
...
+              toolkit
+            }).pipe(Effect.provide(OpenAiLanguageModel.model(model)))
+
+            const toolCall = result.content.find((part) => part.type === "tool-call")
+            assert.isDefined(toolCall)
+            if (toolCall?.type === "tool-call") {
+              deepStrictEqual(toolCall.params, {
+                action: { type: "search", query: "Effect TypeScript" }
+              })
+            }
```
IMax153: Are these if statements really required? We should be able to narrow the type of the part using a refinement in the `.find` above and then directly use `assert.deepStrictEqual` instead of the `deepStrictEqual` helper, which is stricter on types.

Same comment on the other test added in this PR.

packages/ai/openai/test/OpenAiLanguageModel.test.ts
```diff
@@ -1099,6 +1132,64 @@ describe("OpenAiLanguageModel", () => {
...
+          prompt: "Search the web",
+          toolkit,
+          disableToolCallResolution: true
+        }).pipe(
+          Stream.runCollect,
+          Effect.provide(OpenAiLanguageModel.model("gpt-4o-mini")),
+          Effect.provide(makeStreamTestLayer(streamEvents))
+        )
+
+        const parts = globalThis.Array.from(partsChunk)
```
IMax153: This should not be necessary as the `partsChunk` should already be an array.

mrtdurdenthe2:
@IMax153 Sorry, not sure if you get notifications for resolving the reviews.

Also thanks for the reviews :smiley:

IMax153 (approved):

PR #6744, opened 2026-07-29
Render bundle size comments as tables

IMax153:
## Summary

- stage bundle statistics in a bounded temporary file and validate the same bytes that are emitted
- render validated bundle statistics as a Markdown table instead of a fenced text block
- reject symlinks, non-regular artifacts, unexpected rows, injected content, and oversized reports
- replace invalid artifacts with fixed trusted messages

This preserves the workflow hardening from #6736 while restoring table rendering.

## Validation

- validated the report from the malformed comment through the bounded temporary-file path
- verified an injected HTML/script line is rejected
- parsed `.github/workflows/bundle-comment.yml` with Ruby YAML
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automated bundle statistics reporting in comments.
  * Added checks to ensure the stats source is present, valid, and not a symlink.
  * Added byte/line size limits to omit overly large stats with clearer messaging.
  * Added format validation for expected table-like output; show an “invalid format” omission when it doesn’t match.
  * Preserves full valid statistics output without prior truncation or wrapping.

### DIFF
--- a/.github/workflows/bundle-comment.yml
+++ b/.github/workflows/bundle-comment.yml
@@ -33,14 +33,27 @@ jobs:
         run: |
           max_bytes=48000
           max_lines=500
-          stats_file="bundle-stats/stats.txt"
+          stats_source="bundle-stats/stats.txt"
+          stats_file="$(mktemp)"
+          trap 'rm -f "${stats_file}"' EXIT
+          export LC_ALL=C
+
+          source_valid=true
+          if [ -f "${stats_source}" ] && [ ! -L "${stats_source}" ]; then
+            head -c "$((max_bytes + 1))" "${stats_source}" > "${stats_file}"
+          else
+            source_valid=false
+          fi
+
           byte_count="$(wc -c < "${stats_file}")"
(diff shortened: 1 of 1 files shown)


IMax153:
The `tr` command replaced every backtick (octal `140`) with an apostrophe (octal `047`).

The artifact is controlled by PR code. Because its contents were placed inside a fenced code block, a malicious artifact could include triple backticks, close the fence, and inject arbitrary Markdown or HTML into the privileged bot comment. Replacing backticks prevented that escape, but it also changed legitimate filename formatting from inline code to quoted text.

This PR no longer needs that transformation because it:

- bounds the artifact before processing
- validates every header and row against a strict grammar
- allows backticks only as the expected filename delimiters
- restricts filenames to `[[:alnum:]_.-]+`
- rejects additional Markdown, HTML, malformed rows, and fence sequences
- emits the same temporary file that was validated

This keeps inline-code filenames safe while allowing the validated Markdown table to render normally.

PR #6777, opened 2026-07-30
Bound pending SSE decoder state

tim-smart:
## Summary

- bound combined pending SSE line and event data state with a configurable 10 MiB default
- surface overflow as a typed `SseError` across low-level SSE stream contracts
- map provider stream overflows to `AiError.InvalidOutputError`

## Testing

- `pnpm lint-fix`
- `pnpm check`
- `pnpm --filter effect test --run test/unstable/encoding/Sse.test.ts`
- `pnpm --filter @effect/openapi-generator test --run test/OpenApiGenerator.test.ts`
- `pnpm test-types HttpApiClient.tst.ts`
- targeted Anthropic, OpenAI, OpenAI-compatible, and OpenRouter client tests

Closes EFF-218

### DIFF
--- a/packages/effect/src/unstable/encoding/Sse.ts
+++ b/packages/effect/src/unstable/encoding/Sse.ts
@@ -22,6 +22,50 @@ import * as Result from "../../Result.ts"
 import * as Schema from "../../Schema.ts"
 import * as SchemaTransformation from "../../SchemaTransformation.ts"
 
+const SseErrorTypeId = "~effect/encoding/Sse/SseError"
+
+/**
+ * Error raised when decoding a Server-Sent Events stream fails.
+ *
+ * @category errors
+ * @since 4.0.0
+ */
+export class SseError extends Data.TaggedError("SseError")<{
+  readonly reason: "EventTooLarge"
+  readonly maxEventSize: number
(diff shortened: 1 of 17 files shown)


packages/effect/src/unstable/encoding/Sse.ts
```diff
@@ -22,6 +22,50 @@ import * as Result from "../../Result.ts"
...
+const SseErrorTypeId = "~effect/encoding/Sse/SseError"
+
+/**
+ * Error raised when decoding a Server-Sent Events stream fails.
+ *
+ * @category errors
+ * @since 4.0.0
+ */
+export class SseError extends Data.TaggedError("SseError")<{
+  readonly reason: "EventTooLarge"
```
tim-smart: Create another error class for the EventTooLarge error, and use that for the reason.

You can see the correct reason error pattern in other modules.

tim-smart: Implemented in 5e5d909e8. EventTooLarge is now a dedicated tagged error reason, SseError wraps SseErrorReason and delegates its message, and the overflow tests assert the nested reason.

PR #6778, opened 2026-07-30
Bound pending SSE decoder state on v3

tim-smart:
## Summary

- bound combined pending SSE line and event data state with a configurable 10 MiB default
- return a typed `SseError` from the parser and terminate channel decoding on overflow
- cover both overflow paths and normal boundary/multi-chunk parsing

## Testing

- `pnpm lint-fix`
- `pnpm test run packages/experimental/test/Sse.test.ts`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`

Closes EFF-224

### DIFF
--- a/packages/experimental/src/Sse.ts
+++ b/packages/experimental/src/Sse.ts
@@ -1,6 +1,7 @@
 /**
  * @since 1.0.0
  */
+import * as Cause from "effect/Cause"
 import * as Channel from "effect/Channel"
 import * as Chunk from "effect/Chunk"
 import * as Data from "effect/Data"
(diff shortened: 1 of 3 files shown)


packages/experimental/src/Sse.ts
```diff
@@ -10,13 +11,62 @@ import * as Mailbox from "effect/Mailbox"
...
+ * @category errors
+ */
+export type SseErrorReason = EventTooLarge
+
+/**
+ * @since 1.0.0
+ * @category errors
+ */
+export class SseError extends Data.TaggedError("SseError")<{
+  readonly reason: SseErrorReason
```
tim-smart: On v3 this can be renamed to `cause`

tim-smart: Renamed the v3 error field to `cause` and updated the assertions in bdf2f6816.

PR #6780, opened 2026-07-30
Emit MCP tool output schemas

### LINKED ISSUE #6753: McpServer: tools/list does not emit outputSchema for tools with a typed success schema
## Summary

MCP (2025-06-18) defines an optional `outputSchema` field on `Tool` so servers can declare the JSON Schema of a tool's structured output. `McpServer.tool` already uses the typed success schema at runtime — call results are encoded to `structuredContent` — but `tools/list` never emits `outputSchema`, so clients cannot validate or type results without out-of-band knowledge.

## Reproduction (`effect@4.0.0-beta.101`)

```ts
import { McpServer, Tool, Toolkit } from "effect/unstable/ai"

const MyTool = Tool.make("my_tool", {
  description: "returns a typed result",
  parameters: { q: Schema.String },
  success: Schema.Struct({ answer: Schema.String }),
})

// Register via McpServer.registerToolkit / layerHttp, then call tools/list:
// each entry has `inputSchema` (JSON Schema for parameters) but no
// `outputSchema`, even though the success schema is known statically and is
// already used to encode `structuredContent` on tools/call responses.
```

`Tool` entries in `McpSchema` currently model `name` / `description` / `inputSchema` / `annotations` only — there is no `outputSchema` field to populate.

## Ask

Emit `outputSchema` on `tools/list` entries for tools whose success schema is defined, mirroring how `inputSchema` is derived (JSON Schema generation from the Effect schema). Clients that ignore it lose nothing; spec-aware clients gain result validation.

tim-smart:
## Summary

- add optional MCP tool output schemas to `tools/list`
- omit `outputSchema` for tools without typed success schemas
- cover typed and untyped tools in the MCP server tests

## Testing

- `pnpm --filter effect test --run test/unstable/ai/McpServer.test.ts`
- `pnpm lint-fix`
- `pnpm check`
- `git diff --check`

Closes EFF-226
Closes #6753

### DIFF
--- a/packages/effect/src/unstable/ai/McpServer.ts
+++ b/packages/effect/src/unstable/ai/McpServer.ts
@@ -888,7 +888,7 @@ export const registerToolkit: <Tools extends Record<string, Tool.Any>>(
       name: tool.name,
       description: Tool.getDescription(tool),
       inputSchema: Tool.getJsonSchema(tool),
-      ...(tool.successSchema === Schema.Void || tool.successSchema === Schema.Unknown ? {} : {
+      ...(SchemaAST.isVoid(tool.successSchema.ast) || SchemaAST.isUnknown(tool.successSchema.ast) ? {} : {
         outputSchema: Tool.getJsonSchemaFromSchema(tool.successSchema)
       }),
       annotations: {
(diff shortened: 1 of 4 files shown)


packages/effect/src/unstable/ai/McpServer.ts
```diff
@@ -888,6 +888,9 @@ export const registerToolkit: <Tools extends Record<string, Tool.Any>>(
       name: tool.name,
       description: Tool.getDescription(tool),
       inputSchema: Tool.getJsonSchema(tool),
+      ...(tool.successSchema === Schema.Void || tool.successSchema === Schema.Unknown ? {} : {
```
tim-smart: This should check the ast instead of comparing against the schema instances

tim-smart: Actually I think every tool has a valid success schema, so we could always include `outputSchema`. Validate this against the mcp spec first before changing anything.

PR #6799, opened 2026-07-30
Upgrade Changesets and restore release queues

IMax153:
## Summary

- upgrade Changesets to `@changesets/cli@3.0.0-next.11` and `@changesets/changelog-github@1.0.0-next.9`
- migrate 925 already-versioned prerelease changesets into `.changeset/pre/`, preserving their contents byte-for-byte
- retain commit-author attribution and deduplicate GitHub lookups in the updated changelog dependency patch
- remove an accidental upstream debug log from `@changesets/read`
- keep all 30 publishable packages, including the new `@effect/doctest`, in the fixed release group
- restore stacked `v4/next-minor` and `v4/next-major` release queues while keeping publishing on `main`
- run the normal check workflow for both queue branches

## Rollout

Before merging, verify bot access and protections for the queue branches. Fork PR approvals reuse the existing `fork` environment. Changesets already on `main` remain in the current Version Packages release and will not be routed retroactively.

## Validation

- `pnpm changeset status`
- `pnpm lint-fix`
- `pnpm check`
- verified all 925 prerelease changesets were migrated without content changes
- verified the fixed group covers all 30 publishable packages
- parsed the affected workflow YAML files
- verified GitHub lookup deduplication and release-plan parity


patches/@changesets__assemble-release-plan.patch
```diff
@@ -1,26 +0,0 @@
-diff --git a/dist/changesets-assemble-release-plan.cjs.js b/dist/changesets-assemble-release-plan.cjs.js
-index e07ba6e793021b6cfdec898afca517e293386ddb..88d80a95fbe739996918ef4883601b4388926123 100644
---- a/dist/changesets-assemble-release-plan.cjs.js
-+++ b/dist/changesets-assemble-release-plan.cjs.js
-@@ -215,7 +215,7 @@ function determineDependents({
-             preInfo,
-             onlyUpdatePeerDependentsWhenOutOfRange: config.___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH.onlyUpdatePeerDependentsWhenOutOfRange
-           })) {
--            type = "major";
-+            type = "minor";
```
tim-smart, line 10: Does the new changeset version support this by default now?

IMax153: Yeah, according to my agent, the new version no longer has the logic that caused the unwanted major bump.


> I validated it by generating the release plan twice from the current changesets:
>
> 1. Using the existing Changesets version with this patch applied.
> 2. Using the upgraded Changesets version with the patch removed.
> 
> I compared each package name, release type, and resulting version between the two plans and there were no differences. I also ran pnpm changeset status with the upgraded setup and confirmed all 29 publishable packages are included in the fixed release group.

IMax153: Found it upstream:
- PR: changesets/changesets#2090 (https://github.com/changesets/changesets/pull/2090)
- Commit: 3aae903 (https://github.com/changesets/changesets/commit/3aae903a668315c495ba09248bad2b9c63424449)
- Released in @changesets/assemble-release-plan@7.0.0-next.7; this PR installs next.9.

.github/workflows/release-queue.yml
```diff
@@ -0,0 +1,44 @@
...
+  push:
+    branches: [main, v4/next-minor, v4/next-major]
+
+permissions: {}
+
+jobs:
+  approval-gate:
+    if: github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: ubuntu-latest
+    environment: release-queue
```
tim-smart:
```suggestion
    environment: fork
```

We can use the same fork environment as the preview builds. Hopefully it will only require one approval per PR.

tim-smart (approved):

PR #6802, opened 2026-07-31
Cap RPC streaming decoder buffers

tim-smart:
## Summary

- cap incomplete NDJSON and MessagePack frames with a configurable 16 MiB default
- preserve the existing `ndjson`, `layerNdjson`, and `layerMsgPack` values while adding `makeNdjson`, `layerNdjsonWith`, and `layerMsgPackWith` for custom limits
- support `maxBufferSize: "unbounded"` as an explicit opt-out
- forward serialization limits through Node, Bun, and Deno cluster layers
- close socket transports with code 1009 when the decoder limit trips
- add regression tests and patch changesets

## Root cause

The framed decoders retained incomplete NDJSON and MessagePack input in per-connection closure state without any upper bound. Socket transports caught decoder failures as ordinary defects, wrote a response, and continued reading from the same connection.

## Impact

Incomplete frames can no longer grow a connection's parser state without bound. Applications can tune or explicitly disable the limit per serializer and through cluster layers, while existing value-based call sites retain their source-compatible API and receive the safe default.

## Validation

- `nix develop -c pnpm check`
- `nix develop -c pnpm test packages/effect/test/rpc packages/effect/test/unstable/ai/McpServer/McpServer.test.ts --run` (45 tests)
- `nix develop -c pnpm test packages/platform-node/test/RpcServer.test.ts packages/platform-node/test/cluster/SocketRunner.test.ts --run` (97 tests)
- targeted `dprint` and `oxlint` checks

Closes EFF-258

### DIFF
--- a/packages/effect/src/unstable/rpc/RpcSerialization.ts
+++ b/packages/effect/src/unstable/rpc/RpcSerialization.ts
@@ -69,7 +69,7 @@ export class MaxBufferSizeExceeded extends Data.TaggedError("MaxBufferSizeExceed
  * @category serialization
  * @since 4.0.0
  */
-export interface StreamingOptions {
+export interface StreamOptions {
   /**
    * Maximum number of bytes or string code units retained for an incomplete frame.
    * The default is 16 MiB. Use `"unbounded"` to disable the limit.
@@ -114,7 +114,7 @@ export const json: RpcSerialization["Service"] = RpcSerialization.of({
  * @category serialization
  * @since 4.0.0
  */
-export const makeNdjson = (options?: StreamingOptions): RpcSerialization["Service"] => {
+export const makeNdjson = (options?: StreamOptions): RpcSerialization["Service"] => {
   const maxBufferSize = options?.maxBufferSize ?? defaultMaxBufferSize
   return RpcSerialization.of({
     contentType: "application/ndjson",
@@ -504,7 +504,7 @@ type JsonRpcMessage = JsonRpcRequest | JsonRpcResponse
  * @since 4.0.0
  */
 export const makeMsgPack = (
-  options?: (Msgpackr.Options & StreamingOptions) | undefined
+  options?: (Msgpackr.Options & StreamOptions) | undefined
 ): RpcSerialization["Service"] => {
   const { maxBufferSize = defaultMaxBufferSize, ...msgpackOptions } = options ?? {}
   return RpcSerialization.of({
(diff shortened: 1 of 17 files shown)


packages/effect/src/unstable/rpc/RpcServer.ts
```diff
@@ -1465,6 +1465,9 @@ const makeSocketProtocol: Effect.Effect<
           step: constVoid
         })
       } catch (cause) {
+        if (cause instanceof RpcSerialization.MaxBufferSizeExceeded) {
```
tim-smart: Can we use Predicate.isTagged instead

tim-smart: Done in 41a94a0c5 — switched the buffer overflow guard to `Predicate.isTagged`. The RPC suite passes (25 tests), and the full TypeScript check is clean.

packages/effect/src/unstable/rpc/RpcSerialization.ts
```diff
@@ -515,7 +575,7 @@ export const layerJson: Layer.Layer<RpcSerialization> = Layer.succeed(RpcSeriali
  * @category serialization
  * @since 4.0.0
  */
-export const layerNdjson: Layer.Layer<RpcSerialization> = Layer.succeed(RpcSerialization)(ndjson)
+export const layerNdjson: Layer.Layer<RpcSerialization> = Layer.succeed(RpcSerialization)(ndjson())
```
tim-smart: This layer should be updated to take options.

tim-smart: Done in f44d82c51 — `layerNdjson` now accepts `StreamingOptions`, and I applied the same option forwarding to `layerMsgPack` while preserving its existing default `useRecords: true` behavior. Added layer-level regression tests proving custom `maxBufferSize` values reach both decoders. Validation: 27 Effect RPC tests, 97 Node RPC/cluster tests, and the full TypeScript check all pass.

tim-smart:
Aligned the v4 API with merged #6803 in 9cdd88272: restored `ndjson`, `layerNdjson`, and `layerMsgPack` as values; added the matching `makeNdjson`, `layerNdjsonWith`, and `layerMsgPackWith` APIs; added the `"unbounded"` sentinel and shared limit helper; and forwarded `serializationMaxBufferSize` through all Node/Bun cluster layers plus the two v4-only Deno counterparts. Added the platform changeset and unbounded/layer-forwarding coverage.

Validation is clean: 45 Effect RPC/MCP tests, 97 Node RPC/cluster tests, targeted formatting/lint, and the full TypeScript check.

PR #6803, opened 2026-07-31
Cap RPC streaming decoder buffers

tim-smart:
## Summary

- cap retained NDJSON and MessagePack decoder buffers with a 16 MiB default
- expose configurable limits and an explicit `"unbounded"` opt-out, including layer constructors
- propagate the NDJSON cap through NDJSON-RPC framing
- close socket transports with code 1009 when the typed buffer overflow error is raised
- add regression coverage and a patch changeset for `@effect/rpc`

## Why

Incomplete MessagePack frames and newline-free NDJSON streams could retain and repeatedly grow per-connection parser state without a bound. The configurable cap limits that retained state while allowing applications with legitimate large frames to raise or disable the limit.

## Validation

- `pnpm test run packages/rpc/test/RpcSerialization.test.ts`
- `pnpm lint-fix`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`

Closes EFF-259

### DIFF
--- a/packages/rpc/src/RpcSerialization.ts
+++ b/packages/rpc/src/RpcSerialization.ts
@@ -35,14 +35,10 @@ export interface StreamOptions {
   readonly maxBufferSize?: number | "unbounded" | undefined
 }
 
-const ensureBufferSize = (bufferSize: number, maxBufferSize: number | "unbounded") => {
-  if (maxBufferSize === "unbounded" || bufferSize <= maxBufferSize) return
-  throw new RpcSerializationError({
-    reason: "BufferSizeExceeded",
-    maxBufferSize,
-    bufferSize
-  })
-}
+const isBufferSizeExceeded = (
+  bufferSize: number,
+  maxBufferSize: number | "unbounded"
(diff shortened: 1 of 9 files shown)


packages/rpc/src/RpcSerialization.ts
```diff
@@ -2,11 +2,48 @@
...
+
+/**
+ * @since 1.0.0
+ * @category serialization
+ */
+export interface StreamOptions {
+  readonly maxBufferSize?: number | "unbounded" | undefined
+}
+
+const ensureBufferSize = (bufferSize: number, maxBufferSize: number | "unbounded") => {
```
tim-smart: I think this should just return a boolean and the call site can throw the error.

packages/rpc/src/RpcServer.ts
```diff
@@ -1484,6 +1484,9 @@ const makeSocketProtocol = Effect.gen(function*() {
           step: constVoid
         })
       } catch (cause) {
+        if (cause instanceof RpcSerialization.RpcSerializationError) {
```
tim-smart: This should use Predicate.isTagged

PR #6805, opened 2026-07-31
Fix PubSub replay value retention

### LINKED ISSUE #6804: PubSub replay window pins the replay list — SubscriptionRef leaks every published value

4.0.0-beta.101 (still present in 4.0.0-beta.102 — `ReplayWindowImpl` is identical). The underlying bug is also latent in v3 (verified on 3.19.8, same `ReplayWindowImpl`), but there it only affects PubSubs created explicitly with `{ replay: n }`. v4 made it universal for `SubscriptionRef` users because `SubscriptionRef.make` now builds on `PubSub.unbounded({ replay: 1 })`, whereas v3's `SubscriptionRef` used a plain `PubSub.unbounded()` and prepended the current value with `Stream.concat` — v3 `SubscriptionRef` is unaffected.

Each subscription to a replay-enabled PubSub creates a `ReplayWindowImpl` that captures `buffer.head` at subscribe time and only advances it while draining the replay. `ReplayBuffer.offer` appends every published value to the same (append-only) list, so once the replay is drained the window's stale node pins the list and transitively retains every value published after the drain, for the life of the subscriber (a subscriber that never drains pins from subscribe time instead). One long-lived `changes` subscriber → unbounded heap growth, even though the subscriber consumes everything promptly.

Run with `node --expose-gc`; heap grows ~400 KB/s (measured on Node 20/22, macOS). Remove the subscriber (or apply the fix below) and it is flat. The same repro against effect 3.19.8 (using `ref.changes` / `Effect.forkDaemon`) is flat, and an explicit `PubSub.unbounded({ replay: 1 })` + one draining subscriber leaks identically on 3.19.8:

```ts
import { Effect, Stream, SubscriptionRef, Duration, Schedule } from "effect"

const fat = (i: number) => ({
  seq: i,
  units: Array.from({ length: 40 }, (_, k) => ({ id: `u${k}`, state: { seq: i } }))
})

  const ref = yield* SubscriptionRef.make(fat(0))
  // one healthy, constantly-draining subscriber is enough
  yield* Effect.forkDetach(SubscriptionRef.changes(ref).pipe(Stream.runForEach(() => Effect.void)))
  yield* Effect.forkDetach(Effect.gen(function* () {
    let i = 0
    while (true) {
      yield* SubscriptionRef.set(ref, fat(++i))
      yield* Effect.sleep(Duration.millis(10))
    }
  }))
  yield* Effect.repeat(Effect.sync(() => {
    ;(globalThis as any).gc?.()
    console.log(`heap=${(process.memoryUsage().heapUsed / 1048576).toFixed(1)}MB`)
  }), Schedule.spaced(Duration.seconds(5)))
})

Effect.runPromise(program)
```

Heap-snapshot retainer chain for a leaked value:

```
value ← {value,next} replay node ← (.next × N) ← ReplayWindowImpl.head ← SubscriptionImpl.replayWindow
```

We hit this in production the day we deployed our v3→v4 migration: a backend publishing device state into `SubscriptionRef`s at telemetry rate leaked ~28 MB/min until it hit `--max-old-space-size`. The same workload on v3 was flat.

Consumed (or slid-out) replay values become collectable; a subscription that stays open indefinitely retains at most the replay window's own values.

### Suggested fix

The window doesn't need to store a node reference at all — derive the cursor as `buffer.head` advanced by the number of values already taken (walk bounded by replay capacity, i.e. 1 for `SubscriptionRef`). Slid-out values are skipped automatically because `buffer.head` itself advances, which also removes the `fastForward` bookkeeping:

```ts
class ReplayWindowImpl<A> implements PubSub.ReplayWindow<A> {
  taken = 0
  remaining: number
  readonly buffer: ReplayBuffer<A>

  constructor(buffer: ReplayBuffer<A>) {
    this.buffer = buffer
    this.remaining = buffer.size
  }
  take(): A | undefined {
    if (this.remaining === 0) return undefined
    let node = this.buffer.head
    for (let i = 0; i < this.taken; i++) node = node.next!
    this.taken++
    this.remaining--
    return node.value as A
  }
  takeN(n: number): Array<A> {
    const len = Math.min(n, this.remaining)
    const items = new Array(len)
    for (let i = 0; i < len; i++) items[i] = this.take()
    return items
  }
  takeAll(): Array<A> {
    return this.takeN(this.remaining)
  }
}
```

We are running exactly this as a pnpm patch in production: the leaking repro goes flat, a late first take on an idle subscription still delivers the latest value (matching the old `fastForward` behavior — values that slid out are skipped), and closed idle subscriptions release their queue backlog as before.

### Platform

- Node v20 / v22, macOS + Linux

tim-smart:
## Summary

- store each subscription's replay values in a bounded snapshot instead of retaining a cursor into the append-only replay list
- preserve sliding PubSub ordering with a bounded ring of values removed from active subscriptions
- release replay snapshots when drained or unsubscribed
- add deterministic object-graph regressions for drained and never-draining replay windows

## Root cause

`ReplayWindowImpl` retained a node from `ReplayBuffer`'s linked list. Future publications extended that same list, so a live subscription could keep every later value reachable even after its replay window was drained. A never-draining subscription retained the same unbounded suffix from its original replay node.

The replay window cannot simply derive every take from the current buffer head: existing replay semantics require a subscription to retain its original replay snapshot, while sliding PubSubs replace only values that are explicitly slid out. This change keeps those semantics while bounding all replay-owned references.

## Validation

- `pnpm lint-fix`
- `pnpm test --run packages/effect/test/PubSub.test.ts packages/effect/test/SubscriptionRef.test.ts` (38 tests)
- `pnpm check`

Closes EFF-262
Closes EFF-264
Closes #6804

### DIFF
--- a/packages/effect/src/PubSub.ts
+++ b/packages/effect/src/PubSub.ts
@@ -2067,7 +2067,7 @@ class BoundedPubSubSingleSubscription<in out A> implements PubSub.BackingSubscri
 
 interface Node<out A> {
   value: A | AbsentValue
-  replayIndex?: number
+  replayIndex: number | undefined
   subscribers: number
   next: Node<A> | null
 }
@@ -2075,6 +2075,7 @@ interface Node<out A> {
 class UnboundedPubSub<in out A> implements PubSub.Atomic<A> {
   publisherHead: Node<A> = {
     value: AbsentValue,
+    replayIndex: undefined,
     subscribers: 0,
     next: null
   }
@@ -2111,11 +2112,9 @@ class UnboundedPubSub<in out A> implements PubSub.Atomic<A> {
     if (subscribers !== 0) {
       const node: Node<A> = {
         value,
+        replayIndex,
         subscribers,
         next: null
-      }
-      if (replayIndex !== undefined) {
-        node.replayIndex = replayIndex
       }
       this.publisherTail.next = node
       this.publisherTail = this.publisherTail.next
--- a/packages/effect/src/PubSub.ts
+++ b/packages/effect/src/PubSub.ts
@@ -139,6 +139,7 @@ export declare namespace PubSub {
     take(): A | undefined
     takeN(n: number): Array<A>
     takeAll(): Array<A>
+    close(): void
     readonly remaining: number
   }
 
@@ -1115,9 +1116,7 @@ const unsubscribe = <A>(self: Subscription<A>): Effect.Effect<void> =>
           Effect.sync(() => {
             self.subscribers.delete(self.subscription)
             self.subscription.unsubscribe()
-            if (self.replayWindow instanceof ReplayWindowImpl) {
-              self.replayWindow.close()
-            }
+            self.replayWindow.close()
             self.strategy.onPubSubEmptySpaceUnsafe(self.pubsub, self.subscribers)
           })
         ),
@@ -1529,6 +1528,7 @@ const makeSubscriptionUnsafe = <A>(
 
 class BoundedPubSubArb<in out A> implements PubSub.Atomic<A> {
   array: Array<A>
+  replayIndices: Array<number>
   publisherIndex = 0
   subscribers: Array<number>
   subscriberCount = 0
@@ -1541,6 +1541,7 @@ class BoundedPubSubArb<in out A> implements PubSub.Atomic<A> {
     this.capacity = capacity
     this.replayBuffer = replayBuffer
     this.array = Array.from({ length: capacity })
+    this.replayIndices = replayBuffer ? Array.from({ length: capacity }) : []
     this.subscribers = Array.from({ length: capacity })
   }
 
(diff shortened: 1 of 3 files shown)


packages/effect/src/PubSub.ts
```diff
@@ -1115,6 +1115,9 @@ const unsubscribe = <A>(self: Subscription<A>): Effect.Effect<void> =>
           Effect.sync(() => {
             self.subscribers.delete(self.subscription)
             self.subscription.unsubscribe()
+            if (self.replayWindow instanceof ReplayWindowImpl) {
```
tim-smart: Add `close` to the interface and remove this check

packages/effect/test/PubSub.test.ts
```diff
@@ -2,6 +2,29 @@ import { assert, describe, it } from "@effect/vitest"
 import { Array, Effect, Exit, Fiber, Latch, PubSub, Stream } from "effect"
 import { pipe } from "effect/Function"
 
+const retains = (root: object, target: object): boolean => {
```
tim-smart: Move this helper to the bottom of the file.

tim-smart:
Benchmark rerun for `9837b66b9` on Node v24.18.0.

Harness: `makeAtomicUnbounded({ replay: 1 })`, one draining subscriber, `publish` + `poll` + `slide`, 5 alternating old/fixed process pairs, 5 timed rounds per process, 5,000,000 operations per round. Per-process minima:

- `7f0d6ffc1`: 18.32, 18.41, 18.71, 18.49, 18.62 ns/op (mean 18.51 ns/op)
- `9837b66b9`: 17.68, 17.51, 17.72, 17.56, 17.84 ns/op (mean 17.66 ns/op)

That is a measured reduction of 0.85 ns/op (4.6%) in this fresh run. The absolute timings are lower than the earlier local experiment, but all paired runs consistently favored the stable node shape.

Validation:

- `pnpm test --run packages/effect/test/PubSub.test.ts packages/effect/test/SubscriptionRef.test.ts` — 38 tests passed
- `pnpm check` — passed

PR #6827, opened 2026-07-31
Add retry and response header options to HttpClient.withRateLimiter

jaipaljadeja:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

Adds two options to `HttpClient.withRateLimiter`:

- `times` limits the number of automatic retries after a `429` response. It counts retries after the initial request, and `times: 0` disables automatic retries. Omitting it preserves the existing unlimited behavior.
- `responseHeaders` allows individual rate-limit response header names to be overridden. Fields that are not configured continue using the existing built-in names.

When the retry limit is reached, the final `429` is returned or failed through its original channel. Response inspection still runs before that happens, allowing the limiter to learn from the final response.

Also updates the public API JSDoc and adds runtime and type-level coverage.

Validation:

- `pnpm lint-fix`
- `pnpm lint`
- `pnpm --filter effect test --run test/unstable/http/HttpClient.test.ts`
- `pnpm test-types packages/effect/typetest/unstable/http/HttpClient.tst.ts`
- `pnpm check`

### DIFF
--- a/packages/effect/src/unstable/http/HttpClient.ts
+++ b/packages/effect/src/unstable/http/HttpClient.ts
@@ -992,7 +992,8 @@ export declare namespace WithRateLimiter {
    *
    * **Details**
    *
-   * They define the backing limiter, initial limit window, keying strategy, algorithm, token cost, and whether response headers update future limits.
+   * They define the backing limiter, initial limit window, keying strategy,
+   * algorithm, token cost, retry count, and response header inspection.
    *
    * @category rate limiting
    * @since 4.0.0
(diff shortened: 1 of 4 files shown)


packages/effect/src/unstable/http/HttpClient.ts
```diff
@@ -1105,11 +1150,11 @@ export const withRateLimiter: {
     const key = resolveKey(request)
     const tokens = Math.max(resolveTokens(request), 1)
     const current = getState(key)
+    const canRetry = options.times === undefined || retries < options.times
     function retry(retryAfter: Duration.Duration | undefined) {
-      if (options.disableResponseInspection) return loop(effect, request)
```
tim-smart, line 1109: Was this intentional?

jaipaljadeja: Yes, this was intentional.

`retry` only receives the value returned by `inspectResponse`. When `disableResponseInspection` is true, `inspectResponse` returns `undefined`, so `retryAfter` is always `undefined`.

In the previous code, this meant the expression below would take the `: loop(effect, request)` branch, which was the same call made by the separate check. So that check was redundant.

In the updated code, the same path now calls `loop(effect, request, retries + 1)` to carry the new retry count forward.

PR #6828, opened 2026-07-31
Use layered storage for Context

tim-smart:
## Summary

- store ordinary services in bounded copy-on-write overlays so `Context.add` is O(1) on the request path
- adaptively flatten read-heavy layered contexts after eight base hits
- memoize cold-path `mapUnsafe` materialization while preserving map ordering, equality, merging, and filtering behavior
- skip fiber derived-state recomputation when a context change preserves the same cache root
- replace per-request context map clones in the Node, Bun, Deno, and HTTP router paths with immutable `Context.add` operations
- preserve scoped `Scope` and `ParentSpan` restoration semantics with focused coverage

`Context.makeUnsafe` retains its trusted input map without copying; callers must not mutate it after construction.

## HTTP throughput benchmark

Current `main` (`5441c8e65`) versus the benchmarked PR implementation (`4c7152d63`) on Node 26.4.0/arm64. `GET /` used the default tracer, 50 keep-alive connections, 8 pipelined requests, a 3-second warmup, and a 10-second measured interval. Each result is the median of three sequential runs on separate worktrees of the same machine. The request contexts were verified through the harness as 16, 46, and 200 entries respectively.

| Request context | `main` | PR | Change |
|---|---:|---:|---:|
| natural (16 services) | 88,581 rps | 100,055 rps | +13.0% |
| +30 inert services (46 services) | 65,233 rps | 101,934 rps | +56.3% |
| +184 inert services (200 services) | 30,196 rps | 109,621 rps | +263.0% |

The PR retains 101.9% of natural throughput at 46 services and 109.6% at 200 services, preserving the flat scaling curve. The adaptive counter was also inspected directly for all three request-context sizes: it remained zero and no request context flattened, confirming that the heuristic is not reached on this short-lived path. Absolute throughput varied between benchmark windows, but there was no service-count-dependent regression. All measured runs completed without errors or timeouts.

## Context read benchmark

Each value is the mean of two process medians. Each process took ten 10-million-read samples per case after a 1-million-read warmup, with case order reversed in the second process. “Depth” is the number of `Context.add` operations. The before/after columns isolate the adaptive-flattening change (`6a4e47686` to `75081721e`); `4c7152d63` reduces successful defined base reads to one map probe and does not touch the overlay-hit path.

| Read case | Before heuristic | After heuristic | Change |
|---|---:|---:|---:|
| flat base hit | 183.7M ops/s | 179.1M ops/s | -2.5% |
| depth-1 base hit | 153.5M ops/s | 163.3M ops/s | +6.4% |
| depth-1 overlay hit | 198.6M ops/s | 199.5M ops/s | +0.4% |
| depth-8 base hit | 73.3M ops/s | 142.8M ops/s | +94.9% |
| depth-8 oldest-overlay hit | 86.1M ops/s | 83.4M ops/s | -3.2% |
| cached hit | 137.7M ops/s | 137.8M ops/s | +0.1% |

After the eighth base hit, the context memoizes and structurally adopts its flat map, so subsequent reads use the original direct-map path without adding an `_flat` check to every overlay lookup. The depth-8 base hit recovers from 73.3M to 142.8M ops/s, above the earlier `main` result of 132.7M. Flat, overlay, and cached cases remain within process variance. Overlay hits do not advance the counter, and adding to the context creates a fresh layered context with a reset counter.

## Validation

- full Effect suite on the benchmarked implementation: 7,546 passed, 3 skipped
- focused Context and docgen parser suites
- Effect documentation generation and example type-checking
- `pnpm check`
- `pnpm lint-fix`

Closes EFF-274
Closes EFF-277
Closes EFF-278

### DIFF
--- a/packages/effect/src/Tracer.ts
+++ b/packages/effect/src/Tracer.ts
@@ -167,7 +167,7 @@ export const ParentSpanKey = "effect/Tracer/ParentSpan"
  * @category services
  * @since 2.0.0
  */
-export class ParentSpan extends Context.Service<ParentSpan, AnySpan>()(ParentSpanKey, { enableCaching: true }) {}
+export class ParentSpan extends Context.Service<ParentSpan, AnySpan>()(ParentSpanKey, { enableFiberCaching: true }) {}
 
 /**
  * Represents a span created outside Effect's tracer, carrying trace and span
@@ -629,7 +629,7 @@ export const TracerKey = "effect/Tracer"
  * @since 2.0.0
  */
 export const Tracer: Context.Reference<Tracer> = Context.Reference<Tracer>(TracerKey, {
-  enableCaching: true,
+  enableFiberCaching: true,
   defaultValue: () =>
     make({
       span: (options) => new NativeSpan(options)
--- a/packages/effect/src/Tracer.ts
+++ b/packages/effect/src/Tracer.ts
@@ -629,7 +629,6 @@ export const TracerKey = "effect/Tracer"
  * @since 2.0.0
  */
 export const Tracer: Context.Reference<Tracer> = Context.Reference<Tracer>(TracerKey, {
-  allocateSlot: true,
   defaultValue: () =>
     make({
       span: (options) => new NativeSpan(options)
--- a/packages/effect/src/Tracer.ts
+++ b/packages/effect/src/Tracer.ts
@@ -168,6 +168,7 @@ export const ParentSpanKey = "effect/Tracer/ParentSpan"
  * @since 2.0.0
  */
 export class ParentSpan extends Context.Service<ParentSpan, AnySpan>()(ParentSpanKey) {}
+Context.unsafeMakeSlot(ParentSpan)
 
 /**
  * Represents a span created outside Effect's tracer, carrying trace and span
--- a/packages/effect/src/Tracer.ts
+++ b/packages/effect/src/Tracer.ts
@@ -167,7 +167,7 @@ export const ParentSpanKey = "effect/Tracer/ParentSpan"
  * @category services
  * @since 2.0.0
  */
-export class ParentSpan extends Context.Service<ParentSpan, AnySpan>()(ParentSpanKey, { slot: true }) {}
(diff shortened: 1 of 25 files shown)


packages/effect/src/Tracer.ts
```diff
@@ -168,6 +168,7 @@ export const ParentSpanKey = "effect/Tracer/ParentSpan"
  * @since 2.0.0
  */
 export class ParentSpan extends Context.Service<ParentSpan, AnySpan>()(ParentSpanKey) {}
+Context.unsafeMakeSlot(ParentSpan)
```
tim-smart: We want to avoid this side effect

tim-smart: Removed in eb899ce90. Slots now come only from Reference construction; the fiber refreshes currentSpan unconditionally in setContext instead, so ParentSpan adds are plain O(1) overlay nodes (no slab copy). Paired benchmark shows no regression.

tim-smart:
`311d47896` reduces the bundle size impact of the new representation (net -35 lines).

Measured with `scripts/bundle-compare.sh` against the merge-base (`c9b56ab5`), minified + gzipped:

| | before (`d6075695c`) | after (`311d47896`) |
|---|---|---|
| `basic.ts` | +0.54 KB (+8.11%) | +0.47 KB (+7.11%) |
| all fixtures | +0.53 to +0.60 KB | +0.45 to +0.52 KB |

What changed:

- The mutable/immutable branches duplicated across `add`, `merge`, `pick`, and `omit` collapsed into one shared `withFlat` helper.
- The eagerly maintained `size` field is gone; `Equal`/`Hash`/`merge` derive it from the memoized flat map. This also removes the presence lookup `add` did on every call to keep `size` accurate.
- Reads no longer consult the slot registry (`slotByKey`) as a fallback: the slab is only a cache over overlay/base, so a slab miss falls through to a correct lookup. Writes keep the fallback since an aliased key object must still trigger the slab copy.

Hot paths got faster as a side effect (2 adds + 2 reads on a 42-service context: ~110ns → ~72ns; non-slotted read: ~27ns → ~21ns). Full effect suite, doctests, and NodeHttpServer tests pass.

PR #6829, opened 2026-08-01
feat: add mcp core for protocol versions

### LINKED ISSUE #6639: Canonical McpSchema with ProtocolAdapter
>I like the idea of having an abstracted McpProtocol - but we will need to come up with a solution for McpSchema as we introduce support for other protocol versions.
>
>[we] will need a robust solution for this as we introduce further protocol versions.
>
>_Originally posted by @IMax153 in https://github.com/Effect-TS/effect/pull/6625#discussion_r3650947178_
            
As part of the effort towards #6617 I want to split out the issue for how the ProtocolAdapter should be able to represent the core or canonical McpSchema that shouldn't change version to versions as well as the version specific shapes and behaviours.

They way I'm seeing it at the moment there are currently 3 supported version in `McpSchema` today. 1 missing and 1 upcoming:

| | |
| --- | --- |
| `v2024-11-05` | historical baseline with the original session-oriented lifecycle and HTTP+SSE transport. |
| `v2025-03-26` | transport modernization. Replaces HTTP+SSE with single-endpoint Streamable HTTP, introduces OAuth 2.1 authorization, adds JSON-RPC batching, audio content, tool annotations, progress messages, and explicit completion capabilities. |
| `v2025-06-18` | structured data and server-initiated interaction. Removes batching, adds structured tool output, outputSchema, resource links, richer metadata and titles, and form-based elicitation. _This is Effect’s current behavioral baseline._ |
| `v2025-11-25` | feature expansion on the existing sessionful architecture. Adds experimental Tasks, URL elicitation, sampling with tools, icons, richer authorization discovery, and JSON Schema 2020-12 as the default dialect. |
| `v2026-07-28.rc` | architectural reset around a stateless core. Removes initialization and protocol sessions, introduces server/discover, moves client/version/capability information into each request, restructures server-to-client interactions as multi-round-trip requests, and adds routing and caching metadata. |

What we are looking for in this is the canonical core of the MCP which will be shared or manipulated across the different versions.  The obvious choices would be `tool`, `prompt`, `resource`, `completion` functionality that hasn't changed, but lifecycle and transport seem to be the main areas of shift across each.

lloydrichards:
## Type
- [x] Feature

## Description
As part of being able to define new protocol adapters, I've added first-class support for multiple MCP protocol revisions based on the existing `McpSchema` (<= v2025-06-18):

- `v2024-11-05`
- `v2025-03-26`
- `v2025-06-18`

So now an MCP server now declares an ordered list of protocol adapters:

```ts
McpServer.layerStdio({
  name: "example",
  version: "1.0.0",
  protocols: [
    McpProtocol.v2025_06_18,
    McpProtocol.v2025_03_26,
    McpProtocol.v2024_11_05
  ]
})
```

The client’s requested protocol is selected before its payload is decoded. Once initialization succeeds, the selected adapter remains associated with that session.

The main separation is between:

- `McpCore`, which owns version-neutral tools, resources, prompts, completions, notifications, and semantic errors
- dated protocol adapters, which own their wire schemas, RPC vocabulary, projections, reverse requests, and transport rules
- `McpServer`, which owns transport lifecycle, sessions, dispatch, cancellation, and notification delivery

Each dated adapter is declared through `McpProtocol.make`:

```ts
export const protocol = McpProtocol.make({
  // Identifies the wire revision selected during initialization.
  protocolVersion: "2025-06-18",

  // Transport behavior can vary between MCP revisions.
  transport: {
    acceptsJsonRpcBatches: false,
    requiresVersionHeader: true
  },

  // The dated client → server and server → client RPC vocabularies.
  clientRpcs: ClientRpcs,
  clientNotificationRpcs: ClientNotificationRpcs,
  serverRequestRpcs: ServerRequestRpcs,
  serverNotificationRpcs: ServerNotificationRpcs,

  // Typed wire handlers translate dated requests into canonical core calls,
  // then project canonical results back into this revision's response types.
  handlerRpcs: AdapterRpcs,
  makeHandlers: (core, lifecycle) =>
    AdapterRpcs.of({
      "tools/list": () =>
        McpServerClient.use((client) =>
          core.tools.list(profileFromClient(client))
        ).pipe(
          Effect.map((tools) => ListToolsResult.make({
            tools: tools.map(projectTool)
          }))
        )
    }),

  // Reverse requests expose one stable server-facing client API even though
  // the wire request and response schemas vary by revision.
  toReverseClient: (profile, client) => ({
    listRoots: /* canonical request ↔ dated request */,
    createMessage: /* canonical request ↔ dated request */,
    elicit: /* supported here; rejected by older adapters */
  }),

  // Server notifications are projected independently for each revision.
  projectNotification,

  // Cancellation remains an adapter concern because its payload must be
  // understood before the transport interrupts an active request.
  normalizeCancellation
})
```

`McpProtocol.make` derives the repetitive mechanics from this declaration:

- payload codecs from the dated RPC schemas
- reverse RPC client construction
- handler installation and service requirement inference

This keeps version-specific decisions at the adapter boundary instead of introducing revision checks throughout `McpServer`.

## How to Review

There is a lot of code here, but also a lot of duplicated patterns since each dated protocol reimplements its own schema and adaptor.  This should make each protocol more resistent to changes as new versions are added, but still adds a lot to look over:

| Area | Added | Deleted |
| --- | ---: | ---: |
| Dated wire schemas | 1,129 | 0 |
| Core and dated adapters | 1,985 | 48 |
| Server integration and selection | 854 | 699 |
| Tests and harnesses | 1,758 | 548 |

I would suggest starting with the commits as they break up the different parts, but can mostly focus on how the v2025-06-18 works since this is semantically the same as how the `McpSchema` currently works.  The other two are mostly about reducing the capabilities or changing how some behaviour works, but nothing too crazy is added and all the conformance suites have been updated to exclude spec tests that didn't apply to older versions.

## Related

- Related Issue #6625 
- Closes #6639


lloydrichards:
The CI failure looks like a flakey test: `test/Persistence.integration.test.ts > Persistence SQL cleanup > deletes expired entries in batches` can you rerun it?

packages/effect/src/unstable/ai/McpProtocol.ts
```diff
@@ -3,41 +3,48 @@
...
+ *
+ * It provides the dated schema and stdio behavior. When supplied to
+ * `McpServer.layerHttp`, the server uses its single-endpoint Streamable HTTP
+ * compatibility transport; it does not implement the historical two-endpoint
+ * HTTP+SSE transport.
+ *
+ * @category protocols
+ * @since 4.0.0
+ */
+export const v2024_11_05 = protocol2024_11_05
```
IMax153: All public exports from the library should have annotated types. If you have the Effect LSP installed, there's a code action for this.

packages/effect/src/unstable/ai/McpSchema.ts
```diff
@@ -1449,6 +1454,26 @@ export class ToolAnnotations extends Schema.Opaque<ToolAnnotations>()(Schema.Str
...
+ *
+ * **Details**
+ *
+ * Property definitions and additional root keywords are constrained to JSON
+ * values. The open root supports generated keywords such as `$defs`.
+ *
+ * @category tools
+ * @since 4.0.0
+ */
+export const ToolJsonSchema = Schema.StructWithRest(
```
IMax153: Type annotation - you can decide how verbose you want it to be (i.e. the full schema type, or just use `Schema.Codec`. The downside of `Schema.Codec` is it erases information, but it's much more concise.

packages/effect/src/unstable/rpc/RpcSerialization.ts
```diff
@@ -260,7 +260,7 @@ function decodeJsonRpcRaw(
     for (let i = 0; i < decoded.length; i++) {
       const message = decodeJsonRpcMessage(decoded[i])
       messages.push(message)
-      if (message._tag === "Request") {
+      if (message._tag === "Request" && !Predicate.isNullish(decoded[i].id)) {
```
IMax153: Does this change have any implications for RPC serialization? Does it allow us to more closely follow the JSON RPC protocol?

IMax153 (changes requested):

packages/effect/src/unstable/rpc/RpcSerialization.ts
lloydrichards: What i found here is that currently the RPC was batching requests that don't need responses:

```jsonc
[
  { "id": 41, "method": "ping" },            // needs a response
  { "method": "notifications/initialized" }, // needs no response
  { "id": 42, "method": "unknown/method" }   // needs a response
]
```
the `decodeJsonRpcMessage` turns this into
```bash
{ _tag: "Request", id: 41, ... }
{ _tag: "Request", id: "", isNotification: true, ... }
{ _tag: "Request", id: 42, ... }
```
And then the condition was counting it was a `batch.size` of 3 instead of 2. So adding the extra condition filters the no response message.  My reading from the [spec](https://www.jsonrpc.org/specification#notification)

see test [v2025_03_26.test.ts#L80-L104](https://github.com/lloydrichards/open_effect/blob/aa171b89137c09b8adf5ed468ef4578bd5b53048/packages/effect/test/unstable/ai/McpServer/v2025_03_26.test.ts#L80-L104):
```txt
 FAIL   effect  test/unstable/ai/McpServer/v2025_03_26.test.ts > Mcp Conformance (2025-03-26) > Transport-specific behavior > MUST accept an operational mixed batch and correlate responses by id
SyntaxError: Unexpected end of JSON input
 ❯ test/unstable/ai/McpServer/v2025_03_26.test.ts:98:26
     96|             readonly error?: { readonly code?: unknown }
     97|           }>
     98|         >(() => response.json())
       |                          ^
     99|         const success = messages.find((message) => message.id === 41)
    100|         const failure = messages.find((message) => message.id === 42)
 ❯ src/internal/effect.ts:1045:24
 ❯ ~effect/Utils/internal src/Utils.ts:208:14
 ❯ MixedScheduler.<anonymous> src/internal/effect.ts:1045:5
 ❯ Object.~effect/Effect/evaluate src/internal/effect.ts:1107:22
```

from the mcp spec, v2025-03-26 spec -> [notification](https://modelcontextprotocol.io/specification/2025-03-26/basic#notifications)

packages/effect/src/unstable/ai/McpProtocol.ts
lloydrichards: added a slimmed down ProtocolAdaptor cause the lsp was trying to add 15k loc with the generated from 😬 

```ts
 export interface ProtocolAdapter<out Version extends ProtocolVersion = ProtocolVersion> {
   readonly [TypeId]: typeof TypeId
   readonly protocolVersion: Version
   readonly transport: {
     readonly acceptsJsonRpcBatches: boolean
     readonly requiresVersionHeader: boolean
   }
 }
```

Since this isn't really used in user-land past just adding them to the `McpServer` config I dont think the specific `RpcGroup` details is super useful

packages/effect/src/unstable/ai/McpSchema.ts
lloydrichards: went with this:
```ts
export type ToolJsonSchema = Schema.JsonObject & {
  readonly type: "object"
  readonly properties?: Readonly<Record<string, Schema.JsonObject>> | undefined
  readonly required?: ReadonlyArray<string> | undefined
}
```
And added a type test for it

packages/effect/src/unstable/ai/McpProtocol.ts
lloydrichards: I'm struggling a bit with this. I tried a more concise version of the public ProtocolAdaptor since its not really useful type info but ended up having to have a `McpProtocolInternal.toServerProtocols` in order to check the TypeId and version.

the commit is a bit of a mess so if you have a suggestion that would be cleaner then i'm all ears 🙏

lloydrichards:
```ts
/**
 * An MCP protocol adapter that can be supplied to `McpServer`.
 *
 * @category models
 * @since 4.0.0
 */
export interface ProtocolAdapter<out Version extends ProtocolVersion = ProtocolVersion>
  extends McpProtocolInternal.AnyProtocolAdapter
{
  readonly protocolVersion: Version
}
```
Sorry, i figured it out 😅 am a bit heat stroked i think

packages/effect/src/unstable/rpc/RpcSerialization.ts
IMax153: cc @tim-smart - you good with this change? Seems reasonable to me.

packages/effect/src/unstable/ai/internal/mcpProtocol/v2025_03_26.ts
```diff
@@ -0,0 +1,360 @@
...
+          return McpSchema.CallToolResult.make({
+            content,
+            isError: result.isError,
+            _meta: result._meta
+          })
+        })
+    }),
+  toReverseClient: (profile, client) => ({
+    listRoots: (request) =>
+      Effect.gen(function*() {
```
IMax153: Minor nit - all these functions that return `Effect.gen` can all be `Effect.fnUntraced`

IMax153 (changes requested):

packages/effect/src/unstable/rpc/RpcSerialization.ts
tim-smart: Yep but it doesn't need `Predicate.isNullish`

packages/effect/src/unstable/ai/internal/mcpProtocol/v2025_03_26.ts
lloydrichards: addressed in the /mcpProtocol and the McpServer 👍

packages/effect/src/unstable/rpc/RpcSerialization.ts
lloydrichards: changed it up to:
```
if (message._tag === "Request" && !!decoded[i].id) {}
```

IMax153: @lloydrichards - if the `id` was set and has a value of `0`, the `!!` boolean coercion would fail here (`!!0 === false`).

I think we can just make assertions on the `message` instead of the decoded element, because the `message` will always fallback to an `id` of `""` if the id is not set.

I think a more robust check here would be:

```ts
if (
  message._tag === "Request" && 
  typeof message.id === "string" &&
  message.id.length !== 0
) { /* ... */ }
```

lloydrichards: `message.id` can also be a `number`.  how about:
```ts
if (message._tag === "Request" && decoded[i].id != null) {}
```

lloydrichards: but then `"" != null` is true.  so we go back to `decoded[i].id != null && decoded[i].id != undefined` which is the same as `Predicate.isNotNullish(decoded[i].id)`

lloydrichards: I've got a better solutions:
```ts
if (message._tag === "Request" && !message.isNotification) {
  batch.size++
  batches.set(message.id, batch)
}
```
This way we don't care about the id and instead focus on the real bahaviour which is no batching notification responses

.changeset/mcp-protocol-versions.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
```
IMax153:
```suggestion
"effect": patch
```

IMax153 (approved):
Really amazing work @lloydrichards ! Will 🚢 after CI runs again!

Issue #5939, opened 2025-12-30
effect/ai: provider web_search tool-result is treated as local tool output and sent to OpenAI /responses API as `function_call_output`


tensor2077:
### What version of Effect is running?

effect/ai 0.33.0

### What steps can reproduce the bug?

1. Enable the provider-defined web_search tool and send a streaming /responses request (gpt5.2).
2. Ask the model to perform a web search (e.g., “Use web search to define AI agent and summarize.”).
3. After the first turn completes, send a **second message** in the same chat session (any text).
4. Observe HTTP 400: No tool call found for function call output with call_id ws_....

### What is the expected behavior?

success response

### What do you see instead?

 {
    "error": {
      "message": "No tool call found for function call output with call_id ws_029e7d51de056f790069536f1df6488196ab10f32c7a3d915e.",
      "type": "invalid_request_error",
      "param": "input",
      "code": null
    }
  }

### Additional information

 After the first turn (a web search + answer), the persisted history looks like:
```json
 {
    "content": [
      {
        "role": "system",
        "content": "You are an AI assistant.",
        "options": {}
      },
      {
        "role": "user",
        "content": "Use web search to look up the concept of an acquisition and summarize briefly.",
        "options": {}
      },
      {
        "role": "assistant",
        "content": [
          {
            "type": "tool-call",
            "id": "ws_029e7d51de056f790069536f1df6488196ab10f32c7a3d915e",
            "name": "web_search",
            "params": {
              "action": {
                "type": "search",
                "query": "acquisition definition",
                "sources": [{ "type": "url", "url": "https://en.wikipedia.org/wiki/Acquisition" }]
              }
            },
            "providerExecuted": true,
            "options": {}
          },
          {
            "type": "text",
            "text": "An acquisition is ...",
            "options": {}
          }
        ],
        "options": {}
      },
      {
        "role": "tool",
        "content": [
          {
            "type": "tool-result",
            "id": "ws_029e7d51de056f790069536f1df6488196ab10f32c7a3d915e",
            "name": "web_search",
            "isFailure": false,
            "result": { "status": "completed" },
            "options": {}
          }
        ],
        "options": {}
      }
    ]
  }
```
So far, everything looks correct.
However, sending any second user message fails with:
```json
  {
    "error": {
      "message": "No tool call found for function call output with call_id ws_029e7d51de056f790069536f1df6488196ab10f32c7a3d915e.",
      "type": "invalid_request_error",
      "param": "input",
      "code": null
[...]

PR #6836, opened 2026-08-01
Route provider-executed tool results into the assistant message in Prompt.fromResponseParts

mkdynamic:
## Summary

Ports #5944 to the v4 line. `Prompt.fromResponseParts` places provider-executed tool results (e.g. OpenAI `web_search`) in a `tool` message like framework results, so round-tripping a conversation sends them back to OpenAI as `function_call_output` items whose corresponding `web_search_call` was dropped from the request (`prepareMessages` skips provider-executed tool calls). OpenAI rejects the request:

```
HTTP 400 from POST /v1/responses
No tool call found for function call output with call_id ws_<...>.
```

The first turn with a provider web search succeeds; every follow-up turn that round-trips the history fails, deterministically. This is exactly #5939, which #5944 fixed on the v3 line (`@effect/ai`) -- neither of that PR's two changes exists in the v4 rewrite.

## Changes

- `Prompt.ToolResultPart` / `ToolResultPartEncoded` / schema gain `providerExecuted` (decoding default `false`), mirroring `ToolCallPart`.
- `fromResponseParts` routes tool results with `providerExecuted: true` into `assistantParts`, keeping the provider call/result pair together in the assistant message. Framework results still form `tool` messages.
- The consumer half already exists in v4: `@effect/ai-openai`'s `prepareMessages` assistant branch has a tool-result case ("Assistant tool-result parts are always provider executed") that skips provider results or emits `item_reference` under `store: true` -- it was dead code until now, since `fromResponseParts` never delivered results to it.
- Mechanical: the now-required field added at the two framework construction sites in `LanguageModel.ts` (`providerExecuted: false`), the JSDoc examples, and existing test fixtures.
- Tests: `fromResponseParts` keeps a provider-executed call/result pair in the assistant message while a framework result still lands in the tool message; and on the Anthropic side, a history built via `fromResponseParts` containing a provider-executed `web_search` pair serializes to `server_tool_use` + `web_search_tool_result` blocks in the assistant turn with no client `tool_result` block (the assistant-branch serializer case this change activates).
- Changeset: `"effect": patch`.

## Validation

- `pnpm test packages/effect/test/unstable/ai packages/ai/openai/test/OpenAiLanguageModel.test.ts packages/ai/openai-compat/test/OpenAiLanguageModel.test.ts packages/ai/anthropic` -- 642/642 across 24 files, run at the branch tip.
- `pnpm check` -- clean.
- `pnpm lint-fix` applied.

## Notes

The `store: true` / `item_reference` path still cannot engage for web search because the streamed and non-streamed `web_search_call` parts carry no `metadata.openai.itemId` (unlike e.g. reasoning and apply_patch parts) and `fromResponseParts` drops response-part metadata for all part types. With this PR the pair is coherently omitted from the request, which unbreaks the round-trip; carrying item ids through for `item_reference` replay would be a follow-up.

### DIFF
--- a/.changeset/provider-executed-tool-results.md
+++ b/.changeset/provider-executed-tool-results.md
@@ -2,6 +2,4 @@
 "effect": patch
 ---
 
-Route provider-executed tool results into the assistant message in `Prompt.fromResponseParts`.
-
-Previously a provider-executed tool result (e.g. OpenAI `web_search`) was placed in a `tool` message like a framework-executed result, so round-tripping the conversation sent it back to OpenAI as a `function_call_output` whose corresponding `web_search_call` had been dropped, failing the request with HTTP 400 `No tool call found for function call output with call_id ws_...`. `Prompt.ToolResultPart` now carries `providerExecuted` (decoding default `false`), and `fromResponseParts` keeps provider-executed results in the assistant message, where the provider packages already expect them. Ports #5944 to the v4 line.
+Route provider-executed tool results into the assistant message in `Prompt.fromResponseParts`
--- a/.changeset/provider-executed-tool-results.md
+++ b/.changeset/provider-executed-tool-results.md
@@ -1,0 +1,7 @@
+---
+"effect": patch
+---
+
+Route provider-executed tool results into the assistant message in `Prompt.fromResponseParts`.
+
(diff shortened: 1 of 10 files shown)


.changeset/provider-executed-tool-results.md
```diff
@@ -0,0 +1,7 @@
+---
+"effect": patch
+---
+
+Route provider-executed tool results into the assistant message in `Prompt.fromResponseParts`.
+
+Previously a provider-executed tool result (e.g. OpenAI `web_search`) was placed in a `tool` message like a framework-executed result, so round-tripping the conversation sent it back to OpenAI as a `function_call_output` whose corresponding `web_search_call` had been dropped, failing the request with HTTP 400 `No tool call found for function call output with call_id ws_...`. `Prompt.ToolResultPart` now carries `providerExecuted` (decoding default `false`), and `fromResponseParts` keeps provider-executed results in the assistant message, where the provider packages already expect them. Ports #5944 to the v4 line.
```
IMax153:
```suggestion
Route provider-executed tool results into the assistant message in `Prompt.fromResponseParts`
```

IMax153 (approved):

PR #6838, opened 2026-08-01
fix(schema): share representation identity across rebuilt containers of one declaration

aryasaatvik:
Stacked on #6837 — the branch includes its commit; only the last commit (`Share representation identity across rebuilt containers of one declaration`) is new here. The two are companion halves of one invariant: #6837 stops `HttpApiSchema.status` from forking the annotations slot; this stops derivation rebuilds from forking the container slots.

## Problem

`SchemaRepresentation` resolves component identity by reference: an AST's identity key is the
composition of the object identities of its slots, with `context` excluded. That works as long as
every reuse of a declared schema reaches the representation as reference-shared structure.

Memoized derivations break that assumption whenever the derivation is non-identity. `toCodecJson`
lowers `Unknown` to `Json` by rewriting the child node, which forces the parent rebuild to allocate
fresh `PropertySignature` wrappers and a fresh `propertySignatures` array — once per top-level
derivation input. A named schema and its context-only copy (`HttpApiSchema.status` after #6837) are
two derivation inputs, so their rebuilt containers can never be reference-equal, and the identity
key diverges at exactly that slot:

```ts
const Widget = Schema.Struct({
  id: Schema.String,
  metadata: Schema.Record(Schema.String, Schema.Unknown)
}).annotate({ identifier: "Widget" })

HttpApiEndpoint.get("get", "/widgets/:id", { params: { id: Schema.String }, success: Widget })
HttpApiEndpoint.post("create", "/widgets", { success: Widget.pipe(HttpApiSchema.status(201)) })

OpenApi.fromApi(Api).components.schemas
// { Widget: …, Widget_1: … } — two identical components for one declared name
// (on 4.0.0-beta.102 this is a `Duplicate identifier: "Widget"` throw instead)
```

Remove the `metadata` field and the pair unifies — the failure is keyed to whether any child's
derivation is non-identity, which makes it look nondeterministic from the API author's seat. Any
schema carrying `Schema.Unknown` (a JSON metadata column, a `Record(String, Unknown)`) hits it in
the default CRUD shape.

## Fix

Container wrappers — `propertySignatures` / `indexSignatures` arrays, `PropertySignature`,
`IndexSignature` — are pure structure: they carry no annotations, checks, or context of their own,
so they have no identity to defend. When two ASTs hold the **same annotations object** — the token
of one user declaration — their container slots now compare by contents (element-wise reference
identity) instead of by container reference.

The gate is what keeps existing semantics intact:

- Anonymous nodes keep pure reference identity — "does not extract structurally equivalent schemas
  with distinct ASTs" is unchanged.
- Two `.annotate({ identifier: "Value" })` calls create two annotations objects, so referentially
  distinct declarations still suffix (`Value` / `Value_1`).
- Two composites sharing a declaration but with genuinely different recursive dependencies still
  split: their containers differ at the child references themselves ("does not deduplicate fallback
  definitions with distinct recursive dependencies" is unchanged).

Scope: one function in `internal/schema/toRepresentation.ts`; no AST or derivation changes.

## Tests

- `toRepresentations.test.ts`: "shares a named reference across context-only copies whose
  derivation rebuilds a container" — pure representation-level regression, fails with
  `["Widget", "Widget_1"]` refs before the fix.
- `OpenApi.test.ts`: "emits one component for a named schema containing Unknown reused with
  HttpApiSchema.status" — end-to-end CRUD shape, both `200` and `201` referencing one `Widget`
  component. Depends on #6837 (without it the copies also differ at the annotations slot).

`packages/effect/test/schema` + `packages/effect/test/unstable/httpapi`: 2220 tests green;
`tsc -b packages/effect` and lint clean.

### DIFF
--- a/.changeset/httpapi-status-key-context.md
+++ b/.changeset/httpapi-status-key-context.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+unstable/httpapi HttpApiSchema: carry `httpApiStatus` in the key context so a named schema reused with `status()` keeps one representation identity — one OpenAPI component referenced by every status — instead of throwing `Duplicate identifier`
--- a/.changeset/representation-annotated-container-identity.md
+++ b/.changeset/representation-annotated-container-identity.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+SchemaRepresentation: compare container slots structurally when two ASTs carry the same annotations object, so a derivation of a context-only copy (e.g. the JSON codec of a schema containing `Unknown` reused with `HttpApiSchema.status`) no longer publishes duplicate components for one declared identifier
--- a/packages/effect/src/internal/schema/toRepresentation.ts
+++ b/packages/effect/src/internal/schema/toRepresentation.ts
@@ -60,6 +60,7 @@ function fromASTs(
   const anonymousReferences = new Map<SchemaAST.AST, string>()
   const referenceOwners = new Map<string, SchemaAST.AST>()
   const valueIds = new Map<unknown, number>()
+  const compositeIds = new Map<string, number>()
   const canonicalByKey = new Map<string, SchemaAST.AST>()
   let nextValueId = 0
   const buildingReferences = new Set<string>()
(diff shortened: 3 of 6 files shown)


aryasaatvik:
Consolidated into #6837 — that PR now carries both commits (the httpapi key-context fix and this representation-identity fix) as companion halves of one invariant.

PR #6843, opened 2026-08-01
Add reproduction for vitest/utils issue

fubhy:
## Summary

- reproduce the false positive in `throws` and `throwsAsync` when the operation succeeds
- keep the "expected to throw" assertion outside the helper's catch block so it cannot be swallowed
- keep the regression coverage in the main `@effect/vitest` test suite

## Verification

```text
pnpm test --run --project @effect/vitest
pnpm lint
pnpm check
```

Closes EFF-292


packages/vitest/test/ThrowsRepro.test.ts
```diff
@@ -0,0 +1,10 @@
+import * as testAssert from "@effect/vitest/utils"
```
tim-smart: Remove this test file and measure the main one.

PR #6853, opened 2026-08-01
Add reproduction for TxQueue issue

fubhy:
## Summary

- fix `TxQueue.offerAll` to materialize input iterables outside the effect
- preserve one-shot iterable values across transaction retries and repeated runs
- add regression coverage for bounded backpressure and effect reuse

## Testing

```text
pnpm test --run packages/effect/test/TxQueue.test.ts -t "offerAll preserves a one-shot iterable across retries"
```

Closes EFF-295

### DIFF
--- a/packages/effect/src/TxQueue.ts
+++ b/packages/effect/src/TxQueue.ts
@@ -608,23 +608,22 @@ export const offerAll: {
   <A, E>(self: TxEnqueue<A, E>, values: Iterable<A>): Effect.Effect<Array<A>>
 } = dual(
   2,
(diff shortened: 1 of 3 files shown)


packages/effect/src/TxQueue.ts
```diff
@@ -609,18 +609,22 @@ export const offerAll: {
 } = dual(
   2,
   <A, E>(self: TxEnqueue<A, E>, values: Iterable<A>): Effect.Effect<Array<A>> =>
-    Effect.gen(function*() {
-      const rejected: Array<A> = []
+    Effect.suspend(() => {
+      const valuesArray = Array.from(values)
```
tim-smart: This should be materialized outside of the effect so it is re-usable between seperate runs.

PR #6857, opened 2026-08-01
Add reproduction for platform-browser/BrowserKeyValueStore issue

fubhy:
## Summary

- wait for IndexedDB write transactions to commit before reporting `KeyValueStore` mutations as successful
- surface transaction errors and aborts as `KeyValueStoreError`
- add a regression test for an abort after the individual request succeeds

## Testing

```text
pnpm test --run packages/platform-browser/test/BrowserKeyValueStore.test.ts
pnpm check
```

Closes EFF-299


packages/platform-browser/test/BrowserKeyValueStoreTransaction.test.ts
```diff
@@ -0,0 +1,73 @@
+import * as BrowserKeyValueStore from "@effect/platform-browser/BrowserKeyValueStore"
```
tim-smart: Can we please remove this test file and merge it into the main one?

PR #6858, opened 2026-08-01
Add reproduction for libsql/LibsqlClient issue

fubhy:
## Summary

Fix `libsql/LibsqlClient` transaction acquisition so a failed transaction begin releases its serialization permit and later operations can retry.

## Root cause

The permit release finalizer was attached to a manually created scope. When the libSQL SDK rejected `transaction("write")`, acquisition failed before that scope was returned to the transaction wrapper, so the scope was never closed and the permit remained consumed.

The acquisition path now closes the scope with the failure cause when beginning the transaction fails.

## Validation

```text
pnpm test --run packages/sql/libsql/test/BeginPermitRepro.test.ts
pnpm lint-fix
pnpm check
```

Closes EFF-320


packages/sql/libsql/test/BeginPermitRepro.test.ts
```diff
@@ -0,0 +1,38 @@
+import { LibsqlClient } from "@effect/sql-libsql"
```
tim-smart: Can we remove this test file and move it into the main one?

tim-smart: Moved the regression into `Client.integration.test.ts` and removed the standalone `BeginPermitRepro.test.ts` file in f34db4125.

PR #6865, opened 2026-08-01
Add reproduction for OtlpResource issue

fubhy:
## Summary

- Decode percent-encoded keys and values from `OTEL_RESOURCE_ATTRIBUTES`.
- Preserve bigint resource attribute precision when mapping values to OTLP int64.

### 1. `unstable-distributed-otlp-resource-attribute-percent-decoding`: Resource environment values are not percent-decoded

Module: `OtlpResource`

Expected contract: OpenTelemetry resource environment parsing must decode percent-encoded keys and values.

Observed result: PASS

Reproduction command:

```text
pnpm test --run packages/effect/test/unstable/observability/OtlpResource.test.ts
```

### 2. `unstable-distributed-otlp-resource-bigint-precision`: Bigint resource attributes lose precision

Module: `OtlpResource`

Expected contract: Accepted bigint attributes mapped to OTLP int64 must preserve their integer value.

Observed result: PASS

Reproduction command:

```text
pnpm test --run packages/effect/test/unstable/observability/OtlpResource.test.ts
```

Closes EFF-304


packages/effect/test/unstable/observability/OtlpResourceBigInt.test.ts
```diff
@@ -0,0 +1,31 @@
+import { assert, it } from "@effect/vitest"
```
tim-smart: Rename this file to OtlpResource.test.ts

PR #6866, opened 2026-08-01
Add reproduction for HttpBody issue

fubhy:
## Summary

Fix partial file-backed HTTP bodies so their advertised `contentLength` matches the bytes selected by `offset` and `bytesToRead`.

Both `HttpBody.file` and `HttpBody.fileFromInfo` now subtract the offset, cap an explicit read length to the bytes available, and avoid negative content lengths. Previously both constructors always copied the full file size into the body metadata, which could produce an incorrect outgoing `Content-Length` header.

## Validation

```text
pnpm test --run packages/effect/test/unstable/http/HttpBody.test.ts
pnpm check
```

Closes EFF-305


packages/effect/test/unstable/http/HttpBody.test.ts
```diff
@@ -0,0 +1,14 @@
+import { assert, it } from "@effect/vitest"
```
tim-smart, line 1: Remove audit from the file name.

PR #6867, opened 2026-08-01
Add reproduction for HttpClientRequest issue

fubhy:
## Summary

- Add a regression test for replacing a fixed-length request body with a stream of unknown length.
- Remove a stale `content-length` header when the replacement body does not provide length metadata.

## Root cause

`HttpClientRequest.setBody` updated `content-length` when the new body had a known length, but left the previous header unchanged when the new length was unknown.

## Validation

```text
pnpm test --run packages/effect/test/unstable/http/HttpClientRequest.test.ts
pnpm lint-fix
pnpm check
```

Closes EFF-306


packages/effect/test/unstable/http/HttpClientRequest.audit.test.ts
```diff
@@ -0,0 +1,10 @@
+import { assert, it } from "@effect/vitest"
```
tim-smart: Merge this test file with the main request tests.

PR #6868, opened 2026-08-01
Add reproduction for Persistence issue

fubhy:
## Summary

Fix Redis-backed persistence clearing so an empty store succeeds without sending a zero-key `DEL` command.

Includes regression coverage in the existing Redis persistence test suite and an `effect` patch changeset.

Closes EFF-307

### 1. `unstable-state-p-3`: Empty Redis clear sends a zero-key DEL

Module: `Persistence`

Expected contract: clear is idempotent and must succeed when the store is empty.

Observed result: A protocol-faithful Redis mock returned a wrong-arity error for DEL with zero keys.

Regression test command:

```text
pnpm test --run packages/effect/test/unstable/persistence/Redis.test.ts
```


packages/effect/test/unstable/persistence/PersistenceRedisClearEmpty.repro.test.ts
```diff
@@ -0,0 +1,24 @@
+import { it } from "@effect/vitest"
```
tim-smart: Can we please remove this test file and merge the tests into the main one?

PR #6869, opened 2026-08-01
Add reproduction for Persistence issue

fubhy:
## Summary

Fix SQL-backed Persistence `getMany` so it returns one positional result for every input key, including repeated keys, matching the memory, Redis, and KVS implementations.

The SQL implementations now decode query rows into a lookup and reconstruct the result in the original input order. The regression case is part of the main SQLite Persistence suite and runs against both SQL backing variants.

Closes EFF-308

## Verification

```text
pnpm test --run packages/sql/sqlite-node/test/Persistence.test.ts
```


packages/sql/sqlite-node/test/PersistenceDuplicateKeys.repro.test.ts
```diff
@@ -0,0 +1,20 @@
+import { SqliteClient } from "@effect/sql-sqlite-node"
```
tim-smart: Remove this test file and merge with the main one.

PR #6871, opened 2026-08-01
Add reproduction for LayerMap issue

fubhy:
## Reproduction only

This PR adds reproduction tests only. No implementation fix is included. CI is expected to fail until the underlying issue is fixed.

### 1. `core-g-r-layermap-preload-options-ignored`: LayerMap preload options have no runtime effect

Module: `LayerMap`

Expected contract: Supplying preloadKeys, or preload: true through fromRecord or Service, builds the selected entries while constructing the map and exposes layer failures through the conditional construction error type.

Observed result: The intended failure was reproduced: acquired keys were [] instead of ["a", "b"].

Reproduction command:

```text
pnpm vitest run packages/effect/test/LayerMap.test.ts -t "make preloads the requested keys"
```

### DIFF
--- a/packages/effect/src/LayerMap.ts
+++ b/packages/effect/src/LayerMap.ts
@@ -158,6 +158,7 @@ export const make: <
   lookup: (key: K) => Layer.Layer<I, EL, RL>,
   options?: {
     readonly idleTimeToLive?: IdleTimeToLiveInput<K> | undefined
+    readonly preloadKeys?: Iterable<K> | undefined
   } | undefined
 ) {
   const context = yield* Effect.context<never>()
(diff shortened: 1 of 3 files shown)


packages/effect/src/LayerMap.ts
```diff
@@ -171,6 +172,12 @@ export const make: <
     idleTimeToLive: options?.idleTimeToLive
   })
 
+  if (options?.preloadKeys) {
+    for (const key of options.preloadKeys) {
+      yield* RcMap.get(rcMap, key)
```
tim-smart:
```suggestion
      yield* Effect.scoped(RcMap.get(rcMap, key))
```

PR #6874, opened 2026-08-01
Add reproduction for Equal issue

fubhy:
## Fix

This PR adds reproduction tests and fixes the underlying `Equal.equals` and `Hash.hash` handling for invalid dates and `DataView` values.

### 1. `core-a-f-equal-dataview-throws`: DataView comparison throws instead of comparing bytes

Module: `Equal`

Expected contract: Equal.equals deeply compares supported objects and never throws; ArrayBuffer views enter the typed-array comparison path.

Observed result: A direct Node 24 source probe reproduced RangeError for invalid dates and TypeError for DataViews.

Reproduction command:

```text
pnpm test --run packages/effect/test/Equal.test.ts
```

### 2. `core-a-f-equal-invalid-date-throws`: Invalid dates violate Equal.equals never-throws contract

Module: `Equal`

Expected contract: Equal.equals returns a boolean and never throws, including for JavaScript Date values.

Observed result: A direct Node 24 source probe reproduced RangeError for invalid dates and TypeError for DataViews.

Reproduction command:

```text
pnpm test --run packages/effect/test/Equal.test.ts
```

Closes EFF-310


packages/effect/test/EqualNeverThrows.test.ts
```diff
@@ -0,0 +1,15 @@
+import { assert, describe, it } from "@effect/vitest"
```
tim-smart: Remove this test file and merge with the main one.

packages/effect/src/Equal.ts
```diff
@@ -323,6 +333,13 @@ function compareTypedArrays(self: Uint8Array, that: Uint8Array): boolean {
   return true
 }
 
+function compareDataViews(self: DataView, that: DataView): boolean {
```
tim-smart: Considering this helper is only used in one place in-line it please.

PR #6875, opened 2026-08-01
Add reproduction for internal/protobuf issue

fubhy:
## Summary

- Add a regression test for negative protobuf `int64` values.
- Encode negative signed integers as ten-byte, 64-bit two's-complement varints.
- Use the package import for the internal OTLP encoder so repository typechecking succeeds.

### `unstable-distributed-otlp-protobuf-negative-int64-varint`

Module: `internal/protobuf`

Negative protobuf `int32` and `int64` values must use ten-byte two's-complement varints.

Reproduction command:

```text
pnpm test --run packages/effect/test/unstable/observability/OtlpSerialization.test.ts
```

Closes EFF-311

### DIFF
--- a/packages/effect/test/unstable/observability/OtlpSerializationNegativeInt64.test.ts
+++ b/packages/effect/test/unstable/observability/OtlpSerializationNegativeInt64.test.ts
@@ -1,8 +1,0 @@
-import { assert, it } from "@effect/vitest"
-import { encodeAnyValue } from "effect/unstable/observability/internal/otlpProtobuf"
-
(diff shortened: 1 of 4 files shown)


packages/effect/test/unstable/observability/OtlpSerializationNegativeInt64.test.ts
```diff
@@ -0,0 +1,8 @@
+import { assert, it } from "@effect/vitest"
```
tim-smart: Remove this test file and merge with the main one.

PR #6879, opened 2026-08-01
Add reproduction for Cause issue

fubhy:
## Summary

Fix `Cause.map` so transforming typed errors preserves annotations attached to `Fail` reasons.

## Root cause

`Cause.map` rebuilt each mapped `Fail` reason with only the transformed error, dropping the original reason's annotations. The mapped reason now receives a copy of those annotations while `Die` and `Interrupt` reasons continue to pass through unchanged.

## Validation

```text
pnpm test --run packages/effect/test/Cause.test.ts packages/effect/test/CauseMapAnnotations.test.ts
pnpm check
pnpm lint
```

Closes EFF-313

### DIFF
--- a/packages/effect/src/internal/effect.ts
+++ b/packages/effect/src/internal/effect.ts
@@ -268,7 +268,7 @@ export const causeMap: {
     const failures = self.reasons.map((failure) => {
       if (isFailReason(failure)) {
         hasFail = true
-        return new Fail(f(failure.error), new Map(failure.annotations))
+        return new Fail(f(failure.error), failure.annotations)
       }
       return failure
     })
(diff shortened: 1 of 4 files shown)


packages/effect/src/internal/effect.ts
```diff
@@ -268,7 +268,7 @@ export const causeMap: {
     const failures = self.reasons.map((failure) => {
       if (isFailReason(failure)) {
         hasFail = true
-        return new Fail(f(failure.error))
+        return new Fail(f(failure.error), new Map(failure.annotations))
```
tim-smart:
```suggestion
        return new Fail(f(failure.error), failure.annotations)
```

PR #6880, opened 2026-08-01
Add HttpApiSchema.WithHeaders for typed response headers

### LINKED ISSUE #5633: From Discord: Add Support for Response Headers Schema in HttpApiEndpoint
# Summary
Users discussed the absence of support for defining schemas for response headers in the HttpApiEndpoint module. Currently, while request headers can be set using setHeaders, there is no provision for response headers in addSuccess or addError methods. A user shared their internal library's approach, where response success/error schemas include a headerParameters field, illustrating a need for this feature. Timsmart indicated that response header support might be added soon, likely as an option alongside status code. Another user highlighted a practical use case involving REST APIs returning 'Location' headers on entity creation and inquired about workarounds or contributing a PR. Overall, the feature is recognized as important with potential upcoming implementation, and community willingness to contribute code.

# Discord thread

https://discord.com/channels/795981131316985866/1348774989029183621

wmaurer:
## Type

- [x] Feature

## Description

- add `HttpApiSchema.WithHeaders`, a marker schema pairing a headers schema with a body schema, usable for successes and errors
- intercept it in `HttpApiEndpoint`, `HttpApiBuilder`, `HttpApiClient`, and `OpenApi`, the same way those modules already intercept the stream markers
- handlers return `{ headers, body }`, generated clients decode the same shape, and the OpenAPI document gains `responses[status].headers`
- reject a `WithHeaders` member that shares a `(status, content-type)` pair with another member of the same endpoint, since the client discriminates on exactly those two signals
- no platform package needed a source change, because `WithHeaders` resolves inside `unstable/httpapi` into an ordinary `HttpServerResponse`

## Motivation

`HttpApi` can declare a schema for _request_ headers, but the success and error sides can describe only a status code and a body. There is no typed way to declare, produce, or consume a _response_ header.

The motivating case is a `POST` that creates an entity and returns `201 Created` with an empty body and a `Location` header, a bog-standard REST pattern. Today `HttpApiClient` gives no typed access to that `Location`.

Existing escape hatches all work but are untyped: reading `HttpClientResponse` via `responseMode`, returning an `HttpServerResponse` from a handler (which requires a cast and teaches neither the client types nor the OpenAPI document anything), or setting blanket headers from `HttpApiMiddleware`.

This follows the API shape @tim-smart called for in [the Discord thread](https://discord.com/channels/795981131316985866/1348774989029183621) on 15 Oct 2025:

> I think a `WithHeaders` special schema in `HttpApiSchema`, that accepts schemas as options `{ headers, body }`. Then the other modules can intercept it.
>
> Then the return type for the handlers would be `{ headers, body }` and it would somehow get mapped to a `HttpServerResponse` correctly.
>
> And the reverse on the client etc.

#5633 was closed as stale in Jul 2026 asking for a fresh proposal against the current API shape. The API has moved since that thread: `effect-smol` merged, HttpApi lives in `packages/effect/src/unstable/httpapi/`, the builder-style `addSuccess`/`addError` are gone, and the client's `withResponse: true` is now `responseMode`. This is written against the current shape.

## API

```ts
const create = HttpApiEndpoint.post("create", "/things", {
  success: HttpApiSchema.WithHeaders({
    headers: { location: Schema.String },
    body: HttpApiSchema.Created
  })
})
```

Handlers return `{ headers, body }`; generated clients decode the same shape.

**Strictness follows the schema**, matching request headers: `Schema.String` is required and a missing header is a decode failure surfaced as `Schema.SchemaError`, the same failure type the client already produces for response _body_ decode failures. `Schema.optional(...)` yields `undefined`. No implicit leniency, so the declared type never lies.

**Status and content-type are not re-derived.** The constructor lifts `httpApiStatus` and `~httpApiEncoding` off `body.ast` onto the wrapper's AST, so `HttpApiSchema.Created`, `asText()`, and `HttpApiSchema.status(...)` compose unchanged. The wrapper's AST is authoritative everywhere; an annotation applied to the wrapper itself wins over the body's.

`responseMode` interaction:

| mode                     | returns                                           |
| ------------------------ | ------------------------------------------------- |
| `decoded-only` (default) | `{ headers, body }`                               |
| `decoded-and-response`   | `[{ headers, body }, HttpClientResponse]`         |
| `response-only`          | `HttpClientResponse`, with **no** header decoding |

`response-only` short-circuits before any schema work, as it does today, so a declared-but-missing header cannot fail a `response-only` call.

## Implementation

`WithHeaders` is a **marker schema** carrying `.headers` and `.body` sub-schemas, which the other httpapi modules intercept. That is structurally identical to how `HttpApiSchema.StreamSse` / `StreamUint8Array` are built and detected today via `isStreamSchema`. The interception mechanism already exists, is already threaded through all five modules, and is already tested.

- **`HttpApiSchema.ts`**: the schema, `isWithHeaders`, and internal unwrap helpers.
- **`HttpApiEndpoint.ts`**: `transformResponse` applies codecs (headers via `Schema.toCodecStringTree`, matching how request headers are built), so `disableCodecs: true` stays honest. `validateSuccessResponse` and `getErrorResponse` unwrap before classifying.
- **`HttpApiBuilder.ts`**: the buffered path encodes the body through the untouched `getResponseEncode`, then applies a single trailing `Response.setHeaders`, so all four encodings gain headers uniformly including the empty-body path. The streaming path passes encoded headers to `Response.stream`.
- **`HttpApiClient.ts`**: decodes `response.headers` through the headers codec alongside the body.
- **`OpenApi.ts`**: emits `responses[status].headers`. The OpenAPI `Response.headers` object was previously never emitted, so this is purely additive.

Errors use the same wrapper and the same machinery: `makeErrorSchema` already routes through `toResponseSchema`. The accepted consequence is that the value a handler fails with becomes `{ headers, body: MyError }`, and the client's error channel matches. That is noisier at the `fail`/`catch` site than for successes, but a second, divergent mechanism for errors would be worse and would not handle empty-body error responses cleanly.

## Deliberate decisions worth a reviewer's attention

**Exclusivity rule.** A `WithHeaders` member may not share a `(status, content-type)` pair with any other member of the same endpoint; violations throw at endpoint construction. The client's decode map is keyed by status, and within a status `groupSchemasByContentType` unions each content-type bucket into one codec over the response bytes, so status and content-type are the only runtime signals. If a plain schema and a `WithHeaders` schema shared a bucket, the caller's return shape would vary at runtime depending on which headers the server happened to send. This is a client-side ambiguity only; the server discriminates on the input value's shape.

Its main cost falls on errors, which share a status more often: two `500` JSON errors where only one carries `Retry-After` are rejected and need distinct statuses. A looser rule, permitting a shared bucket when every member is `WithHeaders` with a structurally equal headers schema, was considered and deferred, since loosening later is backwards-compatible while tightening is not.

**Nesting throws in the schema constructor**, not at endpoint construction. That is stricter and earlier than strictly necessary, and it gives a better error site.

**The declaration predicate is shape-based**, matching any `{ headers, body }` value. A plain `Schema.Struct({ headers, body })` declared alongside a `WithHeaders` member at a different status is therefore indistinguishable in the server's encode union, and the first declared member wins. This is pinned by a test (`HttpApiBuilder.test.ts`, "dispatches an ambiguous `{ headers, body }` value to the first declared member") rather than prevented, and it is not new in kind: `streamSchema = Schema.declare(Stream.isStream)` is looser still and backs both stream markers. The difference is that endpoint validation already guarantees at most one stream candidate per status, whereas these two members coexist legally. Happy to close it with an extra construction-time check if you would prefer that.

**Only object-shaped headers schemas reach the OpenAPI document.** OpenAPI models `Response.headers` as a fixed name to schema map, so a schema with dynamic keys (`Schema.Record`) has no representation there. It still encodes and decodes normally; it simply contributes no entries. This mirrors `processParameters`, which bails identically for non-object-shaped request `params`/`query`/`headers` on `main`. Documented in the `WithHeaders` JSDoc and pinned on both sides.

**Two bits of machinery that are not obvious from the diff:**

- `dropUndefinedMember` (`OpenApi.ts`) strips the `undefined` union member that `Schema.optional` injects, so an optional header renders as `{ type: "string" }` rather than an `anyOf`. The same defect is latent in `processParameters` for request parameters; that is pre-existing and left for a follow-up rather than mixed into this changeset.
- The encoding-override overloads on `schemasToResponse` / `toCodecArrayBuffer` (`HttpApiClient.ts`) exist so the wrapper's AST stays authoritative for content-type. The overload is type-level, so a multi-member array cannot be passed together with an override.

## Testing

Runtime (`packages/effect/test/unstable/httpapi/`): the motivating `Location`-on-`201`-with-empty-body round trip, headers on a JSON body, a missing required header failing as `Schema.SchemaError`, an absent optional header yielding `undefined`, error responses with headers, SSE and `Uint8Array` streams with headers, `response-only` not decoding headers, both construction-time validation errors, wrapper-annotation precedence for status/encoding/description, two `WithHeaders` endpoints staying independent, and the OpenAPI `responses[status].headers` output.

Integration (`packages/platform-node/test/HttpApi.test.ts`): `HttpApiTest` runs in memory, so one case was added to the platform-node suite, which serves over a real socket via `NodeHttpServer.layerTest`. It covers the empty-body path, where the headers are the only thing written, and the streaming path, where they must be flushed before the first chunk. It asserts both the raw wire headers and the generated client's decoded shape.

No platform package needed a source change: `WithHeaders` resolves entirely inside `unstable/httpapi` into an ordinary `HttpServerResponse`, and both `NodeHttpServer` and `BunHttpServer` already apply `response.headers` uniformly across every body variant.

Type-level (`packages/effect/typetest/unstable/httpapi/`): handler return type, client return type across all three `responseMode` values, and the error channel for a `WithHeaders` error. The handler-return test carries negative `.not.toBeCallableWith` assertions so it cannot pass vacuously.

## Validation

Every job in `.github/workflows/check.yml` was run locally:

- `pnpm lint`: pass
- `pnpm check` and `pnpm test-types`: pass, 1992 tests and 4986 assertions
- `pnpm build`: pass, with `stripInternal: true` as CI sets it
- `deno check .`: pass, same `stripInternal` patch
- `pnpm test`: 8717 passed across 340 files. Two unrelated flakes, both passing on their own and neither touching httpapi: `@effect/platform-browser` `RpcWorker.test.ts > Stream` asserts on a 2-second sleep, and `@effect/platform-node` `NodeHttpClient.test.ts > jsonplaceholder schemaBodyJson` makes a live network request.
- `pnpm doctest`: pass, 3376 tests
- `pnpm docgen`: pass, 3319 examples typechecked
- `pnpm ai-docgen`: pass, `LLMS.md` unchanged
- `pnpm circular`: pass
- `pnpm bundle-compare 0a0f2f0fb`: 0.00 KB / 0.00% on every fixture. Note there is no httpapi bundle fixture, so this confirms no regression elsewhere rather than measuring this feature's own cost.

Narrower runs worth naming: the httpapi suite is 139/139, and `@effect/platform-node` `HttpApi.test.ts` is 45/45.

A changeset is included (`patch`, matching the repo's pre-release convention).

## Related

- Closes #5633

### DIFF
--- a/.changeset/httpapi-response-header-schemas.md
+++ b/.changeset/httpapi-response-header-schemas.md
@@ -1,5 +1,5 @@
 ---
-"effect": minor
+"effect": patch
 ---
 
 Add `HttpApiSchema.WithHeaders` for typed HTTP response headers.
--- a/.changeset/httpapi-response-header-schemas.md
+++ b/.changeset/httpapi-response-header-schemas.md
@@ -1,0 +1,30 @@
+---
+"effect": minor
+---
+
+Add `HttpApiSchema.WithHeaders` for typed HTTP response headers.
+
+Response schemas can now declare headers alongside a body:
+
+```ts
+import { Schema } from "effect"
+import { HttpApiEndpoint, HttpApiSchema } from "effect/unstable/httpapi"
+
+const create = HttpApiEndpoint.post("create", "/things", {
+  success: HttpApiSchema.WithHeaders({
+    headers: { location: Schema.String },
+    body: HttpApiSchema.Created
+  })
+})
+```
+
+Handlers return `{ headers, body }` and generated clients decode the declared
+headers into the same shape. Header strictness follows the schema: a required
+header that is missing from a response is a decode failure, while
+`Schema.optional` yields `undefined`. `WithHeaders` works for successes, errors,
+and streaming responses, and the declared headers are emitted in the generated
+OpenAPI document as `responses[status].headers`.
+
+A `WithHeaders` response may not share a status and content-type with any other
+response of the same endpoint, since the client discriminates response
+alternatives on exactly those two signals.
--- a/.changeset/httpapi-response-header-schemas.md
+++ b/.changeset/httpapi-response-header-schemas.md
@@ -18,13 +18,6 @@ const create = HttpApiEndpoint.post("create", "/things", {
 })
 ```
 
-Handlers return `{ headers, body }` and generated clients decode the declared
-headers into the same shape. Header strictness follows the schema: a required
-header that is missing from a response is a decode failure, while
-`Schema.optional` yields `undefined`. `WithHeaders` works for successes, errors,
-and streaming responses, and the declared headers are emitted in the generated
-OpenAPI document as `responses[status].headers`.
+Handlers return `{ headers, body }` and generated clients decode the declared headers into the same shape. Header strictness follows the schema: a required header that is missing from a response is a decode failure, while `Schema.optional` yields `undefined`. `WithHeaders` works for successes, errors, and streaming responses, and the declared headers are emitted in the generated OpenAPI document as `responses[status].headers`.
 
-A `WithHeaders` response may not share a status and content-type with any other
-response of the same endpoint, since the client discriminates response
-alternatives on exactly those two signals.
+A `WithHeaders` response may not share a status and content-type with any other response of the same endpoint, since the client discriminates response alternatives on exactly those two signals.
--- a/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
+++ b/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
@@ -927,7 +927,7 @@ function makeStreamEncoder(endpoint: HttpApiEndpoint.Top): StreamEncoder | undef
 
   const streamSchema = streamSuccess.stream
   const hasBuffered = hasBufferedSuccess(endpoint)
-  const status = HttpApiSchema.getStatusStream(streamSchema)
+  const status = streamSuccess.status
   const contentType = streamSchema.contentType
   const encodeHeaders = streamSuccess.headers === undefined
     ? undefined
@@ -989,6 +989,7 @@ function makeStreamEncoder(endpoint: HttpApiEndpoint.Top): StreamEncoder | undef
 
 interface StreamSuccess {
   readonly stream: HttpApiSchema.StreamSchema
+  readonly status: number
   readonly headers: Schema.Top | undefined
 }
 
@@ -998,6 +999,10 @@ function getStreamSuccessSchema(endpoint: HttpApiEndpoint.Top): StreamSuccess | 
     if (HttpApiSchema.isStreamSchema(body)) {
       return {
         stream: body,
+        // read the status off the outer schema: for a `WithHeaders` wrapper an
+        // annotation applied to the wrapper wins over the body's lifted one, and
+        // for a bare stream the two schemas are the same object
+        status: HttpApiSchema.getStatusSuccess(schema.ast),
         headers: HttpApiSchema.isWithHeaders(schema) ? schema.headers : undefined
       }
     }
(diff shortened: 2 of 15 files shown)


tim-smart:
I'm not too sure about the experience of defining error headers yet. I'll have a look into it shortly.

tim-smart:
> I'm not too sure about the experience of defining error headers yet. I'll have a look into it shortly.

I think this might be a good approach: https://github.com/Effect-TS/effect/pull/6934/changes#diff-dd36f5096e4fc3b43f5cba36ed8c34e8c236dc9967912db8e6e71d883c5a8511R1648

wmaurer:
`encodeToWithHeaders` is the better solution. In my version, an error schema wrapped in `WithHeaders` also changes the error channel to `{ headers, body }`. So a tagged error is no longer at the top level, and `catchTag` stops working on it. With your version the error channel stays the domain error, and the `{ body, headers }` pair only exists on the encoded side.

The branded response value is also better than my structural check.

Closing this in favour of #6934. I left one note there.

PR #6682, opened 2026-07-28
feat: DenoFileSystem

tim-smart:
## Summary

- extract the FileSystem tests into a reusable conformance suite
- preserve the existing Node assertions and fixture coverage
- add documented opt-outs for the accepted Deno divergences

## Testing

- `pnpm lint-fix`
- `pnpm --filter @effect/platform-node-shared test --run test/NodeFileSystem.test.ts`
- `pnpm check`

Closes EFF-139
Closes EFF-159

## Summary by CodeRabbit

* **New Features**
  * Added a Deno-backed `FileSystem` implementation and exposed it via the platform-deno package.
* **Tests**
  * Introduced a reusable, configurable filesystem test harness and migrated the Node `FileSystem` suite to use it.
  * Added dedicated Deno `FileSystem` tests and expanded fixture, temp directory/file (including scoped cleanup), read/write/append, truncation, and cursor-behavior coverage.
  * Updated the text fixture used for read validation.
* **Bug Fixes**
  * Improved Deno error mapping to consistent platform error reasons.
* **Chores**
  * Added a Deno `@std/fs` dependency and published a patch release note.

PR #6885, opened 2026-08-02
feat: FileSystem support for copyFile flags

lika85456:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Support for FileSystem.copyFile mode flags:
- COPYFILE_EXCL: Fail if the destination already exists.
- COPYFILE_FICLONE: Attempt copy-on-write cloning, falling back to a regular copy.
- COPYFILE_FICLONE_FORCE: Require copy-on-write cloning; fail if unsupported.

Node and Bun implementations already support this, I just added the mode.

Deno however doesn't seem to support these flags, so my current implementation for Deno fails on the EXCL and FICLONE_FORCE flags, on FICLONE it does nothing.

## Related

- PR #6682

### DIFF
--- a/packages/platform-deno/src/DenoFileSystem.ts
+++ b/packages/platform-deno/src/DenoFileSystem.ts
@@ -67,8 +67,20 @@ const copy: FileSystem.FileSystem["copy"] = (fromPath, toPath, options) =>
       preserveTimestamps: options?.preserveTimestamps ?? false
     }))
 
-const copyFile: FileSystem.FileSystem["copyFile"] = (fromPath, toPath) =>
-  tryPromise("copyFile", fromPath, () => Deno.copyFile(fromPath, toPath))
+const copyFile: FileSystem.FileSystem["copyFile"] = (fromPath, toPath, options) => {
+  const mode = options?.mode ?? 0
+  const { COPYFILE_EXCL, COPYFILE_FICLONE_FORCE } = FileSystem.CopyFileFlag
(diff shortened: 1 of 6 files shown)


packages/platform-deno/src/DenoFileSystem.ts
```diff
@@ -67,8 +67,20 @@ const copy: FileSystem.FileSystem["copy"] = (fromPath, toPath, options) =>
       preserveTimestamps: options?.preserveTimestamps ?? false
     }))
 
-const copyFile: FileSystem.FileSystem["copyFile"] = (fromPath, toPath) =>
-  tryPromise("copyFile", fromPath, () => Deno.copyFile(fromPath, toPath))
+const copyFile: FileSystem.FileSystem["copyFile"] = (fromPath, toPath, options) => {
+  const mode = options?.mode ?? 0
+  const { COPYFILE_EXCL, COPYFILE_FICLONE_FORCE } = FileSystem.CopyFileFlag
+  if ((mode & COPYFILE_EXCL) !== 0 || (mode & COPYFILE_FICLONE_FORCE) !== 0) {
```
tim-smart, line 73: We want to avoid adding options that are unsupported on some platforms.

lika85456: Why is DenoFileSystem not implemented with NodeFileSystem like BunFileSystem? `node:fs` has the interface for these flags and Deno also implements it - however not correctly (EXCL is not atomic and COPYFILEs ignored).

Issue #6887, opened 2026-08-02
Allow Atom runtime memoization to follow AtomRegistry lifetime


andrskr:
## Problem

`Atom.context` accepts a concrete `Layer.MemoMap`. That map belongs to the runtime factory, even when the runtime is used by several `AtomRegistry` instances.

This is fine when the factory and registry have the same lifetime. It becomes a problem when registries are expected to be isolated but their Effect layers are still shared through the factory memo map.

## How we hit this

We use module-level `AtomRpc` queries and mutations with a separate `AtomRegistry` for each SSR request.

We want to keep module-level atoms. They give us stable atom identity, normal `AtomRpc` ergonomics, and predictable serialization and hydration.

In our focused reproduction, we mounted the same reactive query in two registries. A mutation in registry A also refreshed the query in registry B. We expected three reads and received four.

Creating another runtime with:

```ts
Atom.context({
  memoMap: Layer.makeMemoMapUnsafe()
})
```

does not solve this. It creates another factory-scoped map, not a registry-scoped one.

Recreating runtimes and atoms for every registry can isolate them, but then consumers must stop using normal module-level atoms. `Layer.fresh` can also avoid some sharing, but it rebuilds more than necessary and does not model the ownership we need.

We eventually narrowed our application so it does not currently depend on this cross-registry mutation behavior. That is a valid local trade-off, but not a general solution.

## Why this is broader than our setup

The same problem can appear anywhere one Atom runtime is used with multiple registries:

- concurrent SSR requests
- multiple React roots
- tenant or session-scoped registries
- isolated test registries
- applications using registry-local stateful layers

The registry looks isolated, but the runtime layer memoization may not be.

## Expected behavior

We need an opt-in way to make memoization follow the registry:

- layers are shared inside one registry
- layers are not shared between different registries
- disposing one registry releases only its resources
- existing factory-scoped behavior remains unchanged
- module-level atoms remain module-level atoms

## Possible direction

The cleanest direction we found is allowing `Atom.context` to accept either a concrete `Layer.MemoMap` or an `Atom<Layer.MemoMap>`:

```ts
const registryMemoMap = Atom.make(() => Layer.makeMemoMapUnsafe())

const runtime = Atom.context({
  memoMap: registryMemoMap
})
```

The memo-map atom would be resolved through the active registry, giving each registry its own map while preserving the current behavior for callers passing a concrete map.

[...]

PR #6888, opened 2026-08-02
Allow registry-scoped Atom memo maps

andrskr:
## Summary

- allow `Atom.context` to accept a concrete `Layer.MemoMap` or an `Atom<Layer.MemoMap>`
- resolve atom-backed memo maps through the active `AtomRegistry`
- keep concrete memo maps factory-scoped, preserving the current behavior

Related to #6887.

## Why

One Atom runtime can be used by several registries, but its concrete memo map is shared by the whole factory. This means registries can hold different atom values while still sharing runtime layers and `Reactivity` through the same map.

We hit this with module-level `AtomRpc` atoms and request-scoped SSR registries. A mutation in one registry also refreshed a reactive query in another registry.

Recreating runtimes and atoms per request avoids the sharing, but gives up stable module-level atoms. `Layer.fresh` rebuilds more than the ownership boundary requires.

## Proposed API

```ts
const registryMemoMap = Atom.make(() => Layer.makeMemoMapUnsafe())

const runtime = Atom.context({
  memoMap: registryMemoMap
})
```

The memo-map atom is resolved through the active registry. A concrete map continues to work as before.

This PR is intentionally a draft. I am not sure an atom is the best public API; an explicit scope option or another composition may be better. The goal is to give #6887 a concrete implementation and proof for discussion.

## Proof

- concrete maps remain shared across registries
- atom-backed maps share layers inside one registry and isolate different registries
- each registry finalizes only its own scoped resources
- a reactive mutation refreshes queries only in its active registry

## Validation

- `pnpm lint-fix`
- `pnpm lint`
- `pnpm test --run packages/effect/test/reactivity/Atom.test.ts`
- `pnpm check`
- `pnpm exec changeset status --since upstream/main`

### DIFF
--- a/.changeset/quiet-maps-share.md
+++ b/.changeset/quiet-maps-share.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Allow Atom runtime memo maps to be resolved from the active `AtomRegistry`.
--- a/packages/effect/src/unstable/reactivity/Atom.ts
+++ b/packages/effect/src/unstable/reactivity/Atom.ts
@@ -694,7 +694,7 @@ export interface AtomRuntime<R, ER = never> extends Atom<AsyncResult.AsyncResult
 }
 
 /**
- * Factory for `AtomRuntime` values that share a `Layer.MemoMap` and a set of global layers.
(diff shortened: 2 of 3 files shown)


andrskr:
Closing in favor of https://github.com/Effect-TS/effect/pull/7082/

PR #6898, opened 2026-08-03
Add HttpServer response compression middleware

tim-smart:
Implements the response compression spec for the v4 HttpServer.

## Shared middleware (`effect/unstable/http`)

- `HttpMiddleware.compression(options?)` with `algorithms` (default `["br", "gzip", "deflate"]`, zstd opt-in), `minSize` (default 1024), `compressible` predicate, and per-algorithm `levels` (gzip 6, deflate 6, br 4, zstd 3).
- Accept-Encoding negotiation per RFC 9110: q-values decide acceptability only, server order decides ranking, `*` matches unlisted codings, malformed headers are treated as absent, and no acceptable coding sends identity with 200 (never 406).
- Skip conditions: 1xx/204/206/304, existing `Content-Encoding` (with `identity` stripped as the opt-out), `Cache-Control: no-transform`, non-compressible or absent content type, known length below `minSize`, Empty/FormData bodies. HEAD requests mirror GET headers.
- `Vary: Accept-Encoding` is appended (no duplicates, `Vary: *` untouched) on every eligible response, including ones skipped by negotiation or `minSize`.
- Applied via the pre-response handler mechanism (same as `cors`) so it composes with `HttpEffect.toWebHandler`, `HttpRouter.serve`, and the platform servers.

## `HttpPlatform.compression` primitive

- New effectful `compression: { algorithms, compressResponse }` member on the service; `HttpPlatform.make` wraps the platform transform with shared header logic (`Content-Encoding`, `Vary`, strong ETags weakened to `W/...`).
- All `Uint8Array` bodies on Node.js, Bun, and Deno use the asynchronous `node:zlib` callback APIs through `Effect.callback`. The exact compressed `Content-Length` comes from `result.byteLength`; there is no sync path or size-based switch to streaming.
- `Stream` and `Raw` bodies remain streaming transforms and drop `Content-Length`.
- Platform coverage matches the spec table:
  - **Node**: all four via `node:zlib` (zstd gated on runtime support); async one-shot APIs for `Uint8Array` bodies and zlib transforms with per-chunk flush for `Stream`/`Raw` bodies.
  - **Bun**: all four via Bun's native `node:zlib` async implementations for one-shot bodies and extended `CompressionStream` for streaming, including `BunFile` responses. No `Bun.*Sync` one-shot branch is used.
  - **Deno**: all four via `node:zlib` compatibility one-shot APIs; streaming gzip/deflate use `CompressionStream`, while br/zstd use `node:zlib` compatibility streams.
  - **Web**: gzip/deflate via `CompressionStream`; byte arrays are necessarily sent through a single-chunk stream because `node:zlib` is unavailable.

## Tests

- Shared conformance coverage exercises negotiation, q-values, wildcard/identity, malformed headers, minSize, content types, opt-outs, no-transform, status handling, HEAD, incremental streaming, ETag weakening, Vary merging, and cookies.
- Node, Bun, and Deno tests assert exact `Content-Length` for async one-shot bodies and continue to assert absent `Content-Length` for stream/file responses.
- Runtime suites cover gzip, deflate, brotli, zstd where supported, file response cleanup, and incremental Node streaming.

Validation:

- `pnpm lint-fix`
- `pnpm check`
- `pnpm --dir packages/platform-bun check`
- `pnpm --dir packages/platform-deno check`
- Node/shared targeted Vitest suites: 47 tests passed
- Bun platform Vitest suite: 10 tests passed
- Deno compression Vitest suite: 8 tests passed

Closes EFF-332
Closes EFF-352


packages/effect/src/unstable/http/HttpMiddleware.ts
```diff
@@ -403,3 +406,157 @@ export const cors = (options?: {
...
     })
 }
+
+/**
+ * Options for the `compression` middleware.
+ *
+ * @category compression
+ * @since 4.0.0
+ */
+export interface CompressionMiddlewareOptions {
```
tim-smart: We should inline these options so you don't have to jump to definition.

packages/effect/src/unstable/http/HttpMiddleware.ts
```diff
@@ -403,3 +406,157 @@ export const cors = (options?: {
...
+   */
+  readonly levels?: {
+    readonly gzip?: number | undefined
+    readonly deflate?: number | undefined
+    readonly br?: number | undefined
+    readonly zstd?: number | undefined
+  } | undefined
+}
+
+const defaultAlgorithms: ReadonlyArray<CompressionAlgorithm> = ["br", "gzip", "deflate"]
```
tim-smart: Move all these helpless underneath the middleware definition.

packages/effect/src/unstable/http/HttpPlatform.ts
```diff
@@ -21,6 +21,85 @@ import * as Etag from "./Etag.ts"
...
+ * @category compression
+ * @since 4.0.0
+ */
+export const makeCompression: (options: {
+  readonly algorithms: Iterable<CompressionAlgorithm>
+  readonly transform: (
+    algorithm: CompressionAlgorithm,
+    options?: CompressionOptions | undefined
+  ) => (stream: ReadableStream<Uint8Array>) => ReadableStream<Uint8Array>
+  readonly compressSync?:
```
tim-smart: And benchmarks the sync APIs usually perform worse than async. Can you confirm and remove the sync path if true.

.changeset/eff-332-http-response-compression.md
```diff
@@ -0,0 +1,20 @@
...
+skip logic (status, existing `Content-Encoding`, `Cache-Control: no-transform`,
+content type, `minSize`), and manages the `Vary` header. The `HttpPlatform`
+service gains a `compression` primitive that advertises supported algorithms
+(gzip, deflate, br, zstd) and transforms response bodies using the fastest
+native API on each platform: `node:zlib` on Node.js, `Bun.gzipSync` and Bun's
+extended `CompressionStream` on Bun, `CompressionStream` plus `node:zlib`
+compatibility streams on Deno, and `CompressionStream` (gzip/deflate only) for
+web handlers. One-shot bodies keep an exact `Content-Length`, streaming bodies
+compress incrementally, and strong ETags are weakened on transformed
+responses.
```
tim-smart:
```suggestion
Add HTTP response compression support
```

packages/effect/src/unstable/http/HttpMiddleware.ts
tim-smart: Addressed in 6e46b411b: the compression options are now inlined in the middleware signature, so there is no separate interface to jump to.

packages/effect/src/unstable/http/HttpPlatform.ts
tim-smart: Benchmarked Node 26 gzip level 6 before changing this. At concurrency 16, the native async callback API delivered roughly 105–120k vs 89–95k ops/s at 4 KiB, 40–42k vs 12.7–13.4k at 64 KiB, and 3.13k vs 0.86k at 1 MiB. The actual createGzip stream path has setup overhead at 4 KiB (44k vs 84k), but wins 2.6× at 64 KiB and 3.7× at 1 MiB while avoiding event-loop blocking. I removed the synchronous one-shot API/path across the platforms in 6e46b411b; Uint8Array responses now use the async streaming transforms and drop Content-Length.

packages/effect/src/unstable/http/HttpMiddleware.ts
tim-smart: Addressed in 6e46b411b: moved the compression helper constants/functions below the middleware definition.

packages/platform-node/src/NodeHttpPlatform.ts
```diff
@@ -92,9 +71,17 @@ const compression: Platform.Compression = {
     const body = response.body
     switch (body._tag) {
       case "Uint8Array": {
-        return ServerResponse.setBody(
-          response,
-          HttpBody.uint8Array(compressSync(body.body, algorithm, options), body.contentType)
+        return ServerResponse.stream(
```
tim-smart: I think we should use a .raw response for this case

tim-smart: Addressed in efc8462ed: the Uint8Array case now returns ServerResponse.raw around Readable.from([body.body]).pipe(compressTransform(...)), while preserving the response metadata and dropping Content-Length.

packages/effect/src/unstable/http/HttpPlatform.ts
```diff
@@ -56,6 +128,7 @@ export class HttpPlatform extends Context.Service<HttpPlatform, {
  */
 export const make: (impl: {
   readonly platform: "deno" | "node" | "bun" | "web"
+  readonly compression?: Compression | undefined
```
tim-smart: This should be a required field

tim-smart: Addressed in b42fe56e5: `HttpPlatform.make` now requires `compression`, the implicit fallback was removed, and the portable web layer supplies `internal.compressionWeb()` explicitly. The Node, Bun, and Deno constructors already provide their native implementations.

PR #6914, opened 2026-08-03
Fix NodePath file URL conversion to follow the layer's platform flavor

### LINKED ISSUE #6913: NodePath.layerWin32 / layerPosix use host-platform file URL conversion instead of the layer's flavor
**Which version:** `effect@4.0.0-beta.101` / `@effect/platform-node` on the v4 line; verified against `main` at `b75884413`.

`NodePath.layerWin32` provides win32 path arithmetic but **host-flavored** `fromFileUrl` / `toFileUrl`, so on a non-Windows host the layer silently disagrees with itself: `join` / `dirname` / `sep` are win32, URL conversion is POSIX.

`packages/platform-node-shared/src/NodePath.ts:18-38` defines the pair once over `node:url`'s `fileURLToPath` / `pathToFileURL`, both of which branch on the *host* platform, and spreads that same pair into `layerPosix`, `layerWin32` and `layer` alike.

### Repro

macOS, `@effect/vitest`:

```ts
import * as NodePath from "@effect/platform-node/NodePath"
import { assert, it } from "@effect/vitest"
import { Effect, Path } from "effect"

it.layer(NodePath.layerWin32)("win32 file URLs", (it) => {
  it.effect("round-trips win32 paths", () =>
    Effect.gen(function*() {
      const path = yield* Path.Path
      assert.strictEqual(path.sep, "\\") // passes
      assert.strictEqual(
        (yield* path.toFileUrl("C:\\Users\\me\\file.txt")).href,
        "file:///C:/Users/me/file.txt"
      ) // fails
      assert.strictEqual(
        yield* path.fromFileUrl(new URL("file:///C:/Users/me/file.txt")),
        "C:\\Users\\me\\file.txt"
      ) // fails
    }))
})
```

Actual on darwin:

```text
toFileUrl("C:\\Users\\me\\file.txt")       -> file:///<cwd>/C:%5CUsers%5Cme%5Cfile.txt
fromFileUrl(file:///C:/Users/me/file.txt)  -> "/C:/Users/me/file.txt"
```

The rest of the layer is correct off-Windows — I confirmed `sep === "\\"`, `delimiter === ";"`, `join("tmp", "file.txt") === "tmp\\file.txt"`, `dirname("C:\\a\\b\\c.txt") === "C:\\a\\b"`, `normalize("C:/a/../b") === "C:\\b"`, `parse("C:\\a\\b.txt").root === "C:\\"` and `relative("C:\\a\\b", "C:\\a\\c") === "..\\c"` all pass. It is only the URL conversion that leaks the host.

On a Windows host the same layer returns the expected values, so this is only observable off-Windows — which is exactly where a fixed-flavor layer is meant to be useful.

`layerPosix` has the mirror-image defect: on a Windows host, `node:url` would resolve `/tmp/file.txt` through `path.win32.resolve` and hand back `file:///C:/tmp/file.txt`, and `fileURLToPath(new URL("file:///tmp/file.txt"))` would give `\tmp\file.txt`. I have not executed that half — no Windows host here — so treat it as read from `node:url`'s platform branch rather than measured.

### Prior art in-tree

`DenoPath` gets this right. `packages/platform-deno/src/DenoPath.ts:61,76` thread `fileUrlOps(DenoPathPosix)` / `fileUrlOps(DenoPathWin)` so the conversion follows the layer's flavor, and `packages/platform-deno/test/DenoPath.test.ts:18,28` asserts both directions for both flavors. `platform-bun` re-exports the node implementation and so inherits the bug.

### Note on the fix

`node:url` exposes no win32/posix variants, so this can't be a re-export — the flavored conversion has to be implemented. Deno's `@std/path` version is the reference, and `effect/Path` already vendors the POSIX equivalent at `packages/effect/src/Path.ts:349,437`. Happy to open a PR along those lines, with the `DenoPath` file-URL tests mirrored into `platform-node-shared`.

Found while checking #6710, which was my own mistaken premise — `NodePath.layerWin32` does exist and does cover the separator testing I was asking for there.

spencerbeggs:
Closes #6913.

`NodePath` built `fromFileUrl` / `toFileUrl` once from `node:url`'s host-dependent defaults and spread that same pair into all three layers, so `layerWin32` disagreed with itself off Windows: `join` / `dirname` / `sep` were win32 while URL conversion stayed POSIX. `layerPosix` had the mirror defect on a Windows host.

`node:url` has exposed the flavor as an option since v22.1.0, so this threads it through rather than reimplementing the conversion: `fileUrlOps(false)` for `layerPosix`, `fileUrlOps(true)` for `layerWin32`, and `fileUrlOps(undefined)` for `layer`, which resolves to the host default and leaves the host layer's behavior unchanged.

The new test mirrors the existing `packages/platform-deno/test/DenoPath.test.ts` file-URL tests, which already assert this for `DenoPath.layerPosix` / `layerWin32`. Before the change the Windows case failed on darwin with `expected '/C:/Users/me/file.txt' to equal 'C:\Users\me\file.txt'`.

Two things worth a maintainer's judgement:

- `packages/platform-node-shared/package.json` declares `engines.node: ">=18.0.0"`, and on Node < 22.1 the extra options argument is silently ignored, so both fixed-flavor layers fall back to host behavior there. That is no worse than today, but it is not fixed either. Since Node 18 and 20 are both EOL, bumping `engines` seems more honest than vendoring a parallel copy of Node's conversion logic, but I did not want to make that call in this PR.
- The POSIX half of the new test is vacuous on a POSIX host, where `{ windows: false }` and the host agree, and this repo's CI is ubuntu-only (`runs-on: ubuntu-latest` across the whole shard/runtime matrix). So nothing here exercises the `layerPosix`-on-Windows case, and I have no Windows host to check it on either; that half of the fix is reasoned from `node:url`'s platform branch rather than measured. The win32 half is what actually demonstrates the change, and it runs everywhere.

Validation run locally on Node 26.5.0 / darwin: `pnpm lint-fix`, `pnpm check`, `vitest run --project @effect/platform-node-shared` (106 passed), `vitest run --project @effect/platform-node` (277 passed). One run of the shared suite hit a pre-existing flake in `NodeStream.test.ts` (`PassThrough.onEnd`, `src/NodeStream.ts:354`) that did not reproduce across three subsequent runs and is unrelated to this change.

Signed-off-by: C. Spencer Beggs <spencer@beggs.codes>

### DIFF
--- a/.changeset/fix-node-path-file-url-flavor.md
+++ b/.changeset/fix-node-path-file-url-flavor.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform-node-shared": patch
+---
+
+NodePath: `layerPosix` and `layerWin32` now convert between paths and `file:` URLs using their own platform flavor instead of the host's, so `fromFileUrl` and `toFileUrl` agree with the rest of the layer when the host platform differs. Requires Node 22.1 or later, which is where `node:url` gained the `windows` option; on older runtimes both keep the previous host-flavored behavior. Closes #6913.
--- a/.changeset/fix-node-path-file-url-flavor.md
+++ b/.changeset/fix-node-path-file-url-flavor.md
@@ -2,4 +2,4 @@
 "@effect/platform-node-shared": patch
 ---
 
-NodePath: `layerPosix` and `layerWin32` now convert between paths and `file:` URLs using their own platform flavor instead of the host's, so `fromFileUrl` and `toFileUrl` agree with the rest of the layer when the host platform differs. Requires Node 22.1 or later, which is where `node:url` gained the `windows` option; on older runtimes both keep the previous host-flavored behavior. Closes #6913.
+NodePath: `layerPosix` and `layerWin32` now convert between paths and `file:` URLs using their own platform flavor instead of the host's.
--- a/packages/platform-node-shared/src/NodePath.ts
+++ b/packages/platform-node-shared/src/NodePath.ts
@@ -15,27 +15,28 @@ import { BadArgument } from "effect/PlatformError"
 import * as NodePath from "node:path"
 import * as NodeUrl from "node:url"
 
-const fromFileUrl = (url: URL): Effect.Effect<string, BadArgument> =>
-  Effect.try({
-    try: () => NodeUrl.fileURLToPath(url),
-    catch: (cause) =>
-      new BadArgument({
-        module: "Path",
-        method: "fromFileUrl",
-        cause
-      })
-  })
-
-const toFileUrl = (path: string): Effect.Effect<URL, BadArgument> =>
-  Effect.try({
-    try: () => NodeUrl.pathToFileURL(path),
-    catch: (cause) =>
-      new BadArgument({
-        module: "Path",
-        method: "toFileUrl",
-        cause
-      })
-  })
+const fileUrlOps = (windows: boolean | undefined) => ({
+  fromFileUrl: (url: URL): Effect.Effect<string, BadArgument> =>
+    Effect.try({
+      try: () => NodeUrl.fileURLToPath(url, { windows }),
+      catch: (cause) =>
+        new BadArgument({
+          module: "Path",
+          method: "fromFileUrl",
+          cause
+        })
+    }),
+  toFileUrl: (path: string): Effect.Effect<URL, BadArgument> =>
+    Effect.try({
+      try: () => NodeUrl.pathToFileURL(path, { windows }),
+      catch: (cause) =>
+        new BadArgument({
+          module: "Path",
+          method: "toFileUrl",
+          cause
+        })
+    })
+})
 
 /**
  * Provides the `Path` service using Node's POSIX path implementation plus
@@ -47,8 +48,7 @@ const toFileUrl = (path: string): Effect.Effect<URL, BadArgument> =>
 export const layerPosix: Layer.Layer<Path> = Layer.succeed(Path)({
   [TypeId]: TypeId,
   ...NodePath.posix,
-  fromFileUrl,
-  toFileUrl
+  ...fileUrlOps(false)
 })
 
 /**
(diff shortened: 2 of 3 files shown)


spencerbeggs:
So, this is my fix for the issue. There are some failing tests in the Node layer, but those don't seem to be possibly related to my changes. I think they are test timeouts. Let me know what you think and if you would like any changes.

spencerbeggs:
Note on the one red check, so it does not cost a reviewer time.

`Test (1/2, Node)` fails in [this job](https://github.com/Effect-TS/effect/actions/runs/30825286918/job/91725313250) on `packages/effect/test/cluster/ClusterWorkflowEngine.test.ts` — `Workflow.withCompensation runs compensation when the workflow fails` and `Activity.raceAll returns the first activity and interrupts losers`, both reporting `Test timed out in 5000ms`. 181 test files passed, that one failed.

It is not reachable from this diff. The test imports only from `effect` core, and the `effect` package has no dependency on `@effect/platform-node-shared` — the dependency runs the other way. The file also passes locally on this branch, 13/13.

The change itself did get exercised: `Test (2/2, Node)` ran the new file and passed.

```text
✓  @effect/platform-node-shared  test/NodePath.test.ts (2 tests) 11ms
```

For what it is worth, the same base commit (`8f9499f56`) went fully green on #6917, and recent runs across other branches have failed on a different unrelated test most times — `DenoFileSystem`, `HashRing`, `NumberBindingRepro`, `sql-mysql2 Persistence`. 

I can push an empty commit to re-roll it if that is useful, but I did not want to spend a CI run on cosmetics without asking.

tim-smart:
Does deno need updating too?

spencerbeggs:
@tim-smart Deno did it the right way (or this way). This just mirrors how Deno works into the `@effect/platform-node-shared` package with is how Node and Bun use it.

.changeset/fix-node-path-file-url-flavor.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/platform-node-shared": patch
+---
+
+NodePath: `layerPosix` and `layerWin32` now convert between paths and `file:` URLs using their own platform flavor instead of the host's, so `fromFileUrl` and `toFileUrl` agree with the rest of the layer when the host platform differs. Requires Node 22.1 or later, which is where `node:url` gained the `windows` option; on older runtimes both keep the previous host-flavored behavior. Closes #6913.
```
tim-smart:
```suggestion
NodePath: `layerPosix` and `layerWin32` now convert between paths and `file:` URLs using their own platform flavor instead of the host's.
```

PR #6917, opened 2026-08-03
Pass already valid class instances through Schema make unchanged

### LINKED ISSUE #6890: Schema class constructors deep re-construct Array(class) field elements but pass Array(union) elements through — 126x difference, silently divergent semantics
**Which version:** effect@4.0.0-beta.101 (v4 line).

A `Schema.Class` field typed as an array OF a class deep re-constructs every element on `make`, even when every element is already a valid instance of exactly that class — while the same field typed as a union (including a one-member union of the same class) passes valid instances through by identity.

Measured while fixing a parser-performance issue in a markdown AST (all elements pre-built, valid instances):

- `children: Schema.Array(TableRow)` — one `Table.make` over 30,000 pre-built rows: **1137ms**, element identity lost (`result.children[0] !== rows[0]`).
- `children: Schema.Array(Schema.Union([TableRow]))` — same data, same shape: **9ms**, identity preserved.

That is a ~126x difference on a semantics axis that is invisible in the types: both spellings declare the same `.d.ts` shape, decode the same wire form, and reject invalid members — but one re-constructs (allocating a fresh tree and discarding caller identity) and the other does not. A consumer building large documents from pre-validated nodes pays the cost only under the class spelling, and code that relies on element identity (caches, WeakMaps, mutation-by-reference audits) behaves differently between two schemas that look equivalent.

Either direction seems preferable to the current split: pass through already-valid instances in the class path too (a fast path via an instance check, matching what the union path already does), or document the re-construction as the class-field contract so the union spelling is the sanctioned identity-preserving form.

Repro sketch (beta.101):

```ts
class Row extends Schema.Class<Row>("Row")({ v: Schema.String }) {}
class TableA extends Schema.Class<TableA>("TableA")({ rows: Schema.Array(Row) }) {}
const RowU = Schema.Union([Row]);
class TableB extends Schema.Class<TableB>("TableB")({ rows: Schema.Array(RowU) }) {}
const rows = Array.from({ length: 30000 }, (_, i) => Row.make({ v: String(i) }));
// time TableA.make({ rows }) vs TableB.make({ rows }); compare rows[0] identity on both results

```

spencerbeggs:
Closes #6890.

`toConstructorAST` rewrites every class `Declaration` into its constructor form unconditionally, so a field typed `Schema.Array(MyClass)` rebuilds each element on `make` even when the element is already an instance of exactly that class. The result is a fresh tree and the loss of the caller's references.

This registers the constructor-form ASTs it produces, and `makeParser` consults that set: when the input already satisfies the declaration and its checks, it is returned unchanged and the encoding chain is skipped. Only the constructor direction is affected, since `toConstructorAST` is reached solely from `makeEffect`; decoding and encoding are untouched.

The failure from that attempt is deliberately discarded rather than reported. The encoding chain below still runs and produces the authoritative issue, so error messages are byte-identical to before. That detail matters: a first attempt at this modelled the fast path as `Union([declaration, constructorForm])`, which worked and was faster, but made every failed `make` on a class field carry a spurious `Expected B, got {...}` prefix from the union's other branch. That approach is not in this PR.

### Measured

`Table.make({ rows })` over 30,000 pre-built `Row` instances, median of 15 runs after warmup, Node 26.5.0:

| | before | after |
| --- | --- | --- |
| `Schema.Array(Row)` | 9.27ms, identity lost | 0.74ms, identity preserved |

For reference the issue's `Schema.Array(Schema.Union([Row]))` workaround measures 1.84ms, so the class spelling is now comfortably ahead of it. (These numbers are measured on this branch after merging #6912; an earlier revision of this description quoted 8.69ms/1.51ms from the pre-merge base.)

### On the issue's second suggestion

#6890 offered an alternative: document the re-construction as the class contract and treat the union spelling as the sanctioned identity-preserving form. That does not hold up. The union spelling is not an equivalent schema in the constructor direction — it rejects plain-object input outright and silently skips constructor defaults inside its members:

```
class  Array(Row).make({v:"a"})           OK   -> {"v":"a","tag":"DEFAULTED"}
union  Array(Union([Row])).make({v:"a"})  FAIL -> Expected Row, got {"v":"a"}
```

The cause is that `toConstructorAST` has no `case "Union"`, so unions fall through to `default: return ast` and never receive constructor semantics at all. Its identity preservation is a side effect of that gap. I have left that behavior alone here — it looks like a separate bug, and I did not want to widen this PR into it — but it is worth a maintainer's eye.

Also worth correcting from the issue: the reported 126x was a cold single-shot measurement on beta.101. On `main` the steady-state gap is about 5x.

### Validation

`pnpm lint-fix`, `pnpm check`, and `vitest run --project effect` — 244 files, 7658 passed, 3 skipped, including the full schema suite.

On the two new tests: disabling the guard makes `make passes already valid class instances through unchanged` fail with `Values have same structure but are not reference-equal`, so that one discriminates this change directly. The second test does not — it still passes with the guard off, because guard-off is simply the old behavior. It is a regression guard aimed at the `Union` shape described above, which is the mutant it actually catches.

### Behavior change to weigh

`make` no longer returns defensive copies of class inputs. Anyone relying on the previous re-construction to isolate caller-held references will now observe the originals. That is the point of the change, but it is a visible semantic shift and belongs in a maintainer's judgement rather than mine — happy to gate it behind an option instead if you would prefer.

Signed-off-by: C. Spencer Beggs <spencer@beggs.codes>


gcanti:
Closing in favor of https://github.com/Effect-TS/effect/pull/7020

spencerbeggs:
Thanks. That is a better fix!

PR #6933, opened 2026-08-04
Stabilize blackholed query recovery test

tim-smart:
## Summary

- replace the fixed recovery delay with a wait for the timed-out blackholed query to actually exit
- keep the strict single-active-query assertion intact
- avoid starting the recovery query while detached interruption is still pending on a busy runner

## Root cause

The lock-operation deadline intentionally returns before waiting for detached interruption to finish. The test used a 150 ms sleep as a proxy for that cleanup, so a sufficiently delayed CI runner could start the recovery query while the original query was still counted as active.

## Verification

- focused blackholed-query recovery test: 8 consecutive post-fix runs
- complete SqlRunnerStorage integration file: 22 tests passed
- dprint check on the changed file
- oxlint on the changed file
- @effect/platform-node typecheck

Closes EFF-377

### DIFF
--- a/packages/platform-node/test/cluster/SqlRunnerStorage.integration.test.ts
+++ b/packages/platform-node/test/cluster/SqlRunnerStorage.integration.test.ts
@@ -110,7 +110,7 @@ describe("SqlRunnerStorage", () => {
       yield* storage.acquire(runnerAddress1, shards)
       partitioned.current = true
       yield* storage.refresh(runnerAddress1, shards).pipe(Effect.exit, TestClock.withLive)
-      yield* Effect.sleep(150).pipe(TestClock.withLive)
+      yield* waitUntil(() => partitioned.activeQueries === 0)
 
       partitioned.current = false
       expect(


tim-smart (commented):
## Review: sound fix, no new hang risk, strict assertion still meaningful

**Root cause verified against the implementation.** `withDeadline` in `packages/effect/src/unstable/cluster/SqlRunnerStorage.ts:144-155` forks the lock operation, times out the *join*, and interrupts the fiber detached in the layer scope (`Fiber.interrupt` + `forkIn(layerScope)`). So the timed-out `refresh` at line 112 can return while the wedged query's interruption is still pending, and the wrapper's `onExit` (which decrements `partitioned.activeQueries`) hasn't run yet. The fixed 150 ms sleep was a proxy for that detached cleanup; waiting on `activeQueries === 0` waits for the actual condition. Correct fix, not a loosened assertion.

**Hang risk: bounded.** Two layers of protection:

1. `waitUntil` (test file, lines 349-360) is wrapped in `Effect.timeoutOrElse({ duration: 10_000, orElse: () => Effect.die(...) })` — if the cleanup never completes, the test *fails* with a defect after 10 s instead of hanging.
2. In this test `resumePending` is `false`, so the wedged query is `Effect.never`, which is fully interruptible — interruption always completes promptly and `onExit` always runs. The condition is reachable by construction.

Additionally the test itself has a 60 s vitest timeout as a backstop.

**Strict assertion still tests the intended property.** `maxActiveQueries <= 1` at line 123 verifies the storage never runs two lock queries concurrently. By waiting for the original query to fully exit before `partitioned.current = false` and the recovery refresh, any `maxActiveQueries === 2` would now indicate a genuine overlap in `SqlRunnerStorage`, not test-side raciness. The assertion's meaning is preserved (arguably strengthened, since it is no longer polluted by the fixture's own timing).

**Minor, non-blocking observations:**

- The recovery `refresh` relies on the pre-existing `Effect.retry({ times: 5, schedule: Schedule.spaced(20) })` window (~100 ms) for the async connection rebuild to land. Since `waitUntil(activeQueries === 0)` typically resolves *faster* than the old 150 ms sleep, the rebuild has slightly less headroom before the first retry. This was already absorbed by the retries and isn't a regression introduced here, but if this test ever flakes again with a refresh failure rather than the assertion, that retry window is the place to look.
- `waitUntil` dying (rather than failing an assertion) on timeout is fine for a test helper and consistent with its existing uses in this file.

LGTM.

PR #6934, opened 2026-08-04
httpapi: WithHeaders response header schemas

tim-smart:
## Summary

- Add `HttpApiSchema.WithHeaders` and branded `withHeaders` response values for typed success headers across handlers and generated clients.
- Add `HttpApiSchema.encodeToWithHeaders` for folding response headers into domain types such as error classes.
- Encode and decode buffered and streaming response headers in `HttpApiBuilder`, `HttpApiClient`, and `HttpApiTest`, including mixed-success dispatch and user-managed codecs.
- Render declared response headers in OpenAPI and document the public API, stream behavior, and overlapping-union constraint.
- Make explicit `content-type` and `content-length` values applied through `HttpServerResponse.setHeader` or `setHeaders` override body-derived values.

Closes EFF-369
Closes EFF-448

## Validation

- `pnpm check`
- `pnpm lint`
- `pnpm vitest --run --project effect test/unstable/httpapi/` (150 passed)
- `pnpm doctest --run packages/effect/src/unstable/httpapi/HttpApiSchema.ts` (3 passed)
- Root `pnpm docgen` reaches only the pre-existing `HttpStaticServer` example error for the missing `HttpPlatform.compression` field; the new examples typecheck and execute successfully.

### DIFF
--- a/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
+++ b/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
@@ -804,14 +804,19 @@ function handlerToHttpEffect(
       }
       let responseHeaders: unknown | undefined
       if (encodeWithHeaders !== undefined && HttpApiSchema.isWithHeadersValue(response)) {
-        responseHeaders = yield* HttpApiSchemaError.wrap("ResponseHeaders", encodeWithHeaders(response.headers))
+        responseHeaders = response.headers
         response = response.body
       }
       const encoded = yield* HttpApiSchemaError.wrap(
         "Body",
         encodeStream?.(response, context) ?? encodeSuccess(response)
       )
-      return responseHeaders === undefined ? encoded : Response.setHeaders(encoded, responseHeaders as any)
+      if (responseHeaders === undefined) return encoded
+      const encodedHeaders = yield* HttpApiSchemaError.wrap(
+        "ResponseHeaders",
+        encodeWithHeaders!.get(encoded.status)!(responseHeaders)
+      )
+      return Response.setHeaders(encoded, encodedHeaders as any)
     })
   ).pipe(
     Effect.withErrorReporting,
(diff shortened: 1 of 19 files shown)


packages/effect/src/unstable/httpapi/HttpApiClient.ts
```diff
@@ -722,9 +736,43 @@ const compilePath = (path: string) => {
...
+    HttpApiSchema.isWithHeaders(schema) || HttpApiSchema.getWithHeadersAnnotation(schema.ast) !== undefined
+  )
+  const codec = hasWithHeaders
+    ? Schema.Union(schemas.map(toCodecArrayBufferWithHeaders))
+    : toCodecArrayBuffer(schemas)
   const decode = Schema.decodeEffect(codec)
-  return (response: HttpClientResponse.HttpClientResponse) => Effect.flatMap(response.arrayBuffer, decode)
+  return (response: HttpClientResponse.HttpClientResponse) =>
+    Effect.flatMap(
+      response.arrayBuffer,
```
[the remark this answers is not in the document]
tim-smart: Confirmed. This stage is intentionally limited to the buffered `HttpApiClient` path, so client-side `WithHeaders<StreamSse | StreamUint8Array, ...>` integration is deferred to the finalize stage. No streaming behavior change is being made in this stage.

packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
```diff
@@ -797,15 +798,23 @@ function handlerToHttpEffect(
           request.payload = yield* HttpApiSchemaError.wrap("Payload", result)
         }
       }
-      const response = yield* handler(request)
+      let response = yield* handler(request)
       if (Response.isHttpServerResponse(response)) {
         return response
       }
+      let responseHeaders: unknown | undefined
+      if (encodeWithHeaders !== undefined && hasProperty(response, HttpApiSchema.WithHeadersValueTypeId)) {
```
tim-smart: Instead of using hasProperty use a type guard predicate which narrows the type.

packages/effect/src/unstable/httpapi/HttpApiEndpoint.ts
```diff
@@ -1156,6 +1176,9 @@ function getErrorResponse(
     if (HttpApiSchema.isStreamSchema(schema)) {
       throw new Error("Streaming schemas are not supported in error responses")
     }
+    if (HttpApiSchema.isWithHeaders(schema)) {
```
tim-smart: I think WithHeaders should be support for errors, but it just won't be recommended.

tim-smart: Addressed in ebc763218. `WithHeaders` is now supported for non-streaming error schemas (while documented as less convenient than `encodeToWithHeaders`): endpoint construction preserves the wrapper/codecs, the builder encodes the branded error body and headers, and client/OpenAPI status handling is wrapper-aware. Added end-to-end server/client and OpenAPI coverage.

packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
tim-smart: Addressed in ebc763218. I exported and used the existing `isWithHeadersValue` type guard, so this branch now narrows the response value directly and no cast is needed.

packages/effect/src/unstable/httpapi/HttpApiSchema.ts
```diff
@@ -467,6 +484,322 @@ function defaultStreamContentType(mode: StreamMode): string {
...
+ *   `StreamSse` and `StreamUint8Array`; error responses remain non-streaming.
+ *   Nesting `WithHeaders` is rejected at construction.
+ * - `headers` is any schema; integrations encode and decode it via
+ *   `Schema.toCodecStringTree`, so leaves become `string | undefined` on the
+ *   wire and `undefined` leaves are omitted from the response.
+ * - Status and response-encoding annotations are resolved from the wrapper
+ *   first, falling through to the inner schema.
+ * - When a success union has multiple `WithHeaders` members, header encoding
+ *   uses the first matching union member. Keep their header shapes disjoint so
+ *   a value cannot encode against the wrong member.
```
wmaurer: Did you leave this to the caller on purpose, instead of checking it when the schema is built? In #6880 I rejected a `WithHeaders` member when another member had the same `(status, content-type)` pair. Normal unions with the same status still work. It only catches the case where the headers schema is ambiguous, not just the body.

If you want the check, there is one detail. `isWithHeaders` looks for `WithHeadersTypeId`, but schemas from `encodeToWithHeaders` do not have it. They are `decodeTo` schemas with the `~httpApiWithHeaders` annotation. So the check would also need `getWithHeadersAnnotation`, otherwise it misses the error-header case.

tim-smart: Will look into it :)

packages/effect/test/unstable/httpapi/HttpApiBuilder.test.ts
```diff
@@ -122,6 +133,610 @@ it.layer(TestServices)("HttpApiBuilder payload content types", (it) => {
...
+      )
+      const GroupLive = HttpApiBuilder.group(
+        Api,
+        "test",
+        (handlers) =>
+          handlers.handle("result", () =>
+            Effect.succeed(HttpApiSchema.withHeaders({
+              body: { _tag: "B" as const },
+              headers: { "x-source": "value" }
+            }) as any))
```
tim-smart, line 236: This `as any` should not be needed by the user.

PR #6988, opened 2026-08-04
Preserve lexical ordering in streaming template interpolation

fubhy:
## Summary

Template.stream allows later static and fast segments to overtake an earlier delayed interpolation, changing a${slow}b${fast}c into abfastcslow rather than aslowbfastc.

> [!IMPORTANT]

## Streaming interpolation does not preserve template order

**Module:** `Template`  
**Audit ID:** `unstable-http-template-stream-order`  

### What happens

Template.stream allows later static and fast segments to overtake an earlier delayed interpolation, changing a${slow}b${fast}c into abfastcslow rather than aslowbfastc.

### Why it happens

The lexical segment stream is flattened with concurrency set to unbounded, so later segments can emit before an earlier effect or stream interpolation completes.

### Expected behavior

A template stream must emit static and interpolated segments in lexical template order.

### Relevant implementation

- [`packages/effect/src/unstable/http/Template.ts:218-223`](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/effect/src/unstable/http/Template.ts#L218-L223)

<details>
<summary>View problematic code at <code>packages/effect/src/unstable/http/Template.ts:218-223</code></summary>

```ts
  return Stream.flatMap(
    Stream.fromIterable(chunks),
    (chunk) =>
      typeof chunk === "string" ? Stream.succeed(chunk) : Effect.isEffect(chunk) ? Stream.fromEffect(chunk) : chunk,
    { concurrency: "unbounded" }
  )
```

[View exact lines on GitHub](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/effect/src/unstable/http/Template.ts#L218-L223)

</details>

### Reproduction

```sh
pnpm test --run packages/effect/test/unstable/http/Template.test.ts
```

**Observed failure:** Failed deterministically under TestClock with abfastcslow instead of aslowbfastc.

## Implementation handoff

```sh
pnpm test --run packages/effect/test/unstable/http/Template.test.ts
```

## Audit provenance

- Findings: `unstable-http-template-stream-order`

Closes EFF-425

### DIFF
--- a/packages/effect/src/unstable/http/Template.ts
+++ b/packages/effect/src/unstable/http/Template.ts
@@ -218,8 +218,7 @@ export function stream<A extends ReadonlyArray<InterpolatedWithStream>>(
   return Stream.flatMap(
     Stream.fromIterable(chunks),
     (chunk) =>
-      typeof chunk === "string" ? Stream.succeed(chunk) : Effect.isEffect(chunk) ? Stream.fromEffect(chunk) : chunk,
-    { concurrency: "unbounded" }
+      typeof chunk === "string" ? Stream.succeed(chunk) : Effect.isEffect(chunk) ? Stream.fromEffect(chunk) : chunk
   )
 }
 
--- a/packages/effect/src/unstable/http/Template.ts
+++ b/packages/effect/src/unstable/http/Template.ts
@@ -215,10 +215,12 @@ export function stream<A extends ReadonlyArray<InterpolatedWithStream>>(
     buffer = ""
   }
 
-  return Stream.flatMap(
-    Stream.fromIterable(chunks),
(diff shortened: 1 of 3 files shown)


packages/effect/src/unstable/http/Template.ts
```diff
@@ -218,8 +218,7 @@ export function stream<A extends ReadonlyArray<InterpolatedWithStream>>(
   return Stream.flatMap(
```
tim-smart: Maybe we can use mapEffect instead to preserve order and concurrency?

PR #6990, opened 2026-08-04
Generate unique paths for duplicate multipart filenames

fubhy:
## Summary

Two ordinary file parts named same.txt receive the same persisted path, so the later write replaces the bytes referenced by both returned PersistedFile values and silently loses one file.

> [!IMPORTANT]

## Equal multipart filenames overwrite each other during persistence

**Module:** `Multipart`  
**Audit ID:** `unstable-http-multipart-filename-collision`  

### What happens

Two ordinary file parts named same.txt receive the same persisted path, so the later write replaces the bytes referenced by both returned PersistedFile values and silently loses one file.

### Why it happens

Each file path is derived only as join(tempDir, basename(file.name).slice(-128)). Parts with equal filenames are written sequentially to one path, and both returned values point to the final write.

### Expected behavior

toPersisted must preserve every valid multipart file part and return a path containing that part's bytes for the scope lifetime.

### Relevant implementation

- [`packages/effect/src/unstable/http/Multipart.ts:672-708`](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/effect/src/unstable/http/Multipart.ts#L672-L708)
- [`packages/effect/src/unstable/http/Multipart.ts:695`](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/effect/src/unstable/http/Multipart.ts#L695)

<details>
<summary>View problematic code at <code>packages/effect/src/unstable/http/Multipart.ts:672-708</code></summary>

```ts
export const toPersisted = (
  stream: Stream.Stream<Part, MultipartError>,
  writeFile = defaultWriteFile
): Effect.Effect<Persisted, MultipartError, FileSystem.FileSystem | Path.Path | Scope.Scope> =>
  Effect.gen(function*() {
    const fs = yield* FileSystem.FileSystem
    const path_ = yield* Path.Path
    const dir = yield* fs.makeTempDirectoryScoped()
    const persisted: Record<string, Array<PersistedFile> | Array<string> | string> = Object.create(null)
    yield* Stream.runForEach(stream, (part) => {
      if (part._tag === "Field") {
        if (!(part.key in persisted)) {
          persisted[part.key] = part.value
        } else if (typeof persisted[part.key] === "string") {
          persisted[part.key] = [persisted[part.key] as string, part.value]
        } else {
          ;(persisted[part.key] as Array<string>).push(part.value)
        }
        return Effect.void
      } else if (part.name === "") {
        return Effect.void
      }
      const file = part
      const path = path_.join(dir, path_.basename(file.name).slice(-128))
      const filePart = new PersistedFileImpl(
        file.key,
        file.name,
        file.contentType,
        path
      )
      if (Array.isArray(persisted[part.key])) {
        ;(persisted[part.key] as Array<PersistedFile>).push(filePart)
      } else {
        persisted[part.key] = [filePart]
      }
      return writeFile(path, file)
    })
```

[View exact lines on GitHub](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/effect/src/unstable/http/Multipart.ts#L672-L708)

</details>
<details>
<summary>View problematic code at <code>packages/effect/src/unstable/http/Multipart.ts:695</code></summary>

```ts
      const path = path_.join(dir, path_.basename(file.name).slice(-128))
```

[View exact lines on GitHub](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/effect/src/unstable/http/Multipart.ts#L695)

</details>

### Reproduction

```sh
pnpm test --run packages/effect/test/unstable/http/Multipart.test.ts
```

**Observed failure:** Failed as intended because both writes used /tmp/audit/same.txt and the returned paths were equal.

## Implementation handoff

```sh
pnpm test --run packages/effect/test/unstable/http/Multipart.test.ts
```

## Audit provenance

- Findings: `unstable-http-multipart-filename-collision`

Closes EFF-427

### DIFF
--- a/packages/effect/src/unstable/http/Multipart.ts
+++ b/packages/effect/src/unstable/http/Multipart.ts
@@ -678,6 +678,7 @@ export const toPersisted = (
     const path_ = yield* Path.Path
     const dir = yield* fs.makeTempDirectoryScoped()
     const persisted: Record<string, Array<PersistedFile> | Array<string> | string> = Object.create(null)
+    let fileIndex = 0
     yield* Stream.runForEach(stream, (part) => {
       if (part._tag === "Field") {
         if (!(part.key in persisted)) {
@@ -692,7 +693,7 @@ export const toPersisted = (
         return Effect.void
       }
       const file = part
-      const path = path_.join(dir, path_.basename(file.name).slice(-128))
+      const path = path_.join(dir, `${fileIndex++}-${path_.basename(file.name).slice(-128)}`)
       const filePart = new PersistedFileImpl(
         file.key,
         file.name,
--- a/packages/effect/src/unstable/http/Multipart.ts
+++ b/packages/effect/src/unstable/http/Multipart.ts
@@ -678,6 +678,7 @@ export const toPersisted = (
     const path_ = yield* Path.Path
     const dir = yield* fs.makeTempDirectoryScoped()
     const persisted: Record<string, Array<PersistedFile> | Array<string> | string> = Object.create(null)
+    const usedPaths = new Set<string>()
     let fileIndex = 0
     yield* Stream.runForEach(stream, (part) => {
       if (part._tag === "Field") {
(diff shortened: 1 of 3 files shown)


packages/effect/src/unstable/http/Multipart.ts
```diff
@@ -692,7 +693,7 @@ export const toPersisted = (
         return Effect.void
       }
       const file = part
-      const path = path_.join(dir, path_.basename(file.name).slice(-128))
+      const path = path_.join(dir, `${fileIndex++}-${path_.basename(file.name).slice(-128)}`)
```
tim-smart: I think it should only add the prefix when the name is already taken

PR #7003, opened 2026-08-04
Persist permanent entries in KVS setMany operations

fubhy:
## Summary

KVS setMany silently skips every entry without a finite TTL, so values requested as permanent are absent immediately after the operation succeeds.

> [!IMPORTANT]

## KVS setMany skips permanent entries

**Module:** `Persistence`  
**Audit ID:** `unstable-state-p-1`  

### What happens

KVS setMany silently skips every entry without a finite TTL, so values requested as permanent are absent immediately after the operation succeeds.

### Why it happens

unsafeTtlToExpires returns null for an undefined TTL, but the setMany loop treats null as a reason to return Effect.void instead of writing the permanent entry.

### Expected behavior

BackingPersistenceStore.setMany has the same persistence semantics as repeated set calls, and an undefined TTL means no expiration.

### Relevant implementation

- [`packages/effect/src/unstable/persistence/Persistence.ts:1094-1100`](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/effect/src/unstable/persistence/Persistence.ts#L1094-L1100)

<details>
<summary>View problematic code at <code>packages/effect/src/unstable/persistence/Persistence.ts:1094-1100</code></summary>

```ts
          setMany: (entries) =>
            Effect.forEach(entries, ([key, value, ttl]) => {
              const expires = unsafeTtlToExpires(clock, ttl)
              if (expires === null) return Effect.void
              const encoded = JSON.stringify([value, expires])
              return store.set(key, encoded)
            }, { concurrency: "unbounded", discard: true }).pipe(
```

[View exact lines on GitHub](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/effect/src/unstable/persistence/Persistence.ts#L1094-L1100)

</details>

### Reproduction

```sh
pnpm test --run packages/effect/test/unstable/persistence/KeyValueStore.test.ts
```

**Observed failure:** get("key") returned undefined after permanent setMany completed.

## Implementation handoff

```sh
pnpm test --run packages/effect/test/unstable/persistence/KeyValueStore.test.ts
```

## Audit provenance

- Findings: `unstable-state-p-1`

Closes EFF-440

### DIFF
--- a/packages/effect/test/unstable/persistence/KeyValueStore.test.ts
+++ b/packages/effect/test/unstable/persistence/KeyValueStore.test.ts
@@ -1,6 +1,7 @@
-import { afterEach, assert, describe, it } from "@effect/vitest"
+import { afterEach, describe, it } from "@effect/vitest"
 import { assertTrue, deepStrictEqual, strictEqual } from "@effect/vitest/utils"
-import { Effect, Layer, Option, Schema } from "effect"
+import type { Layer } from "effect"
+import { Effect, Option, Schema } from "effect"
 import * as KeyValueStore from "effect/unstable/persistence/KeyValueStore"
 import * as Persistence from "effect/unstable/persistence/Persistence"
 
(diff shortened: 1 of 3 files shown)


packages/effect/test/unstable/persistence/KeyValueStore.test.ts
```diff
@@ -88,6 +88,21 @@ export const testLayer = <E>(layer: Layer.Layer<KeyValueStore.KeyValueStore, E>)
 
 describe("KeyValueStore / layerMemory", () => testLayer(KeyValueStore.layerMemory))
 
+describe("Persistence / layerBackingKvs", () => {
+  it.effect("stores entries without a TTL", () =>
```
tim-smart: Can we move this into the shared test suite

PR #7009, opened 2026-08-04
Share persistence layers across cache tests

tim-smart:
## Summary

- acquire each persistence layer once per PersistedCache suite instead of once per test
- keep container setup and teardown outside the individual 30-second test budget
- remove the timeout wrapper that could turn an unexecuted test body into a passing test
- align the IndexedDB fixture lifetime with the shared suite layer

## Root cause

The failing MySQL test was not blocked by a PersistedCache race. Under CI load, each test independently started and stopped a `mysql:lts` container, with individual cases taking 24–30 seconds. The 10-second timeout wrapper then interrupted the test and returned success while container cleanup continued, so the outer 30-second Vitest timeout could still fire—and nominal passes could skip the assertions entirely.

## Validation

- `pnpm lint-fix`
- `pnpm test --run packages/effect/test/unstable/persistence/PersistedCache.test.ts --reporter=verbose`
- `pnpm test --run packages/platform-browser/test/BrowserPersistencePersistedCache.test.ts --reporter=verbose`
- `pnpm check`

The MySQL integration test could not run locally because this environment has no container runtime; the original Actions log was used to confirm the resource-lifecycle timing pattern.

Closes EFF-445

### DIFF
--- a/packages/effect/test/unstable/persistence/PersistedCacheTest.ts
+++ b/packages/effect/test/unstable/persistence/PersistedCacheTest.ts
@@ -1,6 +1,6 @@
-import { assert, describe, it } from "@effect/vitest"
+import { assert, it } from "@effect/vitest"
 import type { Layer } from "effect"
-import { Context, Data, Effect, Exit, Result, Schema } from "effect"
+import { Context, Data, Effect, Exit, Schema } from "effect"
 import { Persistable, PersistedCache, Persistence } from "effect/unstable/persistence"
 
 class User extends Schema.Class<User>("User")({
@@ -21,7 +21,7 @@ export class TransientError extends Data.TaggedError("TransientError") {}
 class LookupService extends Context.Service<LookupService, { readonly value: string }>()("LookupService") {}
 
 export const suite = (storeId: string, layer: Layer.Layer<Persistence.Persistence, unknown>) =>
-  describe(`PersistedCache (${storeId})`, { timeout: 30_000 }, () => {
+  it.layer(layer, { timeout: "60 seconds" })(`PersistedCache (${storeId})`, (it) => {
     it.effect("smoke test", () =>
       Effect.gen(function*() {
         const persistence = yield* Persistence.Persistence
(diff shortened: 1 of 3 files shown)


tim-smart (commented):
## Review: Share persistence layers across cache tests

Verdict: approve. No correctness blockers. Both concerns from the review request check out.

### Shared layer / state leakage — safe

- `it.layer` builds the layer once per suite in `beforeAll` (memoized, `Effect.cached`) and closes the scope in `afterAll`, with the `{ timeout: "60 seconds" }` applied to those hooks (`packages/vitest/src/internal/internal.ts:243`, `:317-325`). Container setup/teardown is now fully outside the 30s per-test budget, which is exactly the fix the flake needed.
- Within a suite the two tests touch disjoint stores: `users` (smoke test) vs `${storeId}-require-services-at` (lookup test) (`packages/effect/test/unstable/persistence/PersistedCacheTest.ts:35`, `:75`). No shared keys, and tests in a file run sequentially, so no intra-suite leakage.
- The `sql-mysql2-multi` / `sql-mysql2-single` (and pg) suites both use store `users` with the same keys, but each `it.layer` call builds its own layer with its own memo map, and `MysqlContainer.layerClient` is a fresh `acquireRelease` per build (`packages/sql/mysql2/test/utils.ts:33`) — separate containers, separate databases. No cross-suite leakage.
- This pattern is already the established convention: `PersistedQueueTest.suite` has shared its layer the same way (`packages/effect/test/unstable/persistence/PersistedQueueTest.ts:6-12`).
- Browser fixture: nested suite `afterAll` (layer scope close) runs before the file-level `afterAll` that calls `fakeIndexedDb.deleteDatabase`, so the DB is deleted after the layer releases it. Fine for fake-indexeddb.

### Timeout wrapper removal — correct, and strictly better

- The old `flakyTest` wrapper (`Effect.timeoutOrElse({ duration: "10 seconds", orElse: () => Effect.void })`) converted a timeout into a pass, and interrupting mid container start/stop could orphan docker resources. Removing it means a genuinely hanging test now fails at the 30s test timeout instead of silently passing. Good.
- Also removed: `Effect.catchFilter(... TransientError ..., () => Effect.void)`, which converted `TransientError` failures into passes. That was the silent-skip path for the Redis suites when the container fails to start (`NodeRedis.integration.test.ts:20` maps container-start failure to `TransientError`). With `it.layer`, layer build runs in `beforeAll` under `Effect.orDie`, so those suites now fail loudly on container startup failure instead of passing. This matches the existing `PersistedQueueTest.suite` behavior, and integration tests only run under `EFFECT_INTEGRATION_TESTS=1` (`vitest.config.ts:11`, `:51`) where docker is expected — so this is an intentional, consistent behavior change, not a regression. Worth being aware of when running integration tests locally without docker.

### Non-blocking nits

- `vitest.config.ts:163` comment ("MySQL starts a fresh container per integration test") is now stale — it is per suite, not per test.
- `TransientError` remains exported only for the Redis test files; fine as-is.

CI snapshot at review time: all checks pending.

PR #7013, opened 2026-08-05
Vendor find-my-way router

tim-smart:
## Summary

- vendor the `find-my-way-ts` v0.1.6 radix router and query-string parser under Effect's `FindMyWay` module
- preserve the existing public API and Tomas Della Vedova's MIT attribution
- port all seven tests from the TypeScript package plus focused upstream regression coverage
- remove the external package from `effect` and the pnpm lockfile

## Upstream audit

The TypeScript rewrite forked from `delvedor/find-my-way` v7.7.0. Applicable fixes through current upstream were ported for:

- optional parameters at the root path
- safe multi-parameter separator matching and distinct route patterns
- prototype-safe method lookup and the `QUERY` HTTP method
- regex mismatch backtracking, empty captures, and `maxParamLength` fallback

Constraint routing, `off`/`findRoute`, and pretty-print fixes do not apply to this reduced API. The later semicolon-delimiter option was not included because it is an opt-in behavior/API change rather than a bug fix.

## Validation

- `pnpm lint-fix`
- all eight targeted `FindMyWay` test files (53 tests)
- `packages/effect/test/HttpClient.test.ts` and `packages/effect/test/unstable/ai/McpServer/McpServer.test.ts` (24 tests)
- `pnpm check`

Closes EFF-451

### DIFF
--- a/packages/effect/src/unstable/http/FindMyWay.ts
+++ b/packages/effect/src/unstable/http/FindMyWay.ts
@@ -1,15 +1,74 @@
-/**
- * Re-exports the `find-my-way-ts` router package used by the unstable HTTP
- * routing modules.
+/*
+ * MIT License
  *
- * This module keeps the router types and helpers available from the Effect HTTP
- * namespace without wrapping or changing them.
+ * Copyright (c) 2017-2019 Tomas Della Vedova
  *
- * @since 4.0.0
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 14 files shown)


packages/effect/src/unstable/http/FindMyWay.ts
```diff
@@ -1,15 +1,74 @@
-/**
- * Re-exports the `find-my-way-ts` router package used by the unstable HTTP
- * routing modules.
+/*
```
tim-smart, line 8: Move the licence under the imports

tim-smart:
@pullfrog Please review the latest changes addressing the open threads.

PR #7026, opened 2026-08-05
ExecutionPlan: lifecycle events for retries and execution observability

### LINKED ISSUE #6699: Effect AI ExecutionPlan: Add lifecycle hooks for retries and execution observability

ExecutionPlan provides a clean abstraction for retries, backoff strategies, and provider failover, but it currently doesn't expose any lifecycle events for individual execution attempts.

As a result, applications cannot observe or instrument:

Retry count
Per-attempt success/failure
Attempt duration
Provider failover events

This makes it difficult to integrate ExecutionPlan with observability tools.

While `Effect.tap`, `tapError`, and `withSpan` can observe the final outcome of an effect, they cannot observe intermediate attempts performed internally by ExecutionPlan. For example, if an operation fails twice before succeeding on the third attempt, there is currently no public API to record those retries or attach telemetry to them.

Introduce optional lifecycle hooks (or an observer interface) that can be attached to an ExecutionPlan. These hooks would be invoked during the execution lifecycle without changing the existing retry semantics.

Example:

```
const plan = ExecutionPlan.make({
  provide: GeminiFlash,
  attempts: 3,
  schedule: Schedule.exponential("100 millis", 1.5),
  while: (error) => error._tag === "NetworkError",

  observe: {
    onAttemptStart(event) {},
    onAttemptSuccess(event) {},
    onAttemptFailure(event) {},
    onRetry(event) {}
  },
})
```

This would allow users to collect metrics, emit tracing events, and implement custom logging while continuing to use ExecutionPlan as the retry mechanism, rather than having to replace it with manual retry logic solely for observability.

- Using `Effect.tap`, `tapError`, or `tapBoth`: These only observe the final outcome of the effect and do not provide visibility into intermediate retry attempts performed by ExecutionPlan.

- Instrumenting the underlying effect (e.g. `LanguageModel.generateObject`): This works for specific use cases but couples observability to the implementation being retried. It also doesn't provide ExecutionPlan-specific context such as the current step, attempt number, or provider transition.

- Replacing ExecutionPlan with `Effect.retry`: This allows full control over retries and observability, but it means giving up the higher-level ExecutionPlan abstraction for provider failover and execution strategies. The goal is to continue using ExecutionPlan while gaining first-class observability.

- Implementing a custom wrapper around ExecutionPlan: This would require duplicating or depending on Effect's internal implementation, which is not ideal and could become difficult to maintain as ExecutionPlan evolves.

Providing built-in lifecycle hooks would offer a standardized and maintainable way to integrate metrics, tracing, and logging without changing existing execution behavior.

tim-smart:
## Summary

Adds an optional `onEvent` handler to `Effect.withExecutionPlan` and `Stream.withExecutionPlan` so individual plan attempts can be observed from outside the effect, for logging and metrics.

```ts
Effect.withExecutionPlan(program, plan, {
  onEvent: (event) => Effect.log("execution plan event", event)
})
```

The handler receives an `ExecutionPlan.Event<E>`, a tagged union of:

- `AttemptStart` — emitted before each attempt runs
- `AttemptSuccess` — the attempt succeeded (always the final event), with `duration`
- `AttemptFailure` — the attempt failed, with `duration` and the full failure `Cause`

Each event carries `attempt` (cumulative across steps, matching `ExecutionPlan.CurrentMetadata.attempt`), `stepAttempt` (1-based within the current step), and `stepIndex`.

## Design notes

- The hook lives on the combinator rather than on `ExecutionPlan.make`, so a shared plan can be observed differently at different call sites.
- The handler is `(event) => Effect<void, never, R>`: it is awaited inline (events are strictly ordered), cannot fail (observation cannot change the plan's outcome), and its requirements union into the result's `R`.
- Every `AttemptStart` is followed by exactly one terminal event. Terminal events run like finalizers, so an interrupted attempt emits `AttemptFailure` with the interruption cause; keying off `Cause` means defects are reported too, even though the plan does not retry them.
- The per-attempt effect is instrumented before the retry schedule is applied, so each retry within a step produces its own event pair, and the synthetic replay used when falling over to a later step emits no events.
- `Effect.withExecutionPlan` moved from `dual(2, ...)` to `dual(isEffect(args[0]), ...)` to support the optional third argument, matching `Stream.withExecutionPlan`.
- Stream attempts are wrapped with `Stream.onStart` / `Stream.onExit` (scope finalizer), giving streams full parity including defects and interruption, which the existing error-channel `catch_` loop cannot observe.

Closes EFF-467
Closes #6699

## Validation

- `pnpm vitest run packages/effect/test/ExecutionPlan.test.ts` (16 tests, including 10 new event tests asserting full ordered event arrays)
- `pnpm test-types Effect.tst.ts` / `pnpm test-types Stream.tst.ts` (new handler-requirement propagation tests)
- `pnpm doctest --run packages/effect/src/ExecutionPlan.ts packages/effect/src/Effect.ts packages/effect/src/Stream.ts`
- `pnpm check`, `pnpm lint`
- Package-local `pnpm docgen` fails on a pre-existing, unrelated `HttpStaticServer` example error (reproduced on the base commit); the new example typechecks.


packages/effect/src/ExecutionPlan.ts
```diff
@@ -372,3 +374,127 @@ export const CurrentMetadata = Context.Reference<Metadata>("effect/ExecutionPlan
...
+ *
+ * await Effect.runPromise(program) // => "good"
+ *
+ * events // => ["AttemptStart:0", "AttemptFailure:0", "AttemptStart:1", "AttemptSuccess:1"]
+ * ```
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export interface Options<E, R> {
```
tim-smart: I think we can just inline these and remove this type

tim-smart: Done in 0d32de706 — removed `ExecutionPlan.Options` and inlined the option object into both `withExecutionPlan` signatures. The worked example moved to the `Effect.withExecutionPlan` JSDoc.

PR #7037, opened 2026-08-05
Fix failed OTLP exports consume delta checkpoints

fubhy:
## Summary

A delta snapshot advances its baselines while constructing the request body, so an HTTP rejection still consumes the counter, histogram, frequency, and summary values in that request. The next successful export reports only changes after the rejected attempt and starts its interval at the rejected attempt's timestamp.

> [!NOTE]
> This PR includes focused regression coverage and the implementation fix.

## Failed OTLP exports consume delta checkpoints

**Module:** `effect/unstable/observability/OtlpMetrics`  
**Audit ID:** `effect-f60831606fc1abf3`  

### What happens

A delta snapshot advances its baselines while constructing the request body, so an HTTP rejection still consumes the counter, histogram, frequency, and summary values in that request. The next successful export reports only changes after the rejected attempt and starts its interval at the rejected attempt's timestamp.

### Why it happens

The exporter calls `snapshot` from `setBody` before `client.execute` determines the response status. During that call, `previousCounterState`, `previousHistogramState`, `previousFrequencyState`, `previousSummaryState`, and `previousExportTimeNanos` are mutated to the current snapshot; the exporter's failure branch only sets `disabledUntil` and clears its buffer, with no rollback of those metric baselines.

### Expected behavior

For delta temporality, the previous-value maps and previous export time must identify the last successfully exported checkpoint; a failed export must leave that checkpoint unchanged so every observation is included in a later successful export.

### Relevant implementation

- [`packages/effect/src/unstable/observability/OtlpMetrics.ts:146-442`](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/unstable/observability/OtlpMetrics.ts#L146-L442)

<details>
<summary>View problematic code at <code>packages/effect/src/unstable/observability/OtlpMetrics.ts:146-195</code></summary>

```ts
          if (isDelta) {
            const previousCount = previousCounterState.get(metricKey)
            if (previousCount !== undefined) {
              if (typeof currentCount === "bigint" && typeof previousCount === "bigint") {
                reportValue = currentCount - previousCount
                // Handle reset: if current < previous, report current value
                if (reportValue < BigInt(0)) {
                  reportValue = currentCount
                }
              } else {
                const curr = Number(currentCount)
                const prev = Number(previousCount)
                reportValue = curr - prev
                // Handle reset
                if (reportValue < 0) {
                  reportValue = curr
                }
              }
            }
            previousCounterState.set(metricKey, currentCount)
          }

          const dataPoint: INumberDataPoint = {
            attributes,
            startTimeUnixNano: intervalStartTime,
            timeUnixNano: nowTime
          }
          if (typeof reportValue === "bigint") {
            dataPoint.asInt = Number(reportValue)
          } else {
            dataPoint.asDouble = reportValue
          }
          if (metricDataByName.has(state.id)) {
            metricDataByName.get(state.id)!.sum!.dataPoints.push(dataPoint)
          } else {
            addMetricData({
              name: state.id,
              description: state.description!,
              unit,
              sum: {
                aggregationTemporality: aggregationTemporalityEnum,
                isMonotonic: state.state.incremental,
                dataPoints: [dataPoint]
              }
            })
          }
          break
        }
        case "Gauge": {
          // Gauges don't have temporality - they always report current value
```

[View exact lines on GitHub](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/unstable/observability/OtlpMetrics.ts#L146-L442)

_Excerpt truncated. [Open the complete packages/effect/src/unstable/observability/OtlpMetrics.ts:146-442 range.](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/unstable/observability/OtlpMetrics.ts#L146-L442)_

</details>

### Reproduction

```sh
pnpm test --run packages/effect/test/unstable/observability/OtlpMetrics.test.ts
```

**Regression:** The focused contract assertion fails against 17f0b91a and passes with this fix.

## Implementation

The shared exporter now pairs each encoded request with an action that runs only after the filtered, retried HTTP request succeeds. Metrics build candidate delta state while serializing and commit the counter, histogram, frequency, summary, and interval checkpoints in that success action. Log and trace exporters use a no-op success action.

The regression test remains in the module's main test file and covers every affected metric type.

## Audit provenance

- Findings: `effect-f60831606fc1abf3`
- Initial patch: focused reproduction tests; implementation completed in `13a4ee2c0`

Closes EFF-513

### DIFF
--- a/packages/effect/test/unstable/observability/OtlpMetrics.test.ts
+++ b/packages/effect/test/unstable/observability/OtlpMetrics.test.ts
@@ -5,15 +5,16 @@ import { HttpClient, type HttpClientError, HttpClientResponse } from "effect/uns
 import { OtlpExporter, OtlpMetrics, OtlpSerialization } from "effect/unstable/observability"
 
 describe("OtlpMetrics", () => {
-  it.effect("retains a delta after a failed export", () =>
+  it.effect("retains delta checkpoints after a failed export", () =>
     Effect.scoped(Effect.gen(function*() {
-      const bodies = yield* Ref.make<Array<any>>([])
+      const bodies = yield* Ref.make<ReadonlyArray<OtlpExportRequest>>([])
       const attempts = yield* Ref.make(0)
       const client = HttpClient.makeWith(
         Effect.fnUntraced(function*(requestEffect) {
           const request = yield* requestEffect
           if (request.body._tag === "Uint8Array") {
-            yield* Ref.update(bodies, (all) => [...all, JSON.parse(new TextDecoder().decode(request.body.body))])
+            const body = JSON.parse(new TextDecoder().decode(request.body.body)) as OtlpExportRequest
+            yield* Ref.update(bodies, Array.append(body))
           }
           const attempt = yield* Ref.updateAndGet(attempts, (n) => n + 1)
           return HttpClientResponse.fromWeb(request, new Response(null, { status: attempt === 1 ? 400 : 200 }))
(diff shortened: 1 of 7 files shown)


packages/effect/test/unstable/observability/OtlpMetrics.test.ts
```diff
@@ -2,9 +2,77 @@ import { assert, describe, it } from "@effect/vitest"
 import { Array, Context, Effect, Layer, Metric, Predicate, Ref } from "effect"
 import { TestClock } from "effect/testing"
 import { HttpClient, type HttpClientError, HttpClientResponse } from "effect/unstable/http"
-import { OtlpMetrics, OtlpSerialization } from "effect/unstable/observability"
+import { OtlpExporter, OtlpMetrics, OtlpSerialization } from "effect/unstable/observability"
 
 describe("OtlpMetrics", () => {
+  it.effect("retains delta checkpoints after a failed export", () =>
+    Effect.scoped(Effect.gen(function*() {
```
tim-smart: Don't need the Effect.scoped here

PR #7039, opened 2026-08-05
Fix failed ResourceRef rebuilds permanently block waiters

fubhy:
## Summary

A failed resource rebuild leaves the reference in `Acquiring` with its latch closed. Every subsequent `await` blocks because no failure path restores a terminal state or opens the latch.

> [!IMPORTANT]
> This PR includes focused regression coverage and the implementation fix.

## Failed ResourceRef rebuilds permanently block waiters

**Module:** `effect/unstable/cluster/internal/resourceRef`  
**Audit ID:** `effect-bcb23fa6a8f4bb8f`  

### What happens

A failed resource rebuild leaves the reference in `Acquiring` with its latch closed. Every subsequent `await` blocks because no failure path restores a terminal state or opens the latch.

### Why it happens

`rebuildUnsafe` closes the latch, installs `{ _tag: "Acquiring", scope }`, closes the previous scope, and runs the new acquisition. Only the success `flatMap` changes state to `Acquired` and opens the latch; an acquisition failure bypasses that block, leaving both the state and latch unchanged and the candidate scope unclosed.

### Expected behavior

Every rebuild exit must leave `ResourceRef` outside an unresolvable acquisition state: success publishes the new value and wakes waiters, while failure must close the candidate scope and wake waiters with a defined failure or recoverable state.

### Relevant implementation

- [`packages/effect/src/unstable/cluster/internal/resourceRef.ts:66-98`](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/unstable/cluster/internal/resourceRef.ts#L66-L98)

<details>
<summary>View problematic code at <code>packages/effect/src/unstable/cluster/internal/resourceRef.ts:66-98</code></summary>

```ts
  rebuildUnsafe(): Effect.Effect<void, E> {
    const s = this.state.current
    if (s._tag === "Closed") {
      return Effect.interrupt
    }
    const prevScope = s.scope
    const scope = Scope.makeUnsafe()
    this.latch.closeUnsafe()
    MutableRef.set(this.state, { _tag: "Acquiring", scope })
    return Effect.withFiber((fiber) => {
      internalInterruptors.add(fiber.id)
      return Scope.close(prevScope, Exit.void)
    }).pipe(
      Effect.andThen(this.acquire(scope)),
      Effect.flatMap((value) => {
        if (this.state.current._tag === "Closed") {
          return Effect.interrupt
        }
        MutableRef.set(this.state, { _tag: "Acquired", scope, value })
        return this.latch.open
      })
    )
  }

  await: Effect.Effect<A> = Effect.suspend(() => {
    const s = this.state.current
    if (s._tag === "Closed") {
      return Effect.interrupt
    } else if (s._tag === "Acquired") {
      return Effect.succeed(s.value)
    }
    return Effect.flatMap(this.latch.await, () => this.await)
  })
```

[View exact lines on GitHub](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/unstable/cluster/internal/resourceRef.ts#L66-L98)

</details>

### Reproduction

```sh
pnpm test --run packages/effect/test/cluster/ResourceRef.test.ts
```

**Observed failure:** Focused contract assertion failed against 17f0b91a, demonstrating: Failed ResourceRef rebuilds permanently block waiters.

## Implementation

Failed rebuilds now close the candidate scope, publish a recoverable `Failed` state containing the original cause, and open the latch. Waiters observe that cause, and a later rebuild can recover successfully. The failure handler publishes state and opens the latch only while its candidate scope still owns the current `Acquiring` state, so stale rebuilds cannot overwrite a newer acquisition.

Validation:

```sh
pnpm test --run packages/effect/test/cluster/ResourceRef.test.ts
pnpm lint-fix
pnpm check
```

## Audit provenance

- Findings: `effect-bcb23fa6a8f4bb8f`
- Fix: regression coverage plus the `ResourceRef` failure-state implementation

Closes EFF-511

### DIFF
--- a/.changeset/quiet-pandas-rebuild.md
+++ b/.changeset/quiet-pandas-rebuild.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix failed `ResourceRef` rebuilds permanently blocking waiters.
--- a/packages/effect/src/unstable/cluster/internal/resourceRef.ts
+++ b/packages/effect/src/unstable/cluster/internal/resourceRef.ts
@@ -96,10 +96,11 @@ export class ResourceRef<A, E = never> {
         }
         return Scope.close(scope, exit).pipe(
           Effect.ensuring(Effect.sync(() => {
-            if (this.state.current._tag !== "Closed") {
+            const state = this.state.current
+            if (state._tag === "Acquiring" && state.scope === scope) {
               MutableRef.set(this.state, { _tag: "Failed", scope, cause: exit.cause })
+              this.latch.openUnsafe()
             }
-            this.latch.openUnsafe()
           }))
         )
       })
--- a/packages/effect/src/unstable/cluster/internal/resourceRef.ts
+++ b/packages/effect/src/unstable/cluster/internal/resourceRef.ts
@@ -1,3 +1,4 @@
+import type * as Cause from "../../../Cause.ts"
 import * as Effect from "../../../Effect.ts"
 import * as Exit from "../../../Exit.ts"
 import * as Latch from "../../../Latch.ts"
@@ -7,7 +8,7 @@ import * as Scope from "../../../Scope.ts"
 import { internalInterruptors } from "./interruptors.ts"
 
 /** @internal */
-export type State<A> = {
+export type State<A, E> = {
   readonly _tag: "Closed"
 } | {
   readonly _tag: "Acquiring"
@@ -16,6 +17,10 @@ export type State<A> = {
   readonly _tag: "Acquired"
   readonly scope: Scope.Closeable
   readonly value: A
+} | {
+  readonly _tag: "Failed"
+  readonly scope: Scope.Closeable
+  readonly cause: Cause.Cause<E>
 }
 
 /** @internal */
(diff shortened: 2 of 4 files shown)


tim-smart:
This is a real branch regression, not a one-off flaky CI failure.

On head `281a2811485ce17abb8edfe2567aae1c3ccefe85`, both the original Node shard and a job re-run failed in the same two places:

- `SqlRunnerStorage > rebuilds the reserved connection when releasing the previous one hangs` timed out at 60s in both attempts.
- `bounds shard lock operations...` observed ~80ms against a 90ms lower bound in both attempts (80.45ms / 80.18ms), so that assertion is timing-sensitive too.

The stalled-release test passed on this PR's prior head `4dd236145362483e0c1123816f6d8fdee0edb573` (run 31047717472), where the `ResourceRef` implementation was still unchanged; it began failing after the implementation commit. Other contemporary PR runs are green. The focused `ResourceRef.test.ts` passes 100/100 locally, but that test does not cover the detached/interrupted rebuild path used by `SqlRunnerStorage`.

The risky boundary is the new `rebuildUnsafe` failure finalizer: timed-out rebuilds can remain alive while a newer rebuild starts, and the old rebuild's `onExit` can later publish `Failed`/open the shared latch without checking that its scope still owns the current `Acquiring` state. I would not relax or patch around the SQL integration timeout; the implementation needs to preserve stale-rebuild ownership before this is merged.

Local execution of the container-backed SQL test was unavailable on this runner (no usable Testcontainers socket, with unrelated Podman workloads active), so the two same-SHA Actions attempts are the full integration reproduction.

PR #7061, opened 2026-08-05
Fix jSON Schema additionalProperties is applied to declared fields

fubhy:
## Summary

Importing an object with a declared string property and boolean `additionalProperties` rejects `{ a: "ok", extra: true }`. The declared property is incorrectly subjected to the additional-property schema, so a JSON Schema-valid object fails Effect validation.

> [!IMPORTANT]

## JSON Schema additionalProperties is applied to declared fields

**Module:** `effect/internal/schema/fromJsonSchemaDocument`  
**Audit ID:** `effect-8486ba58a19a6ce5`  

### What happens

Importing an object with a declared string property and boolean `additionalProperties` rejects `{ a: "ok", extra: true }`. The declared property is incorrectly subjected to the additional-property schema, so a JSON Schema-valid object fails Effect validation.

### Why it happens

The importer creates a property signature for `a` and also creates an unrestricted string index signature for `additionalProperties`. Effect's object representation applies that index signature to every string key, intersecting the boolean constraint with the declared string constraint on `a` instead of excluding declared names.

### Expected behavior

`SchemaRepresentation.fromJsonSchemaDocument` must preserve Draft 2020-12 object semantics: `additionalProperties` applies only to property names not covered by `properties` or `patternProperties`.

### Relevant implementation

- [`packages/effect/src/internal/schema/fromJsonSchemaDocument.ts:823-875`](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/internal/schema/fromJsonSchemaDocument.ts#L823-L875)

<details>
<summary>View problematic code at <code>packages/effect/src/internal/schema/fromJsonSchemaDocument.ts:823-872</code></summary>

```ts
  function collectProperties(
    schema: JsonSchema.JsonSchema,
    path: Path
  ): Array<SchemaRepresentation.PropertySignature> {
    const properties =
      typeof schema.properties === "object" && schema.properties !== null && !Array.isArray(schema.properties)
        ? schema.properties as Record<string, unknown>
        : {}
    const required = Array.isArray(schema.required)
      ? schema.required.filter((key): key is string => typeof key === "string")
      : []
    const keys = new Set([...Object.keys(properties), ...required])
    return Array.from(keys, (name) => ({
      name,
      type: recur(properties[name], [...path, "properties", name]),
      isOptional: !required.includes(name),
      isMutable: false
    }))
  }

  function collectIndexSignatures(
    schema: JsonSchema.JsonSchema,
    path: Path
  ): Array<SchemaRepresentation.IndexSignature> {
    const signatures: Array<SchemaRepresentation.IndexSignature> = []
    if (
      typeof schema.patternProperties === "object" &&
      schema.patternProperties !== null &&
      !Array.isArray(schema.patternProperties)
    ) {
      for (const [pattern, value] of Object.entries(schema.patternProperties)) {
        signatures.push({
          parameter: {
            _tag: "String",
            checks: [jsonSchemaFilter("effect/schema/isPattern", { source: pattern, flags: "" })]
          },
          type: recur(value, [...path, "patternProperties", pattern])
        })
      }
    }
    if (schema.additionalProperties === undefined || schema.additionalProperties === true) {
      signatures.push({
        parameter: string,
        type: unknown
      })
    } else if (typeof schema.additionalProperties === "object" && schema.additionalProperties !== null) {
      signatures.push({
        parameter: string,
        type: recur(schema.additionalProperties, [...path, "additionalProperties"])
      })
```

[View exact lines on GitHub](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/internal/schema/fromJsonSchemaDocument.ts#L823-L875)

_Excerpt truncated. [Open the complete packages/effect/src/internal/schema/fromJsonSchemaDocument.ts:823-875 range.](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/internal/schema/fromJsonSchemaDocument.ts#L823-L875)_

</details>

### Reproduction

```sh
pnpm test --run packages/effect/test/schema/representation/fromJsonSchemaDocument.test.ts
```

**Observed failure:** Focused contract assertion failed against 17f0b91a, demonstrating: JSON Schema additionalProperties is applied to declared fields.

## Implementation handoff

```sh
pnpm test --run packages/effect/test/schema/representation/fromJsonSchemaDocument.test.ts
```

## Audit provenance

- Findings: `effect-8486ba58a19a6ce5`

Closes EFF-490

### DIFF
--- a/.changeset/clever-wolves-exclude.md
+++ b/.changeset/clever-wolves-exclude.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix imported JSON Schemas so `additionalProperties` ignores declared and pattern-matched properties.
--- a/packages/effect/src/internal/schema/fromJsonSchemaDocument.ts
+++ b/packages/effect/src/internal/schema/fromJsonSchemaDocument.ts
@@ -33,6 +33,40 @@ const never: ImportedJsonSchemaRepresentation = { _tag: "Never", checks: [] }
 const unknown: ImportedJsonSchemaRepresentation = { _tag: "Unknown", checks: [] }
 const string: ImportedJsonSchemaRepresentation = { _tag: "String", checks: [] }
 
+const additionalPropertyKeyFilterId = "effect/schema/isAdditionalPropertyKey"
+
+interface AdditionalPropertyKeyFilterPayload extends Schema.JsonObject {
+  readonly properties: ReadonlyArray<string>
+  readonly patterns: ReadonlyArray<string>
+}
+
+function additionalPropertyKeyFilter(
+  payload: AdditionalPropertyKeyFilterPayload,
+  annotations?: Schema.Annotations.Filter
+): SchemaAST.Filter<string> {
+  const properties = new Set(payload.properties)
+  const patterns = payload.patterns.map((pattern) => new globalThis.RegExp(pattern))
+  return Schema.makeFilter(
+    (key: string) => !properties.has(key) && patterns.every((pattern) => !pattern.test(key)),
+    {
+      representation: {
+        id: additionalPropertyKeyFilterId,
+        payload
+      },
+      ...annotations
+    }
+  )
+}
+
+const additionalPropertyKeyFilterReviver: SchemaRepresentation.FilterReviver<AdditionalPropertyKeyFilterPayload> = {
+  id: additionalPropertyKeyFilterId,
+  payloadSchema: Schema.Struct({
+    properties: Schema.Array(Schema.String),
+    patterns: Schema.Array(Schema.String)
+  }),
+  revive: ({ annotations, payload }) => additionalPropertyKeyFilter(payload, annotations)
+}
+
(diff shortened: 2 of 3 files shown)


gcanti:
Closing this PR because the proposed fix handles the JSON Schema semantics only at runtime, while the resulting constraint cannot be represented faithfully as a TypeScript type:

```ts
// invalid typescript type
type T = {
  a: string
  [key: string]: boolean
}
```

Supporting this correctly would require determining whether declared properties are assignable to `additionalProperties`. A conservative rejection would also exclude valid, representable cases, while a complete solution would require broader schema assignability analysis that is outside the scope of this PR.

For now, I prefer to leave this case unsupported rather than introduce a runtime-only representation that cannot preserve the expected correspondence between `Schema` and TypeScript types.

PR #7063, opened 2026-08-05
Fix tuple.make loses element literal types

fubhy:
## Summary

`Tuple.make("a", 1, true)` is inferred as `[string, number, boolean]` rather than `["a", 1, true]`. The constructor loses exact argument information, preventing its result from satisfying APIs that require those literal tuple positions.

## Tuple.make loses element literal types

**Module:** `effect/Tuple`  
**Audit ID:** `effect-de659caf5c519c20`  
**Severity / confidence:** low / high

### What happens

`Tuple.make("a", 1, true)` is inferred as `[string, number, boolean]` rather than `["a", 1, true]`. The constructor loses exact argument information, preventing its result from satisfying APIs that require those literal tuple positions.

### Why it happens

The rest parameter uses `Elements extends ReadonlyArray<unknown>` without a `const` type parameter. TypeScript therefore widens primitive arguments while inferring `Elements`, and the identity implementation returns that widened generic tuple type.

### Expected behavior

`Tuple.make` must preserve each argument's literal type in the returned tuple, as promised by its constructor contract.

### Relevant implementation

- [`packages/effect/src/Tuple.ts:20-46`](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/Tuple.ts#L20-L46)

<details>
<summary>View problematic code at <code>packages/effect/src/Tuple.ts:20-46</code></summary>

````ts
/**
 * Creates a tuple from the provided arguments.
 *
 * **When to use**
 *
 * Use when you need a properly typed tuple without writing `[a, b, c] as const`
 * or another manual cast.
 *
 * **Details**
 *
 * The returned value has the exact tuple type, with each element's literal type
 * preserved.
 *
 * **Example** (Creating a tuple)
 *
 * ```ts import.meta.vitest
 * import { Tuple } from "effect"
 *
 * Tuple.make(10, 20, "red") // => [10, 20, "red"]
 * ```
 *
 * @see {@link get} – access a single element by index
 * @see {@link appendElement} – append an element to a tuple
 * @category constructors
 * @since 2.0.0
 */
export const make = <Elements extends ReadonlyArray<unknown>>(...elements: Elements): Elements => elements
````

[View exact lines on GitHub](https://github.com/Effect-TS/effect/blob/17f0b91a243ccfe4a38d27debdc983adf434e738/packages/effect/src/Tuple.ts#L20-L46)

</details>

### Reproduction

```sh
pnpm test-types packages/effect/typetest/Tuple.tst.ts
```

**Observed failure:** Focused contract assertion failed against 17f0b91a, demonstrating: Tuple.make loses element literal types.

## Implementation

`Tuple.make` now uses a const type parameter constrained to a readonly array, with mutable tuple spreads for the rest parameter and return type. This preserves each argument's literal type while retaining a mutable tuple result.

## Validation

```sh
pnpm test-types packages/effect/typetest/Tuple.tst.ts
pnpm test packages/effect/test/Tuple.test.ts --run
pnpm lint
pnpm check
```

## Audit provenance

- Findings: `effect-de659caf5c519c20`

Closes EFF-488

### DIFF
--- a/packages/effect/src/Tuple.ts
+++ b/packages/effect/src/Tuple.ts
@@ -43,7 +43,7 @@ import type { Apply, Lambda } from "./Struct.ts"
  * @category constructors
  * @since 2.0.0
  */
-export const make = <Elements extends ReadonlyArray<unknown>>(...elements: Elements): Elements => elements
+export const make = <const Elements extends Array<unknown>>(...elements: Elements): Elements => elements
 
 type Indices<T extends ReadonlyArray<unknown>> = Exclude<Partial<T>["length"], T["length"]>
 
--- a/packages/effect/src/Tuple.ts
+++ b/packages/effect/src/Tuple.ts
@@ -43,7 +43,8 @@ import type { Apply, Lambda } from "./Struct.ts"
  * @category constructors
  * @since 2.0.0
  */
-export const make = <const Elements extends ReadonlyArray<unknown>>(...elements: Elements): Elements => elements
+export const make = <const Elements extends ReadonlyArray<unknown>>(...elements: [...Elements]): [...Elements] =>
+  elements
 
 type Indices<T extends ReadonlyArray<unknown>> = Exclude<Partial<T>["length"], T["length"]>
 
--- a/packages/effect/src/Tuple.ts
+++ b/packages/effect/src/Tuple.ts
@@ -43,7 +43,7 @@ import type { Apply, Lambda } from "./Struct.ts"
  * @category constructors
  * @since 2.0.0
  */
-export const make = <const Elements extends Array<unknown>>(...elements: Elements): Elements => elements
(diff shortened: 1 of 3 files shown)


packages/effect/src/Tuple.ts
```diff
@@ -43,7 +43,7 @@ import type { Apply, Lambda } from "./Struct.ts"
  * @category constructors
  * @since 2.0.0
  */
-export const make = <Elements extends ReadonlyArray<unknown>>(...elements: Elements): Elements => elements
+export const make = <const Elements extends Array<unknown>>(...elements: Elements): Elements => elements
```
tim-smart:
```suggestion
export const make = <const Elements extends ReadonlyArray<unknown>>(...elements: Elements): Elements => elements
```

Issue #5988, opened 2026-01-17
RpcClientError loses socket error details due to hardcoded message and Cause.squash


schickling:
## Problem

When a socket error occurs in `RpcClient.makeProtocolSocket`, the `RpcClientError` is created with:
1. A hardcoded generic message `"Error in socket"`
2. `Cause.squash(cause)` which flattens the full cause tree

This results in error messages like:
```
Error: RpcClientError: Error in socket at ...
```

The underlying socket error details (like `SocketCloseError` with `code`, `closeReason`, or `SocketGenericError` with `reason: "Open" | "Read" | "Write" | "OpenTimeout"`) are buried in the `cause` field and not surfaced in the error message.

## Location

https://github.com/Effect-TS/effect/blob/main/packages/rpc/src/RpcClient.ts#L1010-L1014

```typescript
currentError = new RpcClientError({
  reason: "Protocol",
  message: "Error in socket",  // <-- hardcoded, loses context
  cause: Cause.squash(cause)   // <-- flattens the cause tree
})
```

## Proposed Solution

Extract socket error details and include them in the message:

```typescript
const socketError = Cause.failureOption(cause)
const errorMessage = Option.isSome(socketError)
  ? `Socket ${socketError.value.reason}${
      'code' in socketError.value ? `: code ${socketError.value.code}` : ''
    }${
      'closeReason' in socketError.value && socketError.value.closeReason 
        ? ` (${socketError.value.closeReason})` 
        : ''
    }`
  : "Error in socket"

currentError = new RpcClientError({
  reason: "Protocol",
  message: errorMessage,
  cause: Cause.squash(cause)
})
```

This would produce messages like:
- `Socket Close: code 1006`
- `Socket Open: connection refused`
- `Socket OpenTimeout`

## Additional Consideration

The `RpcClientError` class itself uses the `message` field as a schema property, which shadows `Error.message`. Consider using `note` (or similar) for the schema field and computing `get message()` from the structured fields - similar to the pattern used for other tagged errors to preserve both structured data and a human-readable message.

PR #7068, opened 2026-08-05
Report transient RPC socket open failures

### LINKED ISSUE #7030: RpcClient: retryTransientErrors swallows pre-open dial failures — registered streams hang with no signal, unbounded

4.0.0-beta.103

`RpcClient.makeProtocolSocket` with `retryTransientErrors: true`: when a **re-dial fails before the socket opens** (`SocketOpenError` — server briefly unreachable, wifi not yet up after laptop wake, etc.), the failure is swallowed and **nothing observable happens**: registered requests/streams are not failed, no event fires, no hook exists. The protocol silently re-dials underneath while every in-flight stream hangs. The window is unbounded — if dials keep failing (server actually gone), streams hang forever with zero signal.

The tap in `makeProtocolSocket` (src/unstable/rpc/RpcClient.ts, ~line 1110 at beta.103):

```ts
Effect.tapCause((cause) => {
  const error = Cause.findError(cause)
  const hasError = Result.isSuccess(error)
  if (
    options?.retryTransientErrors && hasError &&
    error.success.reason._tag === "SocketOpenError"
  ) {
    return Effect.void          // swallowed: no broadcast, no signal
  }
  currentError = new RpcClientError({ ... })
  return broadcast({ _tag: "ClientProtocolError", error: currentError })
}),
Effect.retry(options?.retryPolicy ?? defaultRetryPolicy),
```

The `ClientProtocolError` broadcast is the **only** path that fails registered entries (its handler walks `entries.keys()` writing a failure `Exit` per request). The early return skips it, so with `retryTransientErrors` enabled a pre-open dial failure is invisible to the application by construction.

Minimal repro shape (we keep a pinned law test for this exact behavior — [`socketRedialLaws.test.ts`, "re-dial law 2"](https://github.com/juspay/kolu/blob/effect/packages/surface/src/links/socketRedialLaws.test.ts)):

1. Build an `RpcClient` over a websocket protocol with `retryTransientErrors: true` and a URL whose dial fails with `SocketOpenError`.
2. Subscribe a stream RPC; let 3+ dial attempts fail.
3. Observe: the stream receives no failure, no element, and its fiber stays parked (`pollUnsafe() === undefined`) — indefinitely. (A close-after-open, by contrast, broadcasts and fails entries correctly — "law 1".)

Some observable signal for the transient-swallow path — any of:

- fail registered entries (letting the caller's retry policy own re-subscription), or
- still `broadcast` a distinct transient-classified event so protocol consumers can react, or
- an explicit hook (`onDialFailure`?) / a bound on the silent window, so an application can distinguish "reconnecting, be patient" from "hanging forever."

Today the only mitigations available to applications are side-channel: watching socket status independently and force-re-driving all streams, or per-stream first-frame deadlines. We shipped both, but they re-implement knowledge the client already has.

### What do you see instead?

Streams hang silently for as long as dials keep failing. This shipped as a production incident for us: a laptop-wake re-dial raced the network coming up, one pre-open failure was swallowed, the next dial succeeded — and the tab's long-lived subscriptions stayed parked with stale data behind a perfectly healthy socket until a hard reload. Field evidence and analysis: juspay/kolu#2101 (comment: https://github.com/juspay/kolu/pull/2101#issuecomment-5194985065).

### Additional information

Adjacent, possibly-related report in the same `tapCause` block: #5988 (`RpcClientError` detail loss when the broadcast *does* happen). So this region currently has two failure modes: when it speaks it drops detail (#5988), and in the `retryTransientErrors`+`SocketOpenError` arm it doesn't speak at all (this issue).

Happy to distill our law test into a standalone repro against `effect` only if that helps.

tim-smart:
## Summary

- add an `onTransientError` hook to socket protocol options and the socket protocol layer
- report every retried pre-open `SocketOpenError` as a structured `RpcClientError`
- fall back to a terminal `ClientProtocolError` broadcast when a finite retry policy is exhausted
- contain and log hook defects so observational code cannot stop reconnection
- cover continuing retries, finite exhaustion, and hook defects with socket RPC regression tests
- add a patch changeset

## Why

With `retryTransientErrors: true`, the socket protocol intentionally suppresses terminal `ClientProtocolError` broadcasts for pre-open failures so registered requests survive reconnection. That same branch previously emitted no observable signal, leaving applications unable to distinguish reconnection from a silently parked stream.

The hook is the smallest signal that preserves the retry contract. Reusing the existing broadcast for every attempt would fail all registered entries and effectively defeat `retryTransientErrors`. Once a finite retry policy is exhausted, however, there will be no reconnection, so the final error is stored and broadcast to fail registered entries cleanly.

Hook defects are logged and ignored. The hook is observational, and allowing its defect to kill the protocol fiber would recreate the silent-stop failure mode it exists to expose.

Issue #5988 is separable and already addressed by the current v4 shape: `RpcClientError.reason` retains the typed socket reason (including close codes, close reasons, and open causes). This change reuses that structured error for both hook and terminal paths without altering the error model.

## Validation

- `nix develop -c pnpm check`
- targeted dprint and oxlint checks
- focused `RpcClient.test.ts` suite: 6/6 passed

Closes EFF-473
Closes #7030

### DIFF
--- a/packages/effect/test/rpc/RpcClient.test.ts
+++ b/packages/effect/test/rpc/RpcClient.test.ts
@@ -84,7 +84,7 @@ describe("RpcClient", () => {
     }))
 
   it.effect("reports transient socket open errors without failing in-flight streams", () =>
-    Effect.scoped(Effect.gen(function*() {
+    Effect.gen(function*() {
       const requestSent = yield* Deferred.make<void>()
       const threeErrors = yield* Deferred.make<void>()
       const errors: Array<RpcClientError> = []
@@ -123,10 +123,10 @@ describe("RpcClient", () => {
         assert.strictEqual(error.reason._tag, "SocketOpenError")
       }
       assert.isUndefined(streamFiber.pollUnsafe())
-    })))
+    }))
 
   it.effect("fails in-flight streams when transient retries are exhausted", () =>
-    Effect.scoped(Effect.gen(function*() {
+    Effect.gen(function*() {
       const requestSent = yield* Deferred.make<void>()
       const socketError = new Socket.SocketError({
         reason: new Socket.SocketOpenError({
@@ -160,10 +160,10 @@ describe("RpcClient", () => {
 
       assert.instanceOf(error, RpcClientError)
       assert.strictEqual(error.reason._tag, "SocketOpenError")
-    })))
+    }))
 
   it.effect("continues retrying when the transient error hook defects", () =>
-    Effect.scoped(Effect.gen(function*() {
+    Effect.gen(function*() {
       const requestSent = yield* Deferred.make<void>()
       let attempts = 0
       const socketError = new Socket.SocketError({
@@ -204,5 +204,5 @@ describe("RpcClient", () => {
       assert.strictEqual(attempts, 3)
       assert.instanceOf(error, RpcClientError)
       assert.strictEqual(error.reason._tag, "SocketOpenError")
-    })))
+    }))
 })
--- a/packages/effect/test/rpc/RpcClient.test.ts
+++ b/packages/effect/test/rpc/RpcClient.test.ts
@@ -1,5 +1,6 @@
 import { assert, describe, it } from "@effect/vitest"
-import { Cause, Deferred, Effect, Layer, Schedule, Schema, Stream } from "effect"
+import { Cause, Deferred, Effect, Fiber, Layer, Schedule, Schema, Stream } from "effect"
+import { TestClock } from "effect/testing"
 import * as HttpClient from "effect/unstable/http/HttpClient"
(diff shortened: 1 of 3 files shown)


packages/effect/test/rpc/RpcClient.test.ts
```diff
@@ -80,4 +82,127 @@ describe("RpcClient", () => {
       assert.strictEqual(error.reason._tag, "RpcClientDefect")
       assert.strictEqual(error.reason.message, "HTTP response ended before RPC request completed")
     }))
+
+  it.effect("reports transient socket open errors without failing in-flight streams", () =>
+    Effect.scoped(Effect.gen(function*() {
```
tim-smart: Remove this Effect.scoped

PR #7082, opened 2026-08-06
Scope Atom runtime memoization to registries

### LINKED ISSUE #6887: Allow Atom runtime memoization to follow AtomRegistry lifetime

tim-smart:
## Summary

- scope the default `Atom.runtime` memo map to each `AtomRegistry`
- add `Atom.context` overloads for registry-scoped map atoms and explicitly shared concrete maps
- isolate `Reactivity` broadcasts between registries and release registry-owned layers on disposal
- remove `Atom.defaultMemoMap` and document the behavior change in a patch changeset

## Root cause

The module-level runtime factory used one process-wide `Layer.MemoMap`, so independent registries shared layer instances and the same `Reactivity` service. Per-request registries therefore could not independently release their layers, and mutations in one registry refreshed queries in another.

## Validation

- `pnpm vitest packages/effect/test/reactivity/Atom.test.ts --run` (107 tests)
- `pnpm check`
- `pnpm lint`

Closes EFF-523
Closes #6887

### DIFF
--- a/packages/effect/src/unstable/reactivity/Atom.ts
+++ b/packages/effect/src/unstable/reactivity/Atom.ts
@@ -802,9 +802,7 @@ export function context(options?: {
         get.subscribe(atom, (value) => get.setSelf(value))
         return get.once(atom)
       }, { initialValueTarget: atom }) as any as A
-  return isAtom(memoMap)
-    ? factory as RegistryRuntimeFactory
-    : factory as SharedRuntimeFactory
+  return factory as any
 }
 
 /**
--- a/packages/effect/src/unstable/reactivity/Atom.ts
+++ b/packages/effect/src/unstable/reactivity/Atom.ts
@@ -694,7 +694,7 @@ export interface AtomRuntime<R, ER = never> extends Atom<AsyncResult.AsyncResult
 }
 
 /**
- * Factory for `AtomRuntime` values that share a `Layer.MemoMap` and a set of global layers.
+ * Factory for `AtomRuntime` values that share a set of global layers.
  *
  * @category models
  * @since 4.0.0
@@ -705,7 +705,6 @@ export interface RuntimeFactory {
       | Layer.Layer<R, E, AtomRegistry | Reactivity.Reactivity>
       | ((get: AtomContext) => Layer.Layer<R, E, AtomRegistry | Reactivity.Reactivity>)
   ): AtomRuntime<R, E>
-  readonly memoMap: Layer.MemoMap
   readonly addGlobalLayer: <A, E>(layer: Layer.Layer<A, E, AtomRegistry | Reactivity.Reactivity>) => void
 
   /**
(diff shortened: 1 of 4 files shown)


packages/effect/src/unstable/reactivity/Atom.ts
```diff
@@ -775,24 +802,18 @@ export const context: (options: {
         get.subscribe(atom, (value) => get.setSelf(value))
         return get.once(atom)
       }, { initialValueTarget: atom }) as any as A
-  return factory
+  return isAtom(memoMap)
```
tim-smart: It's fine to use `as any` here to avoid the useless predicate check

PR #7085, opened 2026-08-06
Decouple core runtime equality protocols

tim-smart:
## Summary

- remove eager Equal / Hash protocol implementations from Cause, Exit, and Context runtime values
- replace `causeCombine`'s `Array.union` and generic `Equal.equals` paths with a small comparator that preserves payload-and-annotation-identity deduplication for Fail / Die reasons
- add focused equality / deduplication coverage and a patch changeset

## Equality semantics

Cause, Exit, and Context no longer advertise or implement the Equal / Hash protocols. Callers that explicitly import `Equal` or `Hash` can still compare or hash these values through the generic structural fallback.

For Context, generic structural equality reflects the stored representation: contexts built through the same operations compare structurally, while logically equivalent contexts built through different internal representations may not.

`Cause.combine` first preserves exact reason-wrapper identity. Distinct matching `Fail` wrappers deduplicate only when both their `.error` payloads and annotation-map references are `===`; matching `Die` wrappers use the same rule for `.defect`. This retains separately annotated reasons even when they share a payload. Remaining reason pairs may deduplicate through an explicitly implemented `[Equal.symbol]`; otherwise they remain distinct.

Types that still opt into Equal / Hash retain their existing protocol behavior.

## Bundle impact

Measured against the PR's rebased parent with:

```sh
pnpm bundle-compare-selected --base f0be8554d packages/tools/bundle/fixtures/basic.ts
```

| Fixture | Before | After | Difference |
|:--|--:|--:|--:|
| `basic.ts` | 6.92 KB | 5.11 KB | -1.81 KB (-26.15%) |

The post-change `pnpm bundle-analyze packages/tools/bundle/fixtures/basic.ts` output retains 42 rendered bytes from `Equal.js` for the `Equal.symbol` binding. `Hash.js`, `internal/equal.js`, `Array.js`, and the deep comparison implementation remain tree-shaken.

## Validation

- `pnpm lint-fix`
- `pnpm check`
- `pnpm test --run packages/effect/test/Cause.test.ts packages/effect/test/Exit.test.ts packages/effect/test/Context.test.ts` (146 tests)
- `pnpm bundle-analyze packages/tools/bundle/fixtures/basic.ts`
- `pnpm bundle-compare-selected --base f0be8554d packages/tools/bundle/fixtures/basic.ts`

Closes EFF-526

### DIFF
--- a/packages/effect/src/internal/effect.ts
+++ b/packages/effect/src/internal/effect.ts
@@ -5,14 +5,12 @@ import type * as Console from "../Console.ts"
 import * as Context from "../Context.ts"
 import * as Duration from "../Duration.ts"
 import type * as Effect from "../Effect.ts"
-import * as Equal from "../Equal.ts"
 import type * as Exit from "../Exit.ts"
 import type * as Fiber from "../Fiber.ts"
 import * as Filter from "../Filter.ts"
 import { formatJson } from "../Formatter.ts"
 import type { LazyArg } from "../Function.ts"
 import { constant, constFalse, constTrue, constUndefined, constVoid, dual, identity } from "../Function.ts"
-import * as Hash from "../Hash.ts"
 import { toJson, toStringUnknown } from "../Inspectable.ts"
 import * as Iterable from "../Iterable.ts"
 import type * as _Latch from "../Latch.ts"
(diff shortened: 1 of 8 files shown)


packages/effect/src/internal/effect.ts
```diff
@@ -250,13 +236,25 @@ export const causeCombine: {
...
+      const reason = that.reasons[i]
+      if (!reasons.some((current) => runtimeEquals(current, reason))) {
+        reasons.push(reason)
+      }
+    }
+    return reasons.length === self.reasons.length ? self : new CauseImpl(reasons)
   }
 )
 
+const EqualSymbol = "~effect/interfaces/Equal"
```
tim-smart: I think importing Equal.symbol is still fine, as the rest of the module should still tree shake.

PR #7086, opened 2026-08-06
Fix Deno filesystem error context

tim-smart:
## Summary

- preserve `writeFile` method and path context when high-level Deno writes delegate to `FileImpl`
- normalize name-only `Deno.errors.AlreadyExists` failures produced by `@std/fs` copy
- add focused filesystem and internal error-mapping regressions
- add a patch changeset for `@effect/platform-deno`

## Root cause

Non-native `writeFile` flag handling reused `FileImpl.writeAll`, which labeled failures as `writeAll` without the original path. Separately, the copy helper constructs `Deno.errors.AlreadyExists` without an errno-style `code`, while the adapter only recognized `EEXIST`.

## Validation

- `deno task test --run --project @effect/platform-deno` (205 passed, 11 skipped)
- `pnpm --filter @effect/platform-deno check`
- focused `dprint` and `oxlint` checks on the changed files

Closes EFF-527

### DIFF
--- a/packages/effect/test/FileSystem.test-utils.ts
+++ b/packages/effect/test/FileSystem.test-utils.ts
@@ -1,8 +1,9 @@
 import { assert, expect, it } from "@effect/vitest"
-import { Array } from "effect"
+import { Array, Result } from "effect"
 import * as Effect from "effect/Effect"
 import * as Fs from "effect/FileSystem"
 import type * as Layer from "effect/Layer"
+import * as PlatformError from "effect/PlatformError"
 import * as Stream from "effect/Stream"
 
 export interface TestLayerOptions {
@@ -137,8 +138,11 @@ export const testLayer = <E>(layer: Layer.Layer<Fs.FileSystem, E>, options: Test
       const fs = yield* Fs.FileSystem
       const path = yield* fs.makeTempFile()
 
-      yield* fs.writeFileString(path, "data", { flag: "r" }).pipe(Effect.flip)
+      const error = yield* fs.writeFileString(path, "data", { flag: "r" }).pipe(Effect.flip)
 
+      assert(error.reason instanceof PlatformError.SystemError)
+      assert.strictEqual(error.reason.method, "writeFile")
+      assert.strictEqual(error.reason.pathOrDescriptor, path)
       assert.strictEqual(yield* fs.readFileString(path), "")
     })))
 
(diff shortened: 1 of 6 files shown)


packages/platform-deno/test/DenoFileSystem.test.ts
```diff
@@ -1,9 +1,45 @@
...
 
-describe("FileSystem", () =>
+describe("FileSystem", () => {
   testLayer(DenoFileSystem.layer, {
     accessOnDirectory: false,
     tempFileScopedRemovesDirectory: false
-  }))
+  })
+
+  it.effect("reports delegated write errors as writeFile errors", () =>
```
tim-smart: Add these tests to the shared suite instead.

packages/platform-deno/src/DenoFileSystem.ts
```diff
@@ -452,7 +465,7 @@ const writeFile: FileSystem.FileSystem["writeFile"] = (path, data, options) => {
   }
   return Effect.acquireUseRelease(
     tryPromise("writeFile", path, () => Deno.open(path, openOptions(flag, options?.mode))),
-    (file) => new FileImpl(file, flag.startsWith("a")).writeAll(data),
+    (file) => new FileImpl(file, flag.startsWith("a")).writeAll(data, "writeFile", path),
```
tim-smart: I think just map the system error reason instead of notifying FileImpl

packages/effect/test/FileSystem.test-utils.ts
```diff
@@ -137,8 +138,11 @@ export const testLayer = <E>(layer: Layer.Layer<Fs.FileSystem, E>, options: Test
       const fs = yield* Fs.FileSystem
       const path = yield* fs.makeTempFile()
 
-      yield* fs.writeFileString(path, "data", { flag: "r" }).pipe(Effect.flip)
+      const error = yield* fs.writeFileString(path, "data", { flag: "r" }).pipe(Effect.flip)
 
+      assert(error.reason instanceof PlatformError.SystemError)
```
tim-smart: Check the tag instead of using instanceof

packages/platform-deno/src/DenoFileSystem.ts
```diff
@@ -452,7 +452,14 @@ const writeFile: FileSystem.FileSystem["writeFile"] = (path, data, options) => {
   }
   return Effect.acquireUseRelease(
     tryPromise("writeFile", path, () => Deno.open(path, openOptions(flag, options?.mode))),
-    (file) => new FileImpl(file, flag.startsWith("a")).writeAll(data),
+    (file) =>
+      new FileImpl(file, flag.startsWith("a")).writeAll(data).pipe(
+        Effect.mapError((error) =>
```
tim-smart, line 457: Don't use instanceof

PR #7097, opened 2026-08-06
Add Cron.format for cron expression conversion

tim-smart:
## Summary

- add `Cron.format` with compact range and step syntax
- omit the default zero-seconds field while retaining six fields for non-default seconds
- support `{ includeSeconds: true }` to force a six-field expression without changing default output
- document that formatting drops timezone information and the days/weekdays `and` restriction
- add focused coverage for compact arithmetic runs, unrestricted fields, and seconds handling
- add an `effect` patch changeset

## Testing

- `nix develop -c pnpm vitest run packages/effect/test/Cron.test.ts`
- `nix develop -c pnpm --filter effect check`
- `nix develop -c pnpm lint`
- `nix develop -c pnpm doctest packages/effect/src/Cron.ts --run`
- `nix develop -c pnpm changeset status --since origin/main`

Closes EFF-534

### DIFF
--- a/packages/effect/src/Cron.ts
+++ b/packages/effect/src/Cron.ts
@@ -641,7 +641,7 @@ export const parseUnsafe = (cron: string, tz?: DateTime.TimeZone | string): Cron
  * @since 4.0.0
  */
 export const format = (cron: Cron, options?: {
-  readonly includeSeconds?: boolean
+  readonly includeSeconds?: boolean | undefined
 }): string => {
   const segments = [cron.seconds, cron.minutes, cron.hours, cron.days, cron.months, cron.weekdays]
     .map(formatSegment)
--- a/packages/effect/src/Cron.ts
+++ b/packages/effect/src/Cron.ts
@@ -613,7 +613,7 @@ export const parse = (cron: string, tz?: DateTime.TimeZone | string): Result.Res
 export const parseUnsafe = (cron: string, tz?: DateTime.TimeZone | string): Cron => Result.getOrThrow(parse(cron, tz))
 
 /**
- * Formats a `Cron` instance as a six-field cron expression.
+ * Formats a `Cron` instance as a cron expression.
  *
  * **Warning**
  *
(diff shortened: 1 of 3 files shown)


packages/effect/src/Cron.ts
```diff
@@ -612,6 +612,31 @@ export const parse = (cron: string, tz?: DateTime.TimeZone | string): Result.Res
...
+ * Formatting drops the timezone information stored by the `Cron` instance.
+ *
+ * **Example** (Formatting a cron expression)
+ *
+ * ```ts import.meta.vitest
+ * import { Cron } from "effect"
+ *
+ * const cron = Cron.parseUnsafe("23 0-20/2 * * 0", "UTC")
+ *
+ * Cron.format(cron) // => "0 23 0,2,4,6,8,10,12,14,16,18,20 * * 0"
```
tim-smart: Please try to compress this more. This output should be "23 0-20/2 * * 0"

tim-smart: Addressed in aaf616c680. Cron.format now emits the compact form `23 0-20/2 * * 0`, compacts multiple arithmetic runs within a field, and retains six fields only for non-default seconds.

packages/effect/test/Cron.test.ts
```diff
@@ -218,6 +218,26 @@ describe("Cron", () => {
...
+      Cron.format(Cron.make({
+        minutes: [0, 1, 2, 10, 20, 30],
+        hours: [],
+        days: [],
+        months: [],
+        weekdays: []
+      })),
+      "0-2,10-30/10 * * * *"
+    )
+  })
```
tim-smart, line 246: Could we add some more tests to see if we cover off all edge cases

tim-smart: Expanded the focused coverage in e7a9057e61. The tests now cover non-uniform values that must remain comma-separated, default seconds omission, non-default seconds retention, unrestricted seconds, and the fully unrestricted six-field expression. The existing tests continue to cover compact stepped and multi-run formatting.

packages/effect/src/Cron.ts
```diff
@@ -612,6 +612,57 @@ export const parse = (cron: string, tz?: DateTime.TimeZone | string): Result.Res
...
+const formatSegment = (values: ReadonlySet<number>): string => {
+  if (values.size === 0) {
+    return "*"
+  }
+  const array = Array.from(values)
+  const segments: Array<string> = []
+  let index = 0
+  while (index < array.length) {
+    const start = array[index]!
+    const step = array[index + 1]! - start
```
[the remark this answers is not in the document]
tim-smart: Reviewed and intentionally left unchanged. On the final element the read produces `undefined` and `step` becomes `NaN`, but `index + 2 < array.length` immediately short-circuits and the value is discarded; it cannot affect output or cause an exception. There is no correctness or runtime-safety issue warranting an implementation change.

packages/effect/src/Cron.ts
```diff
@@ -612,6 +612,57 @@ export const parse = (cron: string, tz?: DateTime.TimeZone | string): Result.Res
  */
 export const parseUnsafe = (cron: string, tz?: DateTime.TimeZone | string): Cron => Result.getOrThrow(parse(cron, tz))
 
+/**
+ * Formats a `Cron` instance as a cron expression.
+ *
+ * **Warning**
```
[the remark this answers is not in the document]
tim-smart: Addressed in 46d0b78597: the caveat now uses the repository-standard `**Gotchas**` heading and documents that formatting drops both timezone information and the days/weekdays `and` restriction, so reparsing is not guaranteed to preserve schedule equivalence.

packages/effect/src/Cron.ts
```diff
@@ -612,6 +612,59 @@ export const parse = (cron: string, tz?: DateTime.TimeZone | string): Result.Res
...
+ *
+ * const cron = Cron.parseUnsafe("23 0-20/2 * * 0", "UTC")
+ *
+ * Cron.format(cron) // => "23 0-20/2 * * 0"
+ * ```
+ *
+ * @category getters
+ * @since 4.0.0
+ */
+export const format = (cron: Cron): string => {
```
tim-smart: I think we should add an option that allows including seconds.

tim-smart: Addressed in 3e3cee9401. `Cron.format` now accepts `{ includeSeconds: true }` to force a six-field expression, while the default behavior remains unchanged: the default `0` seconds field is omitted, and non-default or unrestricted seconds are still preserved. Added focused coverage for the opt-in path and updated the JSDoc example/details and patch changeset. Cron tests, doctests, the `effect` typecheck, lint, and changeset validation pass.

packages/effect/src/Cron.ts
```diff
@@ -612,6 +612,69 @@ export const parse = (cron: string, tz?: DateTime.TimeZone | string): Result.Res
...
+ *
+ * Cron.format(cron) // => "23 0-20/2 * * 0"
+ * Cron.format(cron, { includeSeconds: true }) // => "0 23 0-20/2 * * 0"
+ * ```
+ *
+ * @category getters
+ * @since 4.0.0
+ */
+export const format = (cron: Cron, options?: {
+  readonly includeSeconds?: boolean
```
tim-smart:
```suggestion
  readonly includeSeconds?: boolean | undefined
```

PR #7139, opened 2026-08-07
Fix PgClient pg_cancel_backend string interpolation

fubhy:
## Summary

`PgClient.pg_cancel_backend` uses string interpolation instead of a parameterized query at `packages/sql/pg/src/PgClient.ts:771`. While the `processId` is a numeric value from database internals (not directly user-controlled), the pattern is unsafe and should use parameterized queries.

## Module

`packages/sql/pg/src/PgClient.ts:771`

### What happens

```
pool.query(\`SELECT pg_cancel_backend(${processId})\`, () => { ... })
```

The `processId` comes from `(client as any).processID`.

### Expected behavior

```
pool.query("SELECT pg_cancel_backend($1)", [processId], () => { ... })
```

### Reproduction

Code inspection at `packages/sql/pg/src/PgClient.ts:771`.

## Implementation handoff

1. Replace string interpolation with parameterized query using `$1`
2. Run `pnpm lint-fix` and targeted tests

### DIFF
--- a/.changeset/four-pens-wave.md
+++ b/.changeset/four-pens-wave.md
@@ -1,0 +1,5 @@
+---
+"@effect/sql-pg": patch
+---
+
+Fix pg_cancel_backend to use parameterized query instead of string interpolation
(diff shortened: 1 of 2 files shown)


fubhy:
Closing this audit PR because the injection claim is not reproducible. processID is a numeric value owned by the pg driver, not caller-controlled input; string interpolation here is not an SQL injection trigger. The submitted test contained no executable assertion.

PR #7145, opened 2026-08-07
Fix CORS single-origin mode returns configured origin without validating request Origin

fubhy:
## Summary

The CORS middleware single-origin branch (`allowedOrigins` with exactly 1 element) returns the configured origin without checking whether the request `Origin` header matches. Any cross-origin request to an endpoint configured with a single allowed origin receives a valid `Access-Control-Allow-Origin` response.

## Module

`packages/effect/src/unstable/http/HttpMiddleware.ts:329-332`

### What happens

The `allowOrigin` function uses a ternary chain:
- `length > 1`: validates via `isAllowedOrigin` call (correct)
- `length === 0`: returns `*` (permissive but expected)
- `length === 1`: returns `constant(opts.allowedOrigins[0])` ignoring request Origin

### Expected behavior

Single-origin mode should call `isAllowedOrigin` to validate the request Origin before returning the ACAO header.

### Reproduction

```sh
pnpm --filter effect test --run test/unstable/http/HttpMiddleware.test.ts -t "rejects unlisted origins in single-origin mode"
```

### Implementation handoff

1. Replace `constant(opts.allowedOrigins[0])` with a call to `isAllowedOrigin`
2. Run `pnpm lint-fix` and confirm the reproduction test passes

### DIFF
--- a/.changeset/fast-ligers-fetch.md
+++ b/.changeset/fast-ligers-fetch.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+CORS: validate request Origin before reflecting in single-origin mode
--- a/packages/effect/src/unstable/http/HttpMiddleware.ts
+++ b/packages/effect/src/unstable/http/HttpMiddleware.ts
@@ -326,9 +326,12 @@ export const cors = (options?: {
     ? constant({
       "access-control-allow-origin": "*"
     })
-    : constant({
-      "access-control-allow-origin": opts.allowedOrigins[0],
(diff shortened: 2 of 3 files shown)


fubhy:
Closing this audit PR because the CORS claim is incorrect. With one configured origin, returning that fixed ACAO value on all responses does not authorize a different requesting origin: browsers compare ACAO with the request Origin and reject the mismatch. The submitted test asserted a stricter header-omission policy, not a CORS security contract.

PR #7147, opened 2026-08-07
Release RPC worker pool entries after worker failure

tim-smart:
## Summary

- remove in-flight entries owned by a worker whose receive loop fails
- open each affected request latch so its scoped pool checkout is released
- preserve the existing worker retry/respawn behavior and leave entries on healthy workers untouched
- add a concurrency-1 regression test that proves a second request can acquire the pool after a worker failure

## Root cause

Worker responses and explicit interrupts removed entries and opened their latches, but the worker-run failure path only broadcast `ClientProtocolError`. At the protocol boundary, an in-flight request could therefore retain its closed latch and scoped pool checkout indefinitely after an externally terminated worker.

## Validation

- `pnpm vitest run packages/effect/test/rpc` (35 tests)
- `pnpm check`
- `pnpm lint`

Closes EFF-567

### DIFF
--- a/packages/effect/test/rpc/RpcClient.test.ts
+++ b/packages/effect/test/rpc/RpcClient.test.ts
@@ -6,6 +6,8 @@ import * as HttpClientResponse from "effect/unstable/http/HttpClientResponse"
 import { Rpc, RpcClient, RpcGroup, RpcMessage, RpcSchema, RpcSerialization } from "effect/unstable/rpc"
 import { RpcClientError } from "effect/unstable/rpc/RpcClientError"
 import * as Socket from "effect/unstable/socket/Socket"
+import * as Worker from "effect/unstable/workers/Worker"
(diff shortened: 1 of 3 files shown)


packages/effect/test/rpc/RpcClient.test.ts
```diff
@@ -60,6 +62,78 @@ const assertEmptyResponseFailsRequest = (
...
+      const second = yield* protocol.send(0, request(2)).pipe(Effect.forkChild)
+      const secondSent = yield* Deferred.await(secondRequestSent).pipe(
+        Effect.timeout("1 second"),
+        Effect.forkChild
+      )
+      yield* TestClock.adjust("1 second")
+      yield* Fiber.join(secondSent)
+      yield* Fiber.interrupt(second)
+      assert.deepStrictEqual(sentRequestIds, [1, 2])
+    }).pipe(Effect.scoped))
```
tim-smart: Doesn't need Effect.scoped here

PR #7155, opened 2026-08-09
Fix multipart file parts hanging when parser limits are exceeded mid-file

### LINKED ISSUE #6284: @effect/platform-bun multipart.toPersisted hangs when maxTotalSize is exceeded mid-file
## Bug

`Multipart.toPersisted` (and `Multipart.persisted` in `@effect/platform-bun`) hangs indefinitely when the multipart parser's `maxTotalSize` limit is exceeded mid-file during upload.

The request times out at 60s because the handler never completes — no response is ever sent.

## Root Cause

This is caused by a bug upstream in **multipasta** (the multipart parsing library): the `write()` function returns early without forwarding the chunk to the boundary scanner when `maxTotalSize` is exceeded.

- multipasta bug: https://github.com/tim-smart/multipasta/issues/38

The chain:

1. **multipasta's `write()`** (`src/internal/multipart.ts`) checks `maxTotalSize` **before** `split.write()`. When exceeded, it calls `onError` and returns — the boundary scanner never sees the chunk.
2. **multipasta's web wrapper** (`src/web.ts`): the `onFile` callback sets `finished` only when `onChunk(null)` is called. This never happens because the scanner never processes the boundary ending the current file part.
3. **`@effect/platform-bun`'s `persisted`** (`dist/esm/internal/multipart.js:18-44`): calls `reader.readMany()` in a loop waiting for the file's readable stream to close. It never does → **hang**.
4. **`@effect/platform`'s `toPersisted`**: uses `Stream.runForEach` which awaits the writer callback for each part sequentially. So the outer stream's `MultipartError` is never observed — the handler never completes, no response is sent.

## Key Distinction

| Limit | Effect |
|---|---|
| `maxPartSize` (per-file) | Error fires, but file readable **closes normally** — chunk was already forwarded to scanner |
| `maxTotalSize` (whole-request) | Error fires, but file readable **hangs** — scanner never sees trailing boundary |

## Upstream Fix

The fix is in multipasta: https://github.com/tim-smart/multipasta/issues/38 — the `write()` method should still forward the chunk to `split.write()` after calling `onError`, so the boundary scanner can close the current file part's readable.

Once multipasta is fixed, `@effect/platform-bun` will need to bump the multipasta dependency.

tarikermis:
## What & why

`Multipart.toPersisted` (and `Multipart.persisted` in `@effect/platform-bun`) hangs indefinitely when a multipart parser limit such as `maxTotalSize` is exceeded **mid-file** during an upload. The handler never completes and the request eventually times out without any response.

Root cause: when the parser reports a limit violation, it drops the offending chunk without forwarding it to the boundary scanner, so the in-progress file part's terminating boundary is never seen and its content stream never closes. The file part's pull loop in `Multipart.makeChannel` only terminated on that boundary (`finished`), so it kept pumping forever (a busy loop after upstream EOF) without ever observing the parser error. This is the same failure mode that was fixed upstream in multipasta (tim-smart/multipasta#40); the fix there landed in the web-stream wrapper, whose equivalent here is `makeChannel`.

Note: on `main` the external multipasta dependency is gone — the parser is vendored in `effect/unstable/http/MultipartParser` — so instead of a dependency bump this fixes the vendored wrapper. The issue is labeled `3.0`; happy to backport or adjust if a 3.x-line fix (multipasta bump) is wanted separately.

## Changes

- `Multipart.makeChannel`: propagate parser errors (and upstream stream failures) to in-progress file part streams via a `fileExit` exit, checked after buffered chunks drain. A part that **completed** before the parser errored still ends cleanly (`finished` wins), and only the in-progress part fails.
- `toPersisted`'s default file writer: only wrap `PlatformError` as `InternalError`; parser `MultipartError`s now keep their reason (e.g. `BodyTooLarge` → 413 instead of a masked `InternalError` → 500). Same for `File.contentEffect`, which previously wrapped every error as `InternalError`.
- Regression tests (3): `maxTotalSize` mid-file, `maxFileSize` mid-file, and a two-file case asserting a completed first file persists fully before the second file fails.
- Changeset: `effect` patch.

## Reproduction & verification

- New tests **hang** on current `main` (an event-loop-starving busy loop — verified by stashing the fix; no in-process timeout can preempt it) and **pass** with the fix.
- End-to-end under real Bun (1.3.14): `BunMultipart.persisted(request)` with `maxTotalSize: 1024` and a chunked 8 KB upload — **hangs pre-fix** (killed after 15 s), **fails fast with `MultipartError: BodyTooLarge` post-fix**.
- Checks run: `pnpm test --run test/unstable/http/Multipart.test.ts` (13/13), `HttpServerRequest.test.ts` + `HttpEffect.test.ts` (40/40), `@effect/platform-node` `MultipartParser.test.ts` (84/84), `bun node_modules/vitest/vitest.mjs run --project @effect/platform-bun` (13/13), `pnpm check` (tsc -b, clean), `pnpm lint-fix` (0 errors).

## Review & limitations

- Diff reviewed with kiro-cli (claude-opus-5); two rounds of findings addressed (default writeFile path now tested, upstream-halt passes the `Cause` through instead of squashing, multi-file ordering test added). Final pass: LGTM.
- The `@effect/platform-bun` test project has no multipart coverage; Bun verification was done with an ad-hoc script (deleted after use) rather than a committed test.
- No explicit timeout guard in the regression tests: the pre-fix failure mode starves the event loop, so no in-process timer (including vitest's own `testTimeout`) can fire — noted in a comment above the tests.

Closes #6284

### DIFF
--- a/.changeset/fix-multipart-total-size-mid-file.md
+++ b/.changeset/fix-multipart-total-size-mid-file.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fail in-progress multipart file part streams when a parser limit (such as `maxTotalSize`) is exceeded mid-file, instead of hanging indefinitely, and stop masking those parser errors as `InternalError` in `Multipart.toPersisted`.
--- a/packages/effect/src/unstable/http/Multipart.ts
+++ b/packages/effect/src/unstable/http/Multipart.ts
@@ -453,6 +453,7 @@ export const makeChannel = <IE>(headers: Record<string, string>): Channel.Channe
     Effect.map(makeConfig(headers), (config) => {
       let partsBuffer: Array<Part> = []
       let exit = Option.none<Exit.Exit<never, IE | MultipartError | Cause.Done>>()
+      let fileExit = Option.none<Exit.Exit<never, MultipartError>>()
 
       const parser = MP.make({
         ...config,
@@ -463,14 +464,19 @@ export const makeChannel = <IE>(headers: Record<string, string>): Channel.Channe
           let chunks: Array<Uint8Array> = []
           let finished = false
           const pullChunks = Channel.fromPull(
-            Effect.succeed(Effect.suspend(function loop(): Pull.Pull<Arr.NonEmptyReadonlyArray<Uint8Array>> {
-              if (!Arr.isReadonlyArrayNonEmpty(chunks)) {
-                return finished ? Cause.done() : Effect.flatMap(pump, loop)
-              }
-              const chunk = chunks
-              chunks = []
-              return Effect.succeed(chunk)
-            }))
+            Effect.succeed(
+              Effect.suspend(function loop(): Pull.Pull<Arr.NonEmptyReadonlyArray<Uint8Array>, MultipartError> {
+                if (!Arr.isReadonlyArrayNonEmpty(chunks)) {
+                  // a part completed before the parser errored still ends cleanly
+                  if (finished) return Cause.done()
+                  if (Option.isSome(fileExit)) return fileExit.value
+                  return Effect.flatMap(pump, loop)
+                }
+                const chunk = chunks
+                chunks = []
+                return Effect.succeed(chunk)
+              })
+            )
           )
           partsBuffer.push(new FileImpl(info, pullChunks))
           return function(chunk) {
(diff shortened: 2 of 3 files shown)


tim-smart:
Addressed in another PR :)

PR #7158, opened 2026-08-09
Improve Union candidate selection for nested unions and contradicted sentinels

k3dom:
> [!NOTE]
> This PR was authored autonomously by Claude Fable 5, manually reviewed only, and is currently in use as a local patch.

Two refinements to the union candidate index. Both are pure narrowings — a member is only dropped when it provably cannot parse the input — so successful parses are unchanged in both `anyOf` and `oneOf` modes. The goal is the one `getCandidates` already states: fewer members tried, fewer issues reported.

1. **`collectSentinels` learns a `Union` case**: the sentinels of a union are the sentinels common to every member. A union of structs that all fix `kind: "a"` (variants discriminated by a second key) is itself dispatchable by `kind` when it stands in a host union. Members are projected through `toCandidate` first, so a suspended member contributes no sentinels and the intersection collapses to empty — laziness is preserved exactly as #6704 established.

2. **`getCandidates` excludes candidates the input contradicts**: a member fixing `variant: "x"` cannot match input carrying `variant: "y"`, even when another sentinel key (a tag shared between sibling variants) selected it. Previously matches were unioned across sentinel keys, so one wrong discriminant re-ran every sibling and reported issues for members the input plainly did not mean. This makes the multi-key case consistent with the single-key case, where `getCandidates({ _tag: "c" }, ...)` already returns no candidates. A missing sentinel key still does not exclude — the member keeps owing its `MissingKey` error.

Motivation: a CMS blocks union where each block is a union of designs sharing a `blockType` literal and differing on `variant`. One wrong `variant` used to produce issues from every sibling design and every non-discriminated block; with both changes it produces a single issue at the failing block's path.

### DIFF
--- a/.changeset/nested-union-sentinels.md
+++ b/.changeset/nested-union-sentinels.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Improve Union candidate selection: a nested union member is dispatched by the sentinels common to all its members, and candidates whose sentinel the input contradicts are excluded.
--- a/packages/effect/runtimeperf/config.json
+++ b/packages/effect/runtimeperf/config.json
@@ -160,6 +160,34 @@
               ],
               "path": "invalid",
               "size": 100
+            },
+            {
+              "name": "multi-sentinel-100-valid-first-effect",
+              "export": "effectMultiSentinel100ValidFirst",
+              "scenario": "multi-sentinel-100-valid-first",
+              "implementation": "effect",
+              "family": "union",
+              "astTags": [
+                "Union",
+                "Objects",
+                "Literal"
+              ],
+              "path": "valid",
+              "size": 100
+            },
+            {
+              "name": "multi-sentinel-100-invalid-variant-effect",
+              "export": "effectMultiSentinel100InvalidVariant",
+              "scenario": "multi-sentinel-100-invalid-variant",
+              "implementation": "effect",
+              "family": "union",
+              "astTags": [
+                "Union",
+                "Objects",
+                "Literal"
+              ],
+              "path": "invalid",
+              "size": 100
             }
           ]
         },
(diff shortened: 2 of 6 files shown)


gcanti:
I amended the PR while preserving its two core ideas:

1. nested unions expose sentinels shared by all their members;
2. candidates contradicted by the input are removed before parsing.

The original implementation proved the behavior, but its exclusion algorithm scanned every literal bucket for every sentinel key. A 100-member multi-sentinel benchmark showed the trade-off clearly: the invalid path became much faster, while the valid path regressed by approximately 81% compared with `main`.

The candidate index now records, for each sentinel key:

- a direct `Map` from literal values to candidate sets;
- the set of all candidates that declare that key.

When a sentinel key is shared by every discriminated candidate, the key with the largest number of distinct values is used for direct dispatch. Its matching bucket becomes the initial candidate set, and only those candidates are checked against the other observed sentinels. A missing match can reject all discriminated candidates immediately.

When no common sentinel can seed the selection, direct lookups collect the possible candidates. Contradictions are then checked only against the selected candidates, using the per-key ownership sets. This preserves candidates that do not declare every indexed sentinel key without scanning every literal bucket.

The existing direct path for single-sentinel unions is preserved.

This also fixes the distinction between a missing property and an own property whose value is `undefined`:

- during decoding, an own `undefined` value contradicts a literal sentinel;
- in constructor mode, it retains the existing missing-property behavior.

I added runtime-performance fixtures for valid and invalid 100-member multi-sentinel unions. Representative local results were:

Final local results against `main` were:

| Scenario | `main` | Amended PR | Difference |
| --- | ---: | ---: | ---: |
| Valid first member | 2.66 µs | 326.5 ns | -87.79% |
| Invalid variant | 131.22 µs | 130.5 ns | -99.90% |

Lower is better. The complete Union runtimeperf family reported no regressions.

The tests now cover:

- nested-union sentinel propagation;
- contradictions across multiple sentinel keys;
- candidates with incomplete sentinel-key coverage;
- explicit `undefined` versus a missing property;
- constructor-mode semantics;
- end-to-end `oneOf` decoding and narrowed diagnostics.

I also removed one assertion that duplicated existing unknown-discriminant coverage.

PR #7170, opened 2026-08-10
Fix use of Schema.NonEmptyArrayEnsure with strings

thewilkybarkid:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Schema.NonEmptyArrayEnsure` doesn't work with strings. The compiler doesn't recognise that the value being passed to `Array.isNonEmptyReadonlyArray` might not be an array, and the function only checks the `length` property which also exists on strings.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/slimy-swans-repeat.md
+++ b/.changeset/slimy-swans-repeat.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix use of Schema.NonEmptyArrayEnsure with strings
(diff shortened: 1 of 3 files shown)


gcanti (changes requested):
`Array.ensure` changes the existing behavior when the value schema itself accepts arrays. For example, an empty array is a valid single value here:

```ts
const schema = S.NonEmptyArrayEnsure(S.Array(S.String))
```

On `v3`, decoding `[]` produces `[[]]`. With `Array.ensure`, it remains `[]` and is then rejected by the target `NonEmptyArray` schema.

Could we preserve that behavior while also checking that non-empty values are actual arrays?

```ts
decode: (i) =>
  (Array.isArray(i) && array_.isNonEmptyReadonlyArray(i) ? i : array_.of(i)),
```

I would also add this regression test:

```ts
it("decode array value", async () => {
  const schema = S.NonEmptyArrayEnsure(S.Array(S.String))
  await Util.assertions.decoding.succeed(schema, [], [[]])
})
```

This should cover both the string fix and the previous empty-array behavior.

gcanti (approved):

PR #7174, opened 2026-08-10
Fix Queue.await Done completion

### LINKED ISSUE #7171: 🐛 [Queue] await fails with Cause.Done

4.0.0-beta.107

Basically you need to start awaiting a queue before it is done.

```ts
import type * as Cause from "effect/Cause";
import * as Effect from "effect/Effect";
import * as Fiber from "effect/Fiber";
import * as Queue from "effect/Queue";

const main = Effect.gen(function* () {
  const q = yield* Queue.make<never, Cause.Done>();
  const fork = yield* Queue.await(q).pipe(Effect.forkScoped);

  // ensures await is registered before queue ends
  yield* Effect.yieldNow;
  yield* Queue.end(q);

  yield* Fiber.join(fork);
});

// fails with Done, even though Queue.await is supposed to catch those
main.pipe(Effect.scoped, Effect.runPromise);
```

It should succeed with void

### What do you see instead?

It fails with Done error

### Additional information

_No response_

tim-smart:
## Summary

- normalize `Cause.Done` to successful `void` completion for awaiters registered before queue termination
- add a regression test covering `Queue.await` started before `Queue.end`
- add a patch changeset for `effect`

## Root cause

`Queue.await` normalized `Cause.Done` only when the queue was already in the `Done` state. Awaiters registered while the queue was open were resumed by `finalize` with the raw terminal exit instead.

## Validation

- `pnpm lint-fix`
- `pnpm test --run packages/effect/test/Queue.test.ts` (23 tests passed)
- `pnpm check`

Closes EFF-585
Closes #7171

### DIFF
--- a/packages/effect/src/Queue.ts
+++ b/packages/effect/src/Queue.ts
@@ -1993,7 +1993,7 @@ const finalize = <A, E>(self: Enqueue<A, E> | Dequeue<A, E>, exit: Failure<never
   }
   openState.takers.clear()
   for (const awaiter of openState.awaiters) {
-    awaiter(exit)
+    awaiter(Pull.isDoneCause(exit.cause) ? internalEffect.exitVoid : exit)
   }
   openState.awaiters.clear()
 }
--- a/packages/effect/src/Queue.ts
+++ b/packages/effect/src/Queue.ts
@@ -1570,16 +1570,14 @@ export const takeUnsafe = <A, E>(self: Dequeue<A, E>): Exit<A, E> | undefined =>
 
 const await_ = <A, E>(self: Dequeue<A, E>): Effect<void, Exclude<E, Done>> =>
   internalEffect.callback<void, Exclude<E, Done>>((resume) => {
+    const awaiter = (effect: Effect<void, E>) => resume(Pull.catchDone(effect, () => internalEffect.exitVoid))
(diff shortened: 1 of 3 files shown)


packages/effect/src/Queue.ts
```diff
@@ -1993,7 +1993,7 @@ const finalize = <A, E>(self: Enqueue<A, E> | Dequeue<A, E>, exit: Failure<never
   }
   openState.takers.clear()
   for (const awaiter of openState.awaiters) {
-    awaiter(exit)
+    awaiter(Pull.isDoneCause(exit.cause) ? internalEffect.exitVoid : exit)
```
tim-smart: It shouldn't remaps errors for everything. Can we add a regression test to ensure this is the case.

tim-smart: Revised in `33c69569c`: `finalize` now forwards the raw terminal exit unchanged, and `Queue.await` alone applies `Pull.catchDone` at its API boundary. The targeted regression registers both `Queue.take` and `Queue.await` before the same `Queue.end`, then proves `take` retains the `Cause.Done` failure while `await` succeeds with `void`. Validation: `pnpm lint-fix`, all 23 Queue tests, and `pnpm check` pass.

PR #7200, opened 2026-08-12
Support refinements in Schedule.while

mikearnaldi:
## Summary
- add type guard overloads to `Schedule.while`
- narrow schedule input and output types for pipeable and data-first usage
- add type-level coverage and a patch changeset

## Testing
- `pnpm test-types packages/effect/typetest/Schedule.tst.ts`
- `pnpm check`
- `pnpm exec dprint check packages/effect/src/Schedule.ts packages/effect/typetest/Schedule.tst.ts .changeset/narrow-schedule-while.md`
- `NODE_OPTIONS=--experimental-strip-types pnpm exec oxlint packages/effect/src/Schedule.ts packages/effect/typetest/Schedule.tst.ts`

### DIFF
--- a/.changeset/narrow-schedule-while.md
+++ b/.changeset/narrow-schedule-while.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Support narrowing schedule input and output types with type guard predicates passed to `Schedule.while`.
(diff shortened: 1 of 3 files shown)


tim-smart:
Maybe we add back .whileInput / .whileOutput as this feels a bit tricky to use.

mikearnaldi:
> Maybe we add back .whileInput / .whileOutput as this feels a bit tricky to use.

Yes, or do something like `while({ input, output, metadata })`

PR #7203, opened 2026-08-12
fix(schema): keep annotations across encodings

### LINKED ISSUE #7192: Schema.toJsonSchemaDocument drops annotations on Schema.Number
> **Draft.** Filing now to get the observation on record — a minimal, self-contained reproduction repo is coming and I'll follow up on this issue with it. Everything below was observed directly against `effect@4.0.0-beta.107`, but I haven't yet reduced it past the snippet here.

## Summary

`Schema.toJsonSchemaDocument` drops `title` / `description` / `examples` annotations on `Schema.Number`. The annotations are present on the AST; they just don't survive the lowering. Every other shape I tried keeps them somewhere.

## Reproduction

```ts

const j = (s: any) => JSON.stringify(Schema.toJsonSchemaDocument(s).schema)

console.log("Number :", j(Schema.Number.annotate({ description: "d" })))
console.log("String :", j(Schema.String.annotate({ description: "d" })))
console.log("Finite :", j(Schema.Finite.annotate({ description: "d" })))
console.log("Int    :", j(Schema.Int.annotate({ description: "d" })))
console.log("NullOr :", j(Schema.NullOr(Schema.String).annotate({ description: "d" })))
```

Output on `4.0.0-beta.107`:

```text
Number : {"anyOf":[{"type":"number"},{"type":"string","enum":["Infinity","-Infinity","NaN"]}]}
String : {"type":"string","description":"d"}
Finite : {"type":"number","allOf":[{"description":"d"}]}
Int    : {"type":"integer","allOf":[{"description":"d"}]}
NullOr : {"anyOf":[{"type":"string"},{"type":"null"}],"description":"d"}
```

`Number` is the only one where the annotation disappears entirely.

## Notes

The annotation is definitely on the AST — it's lost during lowering, not at annotation time:

```ts
Schema.Number.annotate({ description: "d" }).ast.annotations
// => { description: "d" }
```

A few observations that may help narrow it:

- `NullOr` shows that an `anyOf` result is not inherently the problem — it keeps `description` as a sibling of `anyOf`.
- `Finite` and `Int` (the checked numerics) keep the annotation, but nest it under `allOf` rather than putting it on the node.
- No authoring form I tried recovers it for `Number`: `.annotate()` chained twice, `.pipe(s => s.annotate(...))`, as a `Schema.Struct` property, and under `Schema.optionalKey` all produce the same annotation-free `anyOf`.
- `ToJsonSchemaOptions` has no knob that affects it (`generateDescriptions` and `includeAnnotationKey` are unrelated here), and the result is identical whichever `target` I pass.

The 4-branch `anyOf` emitted by the `Number` case in `internal/schema/toJsonSchemaDocument.ts` also arrives at the caller compacted to 2 branches, so the value being annotated may not be the representation the annotations are attached to. I have not confirmed that — it's a guess, and it's the part the minimal reproduction is meant to settle.

## Impact

Consumers generating JSON Schema for numeric fields lose their documentation silently — no error, no warning, just a missing `description`. In our case it stripped descriptions from 29 numeric fields across an MCP tool surface, which we only caught because one test asserted on it.

`Schema.Finite` is a workable substitute for anyone hitting this (it keeps the annotation and lowers to a plain `number`), provided you also lift the `allOf` nesting.

## Environment

- `effect@4.0.0-beta.107`
- Node 26.6.0, macOS arm64

spencerbeggs:
Fixes #7192.

This is a fix I put together for a problem I ran into; the analysis below is my reading of the code, not something a maintainer has signed off on. I have tried to make the premises, the judgement calls and the gaps easy to check separately, so any one of them can be rejected without discarding the rest. Happy to take this in a different direction.

## What I am proposing

`Schema.toJsonSchemaDocument` keeps `title`, `description`, `examples` and the other JSON Schema annotations on schemas whose encoded form differs from their type. `Schema.Number` was the reported case, and it ends up behaving like `Schema.String` and `Schema.NullOr` already do:

```ts
Schema.toJsonSchemaDocument(Schema.Number.annotate({ description: "d" })).schema
// before: { anyOf: [{ type: "number" }, { type: "string", enum: ["Infinity", "-Infinity", "NaN"] }] }
// after:  { anyOf: [...], description: "d" }
```

## Root cause, as I read it

The annotation is not lost in the JSON Schema compiler — it never reaches it.

`internal/schema/toRepresentation.ts` builds every representation from `SchemaAST.getLastEncoding(input)`, which walks to the final link of the encoding chain and then reads annotations off that node alone. `Schema.Number.toCodecJson()` calls `replaceEncoding(this, [numberToJson(this.checks)])`, so the annotated `Number` node is the *type* side, and the encoded side is a fresh `Union([finite, nonFiniteLiterals])` carrying no annotations. Everything on the type side is dropped before the JSON Schema pass runs.

That also accounts for the other observations in the issue. `String` and `NullOr` have no encoding, so `input === encoded` and nothing is lost. `Finite` and `Int` short-circuit `toCodecJson()` back to `this`, so they are in the same position and keep the annotation. And the 4-branch `anyOf` in `toJsonSchemaDocument.ts` is unrelated: the 2-branch shape the caller sees comes from the encoded `Union`, not from that code path.

## Scope

`Schema.Number` is not the only schema affected — as far as I can tell this applies to every schema with an encoding link. `Schema.BigInt`, `Schema.Date`, `Schema.URL`, `Schema.Option`, `Schema.ReadonlyMap`, `Schema.Unknown`, `Schema.Void`, `Schema.Undefined`, `Schema.ObjectKeyword` and `bigint` literals all dropped their annotations the same way, and six existing tests asserted that behavior. Those are updated here, and the fix covers all of them rather than special-casing `Number`.

If you would rather scope this to `Number` only, say so — that is a smaller change and I am not attached to the broad version.

## The fix, and the two judgement calls in it

`toRepresentation` carries the type-side annotations forward across the encoding chain. Two decisions in there are mine, and either can be reversed cheaply:

**1. Only `jsonSchemaAnnotationKeys` travel.** `representation`, `expected`, `identifier`, the `to*` hooks and everything else stay bound to the node that declared them. Carrying `expected` in particular looked wrong to me: it describes the decoded value, so `Schema.Option(...)` would start emitting `description: "Option"` for its encoded union under `generateDescriptions: true`.

**2. Annotations closer to the encoded side win.** `unstable/ai/internal/structured-output.ts` composes the type-side description into its own (`"Tuple encoded as an object with numeric string keys …; description"`). My first attempt let the type side win and broke four `AnthropicStructuredOutput` / `OpenAiStructuredOutput` tests, which is what convinced me that a link rewriting the shape of the data should get to describe the result. It also lines up with `resolveReferenceIdentifier`, which prefers the encoded identifier and falls back to the input's. If you read the precedence the other way, it is a one-line flip in `withCarriedAnnotations`, but the structured-output description composition would need reworking to match.

## Tests

Assertions were checked by mutation rather than by inspection — each row below is a single-line change to the fix, re-run against the full workspace suite on this branch:

| Mutation | Tests that fail | Where |
| --- | --- | --- |
| Drop the carry entirely | 22 | `toJsonSchemaDocument` (14), `toRepresentation` (4), `OpenApiGenerator` (3), `HttpApi` (1) |
| Flip the precedence to type-side-wins | 5 | `toRepresentation`, `AnthropicStructuredOutput`, `OpenAiStructuredOutput` |
| Carry every annotation key instead of `jsonSchemaAnnotationKeys` | 9 | `toRepresentation`, `toRepresentations`, `toJsonSchemaMultiDocument`, `toJsonSchemaDocument` |
| Walk one step instead of the whole chain | 1 | `toRepresentation` |
| Reverse the precedence between chain links | 1 | `toRepresentation` |
| Skip the carry on the reference path | 3 | `toRepresentation`, `toJsonSchemaDocument`, `HttpApi` |

`test/schema/representation/toRepresentation.test.ts` holds the unit-level assertions (carry, key filtering, precedence, the reference path, a two-link `Unknown > Declaration > String` chain). `test/schema/toJsonSchemaDocument.test.ts` holds the user-visible ones, including that `Schema.Date.annotate({ default: new Date(0) })` still drops the non-JSON value rather than emitting it.

## Known gap: annotations that land on a check

`.annotate()` targets the last **check** when a schema has one, and checks do not travel to the encoded side of a transformation. So this case is still annotation-free and is **not** fixed here:

```ts
Schema.toJsonSchemaDocument(Schema.FiniteFromString.annotate({ description: "d" })).schema
// { type: "string" }  <- description still dropped
```

I left it out on purpose. The obvious fix is to read through `InternalAnnotations.resolve` instead of `ast.annotations` in `carriedAnnotations`. I tried that and measured it, and it trades one class of dropped annotation for another, because `resolve` returns *only* the last check's annotations and does not merge them with the node's:

| Input | This PR | Via `resolve` |
| --- | --- | --- |
| `FiniteFromString.annotate({description})` | `{type:"string"}` ❌ | `{type:"string", description:"d"}` ✅ |
| `Number.annotate({description}).check(isGreaterThan(0))` | `{anyOf:[…], description:"d"}` ✅ | `{anyOf:[…]}` ❌ |
| `BigInt.annotate({description}).check(…)` | `{type:"string", description:"d", pattern:…}` ✅ | `{type:"string", pattern:…}` ❌ |
| `Number.check(isGreaterThan(0)).annotate({description})` | `{anyOf:[{…, description:"d"}, …]}` | `{anyOf:[{…, description:"d"}, …], description:"d"}` (duplicated) |

Worth flagging clearly: **the full suite passes either way.** No test currently defends either the regression in rows 2–3 or the duplication in row 4, so green CI does not settle this one — it is a behavior question, not a test question.

A merge of the two (`{...ast.annotations, ...lastCheckAnnotations}`) would close the gap without the regression, and I am glad to add it. What stops me from doing it unilaterally is row 4: it leaves a `description` both inside the finite branch and at the top level, and deciding whether that is redundant-but-fine or actually wrong is really the question of what a check annotation means on an encoded shape. That reads to me like your call rather than mine, and it is adjacent to #7336. Happy to fold it into this PR, split it into a follow-up, or drop it entirely.

## Interaction with #7336

Rebased on top of the check compaction. One expectation in `toJsonSchemaDocument.test.ts` needed updating: the annotated `Schema.BigInt` assertion is added by this branch, so it was not covered by #7336's sweep and still expected the pre-compaction shape.

```diff
-{ type: "string", description: "a", allOf: [{ pattern: "^-?\\d+$" }] }
+{ type: "string", description: "a", pattern: "^-?\\d+$" }
```

The compaction also supersedes something I claimed in an earlier version of this description: `Schema.Finite.annotate({description})` no longer nests the annotation under `allOf`, it now reads `{ type: "number", description: "d" }`. That paragraph is gone.

## Verification

On the current head of this branch:

- Full workspace suite — 387 files, 9954 passed, 1 expected fail, 55 skipped.
- `pnpm check` — clean.
- `pnpm lint` (oxlint + `dprint check`) — clean.
- `pnpm test-types --target '>=5.9'` — 182 files, 2144 tests, 5270 assertions.
- CI green on all 13 jobs, Node, Deno and Bun.

One snapshot moved and I think it is worth a look rather than a rubber stamp: `HttpApi`'s OpenAPI fixture now emits `"description": "Some description for User"` on the `UserEncoded` component schema. `User` is a `Schema.Class` declaring that description, and the encoded component was dropping it while the OpenAPI *response* objects already carried it through a different path. The two agree now, but if the previous asymmetry was deliberate I have missed the reason.

## Questions for reviewers

1. Is the broad scope right, or would you prefer this narrowed to `Schema.Number`?
2. Is "encoded side wins" the precedence you want? The structured-output composition depends on it, but that is evidence, not a mandate.
3. The check-annotation gap above — fold in, follow-up, or leave alone?


gcanti:
https://github.com/Effect-TS/effect/pull/7349

spencerbeggs:
Thank you! Amazing.

gcanti:
@spencerbeggs I changed my mind. I now think we should not interfere with the artificial transformation by projecting checks and annotations from the source onto the target: when a schema cannot be represented directly in JSON, some information loss is inevitable unless we introduce arbitrary propagation rules.

spencerbeggs:
@gcanti All good! Schema.Finite was the simple solution to my original problem in my [@effected/schemastore](https://github.com/spencerbeggs/effected/tree/main/packages/schemastore) package. The changes in #7349 gives me the tools to close any other gaps myself, but the package is really just some sugar to deal with SchemaStore quirkiness I kept reinventing. Thank you for considering this. Effect is awesome and you all rock for building it. I appreciate playing in your sandbox.

PR #7206, opened 2026-08-12
feat(cluster): bound runner entity residency and storage reads

tim-smart:
Closes EFF-603

Adds admission control to cluster runners so a backlog of persisted messages for many distinct entity ids can no longer materialize unbounded entities and OOM the runner.

## Changes

**`ShardingConfig`** gains two knobs (env config accepts integers only):

- `maxResidentEntities` (default `10_000`): a runner-wide cap on resident entities. The spawn gate sits in front of the entity `ResourceMap`, so the sequential storage read loop and concurrent volatile `sendLocal` share one counter and neither can bypass the cap. Workflow clock entities count toward the same cap. `"unbounded"` (programmatic only) restores the previous spawn behaviour.
- `unprocessedMessageBatchSize` (default `1024`): bounds each storage read. Cap unit is entities, batch unit is messages.

**Storage read loop** is now two-step on each wake:

1. Read due messages for already-resident addresses (limit = batch size) and deliver them, so residents keep moving even at the cap.
2. If spawn budget remains, walk the remaining due rows in `rowid` order, admitting a new address only while a slot is free.

Latch behaviour: a full batch with delivery progress reopens the read latch immediately; at the cap the loop waits for an entity removal (any `entities.remove` opens the latch) or the poll interval instead of spinning. Rows skipped at the cap have their claims released (`resetAddress`) so they are re-read as soon as a slot frees up.

**At-cap send paths**:

- Storage loop skips new addresses and does not start `resumeEntityFromStorage`; the resident-entity `MailboxFull` resume path for a full per-entity mailbox is unchanged.
- Volatile sends to a new address fail with the existing `MailboxFull` error, which now also means runner-full (documented).
- Persisted sends still persist + notify and return success; `waitUntilRead` notifications resolve successfully at the cap instead of hanging or failing.

**`MessageStorage.unprocessedMessages`** (decoded + encoded, SQL + memory + noop) accepts `{ limit, addresses }` and claims (`last_read`) only the rows actually returned; unclaimed rows stay eligible.

## Tests

- `ShardingConfig`: defaults, env int-only parsing, programmatic `"unbounded"`.
- `Sharding`: cap held while a backlog drains via idle reaping, volatile `MailboxFull` at the cap while persisted sends succeed, resident-prefer delivery with a newer-id backlog in front, full-batch drain within one poll interval, no busy-polling at the cap, `"unbounded"` behaviour.
- `MessageStorage` (memory) and `SqlMessageStorage.integration` (pg / mysql / sqlite): `limit`, `addresses` filter, and `last_read` set only on returned rows.
- Node cluster-integration (`Residency.test.ts`, pg + mysql): a single runner with `maxResidentEntities: 3` holds the cap against a backlog of new ids, residents keep receiving messages, and the backlog drains in waves once slots free up.

`pnpm check`, `pnpm lint`, the full `effect` test suite (8135 tests), platform-node cluster tests, and the cluster-integration suite all pass locally.


packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts
```diff
@@ -702,6 +704,23 @@ const ResumeRpc = Rpc.make("resume", {
...
+// Workflow state is durable, so an idle entity (completed or suspended) can
+// be released quickly and is rebuilt from storage when the next message
+// arrives. Cap the idle time so completed and suspended executions do not
+// hold `maxResidentEntities` slots for the full `entityMaxIdleTime`.
+const workflowIdleTimeCap = Duration.seconds(10)
+
+// The cap only ever shortens the configured `entityMaxIdleTime`, so operators
+// tuning it below the cap keep their value. The config is read optionally to
+// keep it out of the public layer requirements; without it the cap applies.
+const workflowMaxIdleTime: Effect.Effect<Duration.Duration> = Effect.map(
```
tim-smart: Remove this. Just keep it at a fixed 10 seconds.

packages/effect/src/unstable/cluster/MessageStorage.ts
```diff
@@ -131,9 +131,15 @@ export class MessageStorage extends Context.Service<MessageStorage, {
...
+   *
+   * The `limit` option bounds the number of messages returned in a single
+   * read, and the `addresses` option restricts the read to the provided
+   * entity addresses. Only the returned messages are claimed; other messages
+   * stay eligible for later reads.
    */
   readonly unprocessedMessages: (
-    shardIds: Iterable<ShardId.ShardId>
+    shardIds: Iterable<ShardId.ShardId>,
+    options?: UnprocessedOptions | undefined
```
tim-smart: In-line the options type so you don't have to jump to definition.

packages/effect/src/unstable/cluster/ShardingConfig.ts
```diff
@@ -95,6 +95,31 @@ export class ShardingConfig extends Context.Service<ShardingConfig, {
...
+  /**
+   * The maximum number of entities that can be resident on this runner at the
+   * same time.
+   *
+   * When the limit is reached, no new entities are spawned: the storage read
+   * loop stops admitting messages for new entity addresses (they stay in
+   * storage until a slot frees up), and volatile sends to new addresses fail
+   * with `MailboxFull`.
+   *
+   * Workflow clock entities (`Workflow/-/DurableClock`) count towards the same
```
tim-smart: Don't mention workflows in this comment.

packages/effect/test/cluster/ClusterWorkflowEngine.test.ts
```diff
@@ -84,6 +84,71 @@ describe.concurrent("ClusterWorkflowEngine", () => {
       expect(yield* EmailWorkflow.poll(executionId)).toEqual(Option.some(new Workflow.Complete({ exit: Exit.void })))
     }).pipe(Effect.provide(TestWorkflowLayer)))
 
+  it.effect("releases idle workflow entities quickly", () =>
```
tim-smart: We don't need tests for the workflow entity idle ttl

packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts
tim-smart: Updated in d8061b5bc: workflow and durable-clock entities now use a fixed 10-second max idle time, and the changeset reflects that fixed behavior.

packages/effect/src/unstable/cluster/MessageStorage.ts
tim-smart: Updated in d8061b5bc: the options object is inline in both decoded and encoded `unprocessedMessages` signatures, and the exported alias is removed.

packages/effect/src/unstable/cluster/ShardingConfig.ts
tim-smart: Updated in d8061b5bc: removed the workflow-specific paragraph from the `maxResidentEntities` comment.

packages/effect/test/cluster/ClusterWorkflowEngine.test.ts
tim-smart: Updated in d8061b5bc: removed both workflow idle-TTL tests.

PR #7208, opened 2026-08-12
Add the Claude 5 family to the Anthropic model capability table

moishinetzer:
## Summary

`getModelCapabilities` in `AnthropicLanguageModel.ts` does not know the Claude 5 family — `claude-sonnet-5`, `claude-opus-5`, `claude-fable-5`, `claude-mythos-5` all fall through to the unknown-model branch: `{ maxOutputTokens: 4096, supportsStructuredOutput: false }`.

The `supportsStructuredOutput: false` half breaks `generateObject` on these models in a silent way: with no toolkit present, `prepareTools` early-returns before the forced-tool fallback is reached, so the request goes out with no `output_config.format`, no tool, and no JSON instruction at all. The model answers in prose and decoding fails with a `StructuredOutputError`.

All four models support native structured outputs and 128K output tokens (see [What's new in Claude Sonnet 5](https://platform.claude.com/docs/en/about-claude/models/whats-new-sonnet-5) and the [model comparison table](https://docs.claude.com/en/docs/about-claude/models/overview#model-comparison-table)), matching the existing Opus 4.6–4.8 / Sonnet 4.6 group, so this PR adds them to that branch — the same shape as #7129. `claude-sonnet-5` and `claude-fable-5`/`claude-mythos-5` are already present in the `Generated.ts` model union; only the capability table was missing them.

No substring collisions: `claude-sonnet-5` does not match `claude-sonnet-4-5` and `claude-opus-5` does not match `claude-opus-4-5`.

### DIFF
--- a/.changeset/anthropic-claude-5-capabilities.md
+++ b/.changeset/anthropic-claude-5-capabilities.md
@@ -2,6 +2,6 @@
 "@effect/ai-anthropic": patch
 ---
 
-Add the Claude 5 family (Sonnet 5, Opus 5, Fable 5, Mythos 5) to the Anthropic model capability table.
+Default new Anthropic models to modern capabilities while preserving the limits of legacy Claude models.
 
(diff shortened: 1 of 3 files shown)


moishinetzer:
Whoops! I pushed some changes at the same time as you by mistake and reverted it back to your commit after. Sorry for the churn @tim-smart

moishinetzer:
I think you need to reenable merge (squash)

PR #7213, opened 2026-08-13
Scope message storage address clearing to shard

tim-smart:
## Summary

- include `shard_id` when clearing messages and replies from SQL storage
- include the encoded shard id when matching addresses in memory storage
- cover cross-shard preservation through the public `MessageStorage` interface in memory, PostgreSQL, MySQL, and SQLite

## Why

`clearAddress` previously matched only entity type and entity id. When the same entity address values existed on multiple acquired shards, clearing one shard could remove another shard's messages and replies.

## Validation

- `pnpm vitest packages/effect/test/cluster/MessageStorage.test.ts --run`
- `EFFECT_INTEGRATION_TESTS=1 pnpm vitest packages/platform/node/test/cluster/SqlMessageStorage.integration.test.ts --run --project @effect/platform-node`
- `pnpm lint`
- `pnpm check`

Closes EFF-616

### DIFF
--- a/.changeset/scope-message-storage-clear-address.md
+++ b/.changeset/scope-message-storage-clear-address.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Scope `MessageStorage.clearAddress` deletions to the addressed shard.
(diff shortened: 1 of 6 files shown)


tim-smart:
Closing without merge: shard routing is deterministic for supported configurations, and shard-scoped cleanup could leave stale rows after routing changes.

PR #7221, opened 2026-08-13
Migrate NodeRedis from ioredis to redis (node-redis)

WikiRik:
Redis has deprecated `ioredis` in favour of `node-redis`, so `@effect/platform-node`'s `NodeRedis` module now builds on the `redis` client. This follows the official migration guide and keeps the peer range compatible with both redis 5 and 6.

## Breaking changes

**Peer dependency** — replace `ioredis` with `redis: >=5.0.0 <7.0.0`. The package continues to support Node >=18; redis 5 remains available for Node 18 users, while redis 6 requires a newer Node release.

**Connection options** — `layer` and `layerConfig` take node-redis' `RedisClientOptions` instead of ioredis' `RedisOptions`. Socket settings move under `socket`, and `db` becomes `database`:

```ts
// before
NodeRedis.layer({ host: "localhost", port: 6379, db: 1 })

// after
NodeRedis.layer({ socket: { host: "localhost", port: 6379 }, database: 1 })
```

A connection string works too: `NodeRedis.layer({ url: "redis://localhost:6379/1" })`.

**Layers can now fail** — node-redis connects explicitly, so both layers connect while being built and carry `RedisError` in their error channel.

**Direct client access** — `NodeRedis.NodeRedis` exposes node-redis's client type. Command methods are camelCase (`client.lRange` rather than `client.lrange`), and arbitrary commands use `client.sendCommand([...])` rather than `client.call(...)`. Protocol selection follows the installed node-redis version's default; callers can explicitly select a protocol through `RESP`, and the exposed types account for protocol-dependent replies.

## Connection lifecycle

- By default, an initial connection error fails layer construction immediately.
- A caller-supplied `socket.reconnectStrategy` is preserved. After the client first emits `ready`, the default strategy uses node-redis' exponential backoff and stops on `SocketTimeoutError`.
- Client `error` events are logged synchronously at warning level; they do not fork a fiber per event.
- Scope finalization uses `close()`. It waits for in-flight commands, including blocking commands, so scope closure can be delayed.

## Testing

- `EFFECT_INTEGRATION_TESTS=1 pnpm vitest run --project @effect/platform-node test/NodeRedis.integration.test.ts` — 13 tests passed, including fail-fast initial connection and protocol-default coverage.
- `pnpm check` — passed.
- `pnpm lint` — passed.
- `pnpm --filter @effect/platform-node check` with redis 5.12.1 — passed in an isolated worktree.

Closes EFF-650

### DIFF
--- a/packages/platform/node/src/NodeRedis.ts
+++ b/packages/platform/node/src/NodeRedis.ts
@@ -16,7 +16,10 @@ import * as Effect from "effect/Effect"
 import * as Fn from "effect/Function"
 import * as Layer from "effect/Layer"
 import * as Redis from "effect/unstable/persistence/Redis"
-import { createClient, type RedisClientOptions, type RedisClientType } from "redis"
+import { createClient, type RedisClientOptions, type RedisClientType, SocketTimeoutError } from "redis"
+
+type NodeRedisClientOptions = RedisClientOptions<{}, {}, {}, 2>
+type NodeRedisClient = RedisClientType<{}, {}, {}, 2>
 
 /**
  * Service tag for the Node Redis integration, exposing the underlying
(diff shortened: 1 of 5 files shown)


tim-smart:
Thanks, just cleaned up the initial connection to prevent Layer hangs.

packages/platform/node/src/NodeRedis.ts
```diff
@@ -14,31 +15,68 @@ import * as Context from "effect/Context"
...
+              if (cause instanceof SocketTimeoutError) return false
+              const jitter = Math.floor(Math.random() * 200)
+              const delay = Math.min(2 ** retries * 50, 2000)
+              return delay + jitter
+            }
+          }
+          : socket
+      })
+    ),
+    (client) => Effect.ignore(Effect.tryPromise(() => client.close()))
```
tim-smart:
```suggestion
    (client) => Effect.ignoreCause(Effect.promise(() => client.close()))
```

PR #7229, opened 2026-08-13
Run boolean flag fallbacks when the flag is absent

WikiRik:
Quick second PR of the day, this one has more tests and does contradict with an earlier commit although I think this is for the better. Feel free to close/adjust if you don't agree.

The following is from the clanker again.

---

`Flag.withFallbackConfig` and `Flag.withFallbackPrompt` are silent no-ops when applied to a boolean flag. Both fallbacks are driven by the `MissingOption` / `MissingArgument` errors, and an absent boolean flag never produces one — the parser resolves it to `false`, which is correct CLI behaviour on its own. The consequence is that the fallback simply never runs.

The JSDoc on `Flag.withFallbackConfig` uses a boolean flag as its only example, so this is the shape people copy:

```ts
const verbose = Flag.boolean("verbose").pipe(
  Flag.withFallbackConfig(Config.boolean("VERBOSE"))
)
```

Nothing fails — a config value is just ignored, which is easy to ship and hard to notice.

## Reproduction

```ts
Command.make("probe", {
  bool: Flag.boolean("bool").pipe(Flag.withFallbackConfig(Config.boolean("PROBE_VALUE"))),
  str: Flag.string("str").pipe(Flag.withFallbackConfig(Config.string("PROBE_VALUE")))
}, Effect.fn(function*({ bool, str }) {
  yield* Console.log(`bool=${bool} str=${str}`)
}))
```

```
$ PROBE_VALUE=true probe
bool=false str=true
```

One environment variable, one combinator, two outcomes depending only on the flag's primitive type.

## The fix

Both combinators now detect the boolean flag a param reads and consult the fallback when that flag is absent from the parsed flags, before parsing resolves it to `false`.

No parser change was needed. The parser already canonicalizes aliases and records `--no-flag` as `"false"` under the canonical flag name, so "no entry in `ParsedArgs.flags`" is exactly "absent from the command line" — the same check `parseFlag` already makes before defaulting to `false`.

Precedence is unchanged everywhere else:

- `--flag` and `--no-flag` both win over the fallback.
- `optional` and `withDefault` supply their own value first, so the fallback is not consulted — matching how they already shadow the fallback for non-boolean flags.
- A param with `orElse` alternatives is left alone, since the alternative may supply the value.
- Missing config still yields `false`; a cancelled prompt still fails with `MissingOption`.

## Open question for maintainers: should `withFallbackPrompt` change too?

The `withFallbackConfig` half is an unambiguous bug fix. The `withFallbackPrompt` half is not, and I'd like your call on it.

It reverses a deliberate decision from `2e3e4b246` ("Add Flag.withFallbackPrompt"), whose spec said:

> Boolean flags that are absent still default to `false` and do not trigger the fallback prompt (consistent with existing parsing rules).

There was a passing test asserting exactly that, which this PR inverts.

The case for changing it: it is the identical no-op, for the identical reason. If `withFallbackConfig` on a boolean flag should do something, it is hard to argue `withFallbackPrompt` should not.

The case against: a boolean flag is inherently optional, so "absent" is not really a missing value the way it is for a required string flag. Prompting on every invocation where `--flag` was omitted is a much more visible behaviour change than reading an environment variable.

Happy to drop that half and keep this to `withFallbackConfig` (the commit is easy to split), or to keep both. If they stay split, the alternative for the prompt side is to document the limitation instead — the JSDoc examples currently promise behaviour that does not happen.

## Testing

`packages/effect/test/unstable/cli/Param.test.ts` — 9 new cases across both combinators: the fallback fires when the flag is absent, `--no-flag` beats the fallback, missing config yields `false`, `withDefault` beats the fallback, an unparseable config value becomes `InvalidValue`, and a cancelled prompt fails with `MissingOption`.

With the source change reverted, 4 of them fail and 26 pass. The other 5 pass either way by design — they pin the precedence rules the fix must not disturb.

```
pnpm vitest run --project effect packages/effect/test/unstable/cli/Param.test.ts
```

`pnpm lint` and `pnpm check` are clean.

### DIFF
--- a/.changeset/cli-boolean-flag-fallbacks.md
+++ b/.changeset/cli-boolean-flag-fallbacks.md
@@ -1,0 +1,15 @@
+---
+"effect": patch
+---
+
+Run `withFallbackConfig` and `withFallbackPrompt` for boolean flags that are absent from the command line.
+
+Both fallbacks were driven by the `MissingOption` error, which a boolean flag never produces because it parses as `false` when absent. Applying either combinator to a `Flag.boolean` therefore did nothing, including in the documented example:
+
+```ts
+const verbose = Flag.boolean("verbose").pipe(
+  Flag.withFallbackConfig(Config.boolean("VERBOSE"))
+)
+```
+
+The fallback now runs whenever the flag is absent, so `VERBOSE=true` is honored. An explicit `--verbose` or `--no-verbose` still wins, and `optional`/`withDefault` still supply the value before any fallback is consulted. When the config is missing the flag falls back to `false` as before; a cancelled prompt fails with `MissingOption`, matching the existing behavior for other flags.
--- a/packages/effect/src/unstable/cli/Param.ts
+++ b/packages/effect/src/unstable/cli/Param.ts
@@ -1357,6 +1357,37 @@ export const withDefault: {
 })
 
 /**
+ * Returns the name of the boolean flag a param reads, if it reads exactly one
+ * and always parses it.
+ *
+ * An absent boolean flag parses as `false` instead of failing with
+ * `MissingOption`, so fallbacks that react to that error never run for one.
+ * Fallback combinators use this to detect the absent case themselves. Params
+ * wrapped by `optional` or `withDefault` already supply their own value, and
+ * alternatives introduced by `orElse` may provide the value instead, so neither
+ * takes part in a fallback.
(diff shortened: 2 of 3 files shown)


tim-smart:
https://github.com/Effect-TS/effect/pull/7296

PR #7241, opened 2026-08-13
Use Context mapUnsafe in less call sites

tim-smart:


packages/ai/openai-compat/src/OpenAiLanguageModel.ts
```diff
@@ -578,10 +578,9 @@ export const make = Effect.fnUntraced(function*({ model, config: providerConfig
 }): Effect.fn.Return<LanguageModel.Service, never, OpenAiClient> {
   const client = yield* OpenAiClient
 
-  const makeConfig = Effect.gen(function*() {
-    const services = yield* Effect.context<never>()
-    return { model, ...providerConfig, ...services.mapUnsafe.get(Config.key) }
-  })
+  const makeConfig = Effect.contextWith((services) =>
```
tim-smart:
```suggestion
  const makeConfig = Effect.contextWith((services: Context.Context<never>) =>
```

PR #7246, opened 2026-08-14
Restore Effect.head

tim-smart:
## Summary

- restore `Effect.head` with v3-compatible first-element and empty-iterable behavior
- preserve the source error and service types while adding `Cause.NoSuchElementError` for empty iterables
- add focused runtime/type tests, API documentation, and a patch changeset

## Validation

- `pnpm --filter effect test --run test/Effect.test.ts -t head`
- `pnpm test-types Effect.tst.ts`
- `pnpm doctest --run packages/effect/src/Effect.ts`
- `pnpm lint-fix`
- `pnpm check`

Closes EFF-656

### DIFF
--- a/.changeset/restore-effect-head.md
+++ b/.changeset/restore-effect-head.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-Restore `Effect.head` for retrieving the first element of an iterable produced by an effect.
+Add `Effect.head` for retrieving the first element of an iterable produced by an effect.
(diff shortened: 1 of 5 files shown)


.changeset/restore-effect-head.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+Restore `Effect.head` for retrieving the first element of an iterable produced by an effect.
```
tim-smart:
```suggestion
Add `Effect.head` for retrieving the first element of an iterable produced by an effect.
```

PR #7249, opened 2026-08-14
Add @effect/atom-livestore, LiveStore bindings for Atom

skusez:
## Summary

Adds `@effect/atom-livestore` under `packages/atom/livestore`, porting [`@effect-atom/atom-livestore`](https://github.com/tim-smart/effect-atom/tree/main/packages/atom-livestore) to Effect v4 alongside the existing atom framework bindings (`atom-react`, `atom-solid`, `atom-vue`).

LiveStore's main branch has migrated to Effect v4 (`effect@^4.0.0-beta.99`, published as `0.0.0-snapshot-*` builds), but the atom bindings it documents ([LiveStore Effect docs](https://docs.livestore.dev/patterns/effect/)) only exist for Effect v3, leaving no working `AtomLivestore` integration on v4. This PR closes that gap.

## API

Same surface as the v3 package, adapted to v4 conventions:

- `AtomLivestore.Tag<Self>()(id, options)` creates a `Context.Service` class for a LiveStore `Store` (was `Context.Tag`/`TagClassShape`), built via `Layer.effect` + `createStore` + `provideOtel` on an `Atom.runtime`
- `store` emits `AsyncResult<Store<S, TContext>>` (v3 `Result` → v4 `AsyncResult`), `storeUnsafe` emits `Store | undefined`
- `makeQuery` / `makeQueryUnsafe` create reactive query atoms; they now accept LiveStore's v4 `Queryable<A>` (superset of `LiveQueryDef<A>`), and use the v4 `store.subscribe(query, onUpdate)` callback signature directly (the v3 `onUpdate.onUpdate` shim is gone)
- `commit` is now typed as `Atom.Writable<void, LiveStoreEvent.Input.ForSchema<S>>` instead of v3's `Writable<void, {}>`, so committing events from another schema is a type error
- New optional `runtime: Atom.RuntimeFactory` option, mirroring `AtomHttpApi.Service` / `AtomRpc`

Implementation notes:

- `otelOptions` are assembled conditionally to satisfy `exactOptionalPropertyTypes`
- `store` is derived via `self.use(Effect.succeed)` rather than passing the service class to `runtime.atom` (the class is a function, which `runtime.atom` would treat as a read callback)
- peer dep on `@livestore/livestore` is `*` since the v4-compatible builds are currently only published under `0.0.0-snapshot-*` versions (which no ordinary range matches); the dev dep pins the snapshot for `livestorejs/livestore@31e8d71`

## Testing

- `pnpm --filter @effect/atom-livestore run check` / `build` pass
- `tsc -b tsconfig.packages.json` passes with the new project reference
- Verified inference end-to-end against a real LiveStore schema (events + materializers + `queryDb`): query atoms infer row types (not `any`), and `commit` rejects foreign events — via a temporary type-level smoke test, not committed
- No runtime tests: exercising a real store needs a platform adapter (`@livestore/adapter-*`), which seemed too heavy a dev dependency for this PR — happy to add if wanted

## Notes for review

- `syncPayloadSchema` typing (third generic on `CreateStoreOptions`) is not threaded through `Options`, matching the v3 surface; can follow up if desired
- `pnpm-workspace.yaml` `minimumReleaseAgeExclude` entries and part of the lockfile churn come from pnpm auto-installing registry copies of `@effect/*` peers for the external `@livestore/*` packages

### DIFF
--- a/.changeset/atom-livestore-package.md
+++ b/.changeset/atom-livestore-package.md
@@ -1,0 +1,7 @@
+---
+"@effect/atom-livestore": patch
+---
+
+Add `@effect/atom-livestore`, porting the LiveStore bindings from `@effect-atom/atom-livestore` to Effect v4.
+
+`AtomLivestore.Tag` creates a `Context.Service` class for a LiveStore `Store` backed by an atom runtime, exposing atoms for accessing the store (`store`, `storeUnsafe`), reactive query helpers (`makeQuery`, `makeQueryUnsafe`), and a writable atom for committing events (`commit`).
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -19,6 +19,7 @@
       "@effect/ai-openai",
       "@effect/ai-openai-compat",
       "@effect/ai-openrouter",
+      "@effect/atom-livestore",
       "@effect/atom-react",
       "@effect/atom-solid",
       "@effect/atom-vue",
--- a/packages/atom/livestore/LICENSE
+++ b/packages/atom/livestore/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023-present The Contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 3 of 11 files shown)


tim-smart:
I don't think this should live in the effect namespace. Maybe it could be part of live store itself.

PR #7305, opened 2026-08-17
Fix Graph cache and pathfinding edge cases

fubhy:
## Summary

- track nested callback-driven graph transformations as graph lifecycle state and reject finalization while one is active
- invalidate CSR once at the outer transformation boundary, then build fresh uncached projections for reentrant reads without redundant nested invalidation
- materialize consistent graph snapshots before Dijkstra, A*, and Bellman-Ford invoke user cost callbacks
- reject Bellman-Ford distance overflow and bound predecessor reconstruction to prevent hangs

## Validation

- `pnpm lint-fix`
- `pnpm --filter effect test --run test/Graph.test.ts`
- `pnpm check` (passes with existing duplicate-package warnings from `tmp/bundle-base`)

### DIFF
--- a/.changeset/curly-graphs-connect.md
+++ b/.changeset/curly-graphs-connect.md
@@ -2,5 +2,4 @@
 "effect": patch
 ---
 
-Add graph snapshots, low-link connectivity analysis, bipartite matching, maximum flow, and minimum cut APIs. Fix
-mutable graph cache consistency and guard weighted pathfinding against inconsistent snapshots and numeric overflow.
+Add graph snapshots, low-link connectivity analysis, bipartite matching, maximum flow, and minimum cut APIs.
--- a/.changeset/curly-graphs-connect.md
+++ b/.changeset/curly-graphs-connect.md
@@ -2,4 +2,5 @@
 "effect": patch
 ---
 
-Add graph snapshots, low-link connectivity analysis, bipartite matching, maximum flow, and minimum cut APIs.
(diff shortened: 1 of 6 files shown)


fubhy:
## Benchmark: Effect Graph vs @statelyai/graph

Ran the existing seeded comparison harness in `.repos/graph/bench/effect-graph-algorithms.bench.ts` against this branch, with an uncontended 1,000 ms sample window and 100 ms warmup per case.

| Workload | Effect mean | Stately mean | Result |
|---|---:|---:|---:|
| Strongly connected components (600 nodes) | 0.0154 ms | 0.2618 ms | Effect 16.98x faster |
| Connected components (800 nodes) | 0.0180 ms | 0.0151 ms | Stately 1.19x faster |
| Bipartite (800 nodes) | 0.0104 ms | 0.1521 ms | Effect 14.60x faster |
| Topological sort (1,000 nodes) | 0.0459 ms | 0.4465 ms | Effect 9.73x faster |
| A* (500 nodes) | 0.0183 ms | 0.0191 ms | Effect 1.04x faster (effectively tied within error) |
| Bellman-Ford (500 nodes) | 0.0670 ms | 0.6105 ms | Effect 9.11x faster |
| Floyd-Warshall (28 nodes) | 0.1295 ms | 0.4848 ms | Effect 3.74x faster |

This harness measures warm algorithm execution on prebuilt graphs. It does not measure graph construction or callback-driven mutation/cache transitions.

Command: `pnpm exec vitest bench --run bench/effect-graph-algorithms.bench.ts`

PR #7312, opened 2026-08-17
Fix completion of choice values containing quotes, colons or spaces

godu:
## Problem

Choice values from `Flag.choice` / `Argument.choice` are interpolated verbatim into the generated completion scripts. A value containing a single quote breaks the script as a whole, not just the affected parameter.

Reproduced with `Flag.choice("mode", ["it's-fine", "node:20"])`:

| Shell | Emitted | Result |
| --- | --- | --- |
| zsh | `:value:(it's-fine node:20)` | `_deploy:14: unmatched '` — script fails to load |
| fish | `-a 'it's-fine node:20'` | `Unexpected end of string, quotes are not balanced` |
| bash | `compgen -W 'it's-fine node:20'` | loads, but `deploy --mode <TAB>` offers nothing |

fish additionally falls back to file completion, so `deploy --mode <TAB>` lists the current directory instead of the choices.

## Why escaping alone isn't enough

Each generator already has an escape helper it applies to descriptions (`escapeZsh`, `escapeForBash`, `escapeFishString`), but applying it to the values only fixes the *outer* quoting. The word list is then parsed a **second** time, and each shell does something different with it:

- **bash** — `compgen -W` re-expands every word, so `it'\''s-fine` came back as `its-fine`.
- **fish** — the `complete -a` list is *fully expanded*, not just re-tokenized. `(whoami)` was **executed as a command substitution** at TAB time and its output offered as the candidate; `#tag` silently emptied the list; `$HOME`, `~`, `*`, `{a,b}` were expanded; `;` split one value into two.
- **zsh** — `_arguments` does `eval ws=( ${action[2,-2]} )` ([`_arguments:435`](https://github.com/zsh-users/zsh/blob/master/Completion/Base/Utility/_arguments)), consuming exactly one backslash level.

Escaping is therefore two-level, and the inner level must **deny by default** — an allowlist of a few characters is what let the fish command execution through.

Separately, bash inserts `COMPREPLY` entries verbatim (it only escapes under `-o filenames`), so a correct candidate list still put `with space` onto the command line as two words.

## Changes

- **bash** — choice params no longer go through `compgen -W`. The script gains a per-command helper (alongside the existing inlined `_init_completion` fallback) that filters an explicitly quoted list, then handles how readline inserts the match:
  - **Unquoted word** — escaped with `printf %q`.
  - **Inside a quote the user opened** — bash closes the quote but escapes nothing, so the match is escaped for that context: the expandable characters in a `"` context, and the `'\''` splice in a `'` context, since a single quote cannot be escaped inside single quotes.
  - **Replacement boundary** — readline replaces only the text after the last `COMP_WORDBREAKS` character, so that head is trimmed from each match. Wordbreaks that are backslash-escaped or quoted do *not* split the word and must not be treated as the boundary.

  The typed prefix is dequoted before matching so re-completion works after an insertion. The helper is named `` _<name>--choices ``: `sanitizeFunctionName` maps every character outside `[a-zA-Z0-9_]` to `_`, so no generated command function can contain `-` and the name cannot collide.
- **fish** — value-taking flags now drop out of the candidate list once typed, as boolean flags already did; the dedup condition excepts the moment fish is completing the flag's value, which is why they were skipped before (the same `complete` entry provides both the flag name and its value list).
- **fish** — `escapeFishChoice` escapes for both rounds, denying by default (`[^A-Za-z0-9_.,/@%+-]`); `escapeFishString` now escapes backslashes before quotes, which also affected descriptions.
- **zsh** — `escapeZshChoice` escapes for both rounds with the same character class, producing the backslash-escaped form zsh's own completions use (cf. `_quilt`: `a\:all\ headers`): `(it\'\''s-fine node\:20 with\ space)`.

Values containing spaces now complete correctly too, since sidestepping re-expansion also sidesteps IFS splitting.

## Verification

Beyond the unit tests, the generated scripts were syntax-checked and driven through real completion runs: fish via `complete -C`, and bash both non-interactively (invoking the function with `COMP_WORDS`/`COMP_CWORD`) and through **real TAB presses in a pty**, with the completed command echoing its `argv` so the check is what the program actually receives. The bash matrix covers every value unquoted, inside `'…'`, and inside `"…"`. readline's actual replacement boundary was measured directly (returning a fixed marker and observing what it replaced) rather than assumed. Each shell was also run against a plain-value control to prove the harness itself works.

Hostile value set: `it's-fine`, `node:20`, `with space`, `(whoami)`, `#tag`, `$HOME`, `{x,y}`, `a;b`, `a*b`, `say"hi"`, `back\slash`.

| Check | Before | After |
| --- | --- | --- |
| `bash -n` / `zsh -n` / `fish -n` | zsh + fish syntax errors | all clean |
| bash `deploy --mode <TAB>` | `[]` | all 11 values, exact |
| fish `deploy --mode <TAB>` | `godu` (whoami ran) + 3 values | all 11 values, exact |
| bash insertion of `with space` | two argv words | one argv word, `with space` |
| bash `--mode node:<TAB>` | `node:node:20` | `node:20` |
| bash `--mode \(w<TAB>` | `whoami)` | `(whoami)` |
| bash `--mode semi\;c<TAB>` | `colon` | `semi;colon` |
| bash `--mode "$H<TAB>` | `/Users/…` (expanded) | `$HOME` |
| bash `--mode 'it<TAB>` | unterminated quote | `it's-fine` |
| bash `--mode 'with<TAB>` | `with\ space` | `with space` |

`pnpm vitest run packages/effect/test/unstable/cli` → 360 passed, `pnpm lint` clean, `pnpm check` clean.

Pre-existing bash assertions were updated to the new emission (`compgen -W 'dev staging prod'` → `_deploy--choices "$cur" "$_comp_word" 'dev' 'staging' 'prod'`).

## Notes

- The bash prefix dequoting is best effort: it strips one opening quote and any backslash escapes, so a value whose own text contains a backslash will not match once the user types it escaped. Documented in the code.
- `complete -a` is also used for subcommand names (`fish.ts`) and flag names, which go through the same expansion. Not changed here — those are constrained identifiers rather than arbitrary user strings.
- The bash candidate *menu* now shows the escaped form (`with\ space`) rather than the bare value, since the escaping lives in `COMPREPLY`. Insertion is correct; this is cosmetic.
- A value holding `!` is completed as-is inside a user-opened `"` quote: history expansion still applies there and `\!` stays literal, so it cannot be escaped safely. The unquoted and `'` contexts handle `!` correctly.
- bash 3.2's `printf %q` does not escape `~` (5.x does), so on macOS's system bash a choice value shaped like a tilde path (`~/cfg`) was inserted unescaped and tilde-expanded when the command ran. Fixed: the helper escapes a leading `~` itself, a no-op where `%q` already does.
- **Bash 3.2 (was: pre-existing, now fixed here):** the generated script used `local -A` for flag-group dedup, which requires bash 4; on macOS's system bash 3.2 that aborted the function and broke flag and positional completion. The form→group mapping is known at generation time, so the runtime map is now a generated `case` (own changeset). The `case` matches `${words[i]%%=*}`, so `--flag=value` now also marks the flag's group as used — the old exact-key lookup did not.


packages/effect/src/unstable/cli/internal/completions/fish.ts
```diff
@@ -12,7 +12,13 @@ import type * as Completions from "../../Completions.ts"
...
 // ---------------------------------------------------------------------------
 
-const escapeFishString = (s: string): string => s.replace(/'/g, "\\'")
+const escapeFishString = (s: string): string => s.replace(/\\/g, "\\\\").replace(/'/g, "\\'")
+
+/**
+ * Fish re-parses the space-separated list given to `complete -a`, so a value
+ * has to survive two rounds: an inner backslash escape, then string quoting.
+ */
+const escapeFishChoice = (s: string): string => escapeFishString(s.replace(/[\\'\s]/g, "\\$&"))
```
tim-smart: Generated by AI review: This does not survive fish's second expansion round. Characters such as `$`, `(`, `*`, `~`, and `{` remain active: for example, `(id -un)` executes during completion, `$HOME` expands, and an unmatched `a*b` candidate disappears. Please escape every character outside a known-safe set before applying `escapeFishString`, for example:

```suggestion
const escapeFishChoice = (s: string): string => escapeFishString(s.replace(/[^A-Za-z0-9_.,/@%+:-]/g, "\\$&"))
```

I verified that this preserves `$HOME`, `(id -un)`, `a*b`, `~x`, and `{a,b}` verbatim without executing expansions. This is a merge blocker.

packages/effect/src/unstable/cli/internal/completions/bash.ts
```diff
@@ -60,12 +60,22 @@ const buildFlagGroupDeclarations = (
...
 }
 
+/**
+ * `compgen -W` re-expands every word of its list, which mangles values holding
+ * quotes, spaces or glob characters. Filter an explicitly quoted list instead.
+ */
+const choiceCompletion = (values: ReadonlyArray<string>): string => {
+  const words = values.map((value) => `'${escapeForBash(value)}'`).join(" ")
+  return `COMPREPLY=(); local _choice; for _choice in ${words}; do` +
+    ` if [[ $_choice == "$cur"* ]]; then COMPREPLY+=("$_choice"); fi; done`
```
tim-smart: Generated by AI review: The candidates reach `COMPREPLY` intact, but readline inserts raw values for `-F` completions. In real TAB tests, `with space` becomes two arguments, `it's-fine` leaves an unmatched quote, and `back\slash` is received as `backslash`. Please shell-quote each candidate at insertion time, e.g. `COMPREPLY+=("$(printf '%q' "$_choice")")`. I verified that this inserts the three values as `with\ space`, `it\'s-fine`, and `back\\slash`, which execute as their originals. This is a merge blocker unless the claimed bash support for these values is explicitly removed.

tim-smart (changes requested):
Generated by AI review: Request changes. I verified the generated completions against real shell behavior: bash and zsh through TAB presses in PTYs, and fish through `complete -C`. The fish escaping still permits command substitution and other expansion during completion, and bash inserts raw candidates that spaces, quotes, and backslashes make unusable. The zsh approach behaves correctly, including interactive TAB completion. The patch changes only internal generated script text and the patch changeset is appropriate, so API/compatibility risk is low. Please also add regression coverage for backslash and expansion characters in choice values, plus a fish description ending in a backslash; those coverage additions are strongly recommended but are not separate merge blockers.

packages/effect/src/unstable/cli/internal/completions/bash.ts
```diff
@@ -60,12 +78,95 @@ const buildFlagGroupDeclarations = (
...
+  lines.push(`    if ((_escaped)); then _escaped=0; continue; fi`)
+  lines.push(`    case "$_c" in`)
+  lines.push(`      \\\\) _escaped=1 ;;`)
+  lines.push(`      \\"|\\')`)
+  lines.push(`        if [[ -z "$_quote" ]]; then _quote="$_c"`)
+  lines.push(`        elif [[ "$_quote" == "$_c" ]]; then _quote=""`)
+  lines.push(`        fi`)
+  lines.push(`        ;;`)
+  lines.push(`      *)`)
+  lines.push(`        if [[ -z "$_quote" && "$COMP_WORDBREAKS" == *"$_c"* ]]; then _cut=$((_i + 1)); fi`)
```
tim-smart: Generated by AI review: This still doesn't work for values containing `:`. For `deploy --mode node:<TAB>`, bash gives us `cur=:` and `prev=node`, so the `--mode` branch never runs. We should pass `-n "$COMP_WORDBREAKS"` to `_init_completion` and update the fallback to reconstruct the current word from `COMP_LINE`. The current `_cut` block never sees the assembled prefix.

packages/effect/src/unstable/cli/internal/completions/bash.ts
```diff
@@ -60,12 +78,95 @@ const buildFlagGroupDeclarations = (
...
+  lines.push(`        ;;`)
+  lines.push(`      "'")`)
+  lines.push(`        # bare assignment: inside double quotes \\' is not an escape`)
+  lines.push(`        _match=\${_choice//\\'/\\'\\\\\\'\\'}`)
+  lines.push(`        ;;`)
+  lines.push(`      *)`)
+  lines.push(`        printf -v _match '%q' "$_choice"`)
+  lines.push(`        ;;`)
+  lines.push(`    esac`)
+  lines.push(`    COMPREPLY+=("$_match")`)
```
tim-smart, line 106: Generated by AI review: This leaves the quote unterminated when the value ends with the quote the user opened. For example, completing `say"hi"` after `deploy --mode "say` inserts `deploy --mode "say\"hi\"` without the closing quote. We should append `_open` when `_match` already ends with it before adding the match to `COMPREPLY`.

packages/effect/src/unstable/cli/internal/completions/zsh.ts
```diff
@@ -14,6 +14,13 @@ import type * as Completions from "../../Completions.ts"
 
 const escapeZsh = (s: string): string => s.replace(/\\/g, "\\\\").replace(/'/g, "'\\''").replace(/:/g, "\\:")
 
+/**
+ * Values in an `_arguments` action list are backslash-tokenized after the spec
+ * itself is unquoted, so a value has to survive two rounds: an inner backslash
+ * escape, then the surrounding single quotes.
+ */
+const escapeZshChoice = (s: string): string => s.replace(/[^A-Za-z0-9_.,/@%+-]/g, "\\$&").replace(/'/g, "'\\''")
```
tim-smart: Generated by AI review: These regexes need the `u` flag. Without it, non-BMP characters are matched as separate UTF-16 code units and end up as replacement characters in the generated script. Changing both the zsh and fish regexes to `/[^A-Za-z0-9_.,/@%+-]/gu` fixes values such as `🚀` and `a😀b`.

packages/effect/src/unstable/cli/internal/completions/bash.ts
```diff
@@ -60,12 +78,95 @@ const buildFlagGroupDeclarations = (
...
+  lines.push(`  COMPREPLY=()`)
+  lines.push(`  local _choice _match`)
+  lines.push(`  for _choice in "$@"; do`)
+  lines.push(`    [[ "$_choice" == "$_prefix"* ]] || continue`)
+  lines.push(`    case "$_open" in`)
+  lines.push(`      '"')`)
+  lines.push(`        _match="\${_choice//\\\\/\\\\\\\\}"`)
+  lines.push(`        _match="\${_match//\\$/\\\\$}"`)
+  lines.push(`        _match="\${_match//\\\`/\\\\\\\`}"`)
+  lines.push(`        _match="\${_match//\\"/\\\\\\"}"`)
```
tim-smart, line 96: Generated by AI review: We should mention this in the notes. `!` is still history-expanded inside a user-opened double quote, so completing `a!b` can change the command line. I don't think we can safely escape it here because `\!` stays literal inside double quotes. The unquoted and single-quoted cases work correctly.

packages/effect/src/unstable/cli/internal/completions/bash.ts
```diff
@@ -16,6 +16,24 @@ const escapeForBash = (s: string): string => s.replace(/'/g, "'\\''")
 
 const sanitizeFunctionName = (s: string): string => s.replace(/[^a-zA-Z0-9_]/g, "_")
 
+/**
+ * Every function name `generateFunction` will emit. `sanitizeFunctionName` maps
+ * any character to `_`, so a subcommand can produce any name — the shared helper
+ * has to pick one that is provably not among them.
+ */
+const emittedFunctionNames = (
```
tim-smart: Generated by AI review: I think this can be much simpler. Use a helper name such as `_${safeName}--choices`. `sanitizeFunctionName` replaces every non-`[a-zA-Z0-9_]` character with `_`, so `-` cannot collide with a generated command name. Then we can remove `emittedFunctionNames`, the recursive tree walk, and the collision test.

packages/effect/test/unstable/cli/completions/completions.test.ts
```diff
@@ -215,7 +224,47 @@ describe("Bash completions", () => {
...
+    const script = Bash.generate("deploy", desc)
+    assert.include(script, `_deploy__choices "$cur" 'it'\\''s-fine' 'node:20' 'with space' '(whoami)' '#tag' '$HOME'`)
+    assert.include(script, `_deploy__choices "$cur" 'o'\\''clock' 'a:b' '{x,y}'`)
+    assert.notInclude(script, `compgen -W 'it`)
+  })
+
+  it("escapes matches for the quoting context of the word being completed", () => {
+    const desc = fromCommand(withTrickyChoices)
+    const script = Bash.generate("deploy", desc)
+    assert.include(script, `local _prefix="\${_cur#[\\"\\']}"; _prefix="\${_prefix//\\\\/}"`)
```
tim-smart: Generated by AI review: I don't think these assertions are useful. They assert the fixed helper source character for character, so formatting changes break them while incorrect behavior can still pass. Keep the call-site assertions, and use one snapshot of the helper block if we want to detect changes to it.

packages/effect/src/unstable/cli/internal/completions/bash.ts
godu: Fixed in 5027dc25a. `_init_completion` now gets `-n "$COMP_WORDBREAKS"`, and the inline fallback reassembles `words`/`cur`/`prev` from `COMP_LINE` the same way, splitting on whitespace only — so `deploy --mode node:<TAB>` sees `prev=--mode` and the assembled `cur=node:20` prefix. Covered by the `splits words on whitespace only` and fallback tests.

packages/effect/src/unstable/cli/internal/completions/bash.ts
godu: Fixed in 5027dc25a: when the escaped match already ends with the quote the user opened, the helper appends `_open` before adding it to `COMPREPLY`, so `deploy --mode "say<TAB>` now inserts `"say\"hi\""` with the closing quote.

packages/effect/src/unstable/cli/internal/completions/zsh.ts
godu: Fixed in 3c93fc53d: both the zsh and fish regexes now carry the `u` flag, so escaping applies to whole code points. Tests cover `🚀` as a value and `a😀b` in the argument choices for all three shells.

packages/effect/src/unstable/cli/internal/completions/bash.ts
godu: Documented in 5027dc25a (helper doc comment) and now in the PR notes: `!` inside a user-opened `"` quote is left as-is, since history expansion still applies there and `\!` would stay literal. Unquoted and single-quoted contexts handle it correctly; `a!b` is in the test value set.

packages/effect/src/unstable/cli/internal/completions/bash.ts
godu: Simplified in 5027dc25a exactly as suggested: the helper is `_${safeName}--choices`, and `emittedFunctionNames`, the tree walk, and the collision test are gone.

packages/effect/test/unstable/cli/completions/completions.test.ts
godu: Replaced in 5027dc25a: the per-line assertions are now a single `toMatchInlineSnapshot` of the helper block, with the call-site assertions kept separately.

godu:
Independent re-verification on macOS with `/bin/bash` 3.2.57, bash 5.3.15 (homebrew), zsh 5.9, and fish 4.8.1. Method: scripts generated straight from this branch's generators for a fixture carrying the full hostile value set; bash and zsh driven by **real TAB keypresses in a pty**, with the completed command defined as a function echoing its `argv` — so every assertion is on what the program receives, not on screen rendering; fish via `complete -C` (which evaluates `-n` conditions). Each shell also ran a plain-value control CLI to validate the harness itself.

*(Updated: initially run at 2330ab984; re-run after 2017ef504 and c99b79381, which turned the two non-green cells found by the first pass into fixes. bash columns fully re-measured on both versions; zsh generation is untouched by those commits.)*

| # | Case | Expected | bash 3.2 | bash 5.3 | zsh | fish |
| --- | --- | --- | --- | --- | --- | --- |
| 0 | syntax parse (`-n` / `--no-execute`) | clean | ✅ | ✅ | ✅ | ✅ |
| 1 | control CLI: `--env de<TAB>` | argv `<dev>` | ✅ | ✅ | ✅ | ✅ |
| 2 | menu: `--mode <TAB>` | all 14 values offered | ✅ | ✅ | ✅ | ✅ verbatim |
| 3 | `wi<TAB>` | argv `<with space>` (one word) | ✅ | ✅ | ✅ | ✅ |
| 4 | `node:<TAB>` (wordbreak) | argv `<node:20>` | ✅ | ✅ | ✅ | ✅ |
| 5 | `'it<TAB>` (open `'`) | argv `<it's-fine>`, quote closed | ✅ | ✅ | ✅ | ✅ |
| 6 | `'with<TAB>` (open `'`) | argv `<with space>` | ✅ | ✅ | ✅ | ✅ |
| 7 | `"$H<TAB>` (open `"`) | argv `<$HOME>` literal | ✅ | ✅ | 🟨 | 🟨 |
| 8 | `\(w<TAB>` (escaped wordbreak) | argv `<(whoami)>`, nothing executes | ✅ | ✅ | ✅ | ✅ |
| 9 | `say<TAB>` (value ends in `"`) | argv `<say"hi">` | ✅ | ✅ | ✅ | ✅ |
| 10 | `bac<TAB>` | argv `<back\slash>` | ✅ | ✅ | ✅ | ✅ |
| 11 | `~<TAB>` | argv `<~x>` | ✅ | ✅ | 🟨 | 🟨 |
| 12 | prefix filter: `it` | only `it's-fine` | ✅ | ✅ | ✅ | ✅ |
| 13 | flag-name menu: `--<TAB>` | all flags, no error | ✅ | ✅ | ✅ | ✅ |
| 14a | used boolean flag: `--loud --<TAB>` | all its forms gone | ✅ | ✅ | ✅ | ✅ |
| 14b | used value flag: `--mode v --<TAB>` | `--mode` gone | ✅ | ✅ | ✅ | ✅ (c99b79381) |
| 14c | value still completes after dedup: `--mode <TAB>`, retyped `--mode`, `--mode=<TAB>` | all choices | ✅ | ✅ | ✅ | ✅ |
| 15 | subcommand: `stat<TAB>` | argv `<status>` | ✅ | ✅ | ✅ | ✅ |
| 16 | positional: `o<TAB>` | argv `<o'clock>` | ✅ | ✅ | ✅ | ✅ |
| 17 | tilde-path value `~/cfg-a` + Enter | literal argv | ✅ (2017ef504) | ✅ | ✅ inserts `\~/cfg-a` | 🟨 |

Rows 13–14 ran on bash 3.2 specifically to exercise the `local -A` → `case` rewrite, which is what used to abort there.

**🟨 (5 cells, by design):** in rows 7 and 11 zsh and fish route the word to their own engine contexts before any per-command spec runs — `$…` to variable-name completion, `~…` to username completion (zsh offered 133 users, fish 122; fish literally completed `"$HOME/` and moved on to path completion). Same for every completion on those shells (git, kubectl, …), identical on `main`, and not reachable from a generated script; row 17 fish is the same tilde context making the insertion path untestable. The values stay reachable through the empty-prefix menu (row 2) in every shell. The unquoted and single-quoted paths for the same values are green everywhere.

No ✗ and no ⚠️ cells remain.

tim-smart (approved):

PR #7322, opened 2026-08-18
Cluster sharding on Cloudflare Durable Objects

tim-smart:
Closes EFF-698

Shared PR for the Cluster Sharding on Cloudflare Durable Objects effort. Serial slices land here one by one; this first slice adds the package scaffold and the Worker/Durable Object glue.

## What's in this slice

- New package `@effect/platform-cloudflare` (cluster plus the minimum Worker/DO glue; no HttpServer/Crypto/FS parity).
- Four SQLite-backed Durable Object classes for the Worker to re-export and bind: `ClusterEntity` (shared entity class), `ClusterWorkflow`, `ClusterDurableQueue`, and `ClusterSingleton` (placeholders reserving their bindings).
- Cheap entity constructor: open SQLite, ensure the mailbox tables, re-arm the single alarm. User handlers are never built in the constructor.
- Length-prefixed name codec `` `${type.length}:${type}${id}` `` shared by entity/workflow/queue/singleton addresses, with strict decode.
- `CloudflareCluster.layer({ entities, entityNamespace, workflowNamespace, queueNamespace, singletonNamespace })` providing the cluster `Sharding` service. `Entity.client` stays the user API; clients resolve objects via `getByName`; unknown entity types fail at the Worker before contacting a Durable Object. Handlers register per `EntityType` at Worker init.
- `CloudflareCluster.makeRunnerAddress`: synthetic `CurrentRunnerAddress` stub from DO identity, no peer dialing.
- Wrangler example in the package README binding the four classes.

## Tests

- Unit tests against fakes: name encode/decode round-trips and strict decode failures, layer wiring (client resolution, unknown-type failure, duplicate/unbound handler registration), constructor storage helpers (table DDL, alarm re-arm semantics).
- One Miniflare smoke test booting the four classes as SQLite-backed Durable Objects in workerd and verifying the bindings resolve and direct fetch of an object is rejected.

## Handler concurrency slice (EFF-727)

`Entity.toLayer(..., { concurrency })` is now honored inside the entity Durable Object. Storage entry (decode, persist-before-run, dedupe, duplicate resume, alarm arming) keeps the single entry permit; handler execution runs in forked fibers governed by a per-entity semaphore sized from the option (default 1, numeric N, `"unbounded"`). Replayed rows and alarm-due runs draw from the same budget, and a handler's permit is released while a stream chunk waits for its client acknowledgement.

**Why not `RpcServer.makeNoSerialization({ concurrency })`:** it was evaluated and does not fit the Durable Object flow. Its permit is held for a stream handler's whole lifetime, including client-ack waits; on Cloudflare acks arrive from remote callers over separate DO requests, so an unacked or abandoned stream would pin the only default permit and brick the entity (the existing "acknowledges stream chunks without holding the entity lock" test fails under that model). It also owns request lifecycle state that this path keeps in SQLite: persist-before-run and `PrimaryKey` duplicate resume must consult the mailbox before any handler starts, replies must be written through `saveReply` inside `transactionSync`, and `lastSentChunk` sequence resume plus the per-registration `defectRetryPolicy` rebuild live in the hand-rolled runtime. Wiring the server in would mean re-adapting all of that around it, as the classic `entityManager` does, for no net simplification. A scoped permit protocol on the existing loop keeps the locked persist/dedupe semantics untouched.

Later slices (mailbox, DeliverAt, keepAlive, workflow engine, queue, singleton/cron, proxies/metrics/docs) push to this branch.

## Interrupted persisted stream cancellation (EFF-731)

- Explicit cancellation now replaces stored stream chunks with an interrupted terminal reply after the handler fiber stops.
- Cancelled requests no longer replay on the next Durable Object contact, retain mailbox capacity, or leave an unsettled `waitUntil` session.
- A real-SQLite regression test covers interrupt → subsequent invoke and the durable terminal state.

Closes EFF-731

## Durable Object initialization readiness (EFF-863)

- `CloudflareDurableObjects.setInitializer` now gives Worker handlers and all four Durable Object classes one request-context-safe readiness contract.
- Cold requests and persisted alarms wait for registration before consulting module registries; a rejected initializer is cached and propagated to callers and alarm handling.
- Deterministic Miniflare gates cover entity, workflow, queue, singleton, successful cold alarms, and initialization failure without fixture-only readiness subclasses.

Closes EFF-863

## DeliverAt smoke test repair

The Miniflare fixture now dispatches each entity alarm at an explicit timestamp, removing the short wall-clock assertion window. The test retains persisted-deadline, first-delivery, and re-arming checks, and also checks no early delivery and a cleared final alarm. Only the test and its fixture change.

Validation: all 12 Durable Object smoke tests and all 135 Cloudflare tests pass.

Closes EFF-1314


.changeset/cloudflare-workflow-engine.md
```diff
@@ -0,0 +1,16 @@
+---
+"@effect/platform-cloudflare": minor
+"effect": minor
```
tim-smart: There should only be one changeset for this PR. patch release for "effect"

packages/effect/src/unstable/cluster/Entity.ts
```diff
@@ -780,3 +785,8 @@ export const KeepAliveRpc = Rpc.make("Cluster/Entity/keepAlive")
 export class KeepAliveLatch extends Context.Service<KeepAliveLatch, Latch.Latch>()(
   "effect/cluster/Entity/KeepAliveLatch"
 ) {}
+
+class KeepAliveHandler extends Context.Service<
```
tim-smart: This can be exported

packages/platform/cloudflare/src/internal/entityKeepAlive.ts
```diff
@@ -0,0 +1,50 @@
+/** @internal */
+import * as Context from "effect/Context"
+import * as Effect from "effect/Effect"
+import * as Latch from "effect/Latch"
+
+/** @internal */
+export class EntityKeepAliveHandler extends Context.Service<
```
tim-smart: This should be imported from the Entity module

packages/platform/cloudflare/src/internal/entityMailbox.ts
```diff
@@ -0,0 +1,277 @@
...
+
+/** @internal */
+export const persistRequest = (
+  sql: SqlStorage,
+  envelopeText: string,
+  primaryKey: string | null,
+  discard = false,
+  deliverAt: number | null = null,
+  replyTo: string | null = null
+): PersistResult => {
```
tim-smart: This should return an Effect

packages/platform/cloudflare/src/internal/entityMailbox.ts
```diff
@@ -0,0 +1,277 @@
...
+    envelopeText,
+    discard ? 1 : 0,
+    deliverAt,
+    replyTo === null ? null : JSON.stringify([replyTo])
+  )
+  return { _tag: "Success" }
+}
+
+/** @internal */
+export const saveReply = (sql: SqlStorage, replyText: string): void => {
```
tim-smart: This should return an Effect

packages/platform/cloudflare/src/internal/entityMailbox.ts
```diff
@@ -0,0 +1,277 @@
...
+export interface StoredMessage {
+  readonly requestId: string
+  readonly envelope: string
+  readonly lastSentChunk: string | undefined
+  readonly discard: boolean
+  readonly deliverAt?: number | undefined
+  readonly replyTos?: ReadonlyArray<string> | undefined
+}
+
+const decodeReplyTargets = (value: unknown): Array<string> => {
```
tim-smart, line 189: This should return an Effect

packages/platform/cloudflare/src/internal/entityReply.ts
```diff
@@ -0,0 +1,40 @@
+/** @internal */
+import * as Context from "effect/Context"
+
+/** @internal */
+export const CurrentEntityName = Context.Reference<string | undefined>(
```
tim-smart: Instead of a Reference, you can use a Service with the optional access apis.

packages/platform/cloudflare/src/internal/queueRuntime.ts
```diff
@@ -0,0 +1,147 @@
...
+      setWithEviction(delivered, waiter.takerId, item.id, deliveredCapacity)
+      const expiry = now + waiter.leaseMillis
+      if (earliest === undefined || expiry < earliest) earliest = expiry
+      waiter.resolve(item)
+    }
+    return earliest === undefined ? Effect.void : armAlarm(options.alarm, earliest)
+  })
+
+  const mutateAndWake = (mutate: () => void): Promise<void> =>
+    Effect.runPromise(Effect.andThen(Effect.sync(mutate), wakeWaiters))
```
tim-smart: Instead of Effect.andThen it could all be done inside Effect.suspend

packages/platform/cloudflare/src/CloudflareDurableObjects.ts
```diff
@@ -0,0 +1,923 @@
...
+ *
+ * @category durable objects
+ * @since 4.0.0
+ */
+export class ClusterEntity extends DurableObject<unknown> {
+  readonly #state: DurableObjectState
+  readonly #address: EntityAddress.EntityAddress
+  readonly #name: string
+  #runtime: EntityRuntime | undefined
+  #serial: Promise<void> = Promise.resolve()
```
tim-smart: Entities should support concurrency using a semaphore.

.changeset/cloudflare-workflow-engine.md
tim-smart: Consolidated into a single `effect` patch changeset (`.changeset/cloudflare-cluster.md`); the nine per-slice changesets are removed.

packages/effect/src/unstable/cluster/Entity.ts
tim-smart: Exported as `Entity.KeepAliveHandler` with docs, and `@effect/platform-cloudflare` now uses it instead of its own copy.

packages/platform/cloudflare/src/internal/entityKeepAlive.ts
tim-smart: Now imported from `effect/unstable/cluster/Entity`; the duplicated service class is removed.

packages/platform/cloudflare/src/internal/entityMailbox.ts
tim-smart: `persistRequest` now returns `Effect<PersistResult, MailboxFullError | EncodedMessageTooLargeError>`, so the call site drops its `Effect.try` + instanceof dance and just pattern-matches the typed failure.

packages/platform/cloudflare/src/internal/entityMailbox.ts
tim-smart: `saveReply` now returns `Effect<void, EncodedMessageTooLargeError>`.

packages/platform/cloudflare/src/internal/entityMailbox.ts
tim-smart: Every exported mailbox operation now returns an Effect, and the transaction boundary moved into a `withTransaction` helper so call sites can compose them (e.g. `clearReplies` + `saveReply` atomically). `decodeReplyTargets` stays a pure private helper inside those effects.

packages/platform/cloudflare/src/internal/entityReply.ts
tim-smart: Both `CurrentEntityName` and `CurrentReplyRegistry` are plain `Context.Service` classes now, read with `Context.getOption` at the optional access sites.

packages/platform/cloudflare/src/internal/queueRuntime.ts
tim-smart: Folded into a single `Effect.suspend`.

packages/platform/cloudflare/src/internal/entityRuntime.ts
```diff
@@ -0,0 +1,841 @@
...
+            Effect.onExit((exit) =>
+              Effect.suspend(() => {
+                if (!persisted) return Effect.void
+                pendingRuns.delete(requestId)
+                return discard && Exit.isSuccess(exit) ? completeTell(sql, requestId) : Effect.void
+              })
+            )
+          )
+      )
+      options.waitUntil(Fiber.await(fiber))
```
tim-smart, line 488: Are we sure this won't leak fibers?

tim-smart: I traced ownership and termination end to end and don't see a leak.

Termination: the forked effect is `entityRuntime.run`, which wraps the handler in `runWithDefectRetry` (`Effect.exit`, or `retryOrElse` ending in `Effect.succeed(Exit.failCause(...))`), so failures and defects come back as exit values instead of escaping. The respond path is synchronous SQLite work (`saveReply` under `withTransaction`) plus, for scheduled rows, a DO RPC whose promise settles and is `catchCause`d inside `deliverScheduledReply`. The only ways this fiber runs forever are a handler that never returns or an unbounded user `defectRetryPolicy`. Both are user code; the runtime has no unguarded wait on this path.

Cleanup rides on the fiber, not the caller: `pendingRuns.delete` and `completeTell` sit in `Effect.onExit`, so they run on success, failure, and interruption.

Ownership: `forkDetach` moves the fiber to the global scope deliberately, since a persisted tell must survive the RPC that delivered it (persist-before-run). Nothing can interrupt it externally; `manager.interrupt` only reaches `sessions`, the ask path. The final owner is the Cloudflare runtime: this line becomes `ctx.waitUntil(Effect.runPromise(Fiber.await(fiber)))` (`CloudflareDurableObjects.ts:114`). `Fiber.await` succeeds with the exit for every outcome, so the promise handed to `waitUntil` settles exactly when the fiber does and never rejects.

Accumulation: re-forking the same request while it is in flight is blocked by `pendingRuns` here and `sessions` on the ask path, so live fibers are bounded by unprocessed mailbox rows plus concurrent invokes, and the handler semaphore bounds actual execution (permits released in `onExit` / `uninterruptibleMask` in `makeHandlerPermit`). Once a fiber settles, nothing retains it.

Worst case, a handler that genuinely never returns pins the object until Cloudflare evicts the isolate; the fiber dies with the isolate and the unprocessed row replays on the next wake. That is the at-least-once contract rather than a leak, and no memory survives across wakes.

So no code change from my side. Leaving the thread open for you.

packages/platform/cloudflare/test/fixtures/worker.ts
```diff
@@ -0,0 +1,360 @@
...
+    const requestId = JSON.parse(envelope).requestId
+    this.#testState.storage.sql.exec(
+      `INSERT INTO cluster_messages (request_id, message_id, envelope, discard, processed, last_reply_id)
+       VALUES (?, NULL, ?, 0, 0, NULL)`,
+      requestId,
+      envelope
+    )
+  }
+
+  scheduledRows(): Array<Record<string, unknown>> {
```
tim-smart: This can be typed more strongly as we know the columns we are selecting. Take a look for any other opportunities to improve query types.

tim-smart: Fixed in `c8f4b6028`. `scheduledRows` now uses an exact `ScheduledRow` result type. I also typed every consumed `SqlStorage.exec` result across the Cloudflare mailbox, workflow, queue, singleton, entity alarm storage, and the remaining SQL-backed test assertions, removing the casts those types made unnecessary. Verified with 62 focused Cloudflare tests, `pnpm lint-fix`, and `pnpm check`.

PR #7358, opened 2026-08-19
Add RPC HTTP stream backpressure

tim-smart:
## Summary

- bound framed RPC HTTP response queues to 16 items by default
- allow `streamBufferSize` to select a numeric capacity or `"unbounded"`
- keep non-framed buffered responses unbounded
- cover the default, numeric override, unbounded opt-out, and non-framed behavior

The framed HTTP transport previously created an unbounded per-request output queue, so an unread response body could not slow a streaming RPC handler. The stream pump already awaits each queue offer; using a bounded queue supplies the missing backpressure without changing the pump.

## Checks

- `pnpm vitest --run packages/effect/test/rpc/RpcServer.test.ts`
- `pnpm test-types RpcServer.tst.ts`
- `pnpm lint`
- `pnpm check`

Closes EFF-741

### DIFF
--- a/packages/effect/src/unstable/rpc/RpcServer.ts
+++ b/packages/effect/src/unstable/rpc/RpcServer.ts
@@ -806,6 +806,7 @@ export const layerHttp = <Rpcs extends Rpc.Any>(options: {
   readonly spanAttributes?: Record<string, unknown> | undefined
   readonly concurrency?: number | "unbounded" | undefined
   readonly disableFatalDefects?: boolean | undefined
+  readonly streamBufferSize?: number | undefined
 }): Layer.Layer<
   never,
   never,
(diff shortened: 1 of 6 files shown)


packages/effect/src/unstable/rpc/RpcServer.ts
tim-smart: I think we can add the options here and make this a breaking change while we're in beta.

packages/effect/typetest/unstable/rpc/RpcServer.tst.ts
```diff
@@ -16,6 +16,30 @@ describe("RpcServer", () => {
     })
   })
 
+  it("HTTP server options accept streamBufferSize", () => {
```
tim-smart: Remove this test

packages/effect/src/unstable/rpc/RpcServer.ts
```diff
@@ -972,7 +973,9 @@ export const layerProtocolWebsocket = (options: {
  * @category protocols
  * @since 4.0.0
  */
-export const makeProtocolWithHttpEffect: Effect.Effect<
+export const makeProtocolWithHttpEffect: (options: {
+  readonly streamBufferSize?: number | "unbounded" | undefined
+}) => Effect.Effect<
```
tim-smart:
```suggestion
export const makeProtocolWithHttpEffect: (options?: {
  readonly streamBufferSize?: number | "unbounded" | undefined
} | undefined) => Effect.Effect<
```

PR #7382, opened 2026-08-20
Add context-based CLI prompt themes

### LINKED ISSUE #7375: cli/Prompt: an empty prefix still indents the prompt by one column

`renderPrompt` builds the line as `leadingSymbol + " " + message`, so the separating space is emitted even when `prefix` is `""`. A CLI that opted out of the mark still gets a one-column indent it cannot remove, leaving the question misaligned with the choices beneath it.

Emit the separating space only when the prefix is non-empty, so `prefix: ""` renders the message flush at column 0. Prompts with a prefix are unaffected.

- `prefix: " "` — pads to two columns instead of removing the indent; trades one unwanted space for another.
- Padding every `message` to compensate — impossible, since the indent cannot be negative.
- Rebuilding the prompt on `Prompt.custom` — works, but duplicates the built-in's input handling and frame clearing to change one space.

### LINKED ISSUE #7376: cli: let a CLI define its own design language, instead of only its prefix

A CLI's visual language is currently governed in two places, and only one of them is reachable.

Help output and errors go through `CliOutput.Formatter` — and since #7089, so do handler failures. Prompts go through a separate path that draws from an internal `figures` record plus a handful of inline literals, none of which a caller can influence. #7256 added a `prefix` option, but that is one symbol out of many, set per call, on one of the two systems.

The result is that a CLI can style its help and its errors, and cannot make its prompts match them.

Everything a prompt draws today, and where it comes from:

| Symbol | Default | Drawn by |
| --- | --- | --- |
| `pointer` | `❯` | the highlighted choice |
| `pointerSmall` | `›` | the trailing symbol on every prompt line |
| `ellipsis` | `…` | a submitted prompt |
| `tick` | `✔` | a confirmed prompt |
| `arrowUp` / `arrowDown` | `↑` `↓` | paging, when choices overflow |
| `checkboxOn` / `checkboxOff` | `☒` `☐` | multi-select state |
| `prefix` | `?` | the leading mark — configurable since #7256 |

And outside the record, unreachable even if it were exposed: the `- ` that joins a choice to its description, the password mask `*`, the toggle separator `/`, and the file prompt's `(Y/n)` and its fixed English confirmation sentence — the latter also an i18n problem.

`figures` additionally defines `arrowLeft`, `arrowRight`, `radioOn`, `radioOff`, `cross`, and `line`, which no renderer references.

One value that describes how a CLI looks, supplied once rather than per call.

- A theme record carrying the symbol vocabulary above, including `prefix` as an ordinary member rather than a special case, and the literals that currently sit inline: the description separator, the password mask, and the toggle separator.
- Supplied as a `Context.Reference`, defaulting to today's platform-selected set, so existing code is unchanged and a CLI states its language once instead of repeating options at every prompt — with a per-prompt override for exceptions.
- Reachable from where `CliOutput.Formatter` is configured, so one declaration covers help, errors, and prompts rather than two systems disagreeing.
- Any symbol set to `""` renders nothing, including the space that would have separated it (this subsumes #7375).

Only the symbols renderers actually draw need to be in the public type; the six unused entries would otherwise be frozen into the API by exposing the record as it stands.

- Adding `suffix`, `pointer`, and similar options beside `prefix` — every setting stays per-call, so consistency across a CLI remains the author's job and a prompt added later silently reverts to the defaults.
- Passing a shared options object into every prompt (what we do now) — works until someone writes a prompt without it, and it cannot reach the mask, the separator, or the paging arrows at all.
- Rebuilding prompts on `Prompt.custom` — duplicates input handling, paging, and frame clearing to change a few glyphs, and drifts as the built-ins evolve.

tim-smart:
## Summary

- replace per-prompt `prefix` options with a `Prompt.Theme` context reference
- support partial per-prompt theme overrides for symbols, masks, separators, and semantic prompt colors
- preserve the existing platform-specific symbols and ANSI colors as defaults
- keep partially empty or multi-character theme elements aligned and add focused runtime and type coverage

## Validation

- `nix develop -c pnpm lint`
- `nix develop -c pnpm test --run packages/effect/test/unstable/cli` (366 tests)
- `nix develop -c pnpm test-types packages/effect/typetest/unstable/cli/Prompt.tst.ts` (TypeScript 5.9 and 6.0)
- `nix develop -c pnpm check`
- `nix develop -c pnpm changeset status --since origin/main`

Closes EFF-797

### DIFF
--- a/packages/effect/src/unstable/cli/Prompt.ts
+++ b/packages/effect/src/unstable/cli/Prompt.ts
@@ -182,6 +182,16 @@ export interface Theme {
   readonly passwordMask: string
   /** The separator between toggle labels. */
   readonly toggleSeparator: string
+  /** The color used for active prompt elements. */
+  readonly primaryColor: string
+  /** The color used for secondary prompt elements. */
+  readonly mutedColor: string
+  /** The color used for completed prompt markers. */
+  readonly successColor: string
+  /** The color used for validation errors. */
+  readonly errorColor: string
+  /** The color used for submitted values. */
+  readonly submittedColor: string
 }
 
 /**
@@ -597,7 +607,12 @@ const defaultTheme: Theme = {
   pointer: "❯",
   descriptionSeparator: "- ",
   passwordMask: "*",
-  toggleSeparator: "/"
+  toggleSeparator: "/",
+  primaryColor: Ansi.cyanBright,
+  mutedColor: Ansi.blackBright,
+  successColor: Ansi.green,
+  errorColor: Ansi.red,
+  submittedColor: Ansi.white
 }
 
 const windowsTheme: Theme = {
@@ -785,7 +800,8 @@ export const all: <
 }
 
 const annotateLine = (line: string): string => Ansi.annotate(line, Ansi.bold)
-const annotateErrorLine = (line: string): string => Ansi.annotate(line, Ansi.combine(Ansi.italicized, Ansi.red))
+const annotateErrorLine = (line: string, color: string): string =>
+  Ansi.annotate(line, Ansi.combine(Ansi.italicized, color))
 const annotateSymbol = (symbol: string, ...styles: Array<string | Array<string>>): string =>
   symbol.length === 0 ? "" : Ansi.annotate(symbol, ...styles)
 const separateSymbol = (symbol: string, text: string): string => symbol.length === 0 ? text : symbol + " " + text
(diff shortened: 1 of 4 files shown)


packages/effect/src/unstable/cli/Prompt.ts
```diff
@@ -566,47 +597,74 @@ export interface ToggleOptions {
...
-  line: "─",
   pointer: ">"
 }
 
-/** @internal */
-export const platformFigures = Effect.map(
-  Effect.sync(() => process.platform === "win32"),
-  (isWindows) => isWindows ? windowsFigures : defaultFigures
-)
+const mergeTheme = (base: Theme, overrides: Partial<Theme> | undefined): Theme => {
```
tim-smart: I think using a plain object spread instead of this is fine.

PR #7384, opened 2026-08-20
Optimize synchronous Schema parsing

tim-smart:
Preserve completed parser exits when adapting Schema issues to `SchemaError`, avoiding a fiber run for synchronous decode and encode calls.

Use a direct, resumable property loop for structs without index signatures under the common parse options. Suspended child parsers resume from their current property without replaying earlier fields; other option modes keep the existing interpreter.

Adds regression coverage for eager Schema adapters and defect conversion in the struct fast path.

Tests:

- `pnpm lint`
- `pnpm test --run` (388 files, 9,981 passed, 1 expected failure, 55 skipped)

Closes EFF-798

### DIFF
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -1518,6 +1518,13 @@ export function decodeUnknownEffect<S extends Constraint>(schema: S, options?: S
 function fromIssueEffect<A, R>(
   self: Effect.Effect<A, SchemaIssue.Issue, R>
 ): Effect.Effect<A, SchemaError, R> {
+  if (core.isExit(self)) {
+    return self._tag === "Success"
+      ? self as unknown as Effect.Effect<A, SchemaError, R>
+      : Exit_.failCause(
+        Cause_.map(self.cause as Cause_.Cause<SchemaIssue.Issue>, (issue) => new SchemaError(issue))
+      )
+  }
   return Effect.catchCause(
     self,
     (cause) => Effect.failCauseSync(() => Cause_.map(cause, (issue) => new SchemaError(issue)))
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -35,6 +35,7 @@ import * as Graph_ from "./Graph.ts"
 import * as HashMap_ from "./HashMap.ts"
 import * as HashSet_ from "./HashSet.ts"
 import * as core from "./internal/core.ts"
+import { effectIsExit } from "./internal/effect.ts"
 import * as InternalGraph from "./internal/graph.ts"
 import * as InternalRecord from "./internal/record.ts"
 import * as InternalAnnotations from "./internal/schema/annotations.ts"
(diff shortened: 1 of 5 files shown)


packages/effect/src/Schema.ts
```diff
@@ -1518,6 +1518,13 @@ export function decodeUnknownEffect<S extends Constraint>(schema: S, options?: S
 function fromIssueEffect<A, R>(
   self: Effect.Effect<A, SchemaIssue.Issue, R>
 ): Effect.Effect<A, SchemaError, R> {
+  if (core.isExit(self)) {
```
tim-smart: We could use mapErrorEager here instead?

tim-smart: `mapErrorEager` is the right shape but not quite equivalent here. Its eager branch is `exitMapError`, which takes the first `Fail` reason and rebuilds the cause with `exitFail`; the lazy branch (`mapError` -> `catch_`) does the same. So a cause carrying both a schema issue and a defect loses the defect. That is reachable: a transformation returning `Effect.failCause(Cause.combine(Cause.fail(issue), Cause.die(e)))` reaches this adapter with `["Fail", "Die"]` and comes out as `["Fail"]`.

Instead I reused `fromIssueExit`, which already existed a few lines below and does the same eager mapping while preserving the full cause:

```ts
function fromIssueEffect<A, R>(
  self: Effect.Effect<A, SchemaIssue.Issue, R>
): Effect.Effect<A, SchemaError, R> {
  if (core.isExit(self)) {
    return fromIssueExit(self as Exit_.Exit<A, SchemaIssue.Issue>)
  }
  return Effect.catchCause(
    self,
    (cause) => Effect.failCauseSync(() => Cause_.map(cause, (issue) => new SchemaError(issue)))
  )
}
```

Happy to switch to `mapErrorEager` if dropping the sibling reasons is acceptable — it is one line and a touch smaller.

packages/effect/src/Schema.ts
```diff
@@ -1518,6 +1518,9 @@ export function decodeUnknownEffect<S extends Constraint>(schema: S, options?: S
...
```
tim-smart: This should use `effectIsExit`

tim-smart: Updated to use `effectIsExit` in fcbac2397.

tim-smart:
## Before and after runtime benchmark

Measured on Node 26.7.0, Linux x64. Each case was warmed up, calibrated to at least 40 ms of work, and sampled five times in three interleaved process rounds. The table reports medians of medians. Lower is better.

`before` is `main` at `c04c7f8b7`. `after` is the optimized implementation at `931f3e1c5`. The final commit, `fcbac2397`, only replaces the exit predicate with its shared helper and does not change the measured path.

| case | before | after | speedup |
| --- | ---: | ---: | ---: |
| decode `String` | 233.2 ns | 21.7 ns | 10.75x |
| decode flat struct, 8 primitives | 611.8 ns | 232.2 ns | 2.63x |
| decode flat struct, 1 transforming field | 748.7 ns | 308.9 ns | 2.42x |
| decode nested struct | 842.4 ns | 283.8 ns | 2.97x |
| decode array of 200 structs | 47,590 ns | 25,852 ns | 1.84x |
| encode flat struct, 8 primitives | 709.1 ns | 264.9 ns | 2.68x |
| encode array of 200 structs | 47,788 ns | 26,039 ns | 1.84x |
| raw `SchemaParser`, flat struct | 392.1 ns | 248.6 ns | 1.58x |

The adapter exit short-circuit dominates small values, while the direct struct loop supplies the scaling improvement for nested structs and arrays. The mixed transformation case retains nearly the same gain as the primitive-only case.

PR #7387, opened 2026-08-21
Implement schema-aware RPC serialization

tim-smart:
Fills the `unknown` holes in RPC protocol messages with a codec supplied by the serialization, instead of a hardcoded `Schema.toCodecJson`. Envelope framing is untouched: `Parser` still only ever sees already-encoded envelopes.

Closes EFF-804

## Slice 1: RPC module

`RpcSerialization` gains `codecFor`:

```ts
export type HoleCodecFor = <S extends Schema.Top>(
  schema: S
) => Schema.Codec<S["Type"], unknown, S["DecodingServices"], S["EncodingServices"]>
```

`RpcClient.Protocol` and `RpcServer.Protocol` re-pass it, so `RpcClient.make` and `RpcServer` read `codecFor` off the protocol and never require `RpcSerialization` themselves. HTTP, socket, websocket, and stdio protocols forward `serialization.codecFor`. Worker protocols use `Schema.toCodecJson` and keep no `RpcSerialization` dependency, because structured clone never goes through a serialization; a binary worker would be a different protocol, not a flag on these constructors. Worker initial messages stay on `Schema.toCodecJson`.

Client codecs are compiled into a per-client `WeakMap` keyed by `Rpc`; the old module-level cache would have been wrong once two protocols can carry different codecs.

`RpcMessage.ResponseDefectEncoded` and `ResponseExitDieEncoded` now wrap an already-encoded defect. Call sites encode first with `codecFor(Schema.Defect())`; `ResponseExitDieEncoded` takes `encodedDefect` in place of `defect`.

## Slice 2: cluster network

`Envelope.PartialRequest.payload` and `Reply.Encoded` change from `Schema.Any` to a named `Envelope.OpaqueHole`: the identity under `Schema.toCodecJson`, so the outer runner encode never transforms an already-encoded hole, and a later binary codec can compile it as a bytes leaf.

The runner network path uses the transport codec — `Runners` send/notify and reply decode, the runner server's reply encode, and the entity payload decode in `entityManager`. `Message.IncomingRequest` carries the codec that filled its holes, so messages read back from `MessageStorage` (always JSON) and messages arriving off the wire decode correctly side by side. `Message.serializeRequest` / `serialize` / `deserializeLocal` and `Reply.Reply` / `serialize` / `serializeOrDefect` / `serializeLastReceived` take an optional codec, defaulting to JSON for the storage path.

Storage is unchanged: `serializeEnvelope`, `Envelope.PartialJson`, the SQL payload columns, and `MessageStorage` all stay JSON.

## Compatibility

Every built-in serialization (json, ndjson, jsonRpc, ndJsonRpc, msgpack) returns `Schema.toCodecJson`, so no wire format changes.

Breaking for custom `Protocol` implementations, which must now supply `codecFor`. `RunnerServer.layerHandlers` (and therefore `layer` / `layerWithClients`) now requires `RpcSerialization`; the existing socket and HTTP runner layers already provided it, so their declared requirements are unchanged.

The local `simulateRemoteSerialization` round trip still uses JSON. `Runners.makeRpc` builds its client protocol per runner address, so it has no codec at construction time to hand the local path.

## Tests

- `test/rpc/RpcHoleCodec.test.ts` — client/server round trip over a serialization whose `codecFor` writes each hole as a JSON string, covering payload, stream chunks, a typed error exit (via a `Schema.Date` field that cannot survive a structural pass), and a handler defect. Plus an assertion that the built-in serializations JSON-lower.
- `test/cluster/Runners.test.ts` — a runner protocol with the same codec: the entity payload is encoded with it and carried through the opaque hole without being re-encoded.
- `test/cluster/RunnerServer.test.ts` — the runner server decodes the entity payload and encodes the reply with the serialization's codec.
- `test/cluster/Envelope.test.ts` — the payload hole is identity under `Envelope.PartialJson`, including for a non-JSON value.

`pnpm check`, `pnpm lint`, `pnpm circular`, and the `effect` cluster / rpc / ai / eventlog / reactivity / workflow / persistence tests plus `@effect/platform-node` and `@effect/platform-browser` all pass locally.

## Left for the follow-up

SchemaBinary as an `RpcSerialization` layer stays out of scope. That map is a layer swap plus one compile rule: `Envelope.OpaqueHole` must compile as a bytes leaf rather than being JSON-blobbed by a `toCodecJson ?? toCodec` fallback.


packages/effect/src/unstable/cluster/RunnerServer.ts
```diff
@@ -52,6 +57,9 @@ const serializeDefectReply = <R extends Rpc.Any>(
 export const layerHandlers = Runners.Rpcs.toLayer(Effect.gen(function*() {
   const sharding = yield* Sharding.Sharding
   const storage = yield* MessageStorage.MessageStorage
+  // The entity payload and the replies travel inside the runner envelope as
+  // already-encoded holes, so they use the wire format's codec, not JSON.
+  const { codecFor } = yield* RpcSerialization.RpcSerialization
```
tim-smart: I think it should be accessed on the protocol so we don't have to add more dependencies.

packages/effect/src/unstable/rpc/RpcSerialization.ts
```diff
@@ -15,8 +15,29 @@ import * as Data from "../../Data.ts"
...
+ *
+ * The envelope around the hole is framed by {@link Parser}, which never sees the
+ * schema. JSON based serializations return `Schema.toCodecJson`, so the encoded
+ * hole is JSON. Other serializations may encode the hole as bytes; both fit the
+ * `unknown` hole, so `RpcSerialization` has no type parameter.
+ *
+ * @category serialization
+ * @since 4.0.0
+ */
+export type HoleCodecFor = <S extends Schema.Top>(
```
tim-smart: I think just `CodecFor` is a better name

packages/effect/test/rpc/RpcHoleCodec.test.ts
```diff
@@ -0,0 +1,125 @@
+import { assert, describe, it } from "@effect/vitest"
```
tim-smart: This should be in RpcSerialization.test.ts

PR #7388, opened 2026-08-21
Fix MCP initialize rejected over the protocol version header

ebramanti:
### Fix MCP initialize rejected over the protocol version header

`McpServer.layerHttp` validates the `MCP-Protocol-Version` header on every POST, including the `initialize` request. That header reports the version negotiated by an earlier `initialize`, so on a fresh connection a client can only send its own default. Whenever that default is not among the server's registered protocols, the `initialize` returns `400` and never reaches version negotiation, even when the body offers a version the server supports.

The `400` is also hard to diagnose, because the specification gives it a second meaning. Per [Streamable HTTP → Backwards Compatibility](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#backwards-compatibility), a client that receives `400`, `404` or `405` from a POSTed `InitializeRequest` should conclude the endpoint speaks the deprecated two-endpoint HTTP+SSE transport and retry with a `GET`. `layerHttp` answers that `GET` with `405`, correctly, so the client discards the `400` and reports "this server does not support SSE transport" — true, and unrelated to the cause.

This scopes the header check to requests after initialization, where the specification requires it: the client "**MUST** include the `MCP-Protocol-Version` header on all **subsequent** requests", and its value "**SHOULD** be the one negotiated during initialization". An `initialize` now negotiates from the version offered in its body, through the protocol registry, which already falls back to the server's preferred protocol and reports the selection in the response header. Behavior after initialization is unchanged. Patch changeset included.

### Validation

Added `negotiates an initialize request from a client on an unsupported protocol version`: a server registering only `v2025_11_25` accepts an `initialize` from a client on `2025-06-18` and answers `200` with `mcp-protocol-version: 2025-11-25`. It fails against unpatched `main`.

All existing tests are green.


lloydrichards:
The bug looks real, and ignoring the version header during the old initialization seems like the practical fix.

But I want to point out that the spec is a little unclear here, see [modelcontextprotocol/modelcontextprotocol#2721](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2721). 

The TLDR is that its not clear on what to do if the initial request includes a wrong or conflicting header, thus the fix looks good, but the compliance claim should be softer, and we need to make sure it does not weaken the new v2026-07-28 behavior (as this might need to be ported into the `mcpRuntime`.

packages/effect/src/unstable/ai/McpServer.ts
```diff
@@ -1300,6 +1300,12 @@ const mcpStdioSerialization = (
  * remain valid. The surrounding HTTP server remains responsible for binding
  * to an appropriate interface and installing authentication.
  *
+ * The `MCP-Protocol-Version` header is never checked on an `initialize` request.
+ * The header reports the version negotiated by an earlier `initialize`, and
+ * clients read `400`, `404` and `405` on a POSTed `InitializeRequest` as a legacy
+ * HTTP+SSE endpoint and retry with a GET. The version offered in the request body
+ * is negotiated through the protocol registry instead.
+ *
  * `layerHttp` always implements the single-endpoint Streamable HTTP topology.
```
tim-smart, line 1305:
```suggestion
 *
 * `layerHttp` always implements the single-endpoint Streamable HTTP topology.
```

packages/effect/src/unstable/ai/McpServer.ts
```diff
@@ -1377,18 +1383,15 @@ const layerMcpProtocolHttp = (options: {
...
-        protocolVersion !== session.protocol.protocolVersion
-      ) {
-        return Effect.succeed(HttpServerResponse.empty({ status: 400 }))
-      }
+      // Evaluated only once the body shows the request is not an `initialize`. The header
+      // reports the version negotiated by an earlier `initialize`, so it cannot gate the
+      // `initialize` itself, and a client reads 400 on a POSTed InitializeRequest as "this
+      // endpoint speaks the deprecated two-endpoint HTTP+SSE transport" and retries with a
+      // GET. An `initialize` negotiates from the version offered in the body instead.
+      const hasUnsupportedProtocolVersion = (protocolVersion !== undefined &&
```
tim-smart: I think the version check should still happen before the parse error and invalid request paths. Right now an unsupported Mcp-Protocol-Version returns a JSON-RPC parse error for invalid JSON, and an invalid request response for {"hello":"world"}, where main returned 400 for both.

packages/effect/src/unstable/ai/McpServer.ts
```diff
@@ -1377,18 +1383,15 @@ const layerMcpProtocolHttp = (options: {
...
-      ) {
-        return Effect.succeed(HttpServerResponse.empty({ status: 400 }))
-      }
-      if (
-        session?.protocol.transport.requiresVersionHeader === true &&
-        protocolVersion !== session.protocol.protocolVersion
-      ) {
-        return Effect.succeed(HttpServerResponse.empty({ status: 400 }))
-      }
+      // Evaluated only once the body shows the request is not an `initialize`. The header
```
tim-smart: This comment isn't quite right, since `hasUnsupportedProtocolVersion` is still evaluated eagerly. Only the places where it is checked have moved.

I think this would be simpler if we computed `isInitialize` at the top of the decoded onSuccess, returned 400 when `!isInitialize && rejectsProtocolVersionHeader`, and removed the duplicate checks below.

tim-smart (changes requested):

ebramanti:
@tim-smart Addressed comments

@lloydrichards How do we make the compliance claim softer?

tim-smart (approved):

PR #7391, opened 2026-08-21
Add a Postgres wire protocol codec to sql-pg

tim-smart:
Adds a client-side wire codec for `@effect/sql-pg` so a later issue can replace the `pg` dependency. This PR is the codec only: `PgClient` is untouched and still uses `pg` at runtime.

Closes EFF-808
Closes EFF-812

## What is here

Three modules in `packages/sql/pg/src/`, all pure functions with no `Effect`, `Schema`, or `Stream` in their signatures:

**`PgProtocol`** - protocol 3.0 frames.

- Specials with no type byte: `encodeSslRequest`, `decodeSslResponse`, `encodeStartupMessage` (defaults `client_encoding=UTF8`), `encodeCancelRequest`.
- Frontend: Parse, Bind, Execute, Describe, Close, Sync, Flush, Terminate, PasswordMessage, SASLInitialResponse, SASLResponse. Bind always requests format code `1` for parameters and results.
- Backend: the Authentication\* family, ParameterStatus, BackendKeyData, ReadyForQuery, RowDescription, DataRow, CommandComplete, the empty acknowledgements, ParameterDescription, Error/Notice/NotificationResponse, NegotiateProtocolVersion, and the COPY messages. An unrecognised type byte decodes to `{ _tag: "Unknown", type, payload }` rather than failing.
- `makeParser({ maxMessageSize })` is incremental. Partial messages stay in an internal buffer, while malformed input throws a terminal `ParseError`. The default limit is 16 MiB.
- `DataRow` fields stay raw bytes. `ErrorResponse` fields are parsed into a record keyed by name, so `code` maps straight onto the SQLSTATE classification `PgClient` already does.

**`PgTypes`** - binary values by OID.

- `encode(value, oid)` and `decode(bytes, oid, format)` return typed `Result` values; `format = 0` is an error in v1.
- Built-ins: bool, int2/4/8, float4/8, numeric, oid, text/varchar/bpchar/name, bytea, json/jsonb, date/time/timetz/timestamp/timestamptz, uuid, inet/cidr, and 1-dimensional arrays of those. `ndim = 0` decodes to `[]`; anything above 1 dimension is an error.
- Layouts follow rust-postgres' `postgres-types`, including the `DT_NOBEGIN`/`DT_NOEND` infinity sentinels. `timestamp` is treated as UTC, which is documented on the module.
- Constructors carry the OID (`int4(1)`, `timestamptz(ms)`); there is no `typeof` inference. `register`/`unregister` add codecs for other OIDs. An unknown OID decodes to `Uint8Array` and fails to encode.

**`PgAuth`** - MD5 and SCRAM-SHA-256, with the SASL frames left opaque in `PgProtocol`. Its fallible public APIs return typed `Result` values. The caller supplies the SCRAM nonce, so nothing here needs randomness. `SCRAM-SHA-256-PLUS` is out of scope until the TLS socket exists.

The API split is deliberate. Stateless public codec and authentication failures use `Result`. `Parser.push` and its field-reader callback are stateful hot paths and throw, because any failure is terminal for that parser. `PgTypes.makeFieldReader` validates the row shape in a `Result`, then returns the internal throwing reader used by the parser.

## Tests

`pnpm vitest run --project @effect/sql-pg` - 194 tests, no Docker.

The goldens in `test/fixtures/goldens.ts` are real bytes, not hand-written hex. `test/fixtures/regenerate.ts` drives a live PostgreSQL 16 over a raw socket using these modules and records what comes back. During that capture the server accepted the startup message, the SCRAM proof, and every encoded parameter, and returned each value byte-identical to what `PgTypes.encode` produced, including `numeric`, IPv6 `inet`, `timetz`, the infinity sentinels, and arrays with nulls. The MD5 and cleartext fixtures come from the same script against servers started with those auth methods.

Re-running the script reproduces every golden except the four that are unique to a session: the SCRAM salt and server nonce, and the backend pid in `backendKeyData` and `notificationResponse`.

Covered: SSL request and both replies, StartupMessage, the auth frames, a Bind and DataRow for every v1 OID including nulls, the empty array and a 1-D array with nulls, PortalSuspended, ErrorResponse with `code`, messages split across `push` calls and two messages in one chunk, an oversize length prefix, an unknown type byte, timestamp/timestamptz infinity, numeric `"NaN"`, int8 at both ends of the range, and terminal parser behavior after protocol or field-reader failures.

## Performance

These results are from PR head `c97e8bc48`, measured on 2026-08-22 with the repository's documented offline command:

```sh
nix develop -c pnpm --filter @effect/sql-pg benchmark:codec
```

The benchmark runs in one Node process with identical semantic payloads and no database, socket, TLS, query, or server latency. Each task warms for 250 ms and measures for one second. The numbers below are nanoseconds per row from the same run, so lower is better.

| Suite | `@effect/sql-pg` binary | `pg` text / `pg-protocol` | `postgres.js` text |
|---|---:|---:|---:|
| Type encode, 6 columns | 311.46 | 173.56 | 298.39 |
| Type decode, 6 columns | 419.12 | 425.05 | 404.07 |
| `int4[16]` decode | 72.70 | 1700.40 | n/a |
| Bind frame from values | 335.44 | 818.42 | n/a |
| Bind frame from array values | 440.12 | 2527.63 | n/a |
| DataRow frames to values | 470.49 | 872.83 | n/a |
| RowDescription parser | 392.62 | 454.72 | n/a |
| DataRow parser, one chunk | 97.95 | 325.58 | n/a |
| DataRow parser, 64-byte chunks | 154.30 | 449.68 | n/a |

The component type tables measure different work. `pg`'s `prepareValue` produces no wire bytes, while its text parsers receive strings already built by the protocol parser. The complete Bind and DataRow suites are the direct client-path comparisons.

The optimized APIs also run alongside their allocation-heavy fallbacks in the same benchmark:

| Internal path | Optimized | Fallback |
|---|---:|---:|
| Bind frame from values | 335.44 | 475.84 |
| Bind frame from array values | 440.12 | 1468.12 |
| DataRow values, 6 mixed columns | 470.49 | 508.19 |
| DataRow values, 20 `int4` columns | 210.29 | 515.36 |

The narrow parser rollback was also compared against `068f20d2b7` in a reversed-order, same-machine pair. Single-chunk parsing was 97.95 vs 98.09 ns/row, 64-byte incremental parsing was 154.30 vs 155.98, and the wide-row field reader was 210.29 vs 215.41. The earlier 4.6-6.2% parser regression is gone.

Most of the gain comes from avoiding temporary arrays and views:

- The parser's buffer is append-only. A full buffer is replaced rather than compacted in place, so bytes already handed out are never rewritten and `DataRow` fields, `CopyData`, and `Unknown` payloads can be views into it. Decoding a message does not allocate a `Reader`, a `DataView`, or a payload subarray.
- Frontend frames are written back to back into a pooled buffer and returned as views, matching `pg-protocol`'s serializer. `makeBindEncoder(PgTypes.writeParameter)` writes values directly into a Bind frame, including array elements.
- `PgTypes.makeFieldReader` resolves codecs once per result shape and reads fields in place. The wide-row suite shows the payoff when per-column overhead dominates.

A view stays valid for as long as it is held, because neither buffer is ever rewritten. Holding one keeps its whole buffer alive, so anything that has to outlive its message needs a copy. That is documented on the module and on `Parser.push`, and covered by tests.

## Out of scope, as specified

Rewiring `PgClient`; connection, TLS, pooling, session pipelining; simple `Query`; encoding or driving COPY; FunctionCall; GSS/Kerberos/SSPI; `SCRAM-SHA-256-PLUS`; text-format codecs; composites, ranges, geometric types, hstore; `Date`, Node `Buffer`, `DateTime`. Live-Postgres tests wait for the client issue.

## Validation

`pnpm lint`, `pnpm check`, `pnpm vitest run --project @effect/sql-pg`, and `pnpm --filter @effect/sql-pg benchmark:codec`.


.changeset/tall-melons-shave.md
```diff
@@ -0,0 +1,19 @@
+---
+"@effect/sql-pg": minor
```
tim-smart:
```suggestion
"@effect/sql-pg": patch
```

packages/sql/pg/src/PgAuth.ts
```diff
@@ -0,0 +1,254 @@
...
+  for (let i = 0; i < left.length; i++) {
+    result[i] = left[i] ^ right[i]
+  }
+  return result
+}
+
+const toBase64 = (bytes: Uint8Array): string => Encoding.encodeBase64(bytes)
+
+const fromBase64 = (text: string, field: string): Uint8Array => {
+  const decoded = Encoding.decodeBase64(text)
```
tim-smart, line 66: Instead of throwing everywhere use the Result type

tim-smart: Fixed in 3fd46f3284. The fallible PgAuth, PgProtocol, and PgTypes public APIs now return typed Result values, including custom codecs, field readers, and fused Bind writers. Tests, docs, fixtures, and the benchmark use the new contracts; the built-in field/writer hot paths retain their allocation-free success path.

tim-smart: Final split is in c97e8bc48. PgAuth and the public PgTypes codec operations still return typed Results. Parser.push and FieldReader are synchronous throwing hot paths again; any parser or field-reader failure is terminal. PgTypes.makeFieldReader validates the row shape in a Result, then returns the internal reader used by the parser. Regression coverage checks both the typed public failure and terminal parser behavior. The same-machine comparison is back at baseline, and the PR description has the fresh numbers.

PR #7393, opened 2026-08-21
Fix `Prompt.autoComplete` swallowing `j` and `k` while filtering

### LINKED ISSUE #7392: 🐛 `Prompt.autoComplete` cannot filter for any text that contains `j` or `k`
## What version of Effect is running?

`4.0.0-rc.111` (also reproduced on `main` @ `15272a66ad`)

## What steps can reproduce the bug?

`Prompt.autoComplete` is a prompt where you type to filter the choices. But the key handling sends a plain `j` or `k` to the cursor (down and up) before the filter ever receives them. So any text that contains one of these two letters is changed without telling the user.

```ts
import { Prompt } from "effect/unstable/cli"

const branch = Prompt.autoComplete({
  message: "Pick a branch",
  choices: [
    { title: "main", value: "main" },
    { title: "feat/jira-fetch-tool", value: "feat/jira-fetch-tool" },
    { title: "feat/worktree-tool", value: "feat/worktree-tool" }
  ]
})
```

Run it and type `jira`. Or type `worktree`, which contains a `k`.

Here is a version that always gives the same result, because it drives the prompt through a mock `Terminal` (the same one used in `packages/effect/test/unstable/cli/services/MockTerminal.ts`):

```ts
// type the text, press enter, then close the input queue, so that a query
// without any match does not block the run forever
yield* terminal.inputText("worktree")
yield* terminal.inputKey("enter")
yield* terminal.inputKey("c", { ctrl: true })

const result = yield* Effect.result(Prompt.run(branch))
// the frames show `[filter: wortree]` and "No matches", and the run quits
// instead of returning feat/worktree-tool
```

## What is the expected behavior?

Every printable character should reach the filter, so the list becomes shorter as the documentation describes:

| you type | expected filter | expected result |
| --- | --- | --- |
| `jira` | `jira` | `feat/jira-fetch-tool` |
| `worktree` | `worktree` | `feat/worktree-tool` |
| `j` (with `feat/jira-fetch-tool` first in the list) | `j` | `feat/jira-fetch-tool` |

The documentation comment on `autoComplete` (`packages/effect/src/unstable/cli/Prompt.ts:1241`) starts with "Creates a prompt that lets users filter select choices by typing". The current behavior does not do this.

## What do you see instead?

| you type | real filter | real result |
| --- | --- | --- |
| `fetch` (control, no `j`/`k`) | `fetch` | `feat/jira-fetch-tool` ✅ |
| `jira` | `ira` | `feat/jira-fetch-tool`, but only by luck, because `ira` still matches |
| `worktree` | `wortree` | **"No matches"**. After `enter` the prompt beeps and then waits forever |
| `j` (with `feat/jira-fetch-tool` first in the list) | *no change* | **`main`**. The `j` moved the cursor, and `enter` returned the wrong value without any error |
| `kot` on the choices from the documentation | `ot` | `Kotlin`, again only by luck |

Please look at the `worktree` row. This query has nothing to do with vi navigation, but it is still changed, only because it contains a `k`. The third row shows the worst case: the prompt returns a wrong value and says nothing.

The example inside the documentation comment of `autoComplete` has the same problem. Its three choices are TypeScript, Rust and **Kotlin**, and `Kotlin` is exactly the one you cannot type for.

### Cause

`handleAutoCompleteProcess`, `packages/effect/src/unstable/cli/Prompt.ts:3559`:

```ts
const handleAutoCompleteProcess = <A>(options: AutoCompleteOptionsReq<A>) => {
  return (input: Terminal.UserInput, state: AutoCompleteState) => {
    if (input.key.ctrl) {
      if (input.key.name === "u") {
        return processAutoCompleteClear(state, options)
      }
      return Effect.succeed(Action.Beep())   // all other ctrl keys only beep
    }
    switch (input.key.name) {
      case "k":
      case "up": {                            // <-- a plain `k` hides the filter
        return processAutoCompleteCursorUp(state)
      }
      case "j":
      case "down": {                          // <-- a plain `j` hides the filter
        return processAutoCompleteCursorDown(state)
      }
      // ...
      default: {
        return processAutoCompleteInput(Option.getOrElse(input.input, () => ""), state, options)
      }
    }
  }
}
```

The `j` and `k` keys come from `handleSelectProcess`, where they are correct. `select` has no text input, so no printable character can be understood in two ways there. But `autoComplete` does have a text input, and vi style navigation does not work together with it.

### Suggested fix

Remove the plain `j` and `k` cases from the `autoComplete` key handling. Leave `select` and `multiSelect` as they are. Then move the vi and emacs style navigation to the ctrl keys, which today only beep:

```ts
if (input.key.ctrl) {
  switch (input.key.name) {
    case "u": return processAutoCompleteClear(state, options)
    case "p":
    case "k": return processAutoCompleteCursorUp(state)
    case "n": return processAutoCompleteCursorDown(state)
    default:  return Effect.succeed(Action.Beep())
  }
}
switch (input.key.name) {
  case "up":   return processAutoCompleteCursorUp(state)
  case "down": return processAutoCompleteCursorDown(state)
  // ... not changed
}
```

`Ctrl+N` and `Ctrl+P` are the default keys in `fzf` and in readline. So you can still move without the arrow keys, and at the same time every printable character reaches the filter.

### Why not `Ctrl+J`

The normal vi pair would be `Ctrl+J` and `Ctrl+K`. But the `Terminal` layer cannot tell the difference between `Ctrl+J` and `Enter`. `Ctrl+J` is ASCII `0x0A`, and `NodeTerminal` (`packages/platform/node-shared/src/NodeTerminal.ts`) uses `readline.emitKeypressEvents`, which reports it as `{ name: "enter", ctrl: false }`:

```
\x0a (ctrl+j) -> { name: "enter",  ctrl: false }   # the same as Enter
\x0b (ctrl+k) -> { name: "k",      ctrl: true  }
\x0d (enter)  -> { name: "return", ctrl: false }
\x0e (ctrl+n) -> { name: "n",      ctrl: true  }
\x10 (ctrl+p) -> { name: "p",      ctrl: true  }
```

This is why `Ctrl+N` is used for "down". To use `Ctrl+J` you would have to skip `readline` and read raw stdin yourself.

### Test coverage

`packages/effect/test/unstable/cli/Prompt.test.ts:533` (`describe("Prompt.autoComplete", ...)`) has no test for a query with a `j` or a `k` in it. These tests would help:

- typing a query with `j` and `k` makes the list shorter (fails today)
- `Ctrl+P` and `Ctrl+N` move the cursor
- `Ctrl+K` moves the cursor up

### Workaround

The arrow keys and `tab` still bring you to the choice, but typing, which is the main idea of this prompt, does not. The other option is to build the prompt again on `Prompt.custom`, which is public, with a `process` handler that sends every printable character to the filter. `platformFigures` is exported, but the render helpers and `internal/ansi.ts` are not, so you also have to write the rendering yourself.

wmaurer:
Closes #7392

## What is the problem

`Prompt.autoComplete` is a prompt where the user types to filter the choices. But in `handleAutoCompleteProcess` a plain `j` and a plain `k` were sent to the cursor (down and up) before the filter received them. Because of this, any query that contains one of these two letters was changed, and the user was not told about it:

| you type | real filter | real result |
| --- | --- | --- |
| `fetch` (control, no `j`/`k`) | `fetch` | `feat/jira-fetch-tool` ✅ |
| `jira` | `ira` | correct value, but only by luck, because `ira` still matches |
| `worktree` | `wortree` | "No matches". After `enter` the prompt beeps and then waits forever |
| `j` (with `feat/jira-fetch-tool` first in the list) | *no change* | **`main`**, a wrong value returned without any error |
| `kot` on the choices from the documentation | `ot` | `Kotlin`, again only by luck |

Two points are important here:

- The `worktree` row shows that this is not only about vi navigation. A normal query is also changed, only because it contains a `k`.
- The `j` row shows the worst case. The prompt returns a value the user did not choose, and there is no error.

The example inside the documentation comment of `autoComplete` has the same problem. Its three choices are TypeScript, Rust and **Kotlin**, and `Kotlin` is exactly the one you cannot type for.

## What this PR changes

`packages/effect/src/unstable/cli/Prompt.ts`, `handleAutoCompleteProcess`:

- The plain `j` and `k` cases are removed, so every printable character now reaches the filter.
- The `ctrl` branch is now a `switch`. Before, it only handled `Ctrl+U` and let all other ctrl keys beep. Now it also handles navigation:
  - `Ctrl+P` and `Ctrl+K` move the cursor up
  - `Ctrl+N` moves the cursor down
  - `Ctrl+U` clears the query, as before
  - everything else still beeps
- The arrow keys, `tab`, `backspace` and `enter` are not changed.
- The documentation comment of `autoComplete` now has a short **Details** section that names these keys, because the prompt itself does not show them anywhere.

`Ctrl+N` and `Ctrl+P` are the default keys in readline and in `fzf`, so this should feel familiar.

`Prompt.select` and `Prompt.multiSelect` are **not** changed. They have no text input, so a plain `j` or `k` cannot be understood in two ways there, and their vi style keys still work.

### Why not `Ctrl+J`

The normal vi pair would be `Ctrl+J` and `Ctrl+K`. But the `Terminal` layer cannot tell the difference between `Ctrl+J` and `Enter`. `Ctrl+J` is ASCII `0x0A`, and `NodeTerminal` uses `readline.emitKeypressEvents`, which reports it as `{ name: "enter", ctrl: false }`:

```
\x0a (ctrl+j) -> { name: "enter",  ctrl: false }   # the same as Enter
\x0b (ctrl+k) -> { name: "k",      ctrl: true  }
\x0d (enter)  -> { name: "return", ctrl: false }
\x0e (ctrl+n) -> { name: "n",      ctrl: true  }
\x10 (ctrl+p) -> { name: "p",      ctrl: true  }
```

This is why `Ctrl+N` is used for "down".

## Tests

Four new tests in `describe("Prompt.autoComplete", ...)` (`packages/effect/test/unstable/cli/Prompt.test.ts`). All four were written first and all four failed before the fix:

| test | how it failed before |
| --- | --- |
| keeps `j` and `k` in the filter query instead of moving the cursor | wrong filter |
| filters a query that merely contains a `k` | timeout, because no match means beep and then wait |
| moves the cursor with ctrl-n and ctrl-p | returned `alpha`, the keys did nothing |
| moves the cursor up with ctrl-k | returned `delta`, the key did nothing |

Results after the fix:

- `Prompt.test.ts`: 53 tests pass
- full CLI suite: 377 tests pass in 12 files
- full `effect` package: 8644 pass, 47 skipped, 0 failures
- `tsc -b` and lint are clean

## Behavior change for `j` and `k` users

Anyone who used `j` and `k` to move inside an `autoComplete` prompt will notice the difference. But that was the bug, so this is what needs to change. These users can now press `Ctrl+P` and `Ctrl+N`, or `Ctrl+K` for up, or use the arrow keys as before. `select` and `multiSelect` are not affected.

The changeset is a `patch`, which matches how the other changes in the current `4.0.0-rc` series are released.

### DIFF
--- a/.changeset/autocomplete-jk-filter.md
+++ b/.changeset/autocomplete-jk-filter.md
@@ -1,0 +1,7 @@
+---
+"effect": patch
+---
+
+Fix `Prompt.autoComplete` swallowing `j` and `k` while typing a filter query. Closes [#7392](https://github.com/Effect-TS/effect/issues/7392).
+
+The key dispatch bound bare `j`/`k` to cursor down/up before the filter input saw them, so any query containing either letter was silently corrupted — typing `jira` filtered on `ira`, `worktree` on `wortree` — and a single `j` could submit the wrong choice with no error. Every printable character now reaches the filter. Navigation moves to `Ctrl+P` / `Ctrl+N` (as in `readline` and `fzf`), with `Ctrl+K` also moving up; the arrow keys, `tab`, and `Ctrl+U` are unchanged. `Prompt.select` and `Prompt.multiSelect`, which have no text input, keep their vi-style `j`/`k` bindings.
--- a/.changeset/autocomplete-jk-filter.md
+++ b/.changeset/autocomplete-jk-filter.md
@@ -2,6 +2,4 @@
 "effect": patch
 ---
 
-Fix `Prompt.autoComplete` swallowing `j` and `k` while typing a filter query. Closes [#7392](https://github.com/Effect-TS/effect/issues/7392).
-
-The key dispatch bound bare `j`/`k` to cursor down/up before the filter input saw them, so any query containing either letter was silently corrupted — typing `jira` filtered on `ira`, `worktree` on `wortree` — and a single `j` could submit the wrong choice with no error. Every printable character now reaches the filter. Navigation moves to `Ctrl+P` / `Ctrl+N` (as in `readline` and `fzf`), with `Ctrl+K` also moving up; the arrow keys, `tab`, and `Ctrl+U` are unchanged. `Prompt.select` and `Prompt.multiSelect`, which have no text input, keep their vi-style `j`/`k` bindings.
+Fix `Prompt.autoComplete` swallowing `j` and `k` while typing a filter query.
--- a/packages/effect/src/unstable/cli/Prompt.ts
+++ b/packages/effect/src/unstable/cli/Prompt.ts
@@ -1241,6 +1241,12 @@ export const select = <const A>(options: SelectOptions<A>): Prompt<A> => {
 /**
  * Creates a prompt that lets users filter select choices by typing.
  *
+ * **Details**
+ *
+ * Every printable character is appended to the filter query, so navigation is
+ * bound to the arrow keys, `tab`, and the `Ctrl+P` / `Ctrl+N` chords used by
+ * `readline` and `fzf` (`Ctrl+K` also moves up). `Ctrl+U` clears the query.
+ *
  * **Example** (Filtering choices with autocomplete)
  *
  * ```ts import.meta.vitest
@@ -3558,18 +3564,29 @@ const handleSelectProcess = <A>(options: SelectOptionsReq<A>) => {
 
 const handleAutoCompleteProcess = <A>(options: AutoCompleteOptionsReq<A>) => {
   return (input: Terminal.UserInput, state: AutoCompleteState) => {
+    // Navigation is bound to ctrl chords rather than bare `j`/`k`, so that every
+    // printable character reaches the filter input.
     if (input.key.ctrl) {
-      if (input.key.name === "u") {
-        return processAutoCompleteClear(state, options)
+      switch (input.key.name) {
+        case "u": {
+          return processAutoCompleteClear(state, options)
+        }
+        case "p":
+        case "k": {
+          return processAutoCompleteCursorUp(state)
+        }
+        case "n": {
+          return processAutoCompleteCursorDown(state)
+        }
+        default: {
+          return Effect.succeed(Action.Beep())
+        }
       }
-      return Effect.succeed(Action.Beep())
     }
     switch (input.key.name) {
-      case "k":
       case "up": {
         return processAutoCompleteCursorUp(state)
       }
-      case "j":
       case "down": {
         return processAutoCompleteCursorDown(state)
       }
--- a/packages/effect/test/unstable/cli/Prompt.test.ts
+++ b/packages/effect/test/unstable/cli/Prompt.test.ts
@@ -679,6 +679,98 @@ describe("Prompt.autoComplete", () => {
 
       assert.isTrue(findFrame(frames, "No matches") !== undefined)
     }).pipe(Effect.provide(TestLayer)))
+
+  it.effect("keeps `j` and `k` in the filter query instead of moving the cursor", () =>
+    Effect.gen(function*() {
+      const prompt = Prompt.autoComplete({
+        message: "Pick a branch",
+        choices: [
+          { title: "feat/jira-fetch-tool", value: "feat/jira-fetch-tool" },
+          { title: "main", value: "main" },
+          { title: "feat/worktree-tool", value: "feat/worktree-tool" }
+        ]
+      })
+
+      yield* MockTerminal.inputText("jira")
+      yield* MockTerminal.inputKey("enter")
+
+      const result = yield* Prompt.run(prompt)
+      assert.strictEqual(result, "feat/jira-fetch-tool")
+
+      const output = yield* MockTerminal.displayLines
+      const frames = toFrames(output)
+      const filteredFrame = findFrame(frames, "[filter: jira]")
+
+      assert.isTrue(filteredFrame !== undefined)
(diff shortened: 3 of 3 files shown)


.changeset/autocomplete-jk-filter.md
```diff
@@ -0,0 +1,7 @@
+---
+"effect": patch
+---
+
+Fix `Prompt.autoComplete` swallowing `j` and `k` while typing a filter query. Closes [#7392](https://github.com/Effect-TS/effect/issues/7392).
+
+The key dispatch bound bare `j`/`k` to cursor down/up before the filter input saw them, so any query containing either letter was silently corrupted — typing `jira` filtered on `ira`, `worktree` on `wortree` — and a single `j` could submit the wrong choice with no error. Every printable character now reaches the filter. Navigation moves to `Ctrl+P` / `Ctrl+N` (as in `readline` and `fzf`), with `Ctrl+K` also moving up; the arrow keys, `tab`, and `Ctrl+U` are unchanged. `Prompt.select` and `Prompt.multiSelect`, which have no text input, keep their vi-style `j`/`k` bindings.
```
tim-smart:
```suggestion
Fix `Prompt.autoComplete` swallowing `j` and `k` while typing a filter query.
```

PR #7401, opened 2026-08-21
Fix EventLog remote write retries

### LINKED ISSUE #7400: EventLog pending remote writes are not retried on the existing backoff schedule after transient failures

4.0.0-rc.111 and main@436f10d1e

1. Create a local EventLog entry.
2. Register a fake remote whose first two `write` calls fail and subsequent writes succeed; on the affected base, those calls are the initial flush and the existing one-shot scope finalizer.
3. Wait until both failed attempts have occurred, then advance `TestClock` by 200 ms without creating another local entry.
4. Observe that `write` is not called a third time.

A minimal fake remote follows this shape:

```ts
const writes = yield* Queue.unbounded<void>()
let attempts = 0
const remote = EventLogRemote.EventLogRemote.of({
  id: EventJournal.makeRemoteIdUnsafe(),
  changes: () => Queue.unbounded<EventJournal.RemoteEntry, EventLogRemote.EventLogRemoteError>(),
  write: () => {
    const attempt = ++attempts
    return Queue.offer(writes, undefined).pipe(
      Effect.andThen(
        attempt <= 2
          ? Effect.fail(new EventLogRemote.EventLogRemoteError({ method: "write", cause: "offline" }))
          : Effect.void
      )
    )
  },
  whenAuthenticated: (effect) => effect
})

yield* log.write({ schema, event: "UserCreated", payload: { id: "user-1" } })
yield* registry.registerRemote(remote)
yield* Queue.take(writes)
yield* Queue.take(writes)
yield* TestClock.adjust("200 millis")
yield* Effect.yieldNow

// attempts is still 2. The base ran the initial flush and its one-shot
// scope finalizer, then terminated the worker without scheduling a third write.
```

Pending local entries should be retried after the remote becomes writable again, without requiring another local journal change. They should remain pending in the configured local journal while writes fail. Shutdown should interrupt scheduled retries and run only the existing one-shot final write.

### What do you see instead?

Pending entries remain stored in the configured local journal, but the remote becoming writable again does not itself flush them.

- An initial outgoing flush failure terminates the per-remote worker after its existing scope finalizer makes one immediate best-effort write. If that final write also fails, pending entries receive no further attempts.
- After a successful initial flush, a change-triggered failure is logged and the loop returns to waiting for another journal notification.
- Only the latter case is attempted again after a subsequent successful local journal write; a local write cannot restart a worker terminated by the initial failure.

### Additional information

The incoming remote stream already uses an exponential retry schedule capped at 10 seconds. The outgoing EventLog worker has no corresponding general retry policy for transient write failures; `EventLogRemote.makeWith` only retries `Forbidden` responses while re-authenticating.

A patch is ready that:

- applies the existing retry schedule to initial and change-triggered outgoing flushes;
- continuously drains journal notifications into a one-slot wake-up queue so a prolonged outage does not retain entry payloads in the unbounded PubSub subscription;
- keeps the shutdown finalizer as a one-shot write; and
- adds deterministic regression coverage for initial retries, later change-triggered retries, preserving a wake-up that arrives during an in-flight retry, and draining the journal PubSub subscription while a retry is blocked.

gjermundgaraba:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Retry transient EventLog remote write failures with the existing exponential backoff schedule so pending local entries synchronize after recovery.

The remote worker now drains journal notifications into a one-slot wake-up queue, preventing payload retention during prolonged outages while preserving a wake-up that arrives during an in-flight retry. The shutdown finalizer remains a one-shot write.

Regression coverage includes initial and change-triggered failures, in-flight wake preservation, and notification draining while retries are blocked.

## Validation

- `pnpm test --run packages/effect/test/unstable/eventlog/EventLog.test.ts`
- `pnpm lint-fix`
- `pnpm check`
- `git diff --check`

## Related

- Closes #7400


packages/effect/src/unstable/eventlog/EventLog.ts
```diff
@@ -874,11 +874,23 @@ const make = Effect.gen(function*() {
       )
 
       const write = journal.withRemoteUncommited(remote.id, (entries) => remote.write({ identity, entries, storeId }))
+      const writeUntilSuccess = write.pipe(
```
tim-smart, line 877: Maybe add a Effect.logDebug here so the error isn't completely swallowed.

packages/effect/src/unstable/eventlog/EventLog.ts
```diff
@@ -874,11 +874,23 @@ const make = Effect.gen(function*() {
...
+        Effect.retry(Schedule.min([
```
tim-smart: We can pull this schedule out now that it is used in two places.

tim-smart (changes requested):

packages/effect/src/unstable/eventlog/EventLog.ts
gjermundgaraba: missed that! added.

packages/effect/src/unstable/eventlog/EventLog.ts
gjermundgaraba: good call, done!

packages/effect/src/unstable/eventlog/EventLog.ts
tim-smart: I would use .tapCause so you get pretty printed errors

gjermundgaraba: done!

tim-smart (approved):

Issue #6831, opened 2026-08-01
Support OpenAI prompt_cache_breakpoint on Responses API input content


brandon-julio-t:
# Support OpenAI `prompt_cache_breakpoint` on Responses API input content

## Summary

`@effect/ai-openai` currently has no way to mark an explicit OpenAI prompt-cache breakpoint on a Responses API input content block.

OpenAI's GPT-5.6 prompt-caching API supports `prompt_cache_breakpoint` on `input_text`, `input_image`, and `input_file` blocks. This is distinct from `prompt_cache_key`, which is a request-level cache-routing control.

## Version checked

- `@effect/ai-openai@4.0.0-beta.102`
- `main` at [`c9b56ab`](https://github.com/Effect-TS/effect/commit/c9b56ab507f224426ee8388dc450da447ec4715f)
- OpenAI documentation checked on 2026-08-01

## Why this matters

GPT-5.6 and later cache exact prefixes at cache breakpoints. With the default implicit policy, the service places a breakpoint at the latest user or tool message. If a stable system prompt or document is followed by changing content, the implicit breakpoint can repeatedly select a changing prefix instead of the stable reusable prefix.

An explicit breakpoint lets an application mark the end of the stable prefix. Combined with `prompt_cache_options.mode: "explicit"`, it can also prevent writes for changing suffixes.

See OpenAI's current documentation for the behavior and supported blocks:

- [Caching behavior when migrating to GPT-5.6](https://developers.openai.com/api/docs/guides/prompt-caching#caching-behavior-changes-when-migrating-to-gpt-56)
- [Prompt cache breakpoints](https://developers.openai.com/api/docs/guides/prompt-caching#prompt-cache-breakpoints)

## Evidence in Effect

The high-level Responses content schemas have no breakpoint field:

- [`InputTextContent`, `InputImageContent`, and `InputFileContent`](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/ai/openai/src/OpenAiSchema.ts#L76-L94) contain the supported content data but no `prompt_cache_breakpoint`.
- [`InputContent` is the union of those three schemas](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/ai/openai/src/OpenAiSchema.ts#L96-L124).

The prompt translator constructs those blocks without a breakpoint:

[...]

Issue #6830, opened 2026-08-01
Expose OpenAI Responses safety and prompt-cache request options through OpenAiLanguageModel.Config


brandon-julio-t:
# Expose OpenAI Responses safety and prompt-cache request options through `OpenAiLanguageModel.Config`

## Summary

`@effect/ai-openai`'s high-level Responses language model cannot currently set several OpenAI request fields related to end-user safety identification and prompt caching.

In particular, `OpenAiLanguageModel.Config` does not expose:

- `safety_identifier`
- `prompt_cache_key`
- `prompt_cache_retention` for models before GPT-5.6
- `prompt_cache_options` for GPT-5.6 and later

## Version checked

- `@effect/ai-openai@4.0.0-beta.102`
- `main` at [`c9b56ab`](https://github.com/Effect-TS/effect/commit/c9b56ab507f224426ee8388dc450da447ec4715f)
- OpenAI documentation checked on 2026-08-01

## Evidence

`OpenAiLanguageModel.Config` derives its request fields from the handwritten `OpenAiSchema.CreateResponse` schema:

- [`OpenAiLanguageModel.Config` derives from `OpenAiSchema.CreateResponse.Encoded`](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/ai/openai/src/OpenAiLanguageModel.ts#L80-L120)
- [The handwritten `OpenAiSchema.CreateResponse` request schema](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/ai/openai/src/OpenAiSchema.ts#L655-L695) omits all four fields above.
- [The request builder forwards the remaining configuration into the OpenAI request](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/ai/openai/src/OpenAiLanguageModel.ts#L600-L630), so adding the fields to the high-level schema should make the request-level options representable and forwardable.

There is already evidence of schema drift inside the package:

- [The generated OpenAI `CreateResponse` request type includes `safety_identifier`, `prompt_cache_key`, and `prompt_cache_retention`](https://github.com/Effect-TS/effect/blob/c9b56ab507f224426ee8388dc450da447ec4715f/packages/ai/openai/src/Generated.ts#L25752-L25796).
[...]

PR #7403, opened 2026-08-22
`ai/openai`: Minimal patch to add cache breakpoint support for openai-gpt-5.6+ prompts

hsyntax:
## Summary

GPT 5.6 and later models add support for [explicit cache breakpoints](https://developers.openai.com/api/docs/guides/prompt-caching#how-caching-differs-by-model). 

This PR attempts to add a minimal patch to add support for cache breakpoints. Source reference : [here](https://developers.openai.com/api/docs/guides/prompt-caching#add-explicit-cache-breakpoints)

## Existing work

These issues and PR's already existed related to the topic. They seemed broader and unlikely to be merged and there hasn't been any progress on them in a while. 

I made this patch to serve as a minimal focused change since this is currently a blocker for me and my team.

### Issues

- #6830 — expose OpenAI Responses safety and prompt-cache request options
- #6831 — support prompt_cache_breakpoint on Responses API input content

### Pull requests

- #6832 — expose OpenAI prompt cache request options
- #6833 — add OpenAI prompt cache breakpoints
- #6678 — update AI codegen generated files with related generated-schema support

_This PR was raised with support from Codex/GPT 5.6-xhigh_

### DIFF
--- a/packages/ai/openai/src/OpenAiLanguageModel.ts
+++ b/packages/ai/openai/src/OpenAiLanguageModel.ts
@@ -142,6 +142,9 @@ declare module "effect/unstable/ai/Prompt" {
     readonly openai?: {
       /**
        * Marks the system input text as the end of a reusable prompt prefix.
+       *
+       * Requires GPT-5.6 or later. OpenAI may reject requests that use this
+       * option with earlier models.
        */
       readonly promptCacheBreakpoint?: PromptCacheBreakpoint | null
     } | null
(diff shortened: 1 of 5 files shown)


packages/ai/openai/src/OpenAiLanguageModel.ts
```diff
@@ -2910,6 +2954,10 @@ const getEncryptedContent = (
 
 const getImageDetail = (part: Prompt.FilePart): ImageDetail => part.options.openai?.imageDetail ?? "auto"
 
+const getPromptCacheBreakpoint = (
```
IMax153, line 2951: If we return `{ prompt_cache_breakpoint: ... } | undefined` from this method, we could just spread the getter into the content items instead of getting the breakpoint and checking if it's `null`.

IMax153 (changes requested):

packages/ai/openai/src/OpenAiLanguageModel.ts
hsyntax: Done

IMax153 (approved):

PR #7408, opened 2026-08-22
Fix Bun per-message websocket compression

phibr0:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Enables outgoing message compression for WebSockets in `@effect/platform-bun`.

Although `perMessageDeflate` can be enabled on the Bun server since https://github.com/Effect-TS/effect/pull/6691, outgoing messages are still sent without Bun's compression flag. As a result, clients can successfully negotiate `permessage-deflate`, but messages sent by the server are still not compressed.

The test is rather complex because we need low-level access to the raw "wire data", since compression is fully abstracted away otherwise.

## Related

Follow-up to https://github.com/Effect-TS/effect/pull/6691

### DIFF
--- a/packages/platform/bun/src/BunHttpServer.ts
+++ b/packages/platform/bun/src/BunHttpServer.ts
@@ -77,13 +77,27 @@ export type ServeOptions<R extends string> =
  * through, e.g.
  * `BunHttpServer.layer({ port: 3000, websocket: { perMessageDeflate: true } })`.
  *
+ * The `compressionThreshold` option controls the minimum message size in bytes
+ * that is compressed when per-message deflate is negotiated. It defaults to
+ * 1024, matching the default threshold of Node's `ws` server.
+ *
  * @category options
  * @since 4.0.0
  */
-export type WebSocketOptions = Omit<
-  Bun.WebSocketHandler<WebSocketContext>,
-  "open" | "message" | "close" | "drain" | "ping" | "pong" | "data" | "binaryType"
->
+export type WebSocketOptions =
+  & Omit<
+    Bun.WebSocketHandler<WebSocketContext>,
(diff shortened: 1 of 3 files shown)


packages/platform/bun/src/BunHttpServer.ts
```diff
@@ -565,11 +565,11 @@ class BunServerRequest extends Inspectable.Class implements ServerRequest.HttpSe
         const write = (chunk: Uint8Array | string | Socket.CloseEvent) =>
           Effect.sync(() => {
             if (typeof chunk === "string") {
-              ws.sendText(chunk)
+              ws.sendText(chunk, true)
```
tim-smart: Would this override `perMessageDeflate: { compress: false }`?

tim-smart: Also will this still compress small messages? Usually you want a 1kb threshold or so.

tim-smart (changes requested):

packages/platform/bun/src/BunHttpServer.ts
phibr0: > Would this override perMessageDeflate: { compress: false }?

Nope, if that is set but `compress:` is unset or false it doesn't do anything. The second test verifies this.

> Also will this still compress small messages? Usually you want a 1kb threshold or so.

Yes, I thought I checked the node behavior but apparently I forgot about that, fixed now.

The default is 1024 in both.

```ts
NodeHttpServer.layer({ port: 3000 }, {
  websocket: {
    perMessageDeflate: { threshold: 4096 }
  }
})

BunHttpServer.layer({
  port: 3000,
  websocket: {
    perMessageDeflate: true,
    compressionThreshold: 4096
  }
})
```

tim-smart (approved):

PR #7411, opened 2026-08-22
Add optional acquisition APIs to RcMap and LayerMap

altendky:
## Type

Feature

## Description

Adds `RcMap.getOption` with data-first and data-last overloads for atomically retaining an entry only when it is already cached. A missing key returns `Option.none` without invoking lookup or performing capacity acquisition. Existing ready or in-flight entries use the normal reference count and caller-scope finalizer, cached failures propagate unchanged, and closed maps interrupt.

This avoids the race inherent in composing `has` and `get`, where the entry can disappear between operations and `get` can start a new lookup.

Adds `LayerMap.contextEffectOption` as the direct optional counterpart to `contextEffect` on both `LayerMap` instances and generated `LayerMap.Service` classes. `Option.none` means the context is not currently cached on an open map, and missing entries do not build a layer.

The change does not modify existing release, TTL, invalidation, or capacity behavior. It includes API lifecycle coverage for missing entries, ready retention and release, in-flight sharing, cached failures, TTL retention and expiry, invalidation linearization, closed-map interruption, and direct and Service `LayerMap` accessors. It also includes an `effect` patch changeset.

Validation:

- `pnpm lint-fix`
- 21 targeted `RcMap` and `LayerMap` tests passed
- 11 `RcMap` and `LayerMap` doctests passed
- `pnpm check`
- `git diff --check`
- `pnpm jsdocs --check` reports only pre-existing unrelated repository diagnostics; neither changed source module is reported

## Related

No related issue.


packages/effect/src/RcMap.ts
```diff
@@ -381,6 +382,76 @@ export const get: {
...
+ */
+export const getOption: {
+  <K>(key: K): <A, E>(self: RcMap<K, A, E>) => Effect.Effect<Option.Option<A>, E, Scope.Scope>
+  <K, A, E>(self: RcMap<K, A, E>, key: K): Effect.Effect<Option.Option<A>, E, Scope.Scope>
+} = dual(
+  2,
+  <K, A, E>(self: RcMap<K, A, E>, key: K): Effect.Effect<Option.Option<A>, E, Scope.Scope> =>
+    Effect.uninterruptibleMask((restore) => {
+      if (self.state._tag === "Closed") {
+        return Effect.interrupt
```
tim-smart: I think for this case we return None, as no acquisition can happen with this api.

tim-smart (changes requested):

packages/effect/src/RcMap.ts
altendky: updated.  thanks for taking a look and catching this.  i appreciate it.

tim-smart (approved):

PR #7421, opened 2026-08-23
Update repository dependencies

tim-smart:
## Summary

- update repository dependencies while leaving peer dependency ranges unchanged
- migrate the D1 test harness to Miniflare 5's v4 option adapter
- add a patch changeset for packages with production dependency updates

TypeScript compiler API consumers remain on 6.0.3 because the TypeScript 7 package no longer exports that API.

## Validation

- `pnpm check`
- `pnpm lint`
- `pnpm test --run` (392 files, 10,252 passed)
- `pnpm build`
- `pnpm exec changeset status --since origin/main --verbose`

Closes EFF-837


packages/sql/d1/package.json
```diff
@@ -66,12 +66,12 @@
   },
   "devDependencies": {
     "effect": "workspace:^",
-    "miniflare": "^4.20260730.0"
+    "miniflare": "5.20260820.0-alpha"
```
tim-smart: Lets stay on non-alpha versions

tim-smart: Updated to the latest non-alpha release, `4.20260730.0`, and restored the native v4 options. The D1 harness passes all 17 tests.

PR #7424, opened 2026-08-24
Skip empty remote journal write callbacks

tim-smart:
## Summary

- skip `withRemoteUncommited` callbacks when no entries are pending
- preserve callback results for non-empty memory, IndexedDB, and SQL journals
- cover empty and non-empty callback behavior in the memory and SQLite tests

## Validation

- `pnpm test --run packages/effect/test/unstable/eventlog/EventJournal.test.ts`
- `pnpm test --run packages/sql/sqlite-node/test/SqlEventJournal.test.ts`
- `pnpm lint-fix`
- `pnpm check`
- `git diff --check`

Closes EFF-851

### DIFF
--- a/packages/effect/src/unstable/eventlog/EventJournal.ts
+++ b/packages/effect/src/unstable/eventlog/EventJournal.ts
@@ -11,6 +11,7 @@
  *
  * @since 4.0.0
  */
+import * as Arr from "../../Array.ts"
 import type { Brand } from "../../Brand.ts"
 import * as Context from "../../Context.ts"
 import * as Data from "../../Data.ts"
@@ -18,6 +19,7 @@ import * as DateTime from "../../DateTime.ts"
 import * as Effect from "../../Effect.ts"
 import * as Uuid from "../../internal/uuid.ts"
 import * as Layer from "../../Layer.ts"
+import type { Option } from "../../Option.ts"
 import * as Order from "../../Order.ts"
 import * as PubSub from "../../PubSub.ts"
 import * as Schema from "../../Schema.ts"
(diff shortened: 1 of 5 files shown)


packages/effect/src/unstable/eventlog/EventJournal.ts
```diff
@@ -78,12 +78,14 @@ export class EventJournal extends Context.Service<EventJournal, {
...
 
   /**
-   * Return the uncommitted entries for a remote source.
+   * Run an effect with the uncommitted entries for a remote source.
+   *
+   * The effect is not run when there are no uncommitted entries.
    */
   readonly withRemoteUncommited: <A, E, R>(
     remoteId: RemoteId,
     f: (entries: ReadonlyArray<Entry>) => Effect.Effect<A, E, R>
```
tim-smart: This could be a NonEmptyReadonlyArray now.

packages/effect/src/unstable/eventlog/EventJournal.ts
```diff
@@ -78,12 +78,14 @@ export class EventJournal extends Context.Service<EventJournal, {
...
-  ) => Effect.Effect<A, EventJournalError | E, R>
+  ) => Effect.Effect<A | void, EventJournalError | E, R>
```
tim-smart: I think Option<A> would be better

packages/effect/src/unstable/eventlog/EventJournal.ts
tim-smart: Updated the callback parameter to `Arr.NonEmptyReadonlyArray<Entry>` and narrowed each backend with the array guard.

packages/effect/src/unstable/eventlog/EventJournal.ts
tim-smart: Updated the result to `Option<A>`. Empty paths return `None`, and callback results are wrapped in `Some`.

PR #7426, opened 2026-08-24
Replace pg with a native PgProtocol client

tim-smart:
Closes EFF-880
Closes EFF-852
Closes EFF-858
Closes EFF-859
Closes EFF-870
Closes EFF-871
Closes EFF-872

Shared PR for all four native sql-pg client slices. Later slices push to this branch (`eff-854/pg-connection-startup`). Do not merge until slice 4 (PgClient switchover and `pg*` removal) is on the PR.

## Slice 1: connection startup

- `PgConnection.make` connects over TCP, Unix sockets, or a caller-supplied `Duplex`.
- Supports libpq URLs, optional TLS through `SSLRequest`, and trust, cleartext, MD5, or SCRAM-SHA-256 authentication.
- Resolves at `ReadyForQuery`; scope release sends `Terminate` and ends the socket after the frame flushes.
- Keeps `processId` public and the cancel secret private.

## Slice 2: unnamed extended query

- `query` and `queryValues` send `Parse` / `Bind` / `Describe` / `Execute` / `Sync` in one socket write and use binary parameters and result fields.
- One statement runs at a time per connection.
- JavaScript values infer OIDs; branded `PgTypes.Parameter` values override inference.
- `PgTypes.makeRegistry()` supplies isolated custom codecs and optional generic one-dimensional array codecs.
- Ordinary query errors drain through `ReadyForQuery`; parser failures, unexpected `COPY` messages, and protocol desync destroy the socket.
- Native results include command, row count, OID, fields, and object rows. `queryValues` preserves `RowDescription` order.

## Tests

- Live testcontainers coverage for startup, authentication failure, TLS refusal, SELECT, binary parameters, positional results, and connection recovery after a `23505` unique violation.
- In-process protocol tests for fragmented startup responses and connection edge cases.
- Unit coverage for URL validation, parameter branding, registry isolation, and custom array codecs.

## Performance

`benchmark/PgClient.ts` against one long-lived PostgreSQL container, client pinned to one CPU, alternating pairs,
median of the per-pair deltas. `main` still runs on `pg`.

| Workload | `main` | This branch | With `multiplex` |
|---|---:|---:|---:|
| parameterized SELECT (1 row) | 22,589 q/s | 28,297 (+25%) | |
| generate_series (100 rows x 3 columns) | 1,082k rows/s | 1,662k (+54%) | |
| generate_series (100 rows x 20 columns) | 343k rows/s | 571k (+67%) | |
| transaction (1 INSERT) | 9,011 tx/s | 9,395 (+4%) | |
| 20 concurrent parameterized SELECTs | 77,741 q/s | 82,276 (+6%) | 123,309 (+61%) |

The changes:

- **Warmest connection first.** `Pool` released an item to the back of its availability list and borrowers took from
  the front, so a run of sequential borrows rotated over every connection the pool had opened and none was ever the
  hot one - a hundred sequential transactions landed on exactly ten backends, ten each, where node-postgres sent all
  hundred through one. A released item now goes to the front. Freshly acquired ones still go to the back, having no
  use behind them; under saturation the two orders agree.
- **`noDelay` on the session socket.** It had been set on the short-lived `CancelRequest` connection instead of the
  socket that carries every statement. Loopback hides this - a statement is one write and then a wait, so Nagle has
  nothing to hold anything behind - but across a virtual network it costs: on macOS through OrbStack the one-row query
  went from 4.6% behind `main` to 5.6% ahead.
- **Prepared transaction control.** `BEGIN`, `COMMIT`, `ROLLBACK` and the savepoint pair went through
  `executeUnprepared`, so every transaction paid to parse them again. `SqlClient.MakeOptions` now takes
  `prepareTransactionControls`, which PostgreSQL sets and which is off for everyone else, since several databases
  refuse to prepare transaction control at all. Transactions were the one workload behind `main`, by 1.8%; they are
  now at parity.
- **Borrowed connections.** A statement that finishes with the effect that runs it opened a scope, leased a
  connection into it and closed it again. `SqlClient.MakeOptions` now takes an optional `borrower`, and `PgPool.use`
  lends a session for the duration of one effect through `Pool.use`. `stream`, transactions, and `reserve` keep the
  acquirer, whose lease outlives the effect that starts it; a driver that offers no borrower is unchanged. This is
  most of the concurrent workload's gain.
- **One permit per statement.** An ordinary statement entered both the ownership and the wire semaphore, but the wire
  permit is redundant once ownership is held: the other holders reach the session through ownership too, and a stream
  releases the wire before its pin. Pinned callers keep it, since it is what serializes the fibers inside a
  transaction. A permit is a scoped acquisition, and a statement's cost is dominated by how many Effect operations it
  runs.
- **Statement state.** A statement's state machine is one object with its methods on the prototype, rather than the
  seven closures plus a holder that a closure-based machine rebuilt per statement, and a query only builds the result
  shape its caller asked for. The garbage collector was the largest remaining cost on the query path; this halves its
  share of the concurrent run.
- **Row construction.** A `RowDescription` builds its row constructor once, and assignment stores the same property
  descriptor as `Object.defineProperty` an order of magnitude faster; only a `__proto__` column takes the slow
  spelling. `Object.defineProperty` was 18% of all samples in a CPU profile of the 100-row workload.
- **`PgProtocol.Parser.pushEach`.** `readField` was documented as settable from a `RowDescription`, but `push` finished
  the whole socket chunk first, so the reader never arrived before the rows it was meant to decode. Messages are now
  handed over as they are decoded, and a column costs neither a view over the parser buffer nor a second pass.
- **Prepared statements.** A repeated statement is kept under a backend name and costs `Bind`/`Execute`/`Sync` with no
  parsing, planning, or `RowDescription`. Per connection, keyed on SQL text plus inferred parameter OIDs, bounded by
  `preparedStatementCacheSize`, evicting least-recently-used and closing what it evicts. Recovers on its own from a
  name the backend has lost (`26000`) or a plan that no longer matches its columns (`0A000`). `prepare: false` opts
  out.
- **Pipelining.** A `multiplex` session writes the statements its fibers submit in the same tick as one batch and
  routes replies to the oldest cycle on the wire. Every cycle keeps its own `Sync`, so a rejected statement only skips
  its own cycle and nothing needs replay. Pinned work still owns the session, and `pin` waits for the pipeline to
  drain.
- **Spreading.** A multiplexed pool bounds how many statements share a connection. The backend runs a connection's
  pipelined statements in receipt order, so unlimited sharing meant the pool never opened a second connection and one
  slow statement held up everything in flight behind it. Bounding it costs no throughput and gains some.
- Frame tail and bind encoder built once instead of per statement, frames assembled in a pooled `Buffer`, a query
  assembling only the row shape its caller asked for, a pool checkout that only loops when a connection has to be
  replaced, and command tags split by reading their two spaces.

### Also in this PR

Two `Pool` liveness fixes, both found through this work:

- A lease was returned to the availability list only on the exact transition out of saturation, so several leases
  released together resumed one waiter and left the rest parked on an item that had room for them, with no later
  transition to wake them. This is what forced `PgPool` to give a multiplexed pool a concurrency of 2^20, and fixing
  it is what made bounded sharing possible.
- Invalidating an item that still had live leases removed it from circulation without topping the pool back up, so
  anybody already queued waited for an item that was never coming back.

And three in `sql-pg`:

- A statement the server refused to parse - a typo, a missing table - was treated as a fatal protocol error and
  destroyed the connection. The spec's fatal case is a failure of our own parser, not a SQL one. It is now an
  ordinary query error that fails only that statement.
- `stream` reaches a connection through the ordinary checkout and pins it itself, so a multiplexed pool kept handing
  the same connection out and a stream consumer running another statement could be given the connection its own
  stream was holding. Every route to an exclusive session goes through `PgConnection.pin`, so the reservation lives
  there now.
- `PgPool.invalidate` threw on a reserved connection: the pinned view copied its internals record from a base that
  had not assigned it yet.
- A connection dying while it was held left a queued checkout waiting forever, because the dead connection was only
  recorded for the *next* checkout to notice.

### Verified off Linux

Measured on an Apple M5 against the same `main` baseline, five interleaved pairs, PostgreSQL reached across an
OrbStack VM: one-row +5.7%, 100x3 +32.7%, 100x20 +50.6%, transaction +8.1%, 20 concurrent +10.3%. Two defects only
that environment could surface are fixed above - `noDelay` on the wrong socket, and the pool's borrow order.

### Tuning across a network

`multiplexConcurrency` sets how many statements share one connection. The default is low on purpose: pipelined
statements queue behind the slowest of them, and against a server a cheap round trip away, spreading over more
connections costs no throughput and buys tail latency. The further away the server, the more a round trip costs and
the more a deeper pipeline is worth - measured on macOS across an OrbStack VM, 32 was 21% faster than the default, and
16 was 7% faster even against a native server.

### Why `multiplex` is not the default

Pipelined statements execute in the order the backend received them, so sharing a connection trades tail latency for
throughput. With ten connections and a 500ms statement in flight, a non-multiplexed pool delays none of ten concurrent
fast statements; a multiplexed one delays three. Unbounded sharing delayed all ten, which is what the spreading change
fixes, but three is still not zero.


packages/sql/pg/src/PgPool.ts
```diff
@@ -0,0 +1,172 @@
...
+  Effect.gen(function*() {
+    const multiplex = options.multiplex ?? false
+    const connectionTTL = options.connectionTTL !== undefined
+      ? Duration.toMillis(Duration.fromInputUnsafe(options.connectionTTL))
+      : undefined
+    const deadConnections = new Set<PgConnection.PgConnection>()
+    const createdAt = new WeakMap<PgConnection.PgConnection, number>()
+
+    const acquire = Effect.tap(PgConnection.make(options), (connection) =>
+      Effect.map(Clock.currentTimeMillis, (now) => {
```
tim-smart: The Clock access can be lifted `const clock = yield* Clock.Clock` and then you can use currentTimeMillisUnsafe

tim-smart: Addressed in 803774e81. `PgPool.make` now acquires `Clock.Clock` once and uses `currentTimeMillisUnsafe()` for both timestamp reads.

packages/effect/src/unstable/persistence/Persistence.ts
```diff
@@ -394,7 +402,9 @@ export const layerBackingSqlMultiTable: Layer.Layer<
         get: (key) =>
           sql<
             { value: string }
-          >`SELECT value FROM ${table} WHERE id = ${key} AND (expires IS NULL OR expires > ${clock.currentTimeMillisUnsafe()})`
+          >`SELECT value FROM ${table} WHERE id = ${key} AND (expires IS NULL OR expires > ${
+            timestampParameter(clock.currentTimeMillisUnsafe())
```
tim-smart: Why do we need `timestampParameter`? Passing a number should work too.

packages/effect/src/unstable/cluster/SqlMessageStorage.ts
```diff
@@ -125,7 +130,7 @@ export const makeEncoded: (options?: {
     switch (envelope._tag) {
       case "Request":
         return {
-          id: envelope.requestId,
+          id: snowflakeParameter(envelope.requestId),
```
tim-smart: Passing a string here should work with the update pg driver.

packages/sql/pg/src/PgConnection.ts
```diff
@@ -0,0 +1,2260 @@
...
+    })
+  })
+
+const resolveConfigUnsafe = (options: Config): ResolvedConfig => {
+  const url = options.url !== undefined ? parseUrl(Redacted.value(options.url)) : {}
+  const host = options.host ?? url.host ?? "localhost"
+  const port = options.port ?? url.port ?? 5432
+  const username = options.username ?? url.username ?? process.env.USER ?? process.env.USERNAME
+  if (username === undefined) {
+    throw configError("No username configured")
```
tim-smart: This should return an effect instead of throwing. Same with the url functions below

packages/sql/pg/src/PgPool.ts
```diff
@@ -0,0 +1,218 @@
...
+ * Creates a pool of PostgreSQL sessions.
+ *
+ * Connections are opened lazily up to `maxConnections` and released down to
+ * `minConnections` after `idleTimeout` without use. Closing the scope shuts
+ * the pool down and releases every session.
+ *
+ * @category constructors
+ * @since 4.0.0
+ */
+export const make = (options: Config): Effect.Effect<PgPool, SqlError, Scope.Scope> =>
```
tim-smart: This can use Effect.fnUntraced

packages/effect/src/unstable/persistence/Persistence.ts
tim-smart: Removed in 38dae15dd. Parameter inference now widens numbers: safe integers beyond the int4 range bind as int8, so a millisecond timestamp binds without the shim and Persistence.ts is back to plain values.

packages/effect/src/unstable/cluster/SqlMessageStorage.ts
tim-smart: It does now. Strings bind untyped in the text format and the backend derives the type from the statement, so a snowflake string works against a bigint column; the shim is reverted in 38dae15dd. Making this work also surfaced a prepared-cache bug (a statement whose first cycle failed at bind kept a poisoned entry and re-parsed under the same backend name, 42P05); fixed in the same commit with a regression test.

packages/sql/pg/src/PgConnection.ts
tim-smart: Done in 38dae15dd: resolveConfig returns Effect<ResolvedConfig, SqlError>, and parseUrl / decodeComponent / parsePort return Result instead of throwing.

packages/sql/pg/src/PgPool.ts
tim-smart: Done in 38dae15dd.

PR #7427, opened 2026-08-24
Default TCP cluster serialization to SchemaBinary

tim-smart:
Change the Bun, Deno, and Node TCP cluster layers to use SchemaBinary when no serialization option is supplied. Explicit MessagePack and NDJSON selections remain unchanged.

SchemaBinary already rejects declared frames above 16 MiB before buffering their bodies. This change also forwards the cluster `serializationMaxBufferSize` option to that frame limit and supports `"unbounded"`, matching the configuration available to the other streaming serializers.

Cluster payloads use the SchemaBinary direct encoding on the wire. The conflict with JSON-only persistence is resolved at the storage boundary instead of lowering payloads to JSON: `MessageStorage` reply handlers now deliver the reply exactly as it was persisted. When a reply cannot be encoded for JSON storage and falls back to a serializable defect, waiting callers receive that same defect rather than a transport-encoded copy of the original value, so live replies and stored replies can no longer diverge.

Benchmarking the Runner server/client over TCP (see EFF-857) exposed a performance defect: the cluster message paths request `codecFor(payloadSchema)` per message, and deriving a SchemaBinary codec costs about 15 us versus 0.4 us for reuse on a small schema. SchemaBinary now memoizes its regular and direct codec constructors by schema identity and wire mode, rather than caching at the RPC layer. After the fix, SchemaBinary beats msgpack on throughput and latency for small, medium, and large workloads while producing 5-50% smaller frames.

SchemaBinary envelope frames are independently decodable, since encoded frames can outlive a single TCP connection.

Regression coverage verifies the SchemaBinary default, frame limits, parser replacement, codec memoization, persisted reply fallback delivery, persisted cluster workflows, and an explicit MessagePack override.

Tests:

- `nix develop -c env EFFECT_CLUSTER_TESTS=1 pnpm test-cluster`: 107/107 passed on the final run. Two earlier runs each lost a cron timing test to a shard-stabilization deadline; each failed case passed immediately when rerun alone.
- `nix develop -c deno task test --run packages/platform/deno/test/cluster/SocketRunner.test.ts`: 2/2 passed.
- `nix develop -c pnpm vitest run packages/effect/test/cluster`: 111/111 passed, including a new MessageStorage regression test for the persisted defect fallback.
- `nix develop -c pnpm vitest run packages/effect/test/unstable/encoding/SchemaBinary.test.ts`: 174/174 passed.
- `nix develop -c pnpm vitest run packages/effect/test/rpc/RpcSerialization.test.ts`: 39/39 passed.
- `nix develop -c pnpm vitest run packages/platform/node/test/cluster/SocketRunner.test.ts`: 5/5 passed.
- `nix develop -c pnpm check`

Closes EFF-857

### DIFF
--- a/packages/effect/test/unstable/encoding/SchemaBinary.test.ts
+++ b/packages/effect/test/unstable/encoding/SchemaBinary.test.ts
@@ -23,6 +23,7 @@ import {
   SchemaTransformation,
   Stream
 } from "effect"
+import * as FastCheck from "effect/testing/FastCheck"
 import * as SchemaBinary from "effect/unstable/encoding/SchemaBinary"
 
 const encode = <A, I>(schema: Schema.Codec<A, I>, value: A): Uint8Array<ArrayBuffer> =>
@@ -457,6 +458,15 @@ describe("SchemaBinary", () => {
         schemaError(() => Schema.encodeUnknownSync(codec, { disableChecks: true })(1.5)).message,
         /an integer/
       )
+    })
+
+    it("rejects integer varints outside the safe-number range", () => {
+      const codec = SchemaBinary.toCodec(Schema.Int)
+      // Sign-magnitude encoding of Number.MAX_SAFE_INTEGER + 2. Converting it
+      // to number would silently round it down to MAX_SAFE_INTEGER + 1.
+      const unsafe = Uint8Array.of(9, 0x20, 0x82, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x20)
+
+      assert.match(schemaError(() => Schema.decodeUnknownSync(codec)(unsafe)).message, /integer/)
     })
 
     it("rejects malformed number payloads", () => {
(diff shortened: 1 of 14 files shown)


packages/effect/test/unstable/encoding/SchemaBinary.test.ts
```diff
@@ -2333,6 +2343,138 @@ describe("SchemaBinary", () => {
...
+      items: Schema.Array(Schema.Struct({ id: Schema.Int, name: Schema.String })),
+      attributes: Schema.Record(
+        Schema.String,
+        Schema.Union([Schema.String, Schema.Number, Schema.Boolean, Schema.Null])
+      ),
+      event: Event
+    })
+    const generated = Schema.toArbitrary(Generated)(FastCheck)
+
+    it("keeps every encoding entry point equivalent for generated nested values", () => {
```
tim-smart: Instead of using FastCheck directly, prefer using `it.prop` with a Schema directly. FastCheck will be removed from effect soon.

tim-smart: Updated in 4d64ba370. The generated regressions now use Schema-backed `it.live.prop` properties, preserve their run counts and generation bounds, and no longer import or invoke FastCheck directly. The focused suite, lint-fix, and check all pass.

PR #7442, opened 2026-08-24
Redact values in CLI wizard command output

### LINKED ISSUE #7422: CLI wizard mode prints values entered at masked password prompts (redacted flags) in plaintext

effect@4.0.0-rc.111

```ts
import { BunRuntime, BunServices } from "@effect/platform-bun"
import { Command, Flag } from "effect/unstable/cli"

Command.make("demo", { password: Flag.redacted("password") }).pipe(
  Command.withHandler(() => Effect.void),
  Command.run({ version: "0.0.0" }),
  Effect.provide(BunServices.layer),
  BunRuntime.runMain,
)
```

Run it with `--wizard` and type a secret at the Password prompt.

The Redacted primitive prompts through `Prompt.password` but immediately unwraps the result into the displayed command line (`packages/effect/src/unstable/cli/internal/wizard.ts` ~L193-194, `case "Redacted": return Effect.map(Prompt.run(Prompt.password({ message })), Redacted.value)`). The assembled line is echoed after each wizard section (`logCurrentCommand`, wizard.ts ~L66/76/99) and again in the completion block (`Console.log(Wizard.renderCompletion(wizardArgs))` in `Command.ts`).

Input the prompt masks on entry stays masked (or omitted) in the echoed "Current command" / "Command ready" blocks.

### What do you see instead?

The prompt masks the input, then the completion block prints it in plaintext (and with more params the intermediate "Current command" blocks repeat it), leaving the secret in terminal scrollback:

```
? Password (--password) › **************
✔ Password (--password) … **************

Command ready
  $ demo --password hunter2-secret

? Run this command? › yes / no
```

### Additional information

Doesnt seem a v3 to v4 regression since v3's `@effect/cli` wizard had the same unmask-then-echo at its single "Wizard Mode Complete!" message (`packages/cli/src/internal/primitive.ts` `case "Redacted"`/`"Secret"`  -> hidden text prompt; `packages/cli/src/internal/cliApp.ts` `renderWizardArgs`).

v4's reintroduced wizard amplifies the exposure by re-printing the full command line after every prompt step rather than once at the end. Present unchanged on current `main`.

tim-smart:
## Summary

- keep separate execution and display values for wizard-generated arguments
- render redacted prompt values as `<redacted>` in intermediate and final command blocks
- cover both command displays while verifying the handler still receives the actual value

## Testing

- `nix develop -c pnpm test --run packages/effect/test/unstable/cli/Command.test.ts`
- `nix develop -c pnpm check`

Closes EFF-877
Closes #7422

### DIFF
--- a/packages/effect/src/unstable/cli/internal/wizard.ts
+++ b/packages/effect/src/unstable/cli/internal/wizard.ts
@@ -16,17 +16,30 @@ export interface Options {
   readonly prefix?: ReadonlyArray<string> | undefined
 }
 
+export interface Result {
+  readonly args: Array<string>
+  readonly displayArgs: Array<string>
+}
+
+interface CommandLineArg {
+  readonly value: string
+  readonly displayValue: string
+}
+
 export const run: (
   command: Command.Command.Any,
   options?: Options
-) => Effect.Effect<Array<string>, CliError.CliError | Terminal.QuitError, Command.Environment> = Effect.fnUntraced(
+) => Effect.Effect<Result, CliError.CliError | Terminal.QuitError, Command.Environment> = Effect.fnUntraced(
   function*(command, options) {
     const commandPath = options?.commandPath ?? [command.name]
     const selected = getCommandAtPath(command, commandPath)
-    const commandLine = [...(options?.prefix ?? commandPath)]
+    const commandLine = (options?.prefix ?? commandPath).map((value) => commandLineArg(value))
     yield* logCurrentCommand(commandLine)
     yield* promptCommand(selected, commandLine, selected === command ? "ROOT" : selected.name)
-    return commandLine
+    return {
+      args: commandLine.map((arg) => arg.value),
+      displayArgs: commandLine.map((arg) => arg.displayValue)
+    }
   }
 )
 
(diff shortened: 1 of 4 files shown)


packages/effect/src/unstable/cli/internal/wizard.ts
```diff
@@ -109,57 +122,58 @@ const hasWizardSteps = (command: Command.Command.Any): boolean => {
...
-        message: `${renderParamLabel(single)} count`,
-        default: Option.getOrElse(metadata.variadicMin, () => 0),
-        min: Option.getOrElse(metadata.variadicMin, () => 0),
-        ...(Option.isSome(metadata.variadicMax) ? { max: metadata.variadicMax.value } : {})
-      }))
-    const values: Array<string> = []
-    for (let i = 0; i < count; i++) {
-      values.push(yield* promptSingle(single))
-    }
+) => Effect.Effect<Array<CommandLineArg>, CliError.CliError | Terminal.QuitError, Command.Environment> = Effect
```
tim-smart: Remove the type annotation and use `Effect.fn.Return` to improve the code formatting

tim-smart: Updated in 1f24f13fb: `promptParam` now uses `Effect.fn.Return` on the generator, with the explicit function return annotation removed. Formatting, the focused command tests, and `pnpm check` pass.

PR #7443, opened 2026-08-24
Terminate `Stream.fromEventListener` after one item if `once: true`

youngspe:
## Type

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently, the stream returned by `Stream.fromEventListener()` does not terminate after the first item when `once` is set to `true` in the `options` parameter. Though it's not explicitly documented I believe this is a bug because after the first item, the stream will never emit additional items, so you would expect it to end.

This change applies `take(1)` to the stream returned by `Stream.fromEventListener` if `options.once` is `true`.

I've also added a basic test for `fromEventListener` as well as a test that verifies the stream emits one item and terminates.

**Feel free to close this if the existing behavior is intentional.**

### DIFF
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -1411,21 +1411,19 @@ export const fromEventListener = <A = unknown>(
     readonly once?: boolean
     readonly bufferSize?: number | undefined
   } | undefined
-): Stream<A> => {
-  const stream = callback<A>((queue) => {
+): Stream<A> =>
+  callback<A>((queue) => {
+    const once = typeof options === "object" && options.once
+
     function emit(event: A) {
       Queue.offerUnsafe(queue, event)
+      if (once) Queue.endUnsafe(queue)
     }
     return Effect.acquireRelease(
       Effect.sync(() => target.addEventListener(type, emit, options)),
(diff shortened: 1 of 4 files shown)


packages/effect/src/Stream.ts
```diff
@@ -1411,8 +1411,8 @@ export const fromEventListener = <A = unknown>(
     readonly once?: boolean
     readonly bufferSize?: number | undefined
   } | undefined
-): Stream<A> =>
-  callback<A>((queue) => {
+): Stream<A> => {
+  const stream = callback<A>((queue) => {
     function emit(event: A) {
       Queue.offerUnsafe(queue, event)
```
tim-smart, line 1419:
```suggestion
      Queue.offerUnsafe(queue, event)
      if (once) Queue.endUnsafe(queue)
```

I think something like this would be simpler.

tim-smart (changes requested):

packages/effect/src/Stream.ts
youngspe: done

tim-smart (approved):

PR #7447, opened 2026-08-25
Prevent resolving tool calls from incomplete responses

candrewlee14:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Problem

Automatic tool resolution currently starts handlers even when the provider response is incomplete. For example, a model can emit a valid tool call and then finish with `length`; the handler still runs although its arguments may have been truncated. A malformed later response part can likewise fail validation only after the handler has already performed its side effect.

Incremental streaming has a related edge case: after emitting part of the first attempt, an `InvalidRequestError` can trigger the full-prompt fallback. Deferred tool calls from the failed attempt are then mixed into the fallback response.

## Minimal reproduction

```ts
import { Effect, Schema, Stream } from "effect"
import { LanguageModel, Tool, Toolkit } from "effect/unstable/ai"

const Echo = Tool.make("Echo", {
  parameters: Schema.Struct({ value: Schema.String }),
  success: Schema.String
})
const toolkit = Toolkit.make(Echo)
let calls = 0

const model = LanguageModel.make({
  generateText: () => Effect.succeed([
    {
      type: "tool-call",
      id: "call-1",
      name: "Echo",
      params: { value: "possibly truncated" }
    },
    {
      type: "finish",
      reason: "length",
      usage: { inputTokens: {}, outputTokens: {} },
      response: undefined
    }
  ]),
  streamText: () => Stream.empty
})

const program = LanguageModel.generateText({ prompt: [], toolkit }).pipe(
  Effect.provideServiceEffect(LanguageModel.LanguageModel, model),
  Effect.provide(toolkit.toLayer({
    Echo: ({ value }) => Effect.sync(() => {
      calls++
      return value
    })
  }))
)

await Effect.runPromise(program)
console.log(calls)
```

Before this PR, the program prints `1`. The incomplete response caused a user handler to run. After this PR, it prints `0`; the tool-call part remains in the response, but no tool result is synthesized.

The same issue exists in `streamText`: a tool-call part is observable before the terminal finish reason is known, so executing its handler immediately can commit a side effect before the provider reports `length`, `content-filter`, `error`, or a stream failure.

## Fix

- Delay automatic tool resolution until the provider response completes.
- Skip handlers for `length`, `content-filter`, and `error` finishes.
- Structurally validate a complete non-streaming response before side effects, while leaving parameters opaque so Toolkit still reports `ToolParameterValidationError` with the tool name and arguments.
- Run deferred handlers with `Effect.forEach` and the configured concurrency, preserving approvals and tool-result-before-finish ordering.
- Permit incremental full-prompt fallback only before the provider emits its first part.

## Verification

Regression tests cover:

- all three incomplete finish reasons in `generateText` and `streamText`
- malformed response content after a valid tool call
- observing a streamed tool call while proving its handler has not started
- provider failure after emitting a tool call
- incremental failure after partial output
- existing bounded concurrency, approvals, ordering, and parameter-validation semantics

The complete upstream matrix passes on Node, Deno, and Bun, together with lint, types, build, bundle, circular-dependency, and documentation checks.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/unstable/ai/LanguageModel.ts
+++ b/packages/effect/src/unstable/ai/LanguageModel.ts
@@ -14,7 +14,6 @@
 import type * as Cause from "../../Cause.ts"
 import * as Context from "../../Context.ts"
 import * as Effect from "../../Effect.ts"
-import * as FiberSet from "../../FiberSet.ts"
 import { constFalse, identity, pipe } from "../../Function.ts"
 import type * as JsonSchema from "../../JsonSchema.ts"
 import * as Option from "../../Option.ts"
@@ -22,7 +21,6 @@ import * as Predicate from "../../Predicate.ts"
 import * as Queue from "../../Queue.ts"
 import * as Schema from "../../Schema.ts"
 import * as SchemaAST from "../../SchemaAST.ts"
-import * as Semaphore from "../../Semaphore.ts"
 import * as Sink from "../../Sink.ts"
 import * as Stream from "../../Stream.ts"
 import type { Span } from "../../Tracer.ts"
@@ -1203,20 +1201,29 @@ export const make: (params: {
 
     const rawContent = yield* generateWithNonIncrementalFallback()
 
+    // Validate the complete response before tool handlers can perform side
+    // effects. Tool parameters remain opaque here so their validation keeps
+    // using Toolkit's more specific ToolParameterValidationError.
+    yield* Schema.decodeEffect(
+      Schema.mutable(Schema.Array(Response.Part(makeToolkitWithOpaqueParameters(toolkit))))
+    )(rawContent)
+
     // Resolve the generated tool calls
-    const toolResults = yield* resolveToolCalls(
-      rawContent,
-      toolkit,
-      providerOptions.prompt.content,
-      concurrency
-    ).pipe(
-      Stream.filter(
-        (result) =>
-          result.type === "tool-approval-request" ||
-          result.preliminary === false
-      ),
-      Stream.runCollect
-    )
+    const toolResults = hasIncompleteFinish(rawContent)
+      ? []
+      : yield* resolveToolCalls(
+        rawContent,
+        toolkit,
+        providerOptions.prompt.content,
+        concurrency
+      ).pipe(
+        Stream.filter(
+          (result) =>
+            result.type === "tool-approval-request" ||
+            result.preliminary === false
+        ),
+        Stream.runCollect
+      )
 
     const content = yield* Schema.decodeEffect(ResponseSchema)(rawContent)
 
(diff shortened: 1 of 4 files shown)


packages/effect/src/unstable/ai/LanguageModel.ts
```diff
@@ -1272,9 +1279,14 @@ export const make: (params: {
         incrementalPrompt: undefined,
         previousResponseId: undefined
       }
+      let emitted = false
       return requestOptions.incrementalPrompt
         ? params.streamText(requestOptions).pipe(
-          Stream.catchReason("AiError", "InvalidRequestError", (_) => params.streamText(fallbackOptions))
+          Stream.tap(() => Effect.sync(() => emitted = true)),
```
tim-smart: This would set `emitted` for non-content parts?

packages/effect/src/unstable/ai/LanguageModel.ts
```diff
@@ -2112,6 +2118,15 @@ const createDenialResults = (
 // Tool Call Resolution
 // =============================================================================
 
+const hasIncompleteFinish = (
+  content: ReadonlyArray<{ readonly type: string; readonly reason?: unknown }>
+): boolean =>
+  content.some(
+    (part) =>
+      part.type === "finish" &&
+      (part.reason === "length" || part.reason === "content-filter" || part.reason === "error")
```
tim-smart: Maybe we flip this to use an allow list?

```ts
const completeFinishReasons: ReadonlyArray<Response.FinishReason> = ["stop", "tool-calls", "pause"]

const hasIncompleteFinish = (
  content: ReadonlyArray<{ readonly type: string; readonly reason?: unknown }>
): boolean =>
  content.some((part) =>
    part.type === "finish" && !completeFinishReasons.includes(part.reason as Response.FinishReason)
  )
```

packages/effect/src/unstable/ai/LanguageModel.ts
```diff
@@ -1203,20 +1201,29 @@ export const make: (params: {
 
     const rawContent = yield* generateWithNonIncrementalFallback()
 
+    // Validate the complete response before tool handlers can perform side
+    // effects. Tool parameters remain opaque here so their validation keeps
+    // using Toolkit's more specific ToolParameterValidationError.
+    yield* Schema.decodeEffect(
+      Schema.mutable(Schema.Array(Response.Part(makeToolkitWithOpaqueParameters(toolkit))))
```
tim-smart: I think we should memoize these schemas instead of building for every request.

tim-smart:
Will instead try out a lookahead approach to try keep eager tool execution where possible.

PR #7449, opened 2026-08-25
Make PostgreSQL listeners scoped and readiness-aware

candrewlee14:
## Summary

- change PostgreSQL and PGlite `listen` from a lazy `Stream` to a scoped effect that returns a `Queue.Dequeue`
- complete the effect only after the listener is installed, giving callers an explicit readiness boundary
- share one lazy PostgreSQL listener connection while reference-counting channels and fanning notifications out to independent subscriber queues
- preserve empty payloads, propagate terminal connection failures, and close failed or interrupted acquisitions promptly

## Problem

The current stream API has no way to observe when PostgreSQL has acknowledged `LISTEN`:

```ts
const received = yield* sql.listen("events").pipe(
  Stream.runHead,
  Effect.forkScoped
)

// This can run before LISTEN is active, so the notification can be lost.
yield* sql.notify("events", "ready")
```

It also has several lifecycle issues:

- the notification handler is attached after the `LISTEN` query, leaving an additional delivery gap
- empty payloads are dropped by a truthiness check
- setup failures can leave the stream waiting indefinitely
- listener `error` and `end` events are not propagated to subscribers
- two subscribers on the same channel share a connection, but either finalizer runs `UNLISTEN` and silently disables the other
- a rejected or interrupted listener connection attempt can skip `client.end()`

## New API

`listen` returns a queue after PostgreSQL acknowledges the subscription:

```ts
const notifications = yield* sql.listen("events")

yield* sql.notify("events", "ready")
const payload = yield* Queue.take(notifications)
```

The readiness boundary also supports race-free snapshot-then-tail workflows:

```ts
const notifications = yield* sql.listen("events")

// Notifications received while this query runs are retained in the queue.
const currentRows = yield* sql<{ id: number }>`SELECT id FROM events ORDER BY id`
const nextPayload = yield* Queue.take(notifications)
```

The caller's scope owns the returned subscription. PostgreSQL subscriptions use a single per-client handler set and a serialized channel registry. The first subscriber performs `LISTEN`; the final subscriber performs bounded `UNLISTEN`. A terminal connection event fails every current queue and is retained so later subscribers cannot attach to a dead shared client.

Existing stream-based consumers can migrate without a sentinel value:

```ts
const notifications = Stream.unwrap(
  Effect.map(sql.listen("events"), Stream.fromQueue)
)
```

PGlite exposes the same readiness-aware scoped API.

## Verification

- `pnpm lint-fix`
- `pnpm check`
- `pnpm test --run packages/sql/pg/test packages/sql/pglite/test` — 228 tests passed
- `EFFECT_INTEGRATION_TESTS=1 pnpm test --run packages/sql/pg/test/Client.integration.test.ts -t "PgClient listen"` — 5 real PostgreSQL tests passed

The tests cover readiness, buffering during setup, empty and non-empty payloads, setup rejection and retry, interruption cleanup, connection acquisition cleanup, terminal fan-out, late subscription after terminal failure, same-channel reference counting, bounded scope cleanup, and PGlite parity.

### DIFF
--- a/.changeset/quiet-pandas-listen.md
+++ b/.changeset/quiet-pandas-listen.md
@@ -1,0 +1,6 @@
+---
+"@effect/sql-pg": patch
+"@effect/sql-pglite": patch
+---
+
+Make PostgreSQL `listen` subscriptions scoped and readiness-aware. `listen` now returns a dequeue after the subscription is installed, preserves empty payloads, shares a reference-counted listener connection, isolates concurrent subscriptions, and reports terminal listener failures.
--- a/packages/sql/pg/README.md
+++ b/packages/sql/pg/README.md
@@ -8,6 +8,48 @@ An Effect SQL client for PostgreSQL, built on the [`pg`](https://node-postgres.c
 npm install effect@rc @effect/sql-pg@rc
 ```
 
+## LISTEN / NOTIFY
+
+`listen` is a scoped subscription that returns after PostgreSQL has acknowledged
+the `LISTEN` command. The returned queue starts buffering notifications
+immediately, so it is safe to establish the subscription before reading initial
+state:
+
+```ts
+import { PgClient } from "@effect/sql-pg"
+import { Effect, Queue } from "effect"
+
+const program = Effect.gen(function*() {
+  const sql = yield* PgClient.PgClient
+  const notifications = yield* sql.listen("events")
+
+  // LISTEN is active here. Notifications received during this query are queued.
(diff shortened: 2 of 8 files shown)


candrewlee14:
Superseded by #7426, whose native PostgreSQL protocol client now provides the scoped, acknowledgement-aware queue API and listener lifecycle this PR introduced. Carrying the previous node-postgres implementation forward would duplicate the new architecture. The still-relevant PGlite parity change has been extracted into #7456.

Issue #6335, opened 2026-07-02
unstable/ai: tool-call parameter decode failure bypasses failureMode: "return" and kills the whole streamText stream


Necmttn:
## Summary

When a model emits a tool call whose `params` fail the tool's parameter schema, the failure kills the **entire `streamText` stream**, even when the tool is declared with `failureMode: "return"`. There are two independent fatality sites, and both run **before** the `failureMode` routing can see the error. The ecosystem docs describe `ToolParameterValidationError` as "Retryable: Yes", but there is currently no way for the model to see the validation error and retry - the stream is dead.

This is not a theoretical concern: slightly schema-drifted tool calls (an object where a string is expected, a stringified number, a missing key) are one of the most common live-model failure modes. We hit 4 production chat outages of this class in a single day.

## Expected behavior

With `failureMode: "return"`, a parameter validation failure should surface to the model as a failed tool result (`{ result: <validation error>, isFailure: true }`) so the model can correct its arguments on the next step - same contract as handler failures.

## Actual behavior - two sites, both fatal

All line refs at HEAD `f11ce73af60823754dc24194f4ffc561b9ea1c2d` (`packages/effect/src/unstable/ai/`). The same code paths exist in effect `4.0.0-beta.84` (repro below runs against it).

### Site 1: `LanguageModel.streamText` part decode (first to fire)

`streamText` decodes every provider chunk against `Schema.NonEmptyArray(Response.StreamPart(toolkit))` (`LanguageModel.ts:1468-1469`, applied at `:1529`). `Response.StreamPart(toolkit)` embeds each tool's `parametersSchema` into its ToolCallPart member, so schema-violating `params` fail `decodeParts` inside the forked driver fiber. The failure propagates via `Effect.tapCause((cause) => Queue.failCause(queue, cause))` (`LanguageModel.ts:1572`) and the consumer stream fails with `AiError.InvalidOutputError` (`Stream.mapError` at `LanguageModel.ts:976-983`). The toolkit never runs; `failureMode` is never consulted.

### Site 2: `Toolkit.handle` re-decode

[...]

PR #7452, opened 2026-08-25
Make model tool call errors recoverable

### LINKED ISSUE #7353: `effect/unstable/ai` disableToolCallResolution results with decoded params while toolkit.handle expects encoded ones

4.0.0-rc.110

1. Define a tool with a transforming schema in its parameters
2. Call LLM generation with disabled tool call resolution
3. Try to call toolkit.handle and observe error while type checking passes

I prepared small reproduction with mocked LLM client:

https://github.com/pawelblaszczyk5/effect-smoll-reproductions/blob/30d9e3e236104de5335c49297c79a8f9efbda6ea/ai-self-resolving-tools.ts

By running `node ai-self-resolving-tools.ts` you can observe the error

Params handed you out by llm generation can be manually resolved with toolkit.handle

### What do you see instead?

toolkit.handle fails in runtime when called with parameters where Encoded and Decoded type differs. 

### Additional information

This is a v4 equivalent of: https://github.com/Effect-TS/effect/issues/6119

I've decided to open a separate one for easier tracking - since that is tagged as v3 and no activity over there.

I also verified the linked PR and checked whether it could be easily ported by me to v4 but I'm not sure about the approach here. It lies on type level, where it returns you unparsed results typed them as encoded ones.

It feels like there's a mismatch mainly in the fact that toolkit.handle excepts decoded params on type level, but in reality it takes encoded ones and decodes them itself.

### LINKED ISSUE #7416: unstable/ai: constructing ToolParameterValidationError throws when toolParams contain NaN/Infinity, turning a typed tool error into a defect
## Summary

`AiError.ToolParameterValidationError` declares `toolParams: Schema.Json` and validates it at construction. Because `Schema.Json` rejects `NaN`, `Infinity`, and `-Infinity`, constructing the error throws whenever the offending params contain one of those values — so the *diagnostic* path for a bad tool call becomes a **defect** (`Die`) instead of the typed, retryable failure it is meant to be.

The value that triggers this is the very value the error exists to report, which makes the failure mode inverted: the worse the input, the less usable the error.

Line refs at `main` HEAD `4d89bb8ffb4cf567a1d11072246b6161ce638712`:

- `packages/effect/src/unstable/ai/AiError.ts:1017` — `toolParams: Schema.Json`
- `packages/effect/src/unstable/ai/Toolkit.ts:302-304` — the only construction site, inside `Toolkit.handle`'s `Effect.mapError`, passing the raw `params` straight through:

```ts
const decodedParams = yield* schemas.decodeParameters(params).pipe(
  Effect.mapError((cause) =>
    AiError.make({
      module: "Toolkit",
      method: `${name}.handle`,
      reason: new AiError.ToolParameterValidationError({
        toolName: name,
        toolParams: params,        // <- unvalidated, may contain NaN/Infinity
        description: cause.message
      })
    })
  )
)
```

Since the constructor throws synchronously inside `mapError`, the throw escapes as a defect rather than the typed error.

## Scope / severity

This cannot arise from a model's JSON tool call — JSON cannot encode `NaN` or infinities. It arises from programmatic callers of `toolkit.handle` (tests, in-process tool invocation, non-JSON transports, provider adapters that hand through already-decoded JS values). We hit it in a test that used `NaN` as a deliberately-bad tool parameter and got an opaque `Error: Schema validation failed` defect instead of `ToolParameterValidationError`.

So: low frequency, but it converts a diagnostic path into a crash path, and the resulting defect carries none of the tool name or decode description that the typed error would have carried.

## Repro

`effect@4.0.0-rc.109`; `4.0.0-rc.111` and `main` are byte-identical in `AiError.ts` for this declaration.

```ts
import * as Exit from "effect/Exit"
import * as Stream from "effect/Stream"
import * as AiError from "effect/unstable/ai/AiError"
import * as Tool from "effect/unstable/ai/Tool"
import * as Toolkit from "effect/unstable/ai/Toolkit"

// 1. Direct construction of the diagnostic error throws on a non-JSON value.
try {
  new AiError.ToolParameterValidationError({
    toolName: "GetWeather",
    toolParams: { city: NaN },
    description: "Expected string, got number"
  })
  console.log("1. constructed OK")
} catch (e) {
  console.log("1. construction THREW:", (e as Error).message)
}

// 2. End to end through Toolkit.handle.
const GetWeather = Tool.make("GetWeather", {
  parameters: Schema.Struct({ city: Schema.String }),
  success: Schema.String
})
const kit = Toolkit.make(GetWeather)
const layer = kit.toLayer({ GetWeather: () => Effect.succeed("sunny") })

const run = (params: unknown) =>
  Effect.gen(function*() {
    const built = yield* kit
    const exit = yield* Effect.exit(
      Effect.flatMap(built.handle("GetWeather", params as any), Stream.runCollect)
    )
    if (Exit.isSuccess(exit)) return "Success"
    return (exit.cause as any).reasons
      .map((r: any) => `${r._tag}(${r.error?.reason?._tag ?? String(r.defect).split("\n")[0]})`)
      .join(", ")
  }).pipe(Effect.provide(layer))

Effect.runPromise(
  Effect.gen(function*() {
    console.log("2. { city: 123 } ->", yield* run({ city: 123 }))
    console.log("3. { city: NaN } ->", yield* run({ city: NaN }))
  }) as Effect.Effect<void>
)
```

Output:

```
1. construction THREW: Schema validation failed
2. { city: 123 } -> Fail(ToolParameterValidationError)
3. { city: NaN } -> Die(Error: Schema validation failed)
```

Line 2 is the expected shape. Line 3 is the bug: the same class of input, one JSON-representable and one not, produces a typed retryable failure in one case and an untyped defect in the other. `Infinity` and `-Infinity` behave the same as `NaN`.

## Expected

`toolkit.handle` should fail with `AiError` / `ToolParameterValidationError` for *any* parameter value that fails the tool's parameter schema, including values `Schema.Json` cannot represent.

## No caller-side workaround

`ToolParameterValidationError` has no unsafe constructor: `.make` validates identically, there is no `makeUnsafe`, and a second `{ disableValidation: true }` argument is not honoured. The only construction site is internal to `Toolkit.handle`, so a consumer cannot sanitize `params` between the decode failure and the error construction — only before calling `handle`, which defeats the purpose of reporting what was actually passed.

## Suggested direction

Any one of:

1. Type the field as `Schema.Unknown` — it is diagnostic payload, not a wire contract.
2. Keep `Schema.Json` but sanitize at the construction site in `Toolkit.ts` (e.g. encode non-finite numbers to their string sentinels, or fall back to a string rendering of `params`).
3. Validate the field lazily rather than at construction.

(1) seems closest to the field's purpose. Happy to open a PR for whichever direction you prefer.

## Versions

- Observed on `effect@4.0.0-rc.109`
- Verified unchanged on `effect@4.0.0-rc.111` (current `rc` dist-tag) — `packages/effect/src/unstable/ai/AiError.ts` is identical to rc.109
- Verified unchanged on `main` @ `4d89bb8ffb4cf567a1d11072246b6161ce638712`
- Bun 1.x, macOS

Searched open and closed issues for `ToolParameterValidationError`, `toolParams`, `AiError NaN`, `Schema.Json NaN` — the nearest matches are #6335 (parameter-decode failure bypasses `failureMode: "return"`, same code path, different symptom) and #6350 (`Schema.Number` non-finite audit, opposite direction: schemas that are too permissive). Neither covers this.

candrewlee14:
## Summary

- normalize provider tool-call parameters once in `LanguageModel`
- return unknown tools and invalid parameters as `Response.ToolCallErrorPart` values instead of failing the whole model operation
- preserve malformed provider JSON as the same model-visible error shape
- turn tool-call errors into the original assistant call plus a failed tool result in chat history, so the model can correct and retry
- preserve canonical encoded parameters when automatic tool resolution is disabled
- feed only valid, decoded calls into the lookahead and interruption machinery added in #7486
- expose decoded Toolkit execution as a typed composition boundary for integrations which route or decorate resolved toolkits
- discriminate successful and failed `HandlerResult` values with `isFailure`
- keep arbitrary invalid parameters as typed diagnostics while retaining a JSON-safe encoded error representation

## Problem

Given a normal toolkit:

```ts
const GetWeather = Tool.make("GetWeather", {
  parameters: Schema.Struct({ city: Schema.String }),
  success: Schema.String
})

const toolkit = Toolkit.make(GetWeather)

const response = yield* LanguageModel.generateText({
  prompt: "What is the weather?",
  toolkit
})
```

if the provider returns a complete tool call like this:

```ts
{
  type: "tool-call",
  id: "call_1",
  name: "GetWeather",
  params: { city: 42 }
}
```

the current implementation fails `generateText` with a
`ToolParameterValidationError`. The model never receives that retryable error,
so it cannot correct the call. An invented tool name has the same problem, and
several providers fail even earlier when tool arguments contain malformed JSON.

With this change, the operation succeeds and exposes the failure directly:

```ts
response.toolCallErrors[0]?.error._tag
// "ToolParameterValidationError"
```

When response parts are added to a `Chat`, the error becomes an assistant tool
call followed by a failed tool result. Direct `Toolkit.handle` calls remain
strict; only model-originated calls are converted into model-visible feedback.

## Design

Providers translate their wire formats but leave toolkit lookup and Schema
decoding to the shared `LanguageModel` layer. The shared layer normalizes tool
calls and validates the complete response before any handler can perform side
effects. Invalid and unknown calls become error parts; valid calls are decoded
once and executed through the Toolkit's decoded-handler boundary.

For streaming responses, only valid decoded calls enter the one-chunk lookahead
buffer introduced in #7486. Invalid calls are emitted immediately and never
execute, while incomplete responses still interrupt or suppress valid calls and
produce their existing synthesized failure results. Mixed responses preserve
and execute valid sibling calls.

`Toolkit.makeWithHandler` exposes the same encoded and decoded boundaries used
internally so integrations can compose resolved toolkits without rebuilding
them or weakening their types. `HandlerResult` is discriminated by `isFailure`,
which preserves success and failure narrowing through those compositions.

The manual-resolution path additionally re-encodes valid calls through the
original tool Schema, preserving its documented encoded parameter type.

`ToolParameterValidationError.toolParams` accepts the original runtime value
but uses Effect's canonical defect encoder for its JSON representation. This
keeps diagnostic construction total without weakening the encoded contract or
changing ordinary JSON objects during decoding.

OpenAI, OpenAI-compatible, Anthropic, and OpenRouter use one shared secure JSON
constructor for streamed and non-streamed calls. A small leaf module for HTTP
detail schemas keeps the new `Response` / `AiError` relationship free of import
cycles; the existing public `Response.HttpRequestDetails` and
`Response.HttpResponseDetails` exports are unchanged.

## Related issues and work

- Addresses #6335 by keeping model-originated parameter validation failures in the response and making them available as retry feedback. Unlike the narrower approach in #6432, direct `Toolkit.handle` calls remain strict; recovery happens at the model boundary for calls produced by a provider.
- Preserves the encoded manual-resolution contract established by #7362 (which closed #7353, the v4 counterpart of #6119).
- Closes #7416 by allowing non-JSON invalid parameters to remain typed diagnostics rather than becoming defects.

## Prior art

This follows the recovery boundary used by other agent runtimes:

- [Vercel AI SDK preserves failed tool-call parsing and validation as an invalid call](https://github.com/vercel/ai/blob/e21bde74c64351453ac82abeae07e00fe838ee9a/packages/ai/src/generate-text/parse-tool-call.ts#L47-L114), then [adds a matching tool error without executing it](https://github.com/vercel/ai/blob/e21bde74c64351453ac82abeae07e00fe838ee9a/packages/ai/src/generate-text/generate-text.ts#L1251-L1271). It also exposes `repairToolCall` for custom repair before falling back to that representation.
- [pi turns unknown tools and argument validation failures into error tool results](https://github.com/badlogic/pi-mono/blob/a79b3733421ead0dcea3cbe32247ea2464400dcb/packages/agent/src/agent-loop.ts#L600-L668), allowing the agent loop to continue and send the failure back to the model.
- [LangChain agents automatically handle schema mismatches as model-visible tool feedback](https://docs.langchain.com/oss/javascript/migrate/langchain-v1#handling-tool-errors), while keeping runtime tool-error handling customizable.

## Validation

- `pnpm lint-fix`
- `pnpm check`
- `pnpm build`
- 100 focused core AI runtime tests
- focused type tests on TypeScript 5.9.3 and 6.0.3
- downstream Vesper package, static, and test verification (805 tests)

### DIFF
--- a/packages/ai/openai/src/OpenAiLanguageModel.ts
+++ b/packages/ai/openai/src/OpenAiLanguageModel.ts
@@ -19,7 +19,6 @@ import * as Predicate from "effect/Predicate"
 import * as Redactable from "effect/Redactable"
 import * as Schema from "effect/Schema"
 import * as AST from "effect/SchemaAST"
-import * as SchemaIssue from "effect/SchemaIssue"
 import * as Stream from "effect/Stream"
 import type { Span } from "effect/Tracer"
 import type { DeepMutable, Mutable, Simplify } from "effect/Types"
@@ -29,7 +28,7 @@ import * as LanguageModel from "effect/unstable/ai/LanguageModel"
 import * as AiModel from "effect/unstable/ai/Model"
 import { toCodecOpenAI } from "effect/unstable/ai/OpenAiStructuredOutput"
 import type * as Prompt from "effect/unstable/ai/Prompt"
-import type * as Response from "effect/unstable/ai/Response"
+import * as Response from "effect/unstable/ai/Response"
 import * as Tool from "effect/unstable/ai/Tool"
 import type * as HttpClientRequest from "effect/unstable/http/HttpClientRequest"
 import type * as HttpClientResponse from "effect/unstable/http/HttpClientResponse"
@@ -39,8 +38,6 @@ import { OpenAiClient } from "./OpenAiClient.ts"
 import type * as OpenAiSchema from "./OpenAiSchema.ts"
 import { addGenAIAnnotations } from "./OpenAiTelemetry.ts"
 import type * as OpenAiTool from "./OpenAiTool.ts"
-
-const formatIssue = SchemaIssue.makeFormatterDefault()
 
 const ResponseModelIds = Generated.ModelIdsResponses.members[1]
 const SharedModelIds = Generated.ModelIdsShared.members[1]
@@ -1387,32 +1384,12 @@ const makeResponse = Effect.fnUntraced(
 
         case "function_call": {
           hasToolCalls = true
-
-          const toolName = part.name
-
-          const toolParams = yield* Effect.try({
-            try: () => Tool.unsafeSecureJsonParse(part.arguments),
-            catch: (cause) =>
-              AiError.make({
-                module: "OpenAiLanguageModel",
-                method: "makeResponse",
-                reason: new AiError.ToolParameterValidationError({
-                  toolName,
-                  toolParams: {},
-                  description: `Faled to securely JSON parse tool parameters: ${cause}`
-                })
-              })
-          })
-
-          const params = yield* transformToolCallParams(options.tools, part.name, toolParams)
-
-          parts.push({
-            type: "tool-call",
+          parts.push(Response.toolCallPartFromJson({
             id: part.call_id,
-            name: toolName,
-            params,
+            name: part.name,
+            params: part.arguments,
             metadata: { openai: makeItemIdMetadata(part.id) }
-          })
+          }))
           break
         }
 
@@ -1452,24 +1429,21 @@ const makeResponse = Effect.fnUntraced(
             ? (approvalRequests.get(part.approval_request_id) ?? part.id)
             : part.id
 
-          const { toolName, params } = yield* normalizeMcpToolCall({
+          const toolCall = makeMcpToolCallPart({
+            id: toolId,
             toolNameMapper,
-            toolParams: part.arguments,
-            method: "makeResponse"
+            params: part.arguments
           })
 
-          parts.push({
-            type: "tool-call",
-            id: toolId,
-            name: toolName,
-            params,
-            providerExecuted: true
-          })
+          parts.push(toolCall)
+          if (toolCall.type === "tool-call-error") {
+            break
+          }
 
           parts.push({
             type: "tool-result",
             id: toolId,
-            name: toolName,
+            name: toolCall.name,
             isFailure: false,
             providerExecuted: true,
             result: {
@@ -1495,19 +1469,16 @@ const makeResponse = Effect.fnUntraced(
           const approvalRequestId = (part as any).approval_request_id ?? part.id
           const toolId = yield* idGenerator.generateId()
 
-          const { toolName, params } = yield* normalizeMcpToolCall({
+          const toolCall = makeMcpToolCallPart({
+            id: toolId,
             toolNameMapper,
-            toolParams: part.arguments,
-            method: "makeResponse"
+            params: part.arguments
           })
 
-          parts.push({
-            type: "tool-call",
-            id: toolId,
-            name: toolName,
-            params,
-            providerExecuted: true
-          })
+          parts.push(toolCall)
+          if (toolCall.type === "tool-call-error") {
+            break
+          }
 
           parts.push({
             type: "tool-approval-request",
@@ -2136,37 +2107,17 @@ const makeStreamResponse = Effect.fnUntraced(
 
                 hasToolCalls = true
 
-                const toolName = event.item.name
-                const toolArgs = event.item.arguments
-
-                const toolParams = yield* Effect.try({
-                  try: () => Tool.unsafeSecureJsonParse(toolArgs),
-                  catch: (cause) =>
-                    AiError.make({
-                      module: "OpenAiLanguageModel",
-                      method: "makeStreamResponse",
-                      reason: new AiError.ToolParameterValidationError({
-                        toolName,
-                        toolParams: {},
-                        description: `Failed securely JSON parse tool parameters: ${cause}`
-                      })
-                    })
-                })
-
-                const params = yield* transformToolCallParams(options.tools, toolName, toolParams)
-
                 parts.push({
                   type: "tool-params-end",
                   id: event.item.call_id
                 })
 
-                parts.push({
-                  type: "tool-call",
+                parts.push(Response.toolCallPartFromJson({
                   id: event.item.call_id,
-                  name: toolName,
-                  params,
+                  name: event.item.name,
+                  params: event.item.arguments,
                   metadata: { openai: makeItemIdMetadata(event.item.id) }
-                })
+                }))
 
                 break
               }
@@ -2205,24 +2156,21 @@ const makeStreamResponse = Effect.fnUntraced(
                     event.item.id)
                   : event.item.id
 
-                const { toolName, params } = yield* normalizeMcpToolCall({
+                const toolCall = makeMcpToolCallPart({
+                  id: toolId,
                   toolNameMapper,
-                  toolParams: event.item.arguments,
-                  method: "makeStreamResponse"
+                  params: event.item.arguments
                 })
 
-                parts.push({
-                  type: "tool-call",
-                  id: toolId,
-                  name: toolName,
-                  params,
-                  providerExecuted: true
-                })
+                parts.push(toolCall)
+                if (toolCall.type === "tool-call-error") {
+                  break
+                }
 
                 parts.push({
                   type: "tool-result",
                   id: toolId,
-                  name: toolName,
+                  name: toolCall.name,
                   isFailure: false,
                   providerExecuted: true,
                   result: {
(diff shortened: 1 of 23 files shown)


packages/effect/src/unstable/ai/LanguageModel.ts
IMax153: At this point, `toolCall.params` would have already been decoded by `normalizeToolCall`. This decodes it a second time. 

For tool call schemas that contain transformations (i.e. `Schema.FiniteFromString`), the second decode fails and `orElseSucceed(false)` would disable approval. 

Please pass the decoded parameters directly to `needsApproval` and add a transformed-schema regression test to ensure this case is covered moving forward.

packages/effect/src/unstable/ai/LanguageModel.ts
```diff
@@ -2246,6 +2247,208 @@ const makeInterruptedToolResult = (
...
+            toolName: toolCall.name,
+            availableTools: Object.keys(toolkit.tools)
+          })
+        }))
+      }
+    }
+
+    const tool = toolkit.tools[toolCall.name]
+    const parametersSchema = tool.parametersSchema
+    const codec = toolCall.providerExecuted === true || Tool.isDynamic(tool)
```
IMax153: Dynamic tools can use either raw JSON Schema _OR_ an Effect `Schema` for defining parameters. 

Tool params defined with a `Schema` still need to use the provider codec. This bypass should really only be used if `tool.jsonSchema` is defined.

packages/effect/src/unstable/ai/Prompt.ts
```diff
@@ -2137,6 +2137,27 @@ export const fromResponseParts = (parts: ReadonlyArray<Response.AnyPart>): Promp
         break
       }
 
+      // Tool Call Error Parts
+      case "tool-call-error": {
+        assistantParts.push(makePart("tool-call", {
+          id: part.id,
+          name: part.name,
+          params: part.params,
```
IMax153, line 2145: This would copy unchecked parameters into chat history. 

For example, a value like `1n` becomes a model-visible error with these changes, but would then cause `Chat.exportJson` to fail. The preserved parameters need the same JSON-safe treatment as `error.toolParams`.

packages/ai/openai/src/OpenAiLanguageModel.ts
IMax153, line 1093: For malformed calls, `part.params` is already the original argument string. Stringifying it again here would change "{" into "\"{\"", so history no longer reproduces the call that failed. Please preserve the original provider argument representation.

IMax153 (changes requested):
I've done a cursory review and noted some significant bugs that could result from this proposal.

However, I'm not too sure in general about the changes proposed by this PR because they would materially change how invalid model-generated calls are handled by `LanguageModel`.

Instead of being surfaced in the Effect error channel, operations like `generateText` would succeed, which may result in existing callers silently missing the fact that you now need to interrogate `response.toolCallErrors` to get at invalid tool calls.

I think the API needs to be revisited. It may be better to have an opt-in option on tool call declarations to enable this behavior (like we have with `failureMode`), or a different API altogether. 

I don't think I can merge this proposal in its current form.

candrewlee14:
Superseded by #7574.

Correcting this comment: it previously said that PR followed the review by making recovery opt-in per tool through `failureMode`. That was not accurate — it used an operation-level option throughout.

#7574 has since been rebuilt on top of #7588 and does now follow the review where it can: a call naming a tool in the toolkit follows that tool's `failureMode`, with no new API. Only a call naming a tool which is not in the toolkit still needs an operation-level option, because there is no declaration to carry a per-tool flag. Details and the four review points are in that PR's description.

Thanks for the review — it was right, and the corrected shape is much smaller.

PR #7460, opened 2026-08-26
Fix tagsExhaustive Effect.fn handler inference

### LINKED ISSUE #7454: `Match.tagsExhaustive` loses handler parameter types with `Effect.fn`
## Description

Using `Effect.fn` or `Effect.fnUntraced` directly as a `Match.tagsExhaustive` handler causes the generator parameter to be inferred as `any`. A handler returning `Effect.gen` is correctly narrowed to its tagged union member.

## Reproduction

```ts
import { Effect, Match } from "effect";

type Input =
  | { readonly _tag: "A"; readonly a: string }
  | { readonly _tag: "B"; readonly b: number };

declare const input: Input;

Match.value(input).pipe(
  Match.tagsExhaustive({
    A: (value) =>
      Effect.gen(function* () {
        yield* Effect.void;
        // @ts-expect-error Correctly rejected: `value` is the A member.
        value.missing;
        return value.a;
      }),
    B: (value) => Effect.succeed(value.b),
  }),
);

Match.value(input).pipe(
  Match.tagsExhaustive({
    A: Effect.fnUntraced(function* (value) {
      yield* Effect.void;
      return value.missing; // No error: `value` is inferred as `any`.
    }),
    B: Effect.fnUntraced(function* (value) {
      yield* Effect.void;
      return value.b;
    }),
  }),
);
```

The same loss of type safety occurs when `Effect.fn` is used instead of `Effect.fnUntraced`.

## Expected

Each `Effect.fn` and `Effect.fnUntraced` handler parameter is narrowed to its corresponding tagged union member, just like a handler returning `Effect.gen`.

## Actual

The handler parameters are inferred as `any`, allowing invalid property access under `strict` TypeScript settings.

## Versions

- `effect`: `4.0.0-rc.110`
- `typescript`: `6.0.2`

The relevant `Match.tagsExhaustive`, `Effect.fn`, and `Effect.fnUntraced` declarations are also unchanged on the latest `main` branch at the time of writing.

tim-smart:
## Summary

- preserve tagged-member contextual types through nested `Effect.fn` and `Effect.fnUntraced` calls in `Match.tagsExhaustive`, `Match.tags`, `Match.valueTags`, `Match.discriminators`, and `Match.discriminatorsExhaustive`
- keep the public and internal declarations aligned without casts, with shared handler-map aliases and a documented deferred conditional for TypeScript's nested inference limitation
- cover handler narrowing, inferred return types, exhaustiveness, and unknown-tag rejection with type tests
- add an `effect` patch changeset

## Testing

- `pnpm test-types Match.tst.ts`
- `pnpm check`
- `pnpm exec oxlint packages/effect/src/Match.ts packages/effect/src/internal/matcher.ts packages/effect/typetest/Match.tst.ts`
- `pnpm exec dprint check packages/effect/src/Match.ts packages/effect/src/internal/matcher.ts packages/effect/typetest/Match.tst.ts`
- `git diff --check`

Closes EFF-895
Closes #7454

### DIFF
--- a/packages/effect/src/Match.ts
+++ b/packages/effect/src/Match.ts
@@ -1161,11 +1161,17 @@ export const tagsExhaustive: <
     & { readonly [Tag in Types.Tags<"_tag", R> & string]: (_: Extract<R, Record<"_tag", Tag>>) => Ret }
     & { readonly [Tag in Exclude<keyof P, Types.Tags<"_tag", R>>]: never }
 >(
-  fields: P
+  // Preserve contextual types through nested generic calls (microsoft/TypeScript#52864).
+  fields: [P] extends [never] ? {
+      readonly [Tag in Types.Tags<"_tag", R> & string]: (
+        _: Extract<R, Record<"_tag", Tag>>
+      ) => Ret
+    }
+    : P
 ) => <I, F, A, Pr>(
   self: Matcher<I, F, R, A, Pr, Ret>
-) => [Pr] extends [never] ? (u: I) => Unify<A | ReturnType<P[keyof P]>> : Unify<A | ReturnType<P[keyof P]>> =
-  internal.tagsExhaustive
+) => [Pr] extends [never] ? (u: I) => Unify<A | ReturnType<P[keyof P]>> : Unify<A | ReturnType<P[keyof P]>> = internal
+  .tagsExhaustive as any
 
 /**
  * Creates a pattern that excludes a specific value while allowing all others.
--- a/packages/effect/src/Match.ts
+++ b/packages/effect/src/Match.ts
@@ -20,8 +20,21 @@ import type { Unify } from "./Unify.ts"
 
 const TypeId = internal.TypeId
 
-// Work around microsoft/TypeScript#52864 so nested generic calls receive contextual types.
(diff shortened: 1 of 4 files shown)


packages/effect/src/Match.ts
```diff
@@ -1161,11 +1161,17 @@ export const tagsExhaustive: <
...
+      readonly [Tag in Types.Tags<"_tag", R> & string]: (
+        _: Extract<R, Record<"_tag", Tag>>
+      ) => Ret
+    }
+    : P
 ) => <I, F, A, Pr>(
   self: Matcher<I, F, R, A, Pr, Ret>
-) => [Pr] extends [never] ? (u: I) => Unify<A | ReturnType<P[keyof P]>> : Unify<A | ReturnType<P[keyof P]>> =
-  internal.tagsExhaustive
+) => [Pr] extends [never] ? (u: I) => Unify<A | ReturnType<P[keyof P]>> : Unify<A | ReturnType<P[keyof P]>> = internal
```
tim-smart: I think we should update the internal types too, to prevent this cast.

Also do any other signatures need updating?

PR #7463, opened 2026-08-26
Remove the MIME dependency from @effect/platform-node

tim-smart:
## Summary

- replace the `mime` package with a local registry covering common static file types
- retain the public `Mime` class and default registry, with deprecation guidance for comprehensive mappings
- remove the production dependency and add lookup and HTTP response coverage

## Validation

- `nix develop -c pnpm test --run packages/platform/node/test/Mime.test.ts packages/platform/node/test/NodeHttpPlatform.test.ts`
- `nix develop -c pnpm check`

Closes EFF-899


packages/platform/node/src/Mime.ts
```diff
@@ -1,25 +1,187 @@
...
  */
 
-/* oxlint-disable no-named-as-default */
+type TypeMap = { readonly [type: string]: ReadonlyArray<string> }
 
-/**
- * @since 4.0.0
- */
-import Mime from "mime"
+const types: TypeMap = {
```
tim-smart: We should improve the coverage, maybe just copy the "standard" types from the mime package.

Also this module should feel like another effect module, with apis exported at the top level etc

tim-smart: Updated the module to expose top-level getType, getExtension, and getAllExtensions functions. The built-in registry now vendors the standard MIME table with 446 extension mappings; the class and default export remain for compatibility.

packages/platform/node/src/internal/mimeTypes.ts
```diff
@@ -0,0 +1,395 @@
+/**
+ * MIME type data copied from mime 4.1.0.
```
tim-smart: Use mime v5 "standard" types

packages/platform/node/src/Mime.ts
```diff
@@ -1,25 +1,176 @@
...
 
 /**
- * @category re-exports
+ * A MIME type registry.
+ *
+ * @category constructors
  * @since 4.0.0
  */
-export * from "mime"
+export class Mime {
```
tim-smart: Remove this class and instead use plain top level functions

packages/platform/node/src/Mime.ts
```diff
@@ -1,25 +1,176 @@
...
+
+/**
+ * The built-in MIME type registry used by `NodeHttpPlatform`.
+ *
+ * @deprecated Use the top-level functions in this module.
+ *
  * @since 4.0.0
  */
-export default Mime
+export default mime
```
tim-smart: Remove the default export

packages/platform/node/src/internal/mimeTypes.ts
tim-smart: I could not find a mime v5 source: npm's latest dist-tag and published versions, the GitHub tags and branches, and repository main all stop at 4.1.0. Which branch or commit should I vendor?

packages/platform/node/src/Mime.ts
tim-smart: Removed the default export and its compatibility test.

packages/platform/node/src/Mime.ts
tim-smart: Removed the class. The registry is now built once at module initialization and exposed only through plain top-level functions.

packages/platform/node/src/Mime.ts
```diff
@@ -1,25 +1,75 @@
...
 
 /**
- * @category re-exports
+ * Returns every file extension associated with a standard MIME type.
+ *
+ * @category utilities
  * @since 4.0.0
  */
-export default Mime
+export const getAllExtensions = (type: string): Set<string> | null => {
```
tim-smart:
```suggestion
export const getAllExtensions = (type: string): ReadonlySet<string> | null => {
```

packages/platform/node/src/Mime.ts
```diff
@@ -1,25 +1,75 @@
...
+}
 
 /**
+ * Returns the standard MIME type associated with a file name or extension.
+ *
+ * @category utilities
  * @since 4.0.0
  */
-import Mime from "mime"
+export const getType = (path: string): string | null => {
```
tim-smart: These should return Option

packages/platform/node/src/Mime.ts
tim-smart: Updated to Option.Option<ReadonlySet<string>>.

packages/platform/node/src/Mime.ts
tim-smart: All three lookup functions now return Option. NodeHttpPlatform uses Option.getOrElse for the application/octet-stream fallback, and the tests assert Some and None values directly.

PR #7464, opened 2026-08-26
fix(cluster): Activity handoff failing with `EntityNotAssignedToRunner`

hsyntax:
sharding can persist an activity request and then return `EntityNotAssignedToRunner`. `ClusterWorkflowEngine` previously converted that error to a defect, causing the parent workflow to record a terminal `Complete` result. This happens quite with rolling deployments.

### Before

```mermaid
stateDiagram-v2
  [*] --> Running
  Running --> ActivityPersisted: Start activity
  ActivityPersisted --> HandoffError: Owner shuts down
  HandoffError --> TerminalFailure: Capture routing error as defect
  TerminalFailure --> [*]

  note right of TerminalFailure
    The activity is durable,
    but its parent cannot replay.
  end note
```

### After

```mermaid
stateDiagram-v2
  [*] --> Running
  Running --> ActivityPersisted: Start activity
  ActivityPersisted --> HandoffSignal: Owner shuts down
  HandoffSignal --> Replayable: Suspend without saving a reply
  Replayable --> Running: New owner replays workflow
  Running --> Completed: Resume persisted activity
  Completed --> [*]
```

This PR treats `EntityNotAssignedToRunner` from persisted activity dispatch as an annotated handoff interruption. This leaves the durable run unanswered so the replacement owner can replay it.
Before replay, it closes the abandoned workflow instance scope to release owner-local resources while suppressing terminal workflow finalizers; deterministic replay registers those finalizers again.
The replay path excludes stored suspension failures and durable user interrupts.

_This PR was raised with support from Codex ( gpt-5.6-sol x-high )_


tim-smart:
Addressed in https://github.com/Effect-TS/effect/pull/7485

hsyntax:
Thanks tim

PR #7465, opened 2026-08-26
Export Effect Cluster SQL row contracts

ebramanti:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Effect Cluster's SQL storage defines its physical table contracts through migration SQL and private implementation types. Applications whose schema tooling owns table provisioning cannot currently detect contract drift during an Effect upgrade.

This change:

- exports complete message, reply, runner, and optional lock row types for PostgreSQL, MySQL, Microsoft SQL Server, and SQLite
- parameterizes dialect-specific scalar representations so consumers can match their ORM models while preserving exact columns, nullability, and generated-column differences
- derives internal SQL encoder and decoder types from the exported contracts
- adds exact type-level coverage for every dialect and custom scalar mappings
- adds a minor changeset

Validation:

- `pnpm lint`
- `pnpm test-types packages/effect/typetest/unstable/cluster/SqlStorage.tst.ts`
- `pnpm check`
- `EFFECT_INTEGRATION_TESTS=1 pnpm test --run packages/platform/node/test/cluster/SqlMessageStorage.integration.test.ts packages/platform/node/test/cluster/SqlRunnerStorage.integration.test.ts -t sqlite` (21 passed)

PostgreSQL and MySQL integration tests were not run locally because no container runtime is available.

## Related

- Related Issue: N/A
- Closes: N/A

### DIFF
--- a/.changeset/cluster-sql-row-types.md
+++ b/.changeset/cluster-sql-row-types.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Export complete, scalar-parameterized row types for every SQL-backed Cluster message, reply, runner, and optional lock table dialect.
--- a/packages/effect/src/unstable/cluster/SqlMessageStorage.ts
+++ b/packages/effect/src/unstable/cluster/SqlMessageStorage.ts
@@ -38,6 +38,144 @@ import * as ShardId from "./ShardId.ts"
 import type { ShardingConfig } from "./ShardingConfig.ts"
 import * as Snowflake from "./Snowflake.ts"
 
+type MessageTableRow<BigIntValue, IntegerValue, BooleanValue, TimestampValue> = {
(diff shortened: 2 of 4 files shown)


tim-smart:
I'm not convinced we want this - if we want to make internal schema changes it would then become a breaking change to a public interface.

ebramanti:
@tim-smart The argument for this is that even though Effect Cluster manages the schema, the consumer should be able to model a type around it. An example right now is that I maintain a schema in Drizzle that maps to this so I can use my normal query building process. I then match the types up to these as a way for TS compilation to verify changes that may occur in the schema in the future

tim-smart:
This is not something we will officially support at this stage.

PR #7468, opened 2026-08-26
Treat NaN as a non-positive count in Chunk.take and Chunk.drop

xia-chao:
## Summary

`Chunk.take` and `Chunk.drop` used `n <= 0` after `Math.floor`. That condition is false for `NaN`, so they built a slice whose `length` was `NaN`. Calling `Chunk.toArray` then threw `RangeError: Invalid array length`.

`Array.take` already treats `NaN` as a non-positive count (`slice(0, NaN)` is `[]`). This change uses `!(n > 0)` so `Chunk.take` returns empty and `Chunk.drop` is a no-op, matching that behavior.

## Test plan

- [x] `pnpm test --run packages/effect/test/Chunk.test.ts`
- [x] `Chunk.toArray(Chunk.take(Chunk.make(1, 2, 3), Number.NaN))` is `[]`
- [x] `Chunk.toArray(Chunk.drop(Chunk.make(1, 2, 3), Number.NaN))` is `[1, 2, 3]`
- [x] Existing fractional `take`/`drop` cases still pass


gcanti:
Closing in favor of https://github.com/Effect-TS/effect/pull/7478

PR #5500, opened 2025-09-12
update op-sqlite for @effect/sql-sqlite-react-native

subtleGradient:
## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR updates our React Native SQLite client to align with the newer
`@op-engineering/op-sqlite` API. The prior implementation was valid for older peer
dependency versions; this is a routine refactor plus a dependency floor
adjustment.

1) API alignment with newer op-sqlite
- Prior code used an async `db.execute` and read `rows?._array`, which belonged
  to older versions. The current API returns results via `QueryResult.rows`.
- Updated client to call `db.executeSync` for the sync path and
  `db.executeAsync` for the async path, and to consume `QueryResult.rows`.

2) TypeScript NodeNext visibility for `Sqlite.open`
- Under `moduleResolution: "NodeNext"`, TypeScript omits star re‑exports from
  internal subpaths not exposed in the package `exports` map, hiding
  `open`/`openSync`/`openRemote` in `@op-engineering/op-sqlite` types.
- Added a minimal, temporary ambient module augmentation at
  `src/op-sqlite-augmentation.d.ts` to surface `open` and a narrow `DB` type so
  this package type‑checks in NodeNext environments.

Changes
- Use `executeSync`/`executeAsync` and consume `QueryResult.rows`
- Keep value/values helpers aligned with the op-sqlite result shape
- Add `src/op-sqlite-augmentation.d.ts` to expose `open` for NodeNext
- Document removal plan once upstream publishes a types fix
- Raise peer dependency lower bound for `@op-engineering/op-sqlite` to
  >= the first version that includes the updated API, so consumers install a
  compatible version by default

Impact
- Aligns this package with the newer op-sqlite API
- Improves NodeNext TS ergonomics; no runtime behavior changes
- Ensures consumers get a compatible op-sqlite version by default

## Related

- Related: https://github.com/OP-Engineering/op-sqlite/pull/324
- Closes https://github.com/Effect-TS/effect/issues/5499

Issue #5499, opened 2025-09-12
@effect/sql-sqlite-react-native silently fails with latest op-sqlite


subtleGradient:
### What version of Effect is running?

@effect/sql-sqlite-react-native@0.47.1 + @op-engineering/op-sqlite@14.1.4

### What steps can reproduce the bug?

```ts
import * as Reactivity from "@effect/experimental/Reactivity";
import * as RNSqlClient from "@effect/sql-sqlite-react-native/SqliteClient";
import * as Console from "effect/Console";
import * as Effect from "effect/Effect";

Effect.gen(function* () {
	const sql = yield* RNSqlClient.make({ filename: ":memory:" });
	yield* Console.log(`SELECT 123`, yield* sql`SELECT 123`);
})
	.pipe(Effect.provide(Reactivity.layer), Effect.scoped, Effect.runPromise)
	.catch(console.error);
```

### What is the expected behavior?

 LOG  SELECT 123 [{"123": 123}]

### What do you see instead?

 LOG  SELECT 123 []

Silent failure. No warnings. No errors.

### Additional information

HYPOTHESIS: the API changed between @op-engineering/op-sqlite@7.1.0 – @op-engineering/op-sqlite@14.1.4

https://github.com/Effect-TS/effect/blob/ec8a769c1208525c9aec6cc50de69c0c127fdbd1/packages/sql-sqlite-react-native/package.json#L57

## PROBLEM 1:

https://github.com/Effect-TS/effect/blob/ec8a769c1208525c9aec6cc50de69c0c127fdbd1/packages/sql-sqlite-react-native/src/SqliteClient.ts#L124

- `db.execute` is async, this should be `db.executeSync`
- `someRandomPromise.rows?._array` will always be undefined

---

## PROBLEM 2:

https://github.com/Effect-TS/effect/blob/ec8a769c1208525c9aec6cc50de69c0c127fdbd1/packages/sql-sqlite-react-native/src/SqliteClient.ts#L120
https://github.com/Effect-TS/effect/blob/ec8a769c1208525c9aec6cc50de69c0c127fdbd1/packages/sql-sqlite-react-native/src/SqliteClient.ts#L124

- there is no `._array` field on `.rows`
- this is the correct shape:

https://github.com/OP-Engineering/op-sqlite/blob/4089ec4c10afde73700a35d7bdf4056c4e11f555/src/types.ts#L9-L31

```ts
export type QueryResult = {
  insertId?: number;
  rowsAffected: number;
[...]

PR #7469, opened 2026-08-26
Fix React Native SQLite value queries

gjermundgaraba:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

OP-SQLite returns a `RawQueryResult` object from `executeRaw` and `executeRawSync`:

```ts
{
  rowsAffected,
  rawRows,
  columnNames
}
```

The React Native client currently types these results as arrays and passes the wrapper object through unchanged. Consequently, `.values` and `.valuesUnprepared` return the wrong shape instead of the nested row arrays required by the Effect SQL connection contract.

This change unwraps `.rawRows` for synchronous and asynchronous value queries and corrects the internal driver types. `rawRows` has been part of OP-SQLite since 17.1.2, covering the package's declared peer range.

The tests now model OP-SQLite's actual result shape and cover synchronous and asynchronous `.values` and `.valuesUnprepared` queries.

## Related

- Follow-up to #6992, which switched value queries to `executeRaw` / `executeRawSync` but modeled their results as bare arrays.
- Related to #5499. That report targets the v3 client and its ordinary-query symptoms no longer reproduce on main, but it concerns the same OP-SQLite API change. Opening fresh per the close-out of #5500.


packages/sql/sqlite-react-native/src/SqliteClient.ts
```diff
@@ -180,14 +201,17 @@ export const make = (
...
+                try: () => db.executeRaw(sql, params as Array<any>),
+                catch: (cause) =>
+                  new SqlError({ reason: classifyError(cause, "Failed to execute statement (async)", "execute") })
+              }),
+              (result) => normalizeValues(result.rawRows)
+            )
           }
           return Effect.try({
-            try: () => db.executeRawSync(sql, params as Array<any>),
+            try: () => normalizeValues(db.executeRawSync(sql, params as Array<any>).rawRows),
```
tim-smart: I think this PR should just focus on this fix. I don't think we should normalize the values the database drivers return. You can instead use a Schema to transform to the desired shape.

gjermundgaraba: good point, i'll open a separate pr for the blob problem. updated the pr now

tim-smart (approved):

PR #7477, opened 2026-08-26
Support WebSocket clients across runtimes

candrewlee14:
`Socket.fromWebSocket` was tied to the DOM `WebSocket` type, so Node clients and authenticated OpenAI connections needed casts even though they already had the required runtime behavior.

This introduces a small shared WebSocket interface and adds typed handshake headers to the constructor. `Socket.fromWebSocket` also preserves the concrete acquired socket type for lifecycle hooks. Node uses `ws` when headers are needed, Bun forwards its native constructor options, and browsers reject header options with a clear error. The related casts in the Node server and OpenAI client are removed.

The Bun adapter keeps one local type assertion because `bun-types` exposes the DOM constructor overload when `lib.dom` is loaded, hiding Bun custom options from TypeScript.

Type tests cover cross-runtime compatibility and concrete socket type preservation on TypeScript 5.9 and 6.0. A Node integration test verifies that handshake headers reach a real WebSocket server.

### DIFF
--- a/packages/effect/src/unstable/socket/Socket.ts
+++ b/packages/effect/src/unstable/socket/Socket.ts
@@ -594,7 +594,7 @@ export interface WebSocketClientOptions {
    * This is supported by Node and Bun WebSocket clients. Browser constructors
    * cannot set arbitrary handshake headers and must reject this option.
    */
-  readonly headers?: Readonly<Record<string, string>>
+  readonly headers?: Readonly<Record<string, string>> | undefined
 }
 
 /**
@@ -713,7 +713,7 @@ export const fromWebSocket = <RO, WS extends WebSocketLike>(
         let open = false
 
         function onMessage(event: WebSocketEvent) {
-          const data = event.data
+          const data = event.data as Blob | ArrayBuffer | Uint8Array | string
           if (data instanceof Blob) {
             const effect = Effect.flatMap(
               Effect.promise(() => data.arrayBuffer()),
@@ -723,9 +723,6 @@ export const fromWebSocket = <RO, WS extends WebSocketLike>(
               }
             )
             return run(effect)
-          }
-          if (typeof data !== "string" && !(data instanceof ArrayBuffer) && !(data instanceof Uint8Array)) {
-            return
           }
           const result = handler(data instanceof ArrayBuffer ? new Uint8Array(data) : data)
           if (Effect.isEffect(result)) {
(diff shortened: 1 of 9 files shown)


packages/effect/src/unstable/socket/Socket.ts
```diff
@@ -636,23 +712,27 @@ export const fromWebSocket = <RO>(
         const run = yield* Effect.provideService(FiberSet.runtime(fiberSet)<R>(), WebSocket, ws)
         let open = false
 
-        function onMessage(event: MessageEvent) {
-          if (event.data instanceof Blob) {
+        function onMessage(event: WebSocketEvent) {
+          const data = event.data
```
tim-smart:
```suggestion
          const data = event.data as Blob | ArrayBuffer | Uint8Array | string
```

Just so we can remove the unnecessary guard

packages/effect/src/unstable/socket/Socket.ts
```diff
@@ -537,27 +537,98 @@ export const makeChannel = <IE = never>(): Channel.Channel<
...
+ * @since 4.0.0
+ */
+export interface WebSocketClientOptions {
+  /**
+   * Headers to include in the opening handshake.
+   *
+   * This is supported by Node and Bun WebSocket clients. Browser constructors
+   * cannot set arbitrary handshake headers and must reject this option.
+   */
+  readonly headers?: Readonly<Record<string, string>>
```
tim-smart:
```suggestion
  readonly headers?: Readonly<Record<string, string>> | undefined
```

packages/platform/node/src/NodeSocket.ts
```diff
@@ -21,6 +21,13 @@ import * as Socket from "effect/unstable/socket/Socket"
  */
 export * from "@effect/platform-node-shared/NodeSocket"
 
+const makeWebSocketWS: Socket.WebSocketConstructor["Service"] = (url, options) => {
```
tim-smart: I think remove this function as the branches are identical. You can cast types if they don't exactly match

packages/platform/bun/src/BunSocket.ts
```diff
@@ -28,9 +28,14 @@ export * from "@effect/platform-node-shared/NodeSocket"
  */
 export const layerWebSocketConstructor: Layer.Layer<
   Socket.WebSocketConstructor
-> = Layer.succeed(Socket.WebSocketConstructor)(
-  (url, protocols) => new globalThis.WebSocket(url, protocols)
-)
+> = Layer.succeed(Socket.WebSocketConstructor)((url, options) => {
```
tim-smart: I think the previous code was easier to grok. Just rename `protocols` to `options` and do a type cast with a comment explaining why it is safe.

tim-smart (changes requested):

tim-smart (approved):

PR #7484, opened 2026-08-26
Fix readonly SQLite transactions

### LINKED ISSUE #7482: @effect/sql-sqlite-bun: withTransaction fails for readonly query_only clients after BEGIN IMMEDIATE default
## Summary

`@effect/sql-sqlite-bun` documents that explicit transactions on writable connections use `BEGIN IMMEDIATE` and that `readonly: true` clients are unaffected, but the implementation passes `BEGIN IMMEDIATE` unconditionally to `SqlClient.make`.

A connection deliberately configured as hard read-only with SQLite's `PRAGMA query_only = ON` therefore cannot use `withTransaction`, even when the transaction body contains only a `SELECT`. The failure occurs while opening the transaction, before the body runs.

This appears to be a narrow regression from #7162. The writer-locking change is useful for writable clients; read-only clients need deferred `BEGIN`.

## Versions and environment

- `@effect/sql-sqlite-bun@4.0.0-rc.112`
- `effect@4.0.0-rc.112`
- Bun 1.3.14 (`0d9b296a`)
- SQLite 3.54.0 (bundled with Bun)
- macOS 27.0 arm64

The current Effect main branch also still has the unconditional opener: [`beginTransaction: "BEGIN IMMEDIATE"`](https://github.com/Effect-TS/effect/blob/8f5290a807a4a83940ee7a78c8c53e3e1c74d606/packages/sql/sqlite-bun/src/SqliteClient.ts#L233-L239).

## Minimal reproduction

Install the two RC packages:

```sh
bun add effect@4.0.0-rc.112 @effect/sql-sqlite-bun@4.0.0-rc.112
```

Then run:

```ts
import * as SqliteClient from "@effect/sql-sqlite-bun/SqliteClient"
import { Database } from "bun:sqlite"
import { Reactivity } from "effect/unstable/reactivity"

const filename = `/tmp/effect-sqlite-readonly-${crypto.randomUUID()}.sqlite`
const seed = new Database(filename)
seed.exec("CREATE TABLE test (id INTEGER PRIMARY KEY); INSERT INTO test DEFAULT VALUES;")
seed.close(false)

try {
  const rows = await Effect.runPromise(
    Effect.scoped(
      Effect.gen(function*() {
        const sql = yield* SqliteClient.make({
          filename,
          readonly: true
        })
        yield* sql.unsafe("PRAGMA query_only = ON")
        return yield* sql.withTransaction(sql.unsafe("SELECT * FROM test"))
      })
    ).pipe(Effect.provide(Reactivity.layer))
  )
  console.log(rows)
} finally {
  await Bun.file(filename).delete()
}
```

## Actual behavior

The `withTransaction(SELECT)` call fails at `BEGIN IMMEDIATE`:

```text
SqlError
└─ UnknownError
   └─ SQLiteError: attempt to write a readonly database
      code: SQLITE_READONLY
      errno: 8
```

Replacing the transaction opener with deferred `BEGIN` returns:

```text
[{ id: 1 }]
```

## Expected behavior

- A client configured with `readonly: true` should open an explicit read transaction with deferred `BEGIN`.
- Writable clients should retain `BEGIN IMMEDIATE` so they still avoid read-to-write snapshot upgrade failures and acquire or queue for the writer lock at transaction start.

This matches SQLite's transaction semantics: [`BEGIN IMMEDIATE` starts a write transaction immediately](https://www.sqlite.org/lang_transaction.html#deferred_immediate_and_exclusive_transactions), while [`PRAGMA query_only = ON` rejects attempts to change database state with `SQLITE_READONLY`](https://www.sqlite.org/pragma.html#pragma_query_only).

A direct Bun control reaches the same boundary: Bun's default transaction (`BEGIN`) succeeds for a `SELECT` after `query_only`, while its explicit `.immediate()` mode fails with `SQLITE_READONLY`. [Bun documents deferred/default and immediate transactions as distinct modes](https://bun.com/docs/runtime/sqlite#transactions), so this is transaction-policy selection in the Effect adapter rather than a Bun driver defect.

## Regression boundary and test gap

The behavior was introduced by commit [`c30386df`](https://github.com/Effect-TS/effect/commit/c30386df4be367350ee480fa357ddeae7a8f0d08) / #7162 and first published in `4.0.0-beta.107`. `beta.106` used the shared deferred `BEGIN` default; `beta.107` and RC.108 through RC.112 contain the unconditional `BEGIN IMMEDIATE`.

The change's documentation says [writable connections use `BEGIN IMMEDIATE` and `readonly: true` clients are unaffected](https://github.com/Effect-TS/effect/blob/c30386df4be367350ee480fa357ddeae7a8f0d08/packages/sql/sqlite-bun/src/SqliteClient.ts#L4-L10). Its [read-only transaction test](https://github.com/Effect-TS/effect/blob/c30386df4be367350ee480fa357ddeae7a8f0d08/packages/sql/sqlite-bun/test/Client.test.ts#L47-L71) opens with `readonly: true`, but does not enable `PRAGMA query_only`, so it does not cover this failure.

## Downstream evidence from Threadnote

Threadnote has real snapshot readers that:

1. construct an Effect Bun SQLite client with `readonly: true` and `readwrite: false`;
2. enforce `PRAGMA query_only = ON`; and
3. use `withTransaction` as a read-only linearization point.

Relevant source:

- [read-only client construction and connection policy](https://github.com/Kashkovsky/threadnote/blob/be4d5dacc11831d7ab63a01df19c03fb253eb602/src/code_graph/store_session.ts#L71-L98)
- [transactional snapshot-lease validation](https://github.com/Kashkovsky/threadnote/blob/be4d5dacc11831d7ab63a01df19c03fb253eb602/src/code_graph/store_leases.ts#L516-L563)
- [exact RC.112 dependency pin](https://github.com/Kashkovsky/threadnote/blob/be4d5dacc11831d7ab63a01df19c03fb253eb602/package.json#L127-L153)

Threadnote currently carries the narrow package patch:

```ts
beginTransaction: options.readonly === true ? "BEGIN" : "BEGIN IMMEDIATE"
```

See the [package patch](https://github.com/Kashkovsky/threadnote/blob/be4d5dacc11831d7ab63a01df19c03fb253eb602/patches/%40effect%252Fsql-sqlite-bun%404.0.0-rc.112.patch#L5-L23) and [two-sided real-file regression test](https://github.com/Kashkovsky/threadnote/blob/be4d5dacc11831d7ab63a01df19c03fb253eb602/test/unit/effect-sqlite-bun-compat.test.ts#L11-L98).

The test proves both sides of the intended contract:

- after the patch, a read-only + `query_only` transaction succeeds and still rejects writes;
- a writable Effect client attempting a transaction behind a separately held `BEGIN IMMEDIATE` writer still fails with `SQLITE_BUSY` / `database is locked`, proving writable transactions remain immediate.

The same unpatched RC.112 failure also reproduced on Ubuntu in Threadnote's real cross-repository workset smoke: the [GitHub Actions job](https://github.com/Kashkovsky/threadnote/actions/runs/32986627323/job/98234315575) reports `@effect/sql-sqlite-bun` failing with `SQLiteError: attempt to write a readonly database`.

## Suggested fix

Use the client's declared access mode when selecting the opener:

```ts
beginTransaction: options.readonly === true ? "BEGIN" : "BEGIN IMMEDIATE"
```

and extend the existing read-only test to enable `PRAGMA query_only = ON` before calling `withTransaction`.

The Node SQLite adapter has the same unconditional construction; it may merit the corresponding test/fix if the same case reproduces there, but this report and reproduction are specifically for `@effect/sql-sqlite-bun`.

Kashkovsky:
## Summary

- use deferred `BEGIN` for read-only Bun and Node SQLite clients
- retain `BEGIN IMMEDIATE` for writable clients
- cover `PRAGMA query_only = ON` with read-only `withTransaction` calls
- add patch changesets for both SQLite adapters

## Why

#7162 changed explicit SQLite transactions to `BEGIN IMMEDIATE` to acquire the writer lock up front and avoid read-to-write snapshot upgrade failures. That behavior is correct for writable clients, but it also applied to clients opened with `readonly: true`.

When a read-only connection enables SQLite's `PRAGMA query_only = ON`, even `withTransaction(SELECT)` then fails while opening the transaction because `BEGIN IMMEDIATE` requests a write transaction. Bun reports `SQLITE_READONLY`; Node reports SQLite error code 8.

This selects the transaction opener from the client's declared access mode. The existing contention tests continue to prove that writable transactions acquire the writer lock immediately.

Closes #7482.

### DIFF
--- a/.changeset/calm-readers-transact.md
+++ b/.changeset/calm-readers-transact.md
@@ -1,0 +1,7 @@
+---
+"@effect/sql-sqlite-bun": patch
+"@effect/sql-sqlite-node": patch
+---
+
+Allow read-only SQLite clients to use `withTransaction` when `PRAGMA query_only` is enabled. Writable clients continue
+to reserve the writer lock when a transaction starts.
--- a/packages/sql/sqlite-bun/src/SqliteClient.ts
+++ b/packages/sql/sqlite-bun/src/SqliteClient.ts
@@ -235,7 +235,7 @@ export const make = (
         acquirer,
         compiler,
         transactionAcquirer,
-        beginTransaction: "BEGIN IMMEDIATE",
+        beginTransaction: options.readonly === true ? "BEGIN" : "BEGIN IMMEDIATE",
         spanAttributes: [
           ...(options.spanAttributes ? Object.entries(options.spanAttributes) : []),
           [ATTR_DB_SYSTEM_NAME, "sqlite"]
--- a/packages/sql/sqlite-bun/test/Client.test.ts
+++ b/packages/sql/sqlite-bun/test/Client.test.ts
@@ -62,11 +62,13 @@ describe("Client", () => {
 
       const sql = yield* SqliteClient.make({ filename, readonly: true })
       assert.deepStrictEqual(yield* sql`SELECT * FROM test`, [])
-      assert.deepStrictEqual(yield* sql.withTransaction(sql`SELECT * FROM test`), [])
 
       const error = yield* Effect.flip(sql`INSERT INTO test DEFAULT VALUES`)
       assert.strictEqual(error._tag, "SqlError")
       assert(error.reason.cause instanceof Error)
       assert.match(error.reason.cause.message, /attempt to write a readonly database/i)
+
+      yield* sql`PRAGMA query_only = ON`
+      assert.deepStrictEqual(yield* sql.withTransaction(sql`SELECT * FROM test`), [])
     }).pipe(Effect.provide(Reactivity.layer)))
 })
--- a/packages/sql/sqlite-node/src/SqliteClient.ts
+++ b/packages/sql/sqlite-node/src/SqliteClient.ts
@@ -320,7 +320,7 @@ export const make = (
         acquirer,
         compiler,
         transactionAcquirer,
-        beginTransaction: "BEGIN IMMEDIATE",
+        beginTransaction: options.readonly === true ? "BEGIN" : "BEGIN IMMEDIATE",
         spanAttributes: [
           ...(options.spanAttributes ? Object.entries(options.spanAttributes) : []),
           [ATTR_DB_SYSTEM_NAME, "sqlite"]
--- a/packages/sql/sqlite-node/test/Client.test.ts
+++ b/packages/sql/sqlite-node/test/Client.test.ts
@@ -156,6 +156,7 @@ describe("Client", () => {
       )
 
       const sql = yield* SqliteClient.make({ filename, readonly: true })
+      yield* sql`PRAGMA query_only = ON`
       assert.deepStrictEqual(yield* sql.withTransaction(sql`SELECT * FROM test`), [])
     }).pipe(Effect.provide([NodeFileSystem.layer, Reactivity.layer])))
 


Kashkovsky:
@tim-smart, would you mind reviewing this? It narrows the transaction opener introduced in #7162 to preserve `BEGIN IMMEDIATE` for writable clients while restoring deferred transactions for clients declared read-only. #7482 contains the downstream reproduction and SQLite evidence.

tim-smart (approved):

PR #7485, opened 2026-08-26
fix(cluster): never surface transient routing states for persisted messages

tim-smart:
Sharding can persist a request and then fail the caller with `EntityNotAssignedToRunner` during a runner shutdown. For durable workflows this converted a transient routing state into a terminal `Complete(Failure(Die(...)))`, which happens frequently with rolling deployments.

This PR fixes the semantics at the layers they belong to, instead of handling the error in the workflow engine:

- **If an entity moves runners before replying, the caller keeps waiting.** This already worked (`replyFromStorage` waits are requestId-scoped, and rebalance-released waiters re-route into the storage wait) and is unchanged.
- **If the local runner is shutting down while a caller waits, the call is interrupted.** `sendOutgoing`'s abandon path persists the request and then interrupts the caller instead of failing: the message is durable, so the routing state is not an error, this runner just can never observe the reply. Shard release during shutdown now interrupts parked reply waiters for the same reason (`unregisterShardReplyHandlers` gains an `interrupt` option); rebalance release keeps the resume-with-error so re-routing still works. `AckChunk` abandons still fail, since a stream cannot resume under a new owner.
- **A workflow stops with nothing persisted, ready to resume elsewhere.** `Workflow.intoResult` treats an interrupt that is neither a suspend nor a durable interrupt as an abandoned run attempt: it releases owner-local resources without running durable finalizers (compensations), does not resume the parent, and exits with the interrupt cause so no reply is stored. The unanswered run request is then replayed by the replacement owner through the entityManager's existing no-reply-on-interrupt rule.

### Before

```mermaid
stateDiagram-v2
  [*] --> Running
  Running --> ActivityPersisted: Start activity
  ActivityPersisted --> HandoffError: Owner shuts down
  HandoffError --> TerminalFailure: Capture routing error as defect
  TerminalFailure --> [*]

  note right of TerminalFailure
    The activity is durable,
    but its parent cannot replay.
  end note
```

### After

```mermaid
stateDiagram-v2
  [*] --> Running
  Running --> ActivityPersisted: Start activity
  ActivityPersisted --> CallerInterrupted: Owner shuts down
  CallerInterrupted --> Replayable: Abandon run, nothing persisted
  Replayable --> Running: New owner replays workflow
  Running --> Completed: Resume persisted activity
  Completed --> [*]
```

### Tests

- Cluster integration tests cover the abandoned handoff (deterministic fault injection), a durable interrupt racing a handoff, and a real owner shutdown with replay on the replacement runner, for both default and `SuspendOnFailure` workflows, asserting no compensations run and workflow-scoped resources are released exactly once per attempt.
- Two tests pinned the old error contract and now assert interruption for persisted requests; the non-persisted case still asserts `EntityNotAssignedToRunner`.
- Fixed a race in the rebalance test that surfaced now that shutdown unwinds promptly (owner read without waiting for stable assignments).

Alternative to #7464.

Closes EFF-900

_This PR was raised with support from Claude (Fable 5)._

### DIFF
--- a/packages/effect/src/unstable/workflow/WorkflowEngine.ts
+++ b/packages/effect/src/unstable/workflow/WorkflowEngine.ts
@@ -256,9 +256,9 @@ export class WorkflowInstance extends Context.Service<
     interrupted: boolean
 
     /**
-     * Whether the workflow scope is being closed for replay rather than completion.
+     * Whether the current workflow run has been abandoned.
      */
-    replaying: boolean
+    abandoned: boolean
 
     /**
      * When SuspendOnFailure is triggered, the cause of the failure is stored
@@ -286,7 +286,7 @@ export class WorkflowInstance extends Context.Service<
       scope,
       suspended: false,
       interrupted: false,
-      replaying: false,
+      abandoned: false,
       cause: undefined,
       awaitedDeferreds: new Set(),
       activityState: {
--- a/packages/effect/src/unstable/workflow/WorkflowEngine.ts
+++ b/packages/effect/src/unstable/workflow/WorkflowEngine.ts
@@ -256,6 +256,11 @@ export class WorkflowInstance extends Context.Service<
     interrupted: boolean
 
     /**
+     * Whether the workflow scope is being closed for replay rather than completion.
+     */
+    replaying: boolean
+
+    /**
      * When SuspendOnFailure is triggered, the cause of the failure is stored
      * here.
      */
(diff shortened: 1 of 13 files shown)


packages/effect/src/unstable/workflow/WorkflowEngine.ts
```diff
@@ -255,6 +255,11 @@ export class WorkflowInstance extends Context.Service<
      */
     interrupted: boolean
 
+    /**
+     * Whether the workflow scope is being closed for replay rather than completion.
+     */
+    replaying: boolean
```
tim-smart: Maybe `abandoned` is a better name

PR #7486, opened 2026-08-27
Execute streaming tool handlers eagerly with interruption and synthesized results

tim-smart:
Alternative to #7447 that keeps its safety properties without giving up eager tool execution in streaming. Stacked on that PR: the first commit (`17fbd64`) is #7447 itself, the commits after it are the delta.

## Semantics

One rule for both paths: a tool call executes only when the response is known to have moved past it.

- `generateText` validates the complete response before any handler side effect (parameters stay opaque so `Toolkit` still reports `ToolParameterValidationError`) and does not run handlers when the finish reason is incomplete.
- `streamText` uses a one-chunk lookahead: a handler starts once the stream advances past its tool call (the next chunk arrives, or the stream ends with a complete finish). Providers emit a truncating finish back-to-back with the last tool call, so this window is what lets an incomplete finish deterministically prevent handlers from starting, while mid-stream tool calls still overlap the rest of generation (delayed by one inter-token gap).
- Handlers still running when an incomplete finish arrives are interrupted via the `FiberSet`. Stream failure and consumer cancellation interrupt through scope closure as before.
- Every tool call left unresolved by an incomplete finish receives a synthesized `tool-result` failure part (`{ type: "execution-interrupted", reason: ... }`, mirroring the existing `execution-denied` shape), emitted before the finish part. The conversation history therefore never contains a tool call without a matching result, which providers otherwise reject on the next request. Calls awaiting approval keep their approval request and get no synthesized result.
- Incomplete finish detection is an allow-list (`stop`, `tool-calls`, `pause`), so `unknown`, `other`, and future reasons fail safe.
- Incremental full-prompt fallback only triggers before the provider emits its first content part; `response-metadata` no longer counts as content.

## Why not defer handlers until the stream ends

Deferral (the approach in #7447, and what the Vercel AI SDK ships since vercel/ai#19066) is the natural design when tools cannot be cancelled. Effect has real interruption, so handlers can start as soon as the stream moves past their call and be interrupted if the response turns out to be incomplete, keeping the latency overlap that streaming exists for. Handlers are expected to be interruption-safe as everywhere else in Effect (`uninterruptible` for commit sections, `onInterrupt` for compensation).

## Testing

Regression tests cover: handlers not starting when the truncating finish arrives in the same or the next chunk (all five incomplete reasons), interruption of in-flight handlers on incomplete finish and on stream failure, completed handlers keeping their results with no duplicate synthesis, approval requests suppressing synthesis, synthesized results in `generateText` for all incomplete reasons, missing `params` fields, malformed content on both paths, and the existing concurrency, approval, ordering, and fallback semantics.

Full `unstable/ai` suite passes (1052 tests), plus type checking, oxlint, and dprint.

Closes EFF-891


packages/effect/src/unstable/ai/LanguageModel.ts
```diff
@@ -1531,7 +1574,12 @@ export const make: (params: {
           approvalId,
           toolCallId: part.id
         }) as Response.StreamPart<Tools>
-        yield* Queue.offer(queue, approvalPart)
+        yield* Effect.uninterruptible(
```
tim-smart: You can use Queue.offerUnsafe to make this atomic without Effect.uninterruptible. Same for the other locations

packages/effect/src/unstable/ai/LanguageModel.ts
```diff
@@ -1573,26 +1648,54 @@ export const make: (params: {
           if (immediateParts.length > 0) {
             yield* Queue.offerAll(queue, immediateParts)
           }
-          // Fork tool call handlers - use the raw chunk for encoded params
+          // The stream has moved past any previously buffered tool calls, so
+          // start their handlers now - unless a finish part (possibly
+          // recorded just above) reports an incomplete response, in which
+          // case they stay pending and resolve with synthesized failure
+          // results.
+          if (Predicate.isUndefined(findIncompleteFinishReason(deferredFinishParts))) {
```
tim-smart: Instead of using the Predicate module for undefined checks, just use plain comparisons.

packages/effect/src/unstable/ai/LanguageModel.ts
tim-smart: Updated in 6427a3d4de. The approval and final-result paths now use `Queue.offerUnsafe` and update `pendingToolCalls` synchronously, so no uninterruptible region is needed.

packages/effect/src/unstable/ai/LanguageModel.ts
tim-smart: Updated in 6427a3d4de. The new undefined checks now use direct `=== undefined` and `!== undefined` comparisons.

.changeset/tidy-tools-finish.md
```diff
@@ -0,0 +1,7 @@
+---
+"effect": patch
+---
+
+Make automatic tool resolution interruption-safe for incomplete language model responses.
+
+`generateText` now validates the complete response before tool handlers can perform side effects. When the finish reason is incomplete, executable tool calls receive an `execution-interrupted` failure result without starting their handlers. `streamText` holds each tool call for one chunk, starts its handler once the provider advances, and interrupts unresolved handlers when the stream fails or finishes with an incomplete reason. Calls left unresolved by an incomplete finish receive the same failure result. Incremental streaming fallback now occurs only before the provider emits its first content part.
```
tim-smart:
```suggestion
Make automatic tool resolution interruption-safe for incomplete language model responses.
```

PR #7487, opened 2026-08-27
Pull-based Socket API with backpressure

tim-smart:
Closes EFF-932

Redesigns `effect/unstable/socket/Socket` around a pull-based read side so sockets have real end-to-end backpressure, and reconnect is just `Effect.retry`.

## Interface

`Socket` now exposes `reader` and `writer` instead of `run` / `runString` / `runRaw`:

```ts
interface Socket {
  readonly reader: Effect<Pull<NonEmptyReadonlyArray<Uint8Array | string>, SocketError>, SocketError, Scope>
  readonly writer: Effect<Writer, SocketError, Scope>
}

interface Writer {
  readonly write: (chunk: Uint8Array | string | CloseEvent) => Effect<void, SocketError>
  readonly writeAll: (chunks: NonEmptyReadonlyArray<Uint8Array | string>) => Effect<void, SocketError>
}
```

`Socket.make` takes `{ reader, writer }` with no derivation logic. `Socket.readerBytes` / `Socket.readerString` wrap the acquisition and map each pulled batch (bytes fast-path returns the array as-is; the `TextDecoder` is created once per acquisition).

## Semantics

- **Acquiring `reader` dials the connection.** The scope owns the connection lifecycle, so one `Effect.retry` around the scoped consume loop covers connect and mid-stream failures — that is the whole auto-reconnect story.
- **No clean-close concept.** The pull never completes through the Done channel; any close fails with `SocketError` wrapping `SocketCloseError`. `closeCodeIsError`, `defaultCloseCodeIsError`, and `SocketCloseError.filterClean` are gone.
- **`NonEmptyReadonlyArray` is the batching unit.** TCP delivers one concatenated buffer, WebSocket one element per frame.
- `onOpen` is removed: code between `yield* socket.reader` and the first pull runs once per (re)connection.
- The writer stays detached via the latch-on-current-connection pattern; releasing the writer scope half-closes (`end()` / `closeWrite()`).

## Backpressure per transport

- **Node TCP (`fromDuplex`)**: the stream stays paused; each pull is `readable` + `stream.read()`, which drains Node's internal buffer as one chunk. Node's buffer fills to `highWaterMark`, the kernel window closes, the peer blocks. Writes use `write()` return-value backpressure (one `drain` await when full); `writeAll` corks/uncorks.
- **WebSocket, single implementation**: feature-detects `"pause" in ws` at acquisition. `ws` sockets stay paused between pulls (resume on empty-buffer pull, pause and deliver the tick's accumulated frames); browser sockets use the identical buffer/delivery code with no-op pause/resume and an optional `highWaterMark` that fails the socket with `SocketReadError` (default unbounded). `onInitialRun` is gone — DenoHttpServer buffers pre-acquisition events behind a small `WebSocketLike` facade so they come out of the first pull.
- **Server sockets** pause immediately on accept, and `reader` attaches to the existing connection: a second acquisition after close fails instead of reconnecting.

## Integration

- `toChannel` / `toChannelString` now read through the pull (no unbounded queue); `Socket.toStream` added for read-only consumption.
- Migrated consumers: `RpcClient.makeProtocolSocket`, `RpcServer` socket protocol, DevTools client/server (unchanged, still duplex channels), `OpenAiClient` realtime WebSocket, Node/Bun/Deno HTTP server upgrades, Node/Deno socket servers, platform WebSocket layers.
- `SendQueueCapacity` removed; migration annotations updated (`currentSendQueueCapacity`, `defaultCloseCodeIsError`, `fromTransformStream`, `toChannelMap`).

## Validation

- `pnpm check`, `pnpm lint`: clean.
- Tests: NodeSocket, NodeHttpServer, RpcClient, RpcServer (effect + node), DevToolsClient, cluster Runners, OpenAiClient — 193 passed. Cluster integration suite: 112/113 under load, flaky tests pass in isolation (shard-stabilization timing, not socket-related).
- `pnpm jsdocs --check` and `pnpm api-diff --check` have pre-existing failures on main; this branch adds no new violations. `migration/v3-to-v4.md` regeneration deferred because the api-diff check has standing gaps outside this change.


packages/effect/src/unstable/socket/Socket.ts
```diff
@@ -671,152 +719,218 @@ export const makeWebSocket = (url: string | Effect.Effect<string>, options?: {
...
+        resume(Effect.succeed(takeBuffer()))
+      }
+
+      function push(data: Uint8Array | string) {
+        buffer.push(data)
+        bufferSize += typeof data === "string" ? data.length : data.byteLength
+        if (waiter !== undefined) {
+          if (!flushScheduled) {
+            flushScheduled = true
+            queueMicrotask(deliver)
```
tim-smart: This should instead use the Scheduler service

packages/platform/bun/src/BunHttpServer.ts
```diff
@@ -582,49 +578,108 @@ class BunServerRequest extends Inspectable.Class implements ServerRequest.HttpSe
...
-          runRaw,
-          writer
+          })
+        const writer: Socket.Socket["writer"] = Effect.succeed({ write, writeAll })
+
+        const reader: Socket.Socket["reader"] = Effect.gen(function*() {
+          if (ws.data.closeError !== undefined && ws.data.buffer.length === 0) {
+            return yield* Effect.fail(ws.data.closeError)
+          }
+          yield* semaphore.take(1)
```
tim-smart: If the effect is interrupted / yields before the finalizer is added it could lose the permit.

packages/effect/src/unstable/socket/Socket.ts
```diff
@@ -376,16 +417,29 @@ export class SocketError extends Schema.TaggedError<SocketError>(SocketErrorType
...
-export const toChannelMap = <IE, A>(
+const closeError = (code: number, closeReason?: string | undefined) =>
+  new SocketError({ reason: new SocketCloseError({ code, closeReason }) })
+
+const writeChunk = (
+  writer: Writer,
+  chunk: NonEmptyReadonlyArray<Uint8Array | string | CloseEvent>
+): Effect.Effect<void, SocketError> => {
+  for (let i = 0; i < chunk.length; i++) {
+    if (isCloseEvent(chunk[i])) {
```
tim-smart, line 436: Here we should using the index of the close event as the end point instead of chunk.length

packages/effect/src/unstable/rpc/RpcClient.ts
```diff
@@ -1065,48 +1061,73 @@ export const makeProtocolSocket = (options?: {
...
+      parser = serialization.makeUnsafe()
+      pinger.reset()
+      return Effect.gen(function*() {
+        const pull = yield* socket.reader
+        currentError = undefined
+        if (Option.isSome(hooks)) {
+          yield* hooks.value.onConnect
         }
-      }, { onOpen }).pipe(
+        return yield* pull.pipe(
```
tim-smart: This could be a while loop now that we are in a generator

packages/effect/src/unstable/rpc/RpcClient.ts
```diff
@@ -960,17 +955,14 @@ export const makeProtocolHttp = (client: HttpClient.HttpClient): Effect.Effect<
           Effect.flatMap((responses) => {
             if (responses.length === 0) return Effect.void
             hasResponse = true
-            let i = 0
-            return Effect.whileLoop({
-              while: () => i < responses.length,
-              body: () => {
-                const response = responses[i++]
+            return Effect.gen(function*() {
```
tim-smart: This can be reverted to whileLoop

packages/effect/src/unstable/rpc/RpcServer.ts
```diff
@@ -1294,11 +1294,10 @@ export const makeProtocolStdio = Effect.gen(function*() {
       Stream.runForEach((data) => {
         const decoded = parser.decode(data) as ReadonlyArray<FromClientEncoded>
         if (decoded.length === 0) return Effect.void
-        let i = 0
-        return Effect.whileLoop({
-          while: () => i < decoded.length,
-          body: () => writeRequest(0, decoded[i++]),
-          step: constVoid
+        return Effect.gen(function*() {
```
tim-smart: This can be reverted to whileLoop

packages/effect/src/unstable/rpc/RpcServer.ts
```diff
@@ -1471,50 +1460,57 @@ const makeSocketProtocol: Effect.Effect<
...
+    const processData = (data: Uint8Array | string): Effect.Effect<void, Socket.SocketError> => {
       try {
         const decoded = parser.decode(data) as ReadonlyArray<FromClientEncoded>
         if (decoded.length === 0) return Effect.void
-        let i = 0
-        return Effect.whileLoop({
-          while: () => i < decoded.length,
-          body() {
-            const message = decoded[i++]
+        return Effect.gen(function*() {
```
tim-smart: This can be reverted to whileLoop

packages/effect/src/unstable/rpc/RpcClient.ts
```diff
@@ -1065,48 +1053,61 @@ export const makeProtocolSocket = (options?: {
...
-                if (clientId !== undefined) {
-                  if (response._tag === "Exit") {
-                    requestClientMap.delete(requestId)
-                  }
-                  return writeResponse(clientId, response)
+    const processData = (data: Uint8Array | string): Effect.Effect<void> => {
+      try {
+        const responses = parser.decode(data) as Array<FromServerEncoded>
+        if (responses.length === 0) return Effect.void
+        return Effect.gen(function*() {
```
tim-smart: This can be reverted to whileLoop

PR #7493, opened 2026-08-27
Control SQL span propagation with a context reference

utopyin:
Add `Statement.SpanPropagationEnabled`, a public `Context.Reference<boolean>` that defaults to `false`. It is read at execution time, so callers can enable or disable propagation for every client executing in the current scope, including existing statements and streams.

```ts
import { Statement } from "effect/unstable/sql"

query.pipe(Effect.provideService(Statement.SpanPropagationEnabled, true))
```

Nested regions can provide `false`. Overrides restore on exit and remain isolated between concurrent fibers. When enabled, driver calls run under `sql.execute`, including connection acquisition and stream pulls. Both paths retain the tracer-enabled guard and use `provideService(Tracer.ParentSpan)` without stack capture. The constructor option and propagation-specific cache split are removed.

Three focused tracing tests cover default/override behavior, acquisition/execution/pull parenting without stack frames, and the tracing-disabled guard. The three pre-existing Statement tests are preserved. All 51 SQL tests, root typechecking, and lint pass. JSDoc checks report no new diagnostics compared with main.

Benchmarks compare implementation `509541e0db` with main `7e24e3a5ef` in fresh processes pinned to one CPU (AMD EPYC 4344P, Node 26.7.0, Tinybench 6.1.4). Five alternating rounds use 500 ms warmup and 2000 ms measurement per process. Values are median ns per statement or complete one-chunk stream. The reference is provided after client construction.

| Case | Main | Default | Explicit false | True |
|---|---:|---:|---:|---:|
| acquirer, traced | 1387 | 1409 | 1414 | 1865 |
| acquirer, untraced | 986 | 983 | 1002 | 1001 |
| borrower, traced | 1185 | 1223 | 1215 | 1647 |
| borrower, untraced | 782 | 789 | 797 | 803 |
| stream, traced | 3197 | 3227 | 3254 | 5389 |
| stream, untraced | 2500 | 2500 | 2518 | 2510 |

Paired median deltas for default behavior are +20–30 ns per traced statement and -1 to +2 ns with tracing disabled. Enabled propagation adds +451–476 ns per traced statement. Explicit false adds +17–34 ns per traced statement; with tracing disabled, false/true overrides add +14–21 ns.

Stream overhead grows with emitted chunks. Three additional rounds at 5 and 20 chunks give these median ns per complete traced stream:

| Chunks | Main | Default | Enabled |
|---|---:|---:|---:|
| 1 | 3197 | 3227 | 5389 |
| 5 | 3552 | 3557 | 7009 |
| 20 | 4739 | 4796 | 12833 |

The paired enabled-minus-default deltas estimate 1.85 µs fixed setup plus 0.31 µs per emitted chunk. All 138 processes completed; maximum per-process relative margin of error was 4.31%. These stub measurements isolate wrapper overhead, not real database or OpenTelemetry costs. Chunk cost is not row cost.

Closes EFF-1286

This PR includes AI-generated changes.

### DIFF
--- a/.changeset/sql-execute-parent-span.md
+++ b/.changeset/sql-execute-parent-span.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Run SQL connection execution under the `sql.execute` span so driver I/O and nested spans nest beneath the statement instead of the enclosing span.
--- a/.changeset/sql-execute-parent-span.md
+++ b/.changeset/sql-execute-parent-span.md
@@ -1,5 +1,0 @@
----
-"effect": patch
----
-
-Run SQL connection execution under the `sql.execute` span so driver I/O and nested spans nest beneath the statement instead of the enclosing span.
--- a/.changeset/sql-optional-span-propagation.md
+++ b/.changeset/sql-optional-span-propagation.md
@@ -2,4 +2,11 @@
 "effect": patch
 ---
 
-Add a `propagateSpan` option to `SqlClient.make` for parenting driver spans under `sql.execute`. Disabled by default.
+Add `Statement.SpanPropagationEnabled` to scope driver span parenting under `sql.execute` for any SQL client. Disabled by default.
+
+```ts
+import { Effect } from "effect"
+import { Statement } from "effect/unstable/sql"
+
+query.pipe(Effect.provideService(Statement.SpanPropagationEnabled, true))
+```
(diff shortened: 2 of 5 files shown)


tim-smart (changes requested):
This adds a 2-4x performance regression in acquisition so I don't think we want it as the default behavior.

tim-smart (approved):

utopyin:
thank you for making it happen @tim-smart

PR #7496, opened 2026-08-27
Expose final URL on HttpClientResponse

nikhilsnayak:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Expose `HttpClientResponse.url` as the resolved response URL, including query parameters and excluding the hash. Fetch and XMLHttpRequest preserve the underlying response URL after redirects. Node clients report the resolved request URL and follow redirects with `HttpClient.followRedirects`.

Web response fallbacks and server response adapters resolve query parameters too. `HttpApiTest` passes through the originating request. Server responses without a request expose an empty URL, as documented.

Includes a changeset, redirect coverage for Fetch and Node, and mocked XHR coverage for `responseURL` and its fallback. Existing assertions cover query parameters, encoded hashes, and fragment removal.

## Validation

136 targeted tests pass. `pnpm lint-fix` and `pnpm check` pass.

## Related

Closes EFF-1300

### DIFF
--- a/.changeset/resolved-client-urls.md
+++ b/.changeset/resolved-client-urls.md
@@ -1,0 +1,8 @@
+---
+"@effect/platform-browser": patch
+"@effect/platform-node": patch
+"effect": patch
+---
+
+Expose the final response URL on `HttpClientResponse`. Fetch and XMLHttpRequest clients now preserve the URL reported by
+the underlying response after redirects, while other clients report the URL of the request that produced the response.
--- a/.changeset/resolved-client-urls.md
+++ b/.changeset/resolved-client-urls.md
@@ -4,7 +4,5 @@
 "effect": patch
 ---
 
-Expose the resolved response URL on `HttpClientResponse`, including query parameters and excluding the hash. Fetch and
-XMLHttpRequest clients preserve the URL reported by the underlying response after redirects. Other clients report the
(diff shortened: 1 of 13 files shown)


tim-smart (changes requested):
Currently the fetch adaptor will return url parameters in the url, and undici + node:http will drop them. I think we should make them consistent.

I think dropping them everywhere is ok.

nikhilsnayak:
> Currently the fetch adaptor will return url parameters in the url, and undici + node:http will drop them. I think we should make them consistent.
> 
> I think dropping them everywhere is ok.

Agreed that the adapters should be consistent. MDN defines [`Response.url`](https://developer.mozilla.org/en-US/docs/Web/API/Response/url) as the final URL obtained after redirects. Query parameters are part of that URL, while [`XMLHttpRequest.responseURL`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseURL) explicitly strips only the fragment.

Would you be open to preserving query parameters across all adapters by reconstructing the Node response URL from `request.url` and `request.urlParams`? This would align Node with the existing Fetch and XHR behavior.

tim-smart (approved):

PR #7512, opened 2026-08-28
Add NodeSocketServer.makeTls

tim-smart:
Adds `NodeSocketServer.makeTls` / `layerTls` in `@effect/platform-node-shared`, mirroring the existing `make` / `layer`
TCP pair. `@effect/platform-node` and `@effect/platform-bun` pick them up through their re-exports.

- Built with `tls.createServer(options)`; `run` handlers fire on `secureConnection`, so a connection reaches the
  handler only after the handshake completes.
- `tlsClientError` destroys the offending socket and the server keeps listening. It does not fail the `SocketServer`.
- Listen errors and address mapping (Unix vs TCP) behave exactly as in `make`.
- Handlers still get `NodeSocket.NetSocket` for the connection. The `TLSSocket` extends `net.Socket`, so there is no
  new service. SNI, mTLS, and ALPN stay as plain Node options on `TlsOptions`.

The listen / pending / `run` loop moved into a private `makeNetServer` helper at the bottom of the module. `make` and
`makeTls` only pick the create-server call and the connection event. `make` is otherwise unchanged.

`NodeSocket.makeTls` (client side) is out of scope here and already landed in #7511.

## Test

The new echo round-trip lives in the `TLS` block of `packages/platform/node/test/NodeSocket.test.ts` rather than beside
the plaintext `make` echo test, so it reuses the self-signed key/cert fixtures #7511 added to that block instead of
duplicating a PEM blob in the same file. The client is `tls.connect` + `NodeSocket.fromDuplex` with `ca: [cert]`, not
`rejectUnauthorized: false`.

Closes EFF-963

### DIFF
--- a/.changeset/eff-963-node-socket-server-tls.md
+++ b/.changeset/eff-963-node-socket-server-tls.md
@@ -1,0 +1,12 @@
+---
+"@effect/platform-node-shared": patch
+"@effect/platform-node": patch
+"@effect/platform-bun": patch
+---
+
+Add `NodeSocketServer.makeTls` and `NodeSocketServer.layerTls` for TLS socket servers.
+
+These mirror `make` / `layer` but build the server with `tls.createServer`, so they take the full `tls.TlsOptions` set
+alongside the usual listen options: certificates (`cert` / `key`), trust anchors for client certificates (`ca`), ALPN
+protocols, and SNI callbacks. Handlers run once the handshake completes and receive the `tls.TLSSocket` as
+`NodeSocket.NetSocket`. Connections that fail the handshake are destroyed; the server keeps listening.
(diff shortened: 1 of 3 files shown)


.changeset/eff-963-node-socket-server-tls.md
```diff
@@ -0,0 +1,12 @@
...
+"@effect/platform-node": patch
+"@effect/platform-bun": patch
+---
+
+Add `NodeSocketServer.makeTls` and `NodeSocketServer.layerTls` for TLS socket servers.
+
+These mirror `make` / `layer` but build the server with `tls.createServer`, so they take the full `tls.TlsOptions` set
+alongside the usual listen options: certificates (`cert` / `key`), trust anchors for client certificates (`ca`), ALPN
+protocols, and SNI callbacks. Handlers run once the handshake completes and receive the `tls.TLSSocket` as
+`NodeSocket.NetSocket`. Connections that fail the handshake are destroyed; the server keeps listening.
```
tim-smart:
```suggestion
Add `NodeSocketServer.makeTls` and `NodeSocketServer.layerTls` for TLS socket servers.
```

PR #7514, opened 2026-08-28
Add in-place TLS socket upgrade

tim-smart:
Adds in-place TLS wrapping to the `Reader` acquired from `socket.reader`. Each reader exposes `{ pull, upgrade }`, keeping the upgrade operation in the same scoped connection region as its pull. `Socket.make` accepts that complete reader acquisition directly; transports that cannot wrap use `SocketUpgradeError.unsupported`.

The Node adapter keeps the current pull and writer while replacing the live duplex after the handshake. Inbound server sockets use the server TLS role; `makeNet` and raw `fromDuplex` sockets use the client role. Client upgrades can omit `key` and `cert`, and callers that need no other settings can call `upgrade()` without an options object. Server upgrades require both credentials. Node/Bun reject missing or one-sided server credentials with `SocketUpgradeError` before attempting the wrap.

The Deno adapter performs the client-side transition with `Deno.startTls`, transfers the live reader and writer locks, and waits for the TLS handshake before succeeding. Deno Unix connections reject the upgrade before releasing those locks, so the original connection remains usable. Handshake failures retain the original cause on both platforms.

Deno-specific limitations are documented on `fromConn`: `Deno.startTls` cannot send a client certificate, so `key`, `cert`, `passphrase`, and `requestCert` do not affect the upgrade. `rejectUnauthorized: false` maps to Deno's hostname-verification switch; certificate-chain validation still applies. Deno also provides neither an abort signal nor the `TlsConn` until a pending `startTls` resolves. Closing the original `TcpConn` after ownership transfers does not reliably close that pending connection, so this PR intentionally does not claim interruption cleanup for an in-flight Deno handshake.

`readerBytes` and `readerString` keep their pull-only return contract while adapting the new reader internally. Channel and stream adapters, WebSocket and TransformStream adapters, Node/Deno/Bun consumers, OpenAI realtime, RPC, tests, benchmarks, and socket type tests are updated for the new reader shape. The changeset covers `effect`, Node, Bun, and Deno platform packages.

Checks run:

- `nix develop -c pnpm vitest run packages/effect/test/unstable/socket/Socket.test.ts packages/platform/node/test/NodeSocket.test.ts` (35 passed)
- `nix develop -c deno run -A npm:vitest --run --project @effect/platform-deno test/DenoSocket.test.ts test/DenoSocketServer.test.ts` (16 passed)
- `nix develop -c pnpm tstyche --target '>=5.9' packages/effect/typetest/unstable/socket/Socket.tst.ts` (TS 5.9 and 6.0 passed)
- `nix develop -c pnpm check`
- `nix develop -c pnpm jsdocs --check` still reports existing repository-wide diagnostics; the updated `Reader` and `TlsUpgradeOptions` docs parse without diagnostics

Closes EFF-965
Closes EFF-973


packages/platform/node-shared/src/NodeSocket.ts
```diff
@@ -40,6 +42,11 @@ export class NetSocket extends Context.Service<NetSocket, Net.Socket>()(
   "@effect/platform-node/NodeSocket/NetSocket"
 ) {}
 
+const isReadonlyArray = <A>(value: A | ReadonlyArray<A>): value is ReadonlyArray<A> => globalThis.Array.isArray(value)
```
fubhy: I'm summoning Dillon if you leave this in.

fubhy (changes requested):

packages/platform/node-shared/src/NodeSocket.ts
mikearnaldi: Why? Why would you do that?

tim-smart: Removed both helpers. `Arr.ensure` now normalizes scalar and array inputs, and each value goes through `Buffer.from`, which accepts both `string` and `Uint8Array` while producing the type Node TLS expects.

packages/effect/src/unstable/socket/Socket.ts
```diff
@@ -121,24 +127,47 @@ export interface Writer {
...
   readonly reader: Socket["reader"]
   readonly writer: Socket["writer"]
+  readonly upgrade?: Socket["upgrade"] | undefined
 }): Socket =>
   Socket.of({
     [TypeId]: TypeId,
     reader: options.reader,
-    writer: options.writer
+    writer: options.writer,
+    upgrade: options.upgrade ?? (() => Effect.fail(new SocketError({ reason: new SocketUpgradeError({}) })))
```
IMax153: If `upgrade` is required here, shouldn't it just be required in the interface?

tim-smart: You aren't required to implement it, as non-tcp sockets don't support it.

packages/effect/src/unstable/socket/Socket.ts
```diff
@@ -159,22 +165,23 @@ export interface TlsUpgradeOptions {
  * @since 4.0.0
  */
 export const make = (options: {
-  readonly reader: Socket["reader"]
+  readonly reader: Effect.Effect<Reader["pull"], SocketError, Scope.Scope>
```
tim-smart: This should return a `Reader` now, and `upgrade` removed from the make options.

tim-smart: Updated in afba2dc8c: `Socket.make` now accepts a complete `Reader` acquisition and no longer has an `upgrade` option. Node attaches its implementation to the acquired reader; unsupported adapters return readers whose `upgrade` fails with `SocketUpgradeError`.

packages/effect/src/unstable/socket/Socket.ts
```diff
@@ -171,7 +171,15 @@ export const make = (options: {
...
 
-const unsupportedUpgrade: Reader["upgrade"] = () => Effect.fail(new SocketError({ reason: new SocketUpgradeError({}) }))
+/**
+ * A `Reader["upgrade"]` for transports that cannot wrap the connection with
+ * TLS. Always fails with a `SocketUpgradeError`.
+ *
+ * @category constructors
+ * @since 4.0.0
+ */
+export const unsupportedUpgrade: Reader["upgrade"] = () =>
```
tim-smart: I think this can be moved to a static readonly const on the SocketUpgradeError class.

packages/effect/src/unstable/socket/Socket.ts
```diff
@@ -156,12 +196,12 @@ const encoder = new TextEncoder()
 export const readerBytes = (
   self: Socket
 ): Effect.Effect<
-  Effect.Effect<NonEmptyReadonlyArray<Uint8Array>, SocketError>,
+  Reader<Uint8Array>,
```
tim-smart: I think this should still return the pull effect directly, instead of a Reader. Same with readerString etc

tim-smart: Restored in 64cc46b8d. `readerBytes` and `readerString` return the adapted pull effect directly again; channel/stream adapters, platform tests, benchmarks, and type tests are updated to match.

packages/effect/src/unstable/socket/Socket.ts
tim-smart: Updated in 64cc46b8d. The shared implementation is now `SocketUpgradeError.unsupported`, and the WebSocket, TransformStream, Deno, Bun, RPC, and test readers use that static constant.

fubhy (approved):

.changeset/eff-965-socket-upgrade.md
```diff
@@ -0,0 +1,14 @@
...
+"@effect/platform-bun": patch
+"@effect/platform-deno": patch
+---
+
+Add reader-scoped TLS upgrades for wrapping an acquired pull-based socket in place.
+
+Node TCP server connections upgrade with the server TLS role, while client-side Node and Deno TCP sockets upgrade
+with the client role. Unsupported transports and failed TLS handshakes now fail with `SocketUpgradeError`.
+Client upgrades no longer require `key` and `cert`; server upgrades require both and reject missing or incomplete
+credentials with `SocketUpgradeError`. Client upgrades that need no other settings can be called as `upgrade()`.
```
tim-smart:
```suggestion
Add Socket.upgrade, for upgrading tcp sockets using STARTTLS
```

PR #7516, opened 2026-08-28
fix(RcMap): preserve replacement entries during invalidated resource cleanup

### LINKED ISSUE #7515: RcMap active invalidation can remove replacement entries or leak resources (v4)
Invalidating an actively borrowed `RcMap` entry and reacquiring the same key breaks cleanup of the old entry. With zero idle TTL, releasing the old borrower removes the replacement mapping. With infinite idle TTL, the old resource never finalizes, even after both borrowers and the map owner close.

Reproduced on `effect@4.0.0-rc.112` and `main` at `6e3ae7b63`, using a plain string key and trivial resources.

## Reproducer

```ts
import { Effect, Exit, RcMap, Scope } from "effect"

for (const ttl of [0, Infinity]) {
  await Effect.runPromise(Effect.gen(function*() {
    const owner = yield* Scope.make()
    const oldBorrower = yield* Scope.make()
    const newBorrower = yield* Scope.make()
    let acquired = 0
    const released: Array<number> = []
    const map = yield* RcMap.make({
      lookup: (_key: string) => Effect.acquireRelease(
        Effect.sync(() => ++acquired),
        (id) => Effect.sync(() => released.push(id))
      ),
      idleTimeToLive: ttl
    }).pipe(Scope.provide(owner))

    yield* RcMap.get(map, "key").pipe(Scope.provide(oldBorrower))
    yield* RcMap.invalidate(map, "key")
    yield* RcMap.get(map, "key").pipe(Scope.provide(newBorrower))
    yield* Scope.close(oldBorrower, Exit.void)
    const hasReplacement = yield* RcMap.has(map, "key")
    const releasedAfterOld = [...released]
    yield* Scope.close(newBorrower, Exit.void)
    yield* Scope.close(owner, Exit.void)

    console.log({ ttl, acquired, releasedAfterOld, hasReplacement, released })
  }))
}
```

## Actual Behavior

Both cases acquire two resources.

| Idle TTL | Released after old borrower closes | Replacement mapped | Released after all scopes close |
|---|---|---|---|
| `0` | `[1]` | `false` | `[1, 2]` |
| `Infinity` | `[]` | `true` | `[2]` |

## Expected Behavior

The invalidated resource stays alive while borrowed, then finalizes when its last borrower closes. Its cleanup must not remove the replacement entry.

For both cases above, `releasedAfterOld` should be `[1]`, `hasReplacement` should be `true`, and the final `released` array should be `[1, 2]`. The replacement should retain its own configured TTL behavior.

## Related Cleanup Path

An old finite-TTL idle timer can also remove a replacement while that replacement is still borrowed. This occurs when an entry becomes idle, is borrowed again, then invalidated and replaced before its old timer expires.

Borrower release currently checks whether the key exists rather than whether it still maps to the releasing entry. Idle expiry also removes by key without checking entry identity. `LayerMap` delegates these operations to `RcMap`.

kitlangton:
## Why

Invalidating a borrowed resource and reacquiring its key lets old cleanup affect the replacement. Zero TTL removes the replacement mapping; infinite TTL leaves the detached old resource unfinalized. An existing finite idle timer can independently remove the replacement.

Closes #7515.
Closes EFF-974.

## What Changes

| Cleanup event | New behavior |
|---|---|
| Invalidated entry still has borrowers | Keep the old resource alive |
| Last borrower of an invalidated entry closes | Finalize only that resource, regardless of idle TTL |
| Last borrower of the current entry closes | Preserve existing TTL behavior |
| Old idle timer expires after replacement | Leave the replacement untouched |

Release and idle expiry check that the key still maps to the entry being cleaned up before removing it.

## Scope

A focused RcMap lifecycle fix, with regression coverage for zero, finite, and infinite TTL and a surviving old idle timer. Includes an `effect: patch` changeset covering the consumer-visible RcMap and LayerMap behavior. No public API changes.

## Verification

```sh
pnpm test --run packages/effect/test/RcMap.test.ts packages/effect/test/LayerMap.test.ts
pnpm check
pnpm exec oxlint -f unix packages/effect/src/RcMap.ts packages/effect/test/RcMap.test.ts
pnpm exec dprint check packages/effect/src/RcMap.ts packages/effect/test/RcMap.test.ts .changeset/rcmap-invalidation-release.md
pnpm exec changeset status
git diff --check
```

All four new regression cases failed before the fix. Fixing release alone passed the three TTL cases but left the stale-timer regression failing; the expiry identity check resolved that remaining failure.

Final result: 25 tests pass, comprising 19 RcMap and 6 LayerMap tests. Type checking, lint, formatting, and changeset validation pass. The standalone reproducer produces the expected results with the fix and no manual resource cleanup.

The full test suite and OpenCode integration were not run.

### DIFF
--- a/packages/effect/src/RcMap.ts
+++ b/packages/effect/src/RcMap.ts
@@ -274,7 +274,7 @@ export const make: {
         self.state = { _tag: "Closed" }
         return Effect.forEach(
           map,
-          ([, entry]) => Effect.exit(Scope.close(entry.scope, Exit.void))
+          ([, entry]) => Effect.exit(closeEntry(entry))
         ).pipe(
           Effect.tap(() =>
             Effect.sync(() => {
@@ -449,21 +449,26 @@ export const getOption: {
     })
 )
 
+const closeEntry = <A, E>(entry: State.Entry<A, E>) =>
+  entry.fiber
+    ? Fiber.interrupt(entry.fiber).pipe(Effect.andThen(Scope.close(entry.scope, Exit.void)))
+    : Scope.close(entry.scope, Exit.void)
+
 const release = <K, A, E>(self: RcMap<K, A, E>, key: K, entry: State.Entry<A, E>) =>
   Effect.withFiber((fiber) => {
     entry.refCount--
     if (entry.refCount > 0) {
       return Effect.void
     } else if (self.state._tag === "Closed") {
-      return Scope.close(entry.scope, Exit.void)
+      return closeEntry(entry)
     }
 
     const o = MutableHashMap.get(self.state.map, key)
     if (o._tag === "None" || o.value !== entry) {
-      return Scope.close(entry.scope, Exit.void)
+      return closeEntry(entry)
     } else if (Duration.isZero(entry.idleTimeToLive)) {
       MutableHashMap.remove(self.state.map, key)
-      return Scope.close(entry.scope, Exit.void)
+      return closeEntry(entry)
     } else if (!Duration.isFinite(entry.idleTimeToLive)) {
       return Effect.void
     }
@@ -595,8 +600,7 @@ export const invalidate: {
     const entry = o.value
     MutableHashMap.remove(self.state.map, key)
     if (entry.refCount > 0) return
-    if (entry.fiber) yield* Fiber.interrupt(entry.fiber)
-    yield* Scope.close(entry.scope, Exit.void)
+    yield* closeEntry(entry)
   }, Effect.uninterruptible)
 )
 
--- a/packages/effect/src/RcMap.ts
+++ b/packages/effect/src/RcMap.ts
@@ -454,14 +454,15 @@ const release = <K, A, E>(self: RcMap<K, A, E>, key: K, entry: State.Entry<A, E>
     entry.refCount--
     if (entry.refCount > 0) {
       return Effect.void
-    } else if (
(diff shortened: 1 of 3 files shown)


packages/effect/src/RcMap.ts
```diff
@@ -454,14 +454,15 @@ const release = <K, A, E>(self: RcMap<K, A, E>, key: K, entry: State.Entry<A, E>
...
-      if (self.state._tag === "Open") {
-        MutableHashMap.remove(self.state.map, key)
-      }
+    } else if (self.state._tag === "Closed") {
+      return Scope.close(entry.scope, Exit.void)
+    }
+
+    const o = MutableHashMap.get(self.state.map, key)
+    if (o._tag === "None" || o.value !== entry) {
+      return Scope.close(entry.scope, Exit.void)
```
tim-smart: Would this leak any running fibers?

tim-smart: Yes. A detached entry with an already-running finite-TTL timer could leave that fiber asleep until its old deadline after the resource scope was closed. Commit 76ecd2d40 routes external entry shutdown through closeEntry, which interrupts the idle fiber before closing the scope; the timer-expiry path still terminates naturally from inside that fiber. I also extended the stale-timer regression to assert immediate interruption. The assertion fails on the prior PR head and passes with this commit.

PR #7520, opened 2026-08-28
Report MCP toolkit failures before recovery

ebramanti:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`McpServer.registerToolkit` converts tool failures and defects into safe MCP error results before the outer HTTP or RPC reporting boundaries can observe their original `Cause`. As a result, configured `ErrorReporter`s do not receive failures that the MCP server subsequently recovers.

This change:

- reports original causes from tool execution and result serialization immediately before recovery produces a tool error result
- leaves parameter-validation failures to the existing protocol reporting boundary
- runs reporting with the context captured during toolkit registration
- preserves existing public `CallToolResult` messages and protocol behavior
- adds coverage for declared failures, defects, response-serialization defects, and successful calls
- includes a patch changeset

Validation:

- `pnpm test --run packages/effect/test/unstable/ai/McpServer/McpServer.test.ts`
- `pnpm lint-fix`
- `pnpm check`

## Related

- Related: https://github.com/Effect-TS/effect/pull/6607
- Closes EFF-1316

### DIFF
--- a/packages/effect/src/unstable/ai/McpServer.ts
+++ b/packages/effect/src/unstable/ai/McpServer.ts
@@ -1573,9 +1573,6 @@ export const registerToolkit: <Tools extends Record<string, Tool.Any>>(
           Stream.unwrap,
           Stream.run(Sink.last()),
           Effect.flatMap(Effect.fromOption),
-          Effect.provideContext(
-            services as Context.Context<Tool.HandlerServices<Tools[keyof Tools]>>
-          ),
           Effect.map((result) =>
             new CallToolResult({
               isError: false,
@@ -1585,6 +1582,10 @@ export const registerToolkit: <Tools extends Record<string, Tool.Any>>(
                 text: JSON.stringify(result.encodedResult)
               }]
             })
+          ),
+          Effect.withErrorReporting,
+          Effect.provideContext(
+            services as Context.Context<Tool.HandlerServices<Tools[keyof Tools]>>
           ),
           Effect.tapCause(Effect.logError),
           Effect.catch((error) => {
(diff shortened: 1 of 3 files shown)


IMax153:
@tim-smart - is `withErrorReporting` intended to be utilized from within library-level code? 

Just want to make sure before we approve / merge this.

tim-smart:
> @tim-smart - is `withErrorReporting` intended to be utilized from within library-level code?
> 
> Just want to make sure before we approve / merge this.

Yes it is used by our rpc and http server etc

packages/effect/src/unstable/ai/McpServer.ts
```diff
@@ -1590,24 +1591,22 @@ export const registerToolkit: <Tools extends Record<string, Tool.Any>>(
...
+                ? Effect.as(reportCause(cause), toolErrorResult(INTERNAL_TOOL_ERROR_MESSAGE))
+                : Effect.failCause(failure.failure)
             }
-            return ErrorReporter.report(cause).pipe(
-              Effect.provideContext(services),
-              Effect.as(toolErrorResult(message))
-            )
+            const error = failure.success
+            if (AiError.isAiError(error)) {
+              const reason = (error as AiError.AiError).reason
```
tim-smart: Why is this cast need after the `isAiError` check?

tim-smart: Removing the cast alone gives TS2339: the guard leaves `AiError | HandlerError<Tools[keyof Tools]>`, where `HandlerError` is a conditional type over the generic tools. The compiler still rejects `.reason` on that unresolved member.

Removed the cast in `dea9ef71a8` by declaring the local `error` as `unknown`. The existing `isAiError` guard then narrows it to `AiError`, so `error.reason` works without an assertion. No runtime change. All 27 McpServer tests, lint, and typecheck pass.

PR #7521, opened 2026-08-29
feat(bun-test): add @effect/bun-test package

emilienbidet:
## Summary

Adds a new `@effect/bun-test` workspace package that mirrors the `@effect/vitest` API surface (`it.effect`, `it.live`, `layer`, `it.prop`, `flakyTest`, `utils`, …) but runs on Bun's built-in [`bun:test`](https://bun.sh/docs/cli/test) runner instead of Vitest.

This is the replacement for Effect-TS/effect-smol#2204, reopened here after the v4 codebase was merged back into this repository (as requested by @IMax153 in that thread). It addresses the same request as #5964 (and supersedes the v3 attempt in #5973). The code has been re-ported against the current `main` — notably the property-testing layer now uses `effect/unstable/arbitrary/Arbitrary` (`Arbitrary.checkEffect`, `CheckOptions`, Schema + Arbitrary mixed inputs) exactly like the current `@effect/vitest`, instead of the FastCheck API the effect-smol PR used.

## What's in the package

`packages/bun-test/` mirrors `packages/vitest/`:

- `src/index.ts` — public API and types, re-exports of the `bun:test` primitives, plus a small `node:assert`-backed `assert` object covering the chai surface Effect suites use from Vitest
- `src/internal/internal.ts` — Effect runner + tester/layer/prop wrappers (`Effect.fnUntraced`, `Layer.buildWithMemoMap`, `Layer.forkMemoMapUnsafe` for nested layers, `Layer.mergeAll(TestConsole.layer, TestClock.layer())`, `Arbitrary.checkEffect`)
- `src/utils.ts` — the `@effect/vitest/utils` assertion helpers (`vassert.instanceOf` swapped for a plain `instanceof` check)
- `test/index.test.ts` — ported from `packages/vitest/test/index.test.ts`
- workspace wiring: a `tsconfig.packages.json` reference and `tsconfig.tests.json` path aliases; a changeset is included

## Improvement over the effect-smol PR: timeouts interrupt fibers

Bun's own test timeout fails the test but cannot stop the Effect running behind it, so finalizers would never run. The wrapper now owns the timeout: when it fires, the synthesized test context's `AbortSignal` aborts, the fiber is interrupted via `Effect.runPromise(effect, { signal })`, and finalizers run — Bun keeps a slightly larger timeout as a backstop. This let the two timeout-interruption tests from the vitest suite (`interrupts on timeout`, `interrupts property checking on timeout`) be ported instead of dropped.

## Differences from `@effect/vitest` (documented in the package README)

- `addEqualityTesters` is a **no-op** — `bun:test`'s `expect` has no `addEqualityTesters` API.
- **`TestContext`** is synthesized inside the test wrapper (`signal`, `onTestFinished`, `onTestFailed`), because `bun:test` doesn't pass a context object.
- **`export * from "bun:test"`** doesn't work in Bun, so the public primitives are re-exported via `const` bindings.
- `it.skip.each` is reimplemented (with `%s`-style title interpolation) since Bun doesn't chain those registrars.

## Test plan

- [x] `pnpm install` at the workspace root resolves cleanly
- [x] `pnpm --filter=@effect/bun-test check` and `build` (tsc + babel) pass
- [x] `pnpm check` (repo-wide `tsc -b`) passes, including the swept-in test file
- [x] `pnpm exec dprint check` and `pnpm exec oxlint` pass
- [x] `bun test` inside `packages/bun-test`: **40 pass, 5 skip (deliberate skip/todo cases), 0 fail**
- [ ] Maintainers to decide whether CI should grow a Bun job for this package (the root vitest config's explicit project list is untouched, so the vitest run does not pick these tests up)

https://claude.ai/code/session_01TBYAaLaKUsUBsfGKV5jGpn

### DIFF
--- a/.changeset/add-effect-bun-test.md
+++ b/.changeset/add-effect-bun-test.md
@@ -1,0 +1,5 @@
+---
+"@effect/bun-test": patch
+---
+
+Add the `@effect/bun-test` package: the `@effect/vitest` API (`it.effect`, `it.live`, `layer`, `it.prop`, `flakyTest`, `utils`) on Bun's native `bun:test` runner. Wrapper-managed timeouts abort the synthesized test context's `AbortSignal`, so Effect fibers are interrupted and their finalizers run on timeout.
--- a/deno.json
+++ b/deno.json
@@ -23,6 +23,7 @@
     "packages/*/benchmark/",
     "packages/ai",
     "packages/atom",
+    "packages/bun-test/",
     "packages/effect/test/cluster/",
     "packages/opentelemetry/",
     "packages/platform/browser/",
--- a/packages/bun-test/CHANGELOG.md
+++ b/packages/bun-test/CHANGELOG.md
@@ -1,0 +1,1 @@
+# @effect/bun-test
--- a/packages/bun-test/LICENSE
+++ b/packages/bun-test/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Effectful Technologies Inc
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/packages/bun-test/README.md
+++ b/packages/bun-test/README.md
@@ -1,0 +1,59 @@
+# @effect/bun-test
+
+A set of helpers for testing [Effect](https://effect.website) programs with
+Bun's native [`bun:test`](https://bun.sh/docs/cli/test) runner.
+
+The API mirrors [`@effect/vitest`](https://www.npmjs.com/package/@effect/vitest)
+(`it.effect`, `it.live`, `layer`, `it.prop`, `flakyTest`, …) so Effect test
+suites move between the two runners without rewrites.
+
+## Installation
+
+```sh
+bun add -d @effect/bun-test
+```
+
+## Usage
+
+```ts
+import { assert, describe, expect, it, layer } from "@effect/bun-test"
+import { Context, Effect, Layer } from "effect"
+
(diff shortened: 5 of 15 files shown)


emilienbidet:
**A real-world data point for why this package is worth having**: we just finished migrating our production monorepo (tries.care — ~16 workspaces, Effect v4 RC end to end) from Vitest to `bun test`. Numbers from that migration:

- **CI test job: 2m00s → 26s** — and most of the remaining 26s is checkout + dependency install, not tests.
- **Local suites**: the domain package runs 1,927 Effect-heavy tests in 1.1s; two more workspaces run 524 tests in 1.5s and 414 React component tests (happy-dom) in 3s. Vitest's per-run startup and transform cost dominated every one of these; bun test compiles TypeScript natively with no worker spin-up.
- **A whole compatibility surface disappeared with the runner**: `vitest` + `@vitest/coverage-v8` + `jsdom`, the jest-dom type-augmentation shims, per-file `// @vitest-environment` pragmas (replaced by one happy-dom preload per workspace), and a Node `engines` pin we only carried because Node 25's built-in Web Storage broke jsdom.
- **Tests run under the runtime that runs production.** No Node forked workers with shimmed `fetch` / `import.meta.env` / Web APIs — which for Effect code also means the ConfigProvider, fetch-based layers, and platform services behave in tests exactly as they do at runtime.

Trade-offs we accepted, for completeness: bun's coverage is lines + functions only, modules no test imports are invisible to coverage, and one suite that needs the real Vite module graph (MDX pages, `import.meta.glob`) stays on Vitest as a per-file escape hatch.

The missing piece after such a migration is exactly what this PR provides: the `@effect/vitest` API (`it.effect`, `TestClock`, `prop`) on top of `bun:test`. (Correction from this comment's first version: bun *does* ship timer mocking since 1.3.4 — `jest.useFakeTimers` & co. But faking the JS timer globals is not `TestClock`: Effect code sleeps through the `Clock` service across fibers, and only `it.effect` + `TestClock` gives deterministic virtual time with Effect semantics.) Until this lands we're covering that gap with the same API published as `@voila.dev/effect-bun-test` against `effect@4.0.0-rc`.

eugene1g:
> * **A whole compatibility surface disappeared with the runner**: `vitest` + `@vitest/coverage-v8` + `jsdom`, the jest-dom type-augmentation shims, per-file `// @vitest-environment` pragmas (replaced by one happy-dom preload per workspace), and a Node `engines` pin we only carried because Node 25's built-in Web Storage broke jsdom.
> * **Tests run under the runtime that runs production.** No Node forked workers with shimmed `fetch` / `import.meta.env` / Web APIs — which for Effect code also means the ConfigProvider, fetch-based layers, and platform services behave in tests exactly as they do at runtime.

My comment is unrelated to the PR, but I'm compelled to leave a note in case other folks are looking for the same type of optimization - 

Vitest adds a lot of facade for backward compatibility, but if your code is fairly clean (e.g., fullstack Effect), you can tell Vitest to skip all that boilerplate with these options:
* [experimental.viteModuleRunner](https://vitest.dev/guide/cli.html#experimental-vitemodulerunner)= `false`
* [experimental.nodeLoader](https://vitest.dev/guide/cli.html#experimental-nodeloader) = `false`
* [isolate](https://vitest.dev/guide/cli.html#isolate) = `false`
* [pool](https://vitest.dev/guide/cli.html#pool) = `threads`

This makes Vitest do almost no additional processing of your code and run everything in the same process, which drastically speeds up the test suite and makes it run within a negligible time difference vs Bun, so you can stay in the same runtime that you use for production.

Again, the PR is useful on its own, but some folks with clean code might be underestimating the performance of Node 26 + Vitest out of the box because so there are many wrappers enabled by default.

tim-smart:
For now you can maintain it as a community package :) At this stage we will only officially support the vitest library.

emilienbidet:
I will maintain it as community package then

wommy:
have you thought about a codemod to transform it, I'm guessing bun is only a matter of time until support lands esp around v8 <> jsc interop around https://github.com/oven-sh/bun/pull/38282  from at least unblocking it - tagging the real blocker will lead to eyes

PR #7522, opened 2026-08-29
Fix Set-Cookie name validation

gwagjiug:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Cookies.fromSetCookie` validated parsed cookie names with `fieldContentRegExp`, which allowed invalid names containing characters such as spaces. This was inconsistent with `Cookies.makeCookie`, which already validates names using the RFC 6265 token syntax.

This PR:

- uses `cookieNameRegExp` when parsing `Set-Cookie` headers;
- ignores cookies with invalid names;
- preserves valid cookies received alongside invalid ones;
- adds focused regression coverage;
- adds an `effect` patch changeset.

### Validation

- `pnpm test --run packages/effect/test/unstable/http/Cookies.test.ts` — 10 tests passed
- `NODE_OPTIONS=--experimental-strip-types pnpm lint-fix`
- `pnpm check`

## Related

No related issue.

### DIFF
--- a/.changeset/strict-cookie-names.md
+++ b/.changeset/strict-cookie-names.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Ignore `Set-Cookie` headers whose cookie names do not satisfy the RFC 6265 token syntax.
--- a/packages/effect/src/unstable/http/Cookies.ts
+++ b/packages/effect/src/unstable/http/Cookies.ts
@@ -288,7 +288,7 @@ function parseSetCookie(header: string): Cookie | undefined {
     return undefined
   }
   const name = parts[0].slice(0, firstEqual)
-  if (!fieldContentRegExp.test(name)) {
+  if (!cookieNameRegExp.test(name)) {
     return undefined
   }
 
--- a/packages/effect/test/unstable/http/Cookies.test.ts
+++ b/packages/effect/test/unstable/http/Cookies.test.ts
@@ -40,6 +40,18 @@ describe("Cookies", () => {
     })
   })
 
+  describe("fromSetCookie", () => {
+    it("ignores invalid cookie names", () => {
+      const cookies = Cookies.fromSetCookie([
+        "bad name=value",
+        "session=abc"
+      ])
+
+      assertNone(Cookies.get(cookies, "bad name"))
+      assertSome(Cookies.getValue(cookies, "session"), "abc")
+    })
+  })
+
   describe("toSetCookieHeaders", () => {
     const invalidCookie = {
       name: "session",


gwagjiug:
<html><head><meta charset="utf-8"></head><body><h2>PR Change Explanation</h2>
<blockquote>
<p>This PR corrects the validation rule applied to cookie names when converting HTTP response <code>Set-Cookie</code> headers into a <code>Cookies</code> collection. The core change is to apply the same RFC 6265 token syntax already used by <code>Cookies.makeCookie</code> and the serialization path to cookies parsed from responses. As a result, only entries with invalid names are ignored, while valid cookies from the same response are preserved.</p>
</blockquote>
<h3>Background</h3>
<h4>Background for readers new to this area</h4>
<p>HTTP cookies use different headers depending on the direction of communication. A server sends cookies through <code>Set-Cookie</code> response headers, and a client later returns stored values through the <code>Cookie</code> request header. This PR concerns the former boundary: converting string-based <code>Set-Cookie</code> response values into Effect’s structured cookie representation.</p>
<p>An Effect <code>Cookies</code> value wraps a readonly record keyed by cookie name. <code>Cookies.fromSetCookie</code> iterates over one or more <code>Set-Cookie</code> strings, parses each one with the internal <code>parseSetCookie</code> function, and passes successfully parsed <code>Cookie</code> values to <code>fromIterable</code> to build that record. This path is used by the Web <code>Response</code>-based <code>HttpClientResponse.cookies</code>, the Node HTTP client’s <code>cookies</code> getter, and JSON decoding through <code>CookiesSchema</code>.</p>
<p>Cookie names are not arbitrary strings. This module represents the RFC 6265 token character set allowed in cookie names with <code>cookieNameRegExp</code>. <code>Cookies.makeCookie</code>, which creates cookies directly, and <code>serializeCookie</code>, which converts them back into <code>Set-Cookie</code> strings, already used this rule through <code>validateCookie</code>.</p>
<h4>Background specific to this change</h4>
<p>Before this change, only <code>parseSetCookie</code> validated names with <code>fieldContentRegExp</code>. That regular expression accepts a broad range of header characters, including tabs and spaces, so a name containing a space, such as <code>bad name</code> in <code>bad name=value</code>, was successfully parsed.</p>
<p>As a result, the same cookie name behaved differently depending on how it entered the system:</p>
<ul>
<li><code>Cookies.makeCookie("bad name", "value")</code> rejects the name with <code>InvalidCookieName</code>.</li>
<li>Before this change, <code>Cookies.fromSetCookie("bad name=value")</code> added the cookie to the collection.</li>
<li>If that collection was serialized again with <code>toSetCookieHeaders</code>, <code>serializeCookie</code> validated the same name later in the flow.</li>
</ul>
<p>In other words, the input parser was more permissive than the construction and serialization paths, so the cookie-name invariant inside a <code>Cookies</code> collection could vary depending on how the cookie was created.</p>
<h3>Intuition</h3>
<p>The core idea is to apply the same cookie-name rule before a cookie enters the collection, regardless of where it came from. There is no reason for directly created cookies and cookies read from HTTP responses to use different name rules, so the parser now reuses the existing <code>cookieNameRegExp</code>.</p>
<h4>Small example</h4>
<p>Suppose a server sends these two headers:</p>
<pre><code class="language-http">Set-Cookie: bad name=value
Set-Cookie: session=abc
</code></pre>

Step | Before | After
-- | -- | --
Parse bad name=value | Passes the broad fieldContentRegExp | Fails cookieNameRegExp
Parse session=abc | Succeeds | Succeeds
Final collection | Contains both bad name and session | Contains only session


<p>The flow after this change is:</p>
<pre><code class="language-mermaid">flowchart LR
  A[Set-Cookie header list] --&gt; B[Split each header into name and value]
  B --&gt; C{Does the name satisfy the RFC 6265 token syntax?}
  C -- No --&gt; D[Ignore only that entry]
  C -- Yes --&gt; E[Create Cookie]
  E --&gt; F[Store in Cookies collection]
</code></pre>
<p>When <code>parseSetCookie</code> returns <code>undefined</code>, the existing <code>fromSetCookie</code> loop simply does not add that entry. Therefore, this change does not require a new error channel or additional control flow. It reuses the existing behavior of skipping entries that cannot be parsed.</p>
<h3>Code walkthrough</h3>
<h4>1. Unifying the cookie-name invariant at the input boundary</h4>
<p>In <code>packages/effect/src/unstable/http/Cookies.ts</code>, <code>parseSetCookie</code> extracts the cookie name from the portion before the first <code>=</code>. This PR changes the following validation step to:</p>
<pre><code class="language-ts">if (!cookieNameRegExp.test(name)) {
  return undefined
}
</code></pre>
<p>With this one-line change, the response parser uses the same name rule as <code>validateCookie</code>. If validation fails, <code>parseSetCookie</code> returns <code>undefined</code> as before, and its caller, <code>fromSetCookie</code>, does not add that entry to the <code>cookies</code> array. The parsing behavior for values and attributes such as <code>Domain</code>, <code>Path</code>, <code>Max-Age</code>, <code>HttpOnly</code>, <code>Secure</code>, and <code>SameSite</code> remains unchanged.</p>
<h4>2. Giving response consumers and Schema paths the same collection contract</h4>
<p><code>HttpClientResponse.cookies</code> and the Node HTTP client’s <code>cookies</code> getter pass response <code>Set-Cookie</code> values to <code>Cookies.fromSetCookie</code> and cache the result. After this change, entries with invalid names are excluded at the input boundary before these consumers receive the resulting <code>Cookies</code> collection.</p>
<p><code>CookiesSchema</code> also uses <code>fromSetCookie</code> when decoding an array of JSON strings and <code>toSetCookieHeaders</code> when encoding. Because parsing now uses the same <code>cookieNameRegExp</code> as <code>makeCookie</code> and <code>serializeCookie</code>, decoding and subsequent encoding share the same cookie-name validation rule. No public types or function signatures are changed.</p>
<h4>3. The observable contract established by the regression test</h4>
<p>The new <code>fromSetCookie</code> test in <code>packages/effect/test/unstable/http/Cookies.test.ts</code> provides an invalid cookie and a valid cookie in the same input list.</p>
<ul>
<li>It verifies that <code>Cookies.get(cookies, "bad name")</code> returns <code>None</code>, establishing that an invalid cookie name does not enter the collection.</li>
<li>It verifies that <code>Cookies.getValue(cookies, "session")</code> returns <code>"abc"</code>, establishing that one invalid entry does not cause valid entries from the same input to be discarded.</li>
</ul>
<p><code>.changeset/strict-cookie-names.md</code> records this consumer-visible behavior as a patch change for the <code>effect</code> package. No API usage or migration is required; the release note documents that invalid cookie names previously accepted from <code>Set-Cookie</code> headers are now ignored.</p>
<p>&lt;!-- explain-my-pr:v1 key=Effect-TS/effect#7522 head=1d729fc33ce1a9f28f9305efa86eb2e1f71d0e93 part=1/1 --&gt;</p>
</body></html>

tim-smart (approved):

PR #7524, opened 2026-08-30
Add shared network address values

fubhy:
## Summary

- add platform-neutral MAC, IPv4, IPv6, resolved internet, Unix path, and portable socket address values under `effect/unstable/net/NetAddress`
- add `IpInterface` values that preserve host bits and `IpNetwork` values for canonical IPv4 and IPv6 CIDR networks
- store MAC, IPv4, and IPv6 values internally as implementation-private network-order byte arrays, copying constructor inputs, including Node.js Buffers, to preserve immutable value semantics
- add strict parsing, canonical formatting, scope identifiers, common constants and predicates, and checked and throwing construction paths
- add `SocketAddress.Input` with explicit checked and unsafe normalization for existing addresses, numeric IP plus port objects, and Unix path objects
- require canonical `SocketAddress` values at `HttpServer.make` and preserve canonical addresses on HTTP and socket server services
- add `effect/Schema` codecs and migrate HTTP/socket server address models plus Node, Deno, and Bun adapters
- keep native address conversion failures in typed server-open channels and bracket IPv6 authorities in HTTP, WebSocket, cluster, and logging consumers
- preserve address families when test clients replace unspecified listener addresses with loopback addresses
- omit IPv6 scope identifiers from WHATWG URL-based HTTP formatting because scoped IPv6 hosts are not supported by that URL model
- preserve Bun listen-option compatibility while reporting hostname-resolution failures through the typed `ServeError` channel before startup
- migrate PostgreSQL `inet` to `IpInterface` and `cidr` to canonical `IpNetwork` validation; parsing now uses the shared strict IP syntax, including rejection of leading-zero IPv4 octets
- include migration annotations and consumer release changesets

## Validation

- `pnpm lint-fix` and `pnpm lint`
- `pnpm check`
- `pnpm test-types NetAddress.tst.ts`
- `pnpm test --run packages/effect/test/unstable/net/NetAddress.test.ts packages/effect/test/unstable/net/IpInterface.test.ts packages/effect/test/unstable/net/IpNetwork.test.ts` — 46 tests pass, including a Buffer mutation regression verified to fail before the fix
- focused HttpServer tests
- focused Node HTTP and socket tests
- focused Bun HTTP server tests under Bun
- focused Deno HTTP and socket server tests under Deno
- Bun package check
- Deno package check

`pnpm jsdocs --check` remains blocked by existing diagnostics in `packages/effect/src/unstable/http/Multipart.ts` and `packages/effect/src/unstable/process/ChildProcess.ts`; no diagnostics reference the changed NetAddress or HttpServer declarations.


packages/effect/src/internal/netAddress.ts
```diff
@@ -0,0 +1,1110 @@
+/**
```
tim-smart, line 1: NetAddress is a better name I think, as the module doesn't provide any actual network connection apis.

packages/effect/src/unstable/net/Net.ts
```diff
@@ -0,0 +1,1110 @@
...
+const InetV6TypeId = "~effect/net/InetAddressV6"
+const UnixPathTypeId = "~effect/net/UnixPathAddress"
+
+/**
+ * A checked network-address operation failure.
+ *
+ * @category errors
+ * @since 4.0.0
+ */
+export class AddressError extends Data.TaggedError("NetAddressError")<{
```
tim-smart:
```suggestion
export class NetAddressError extends Data.TaggedError("NetAddressError")<{
```

tim-smart: Can it be a Schema.TaggedError?

packages/effect/src/unstable/net/Net.ts
```diff
@@ -0,0 +1,1110 @@
...
+import * as Data from "../../Data.ts"
+import * as Equal from "../../Equal.ts"
+import { dual } from "../../Function.ts"
+import * as Hash from "../../Hash.ts"
+import { NodeInspectSymbol } from "../../Inspectable.ts"
+import * as Option from "../../Option.ts"
+import { hasProperty } from "../../Predicate.ts"
+import * as Result from "../../Result.ts"
+
+const Ipv4TypeId = "~effect/net/Ipv4Address"
```
tim-smart: Just have one type id for all addresses. The _tag is the discriminant once you have an address.

packages/effect/src/unstable/net/Net.ts
```diff
@@ -0,0 +1,1110 @@
...
+/**
+ * An immutable 128-bit IPv6 address without socket scope metadata.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export interface Ipv6Address extends Equal.Equal, Hash.Hash {
+  readonly _tag: "Ipv6Address"
+  readonly [Ipv6TypeId]: typeof Ipv6TypeId
+  toString(): string
```
tim-smart, line 38: The interface itself can have the `bytes`, then you can get rid of the Impl interfaces.

tim-smart (changes requested):

packages/effect/src/unstable/net/Net.ts
fubhy: Not until we refactor the schema module structure with a public facade vs. internal modules.

fubhy: ... If we want to keep Schema.ts as the univeral entrypoint for all module schemas

fubhy: Can do this in a follow-up once Schema has been restructured accordingly (ongoing discussion).

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1091 @@
...
+/**
+ * An immutable 32-bit IPv4 address.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export interface Ipv4Address extends Equal.Equal, Hash.Hash {
+  readonly _tag: "Ipv4Address"
+  readonly [TypeId]: typeof TypeId
+  readonly bytes: Uint8Array
```
tim-smart: Actually, `bytes` needs to stay private. `Object.freeze` doesn't freeze typed array contents, so callers can mutate shared constants and change an address's hash. The `*ToOctets` functions already provide copies.

packages/sql/pg/src/PgTypes.ts
```diff
@@ -738,20 +740,24 @@ const formatIPv6 = (bytes: Uint8Array, offset: number): string => {
...
-  }
-  for (let index = wholeBytes + (partialBits === 0 ? 0 : 1); index < length; index++) {
-    if (bytes[offset + index] !== 0) return true
-  }
-  return false
-}
-
 const encodeInet = (value: unknown, isCidr: boolean): Uint8Array => {
   const text = requireString(value, isCidr ? "cidr" : "inet")
+  if (isCidr) {
```
tim-smart, line 667: This regresses `cidr` encoding for bare addresses. `PgTypes.encode("10.1.2.3", OID.cidr)` used to encode it as `/32`, and Postgres accepts that form. Keep the prefix optional and default it to the address width.

packages/sql/pg/src/PgTypes.ts
```diff
@@ -738,20 +740,24 @@ const formatIPv6 = (bytes: Uint8Array, offset: number): string => {
...
+    const parsed = IpNetwork.fromString(text)
```
tim-smart: Can you migrate the `inet` branch too? Right now `inet` accepts IPv4 octets with leading zeroes while `cidr` rejects them, and the decode paths use separate formatters. Using `NetAddress` for both removes that inconsistency and the duplicated parsers.

packages/platform/bun/src/BunHttpServer.ts
```diff
@@ -115,6 +121,11 @@ export const make = Effect.fnUntraced(
     }
   ) {
     const scope = yield* Effect.scope
+    const requestedIp = "unix" in options
```
tim-smart: `hostname` is still typed as `string`, but this now rejects valid Bun hostnames such as `localhost` at runtime. Either keep hostname binds working or narrow the option type. Also build the reported address from `server.hostname` / `server.port` and keep parse failures typed as `ServeError`.

migration/v3-to-v4.md
```diff
@@ -7124,12 +7124,18 @@ effect/unstable/rpc/Utils (barrel: effect/unstable/rpc)
 
 ### `@effect/platform/HttpServer`
 
+- `HttpServer.Address` -> `effect/unstable/net/Net#SocketAddress`: Replaced by the shared concrete internet-or-Unix socket address union.
```
tim-smart: This generated file still points to `effect/unstable/net/Net`, which no longer exists. Please regenerate it after updating the annotations.

packages/effect/src/Schema.ts
```diff
@@ -67,6 +67,8 @@ import type { Assign, Lambda, Mutable, Simplify } from "./Struct.ts"
 import * as Struct_ from "./Struct.ts"
 import type { RequiredKeys, UnionToIntersection } from "./Types.ts"
 import type { Unify } from "./Unify.ts"
+import * as IpNetwork_ from "./unstable/net/IpNetwork.ts"
```
tim-smart, line 80: I don't think these schemas should live in the stable `Schema` module. This makes stable code depend on `unstable/net` and exports unstable types from a stable module. Move them under `effect/unstable/net`, and add explicit annotations to the derived schemas.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1091 @@
...
+ * A resolved IPv6 internet address, port, flow information, and scope identifier.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export interface InetAddressV6 extends Equal.Equal, Hash.Hash {
+  readonly _tag: "InetAddressV6"
+  readonly address: Ipv6Address
+  readonly port: number
+  readonly flowInfo: number
```
tim-smart: Can we remove `flowInfo`? No adapter sets it and `formatInet` drops it, so the string codec doesn't round-trip values even though the field participates in `Equal` and `Hash`.

packages/effect/src/unstable/net/IpNetwork.ts
```diff
@@ -0,0 +1,421 @@
...
+
+const networkError = (
+  kind: NetworkError["kind"],
+  input: unknown,
+  reason: string
+): Result.Result<never, NetworkError> => Result.fail(new NetworkError({ kind, input, reason }))
+
+const width = (address: NetAddress.IpAddress): 32 | 128 => NetAddress.isIpv4Address(address) ? 32 : 128
+
+const toBytes = (address: NetAddress.IpAddress): ReadonlyArray<number> =>
```
tim-smart, line 96: Add unsafe byte constructors and use them here. These bytes are valid by construction, so converting through tuples and `Result.getOrThrow` is unnecessary. The same applies to the decode path in `PgTypes`.

packages/effect/src/unstable/cluster/HttpRunner.ts
```diff
@@ -31,6 +31,8 @@ import type { RunnerStorage } from "./RunnerStorage.ts"
 import * as Sharding from "./Sharding.ts"
 import type * as ShardingConfig from "./ShardingConfig.ts"
 
+const formatAuthorityHost = (host: string): string => host.includes(":") && !host.startsWith("[") ? `[${host}]` : host
```
tim-smart: Use `NetAddress.formatUrlHost` here instead of duplicating the IPv6 logic. If hostnames are also allowed, move a shared hostname/IP helper next to it.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1091 @@
...
+}
+
+/**
+ * Returns `true` for the all-zero address of either IP version.
+ *
+ * @category predicates
+ * @since 4.0.0
+ */
+export const isUnspecified = (self: IpAddress): boolean => {
+  const bytes = isIpv4Address(self) ? self.bytes : self.bytes
```
tim-smart: Both branches return `self.bytes`. This can just be `self.bytes.every((byte) => byte === 0)`.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1091 @@
...
+
+const bytesEqual = (self: Uint8Array, that: Uint8Array): boolean => {
+  if (self.length !== that.length) return false
+  for (let index = 0; index < self.length; index++) {
+    if (self[index] !== that[index]) return false
+  }
+  return true
+}
+
+const hashBytes = (tag: string, bytes: Uint8Array): number => {
```
tim-smart, line 297: `Hash.array` already does this: `Hash.combine(Hash.string(tag), Hash.array(bytes))`. Also use the data-first `Hash.combine` form for `InetV6Proto` below instead of nesting four curried calls.

tim-smart (commented):
Please also add an `index.ts` barrel for `unstable/net`, matching the other unstable directories.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1091 @@
...
+  readonly onIpv6: (address: Ipv6Address) => B
+}): A | B => isIpv4Address(self) ? options.onIpv4(self) : options.onIpv6(self))
+
+/**
+ * Formats an IP address in canonical numeric form.
+ *
+ * @category encoding
+ * @since 4.0.0
+ */
+export const formatIp = (self: IpAddress): string => {
```
tim-smart, line 690: Use `isIpv4Mapped` here instead of repeating its byte checks. `toCanonical` has the same duplication with `fromIpv4Mapped`.

packages/effect/src/unstable/net/IpNetwork.ts
```diff
@@ -0,0 +1,421 @@
...
+
+const TypeId = "~effect/net/IpNetwork" as const
+
+/**
+ * A checked IP network operation failure.
+ *
+ * @category errors
+ * @since 4.0.0
+ */
+export class NetworkError extends Data.TaggedError("IpNetworkError")<{
```
tim-smart: Can this be `IpNetworkError` so the class name matches its `_tag`?

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1091 @@
...
+    }
+  } else {
+    const separator = input.lastIndexOf(":")
+    if (separator < 0 || input.indexOf(":") !== separator) {
+      return addressError("InetAddress", input, "IPv6 addresses must be bracketed")
+    }
+    host = input.slice(0, separator)
+    portText = input.slice(separator + 1)
+  }
+  if (!/^\d+$/.test(portText)) return addressError("InetAddress", input, "port must be decimal")
```
tim-smart: Reject leading zeroes in ports for consistency with IPv4 octets. Also re-wrap invalid ports as `InetAddress` errors so every failure from this parser keeps the original input.

packages/platform/node/src/NodeHttpServer.ts
```diff
@@ -131,6 +132,12 @@ export const make = Effect.fnUntraced(function*(
   })
 
   const address = server.address()!
+  const effectAddress = typeof address === "string"
+    ? Effect.succeed(NetAddress.unixPathAddress(address))
+    : Effect.fromResult(NetAddress.inetAddressFromIpString(address.address, address.port)).pipe(
```
tim-smart, line 137: Please mention the `::` logging change in the changeset. `withLogAddress` now prints `http://[::]:3000` instead of `http://0.0.0.0:3000`. Also, can Node return a zoned IPv6 address here? `ipFromString` rejects `%`, which would turn a successful bind into a startup error.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1091 @@
...
+  return output as any
+}
+
+/**
+ * Returns the sixteen numeric octets of an IPv6 address in a fresh tuple.
+ *
+ * @category getters
+ * @since 4.0.0
+ */
+export const ipv6ToOctets = (
```
tim-smart, line 598: Can this return `ReadonlyArray<number>`? The 16-element tuple adds a lot of noise, especially when the implementation casts an array through `any`.

packages/effect/src/unstable/net/NetAddress.ts
fubhy: Keeping `bytes` public and mutable as an explicit simplicity tradeoff.

packages/sql/pg/src/PgTypes.ts
fubhy: Fixed. Bare `cidr` addresses now default to `/32` or `/128`, with regression coverage.

packages/sql/pg/src/PgTypes.ts
fubhy: Done. `inet` and `cidr` now share `NetAddress` parsing and formatting, including strict IPv4 input.

packages/platform/bun/src/BunHttpServer.ts
fubhy: Fixed. Bun hostnames resolve before binding, the address comes from `server.hostname` / `server.port`, and failures remain `ServeError`.

migration/v3-to-v4.md
fubhy: Regenerated. The migration guide now points to `effect/unstable/net/NetAddress`.

packages/effect/src/Schema.ts
fubhy: Keeping these exports in `Schema.ts` by design, but added explicit identifiers to the declarations and derived codecs.

packages/effect/src/unstable/net/NetAddress.ts
fubhy: Removed `flowInfo` from the model, constructors, equality, hashing, docs, and tests.

packages/effect/src/unstable/cluster/HttpRunner.ts
fubhy: Done. Added the shared `NetAddress.formatUrlHostString` helper and use it for both transports.

packages/effect/src/unstable/net/IpNetwork.ts
fubhy: Done. Added unsafe IPv4/IPv6 byte constructors and use them in `IpNetwork` and PostgreSQL decoding.

packages/effect/src/unstable/net/NetAddress.ts
fubhy: Done. Byte hashing uses `Hash.array`, and IPv6 hashing uses data-first `Hash.combine`.

packages/effect/src/unstable/net/IpNetwork.ts
fubhy: Renamed to `IpNetworkError` and updated the public result types.

packages/effect/src/unstable/net/NetAddress.ts
fubhy: Fixed. Ports reject leading zeroes, and constructor failures are rewrapped with the original `InetAddress` input.

packages/effect/src/unstable/net/NetAddress.ts
fubhy: Changed to `ReadonlyArray<number>` and removed the cast.

packages/effect/src/unstable/net/NetAddress.ts
fubhy: Done. Formatting reuses `isIpv4Mapped`, and canonicalization reuses `fromIpv4Mapped`.

packages/effect/src/unstable/net/NetAddress.ts
fubhy: Simplified to `self.bytes.every(...)`.

packages/platform/node/src/NodeHttpServer.ts
fubhy: Updated the changeset. Node TCP `server.address()` uses libuv `uv_ip_name`, which omits `sin6_scope_id`, so it cannot return a `%zone` suffix here.

fubhy:
Added the `effect/unstable/net` barrel with `IpNetwork` and `NetAddress`, plus development and published package export-map entries.

packages/effect/src/unstable/net/Net.ts
fubhy: Done. All address variants now share one type ID and use `_tag` as the discriminant.

packages/effect/src/internal/netAddress.ts
fubhy: Renamed the module to `NetAddress`.

packages/effect/src/unstable/net/Net.ts
fubhy: Done. `bytes` is on the public interfaces and the implementation interfaces were removed.

packages/effect/src/unstable/net/NetAddress.ts
fubhy: This was why I originally had a `*Impl` with private bytes. If you want that back, lmk.

packages/effect/src/unstable/net/IpInterface.ts
```diff
@@ -0,0 +1,284 @@
...
+  return Object.freeze(self)
+}
+
+/**
+ * Creates an interface address while preserving all address bits.
+ *
+ * @category constructors
+ * @since 4.0.0
+ */
+export const make =
```
tim-smart: This signature is not consistent with other modules

packages/effect/src/unstable/net/IpInterface.ts
```diff
@@ -0,0 +1,284 @@
...
+  return make(address.success, parts.success[1] ?? width(address.success))
+}
+
+/**
+ * Creates a trusted interface address, throwing when its prefix length is invalid.
+ *
+ * @category unsafe
+ * @since 4.0.0
+ */
+export const makeUnsafe =
```
tim-smart: Another weird signature

packages/effect/src/unstable/net/IpInterface.ts
```diff
@@ -0,0 +1,284 @@
...
+export const format = (self: IpInterface): string => `${NetAddress.formatIp(self.address)}/${self.prefixLength}`
+
+/**
+ * Returns the canonical network containing an interface address.
+ *
+ * @category conversions
+ * @since 4.0.0
+ */
+export const network =
+  ((self: IpInterface): IpNetwork.IpNetwork => IpNetwork.fromAddressUnsafe(self.address, self.prefixLength)) as {
```
tim-smart: And again

packages/effect/src/unstable/net/IpNetwork.ts
```diff
@@ -0,0 +1,413 @@
...
+ *
+ * **Gotchas**
+ *
+ * Use {@link fromAddress} when a host address should be truncated to its containing network.
+ *
+ * @category constructors
+ * @since 4.0.0
+ */
+export const make =
+  ((address: NetAddress.IpAddress, prefixLength: number): Result.Result<IpNetwork, IpNetworkError> => {
```
tim-smart: Weird signature, I think you get the idea now :)

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1156 @@
...
+   *
+   * **Details**
+   *
+   * String addresses must be numeric IPv4 or IPv6 literals.
+   *
+   * @category models
+   * @since 4.0.0
+   */
+  export type Input =
+    | SocketAddress
```
tim-smart, line 140: Would be nice to support plain strings too "127.0.0.1:3000"

tim-smart (changes requested):
Will get AI to do another review too

packages/effect/src/unstable/http/HttpServer.ts
```diff
@@ -103,9 +69,12 @@ export const make = (
       httpEffect: Effect.Effect<HttpServerResponse, unknown, HttpServerRequest | Scope.Scope>,
       middleware?: Middleware.HttpMiddleware
     ) => Effect.Effect<void, never, Scope.Scope>
-    readonly address: Address
+    readonly address: NetAddress.SocketAddress.Input
   }
-): HttpServer["Service"] => options
+): HttpServer["Service"] => ({
+  ...options,
+  address: NetAddress.socketAddressFromInputUnsafe(options.address)
```
tim-smart: Generated by AI review: `HttpServer.make` now throws for inputs like `{ address: "localhost", port: 3000 }`, even though the type accepts them and the signature has no error channel. Can we keep this total, or return a `Result`? Would also be good to test the rejected input.

packages/effect/src/unstable/http/HttpServer.ts
```diff
@@ -207,18 +176,19 @@ export const serveEffect: {
...
+export const formatAddress = (address: NetAddress.SocketAddress): string => {
   switch (address._tag) {
-    case "UnixAddress":
+    case "UnixPathAddress":
       return `unix://${address.path}`
-    case "TcpAddress":
-      return `http://${address.hostname}:${address.port}`
+    case "InetAddressV4":
+    case "InetAddressV6":
+      return `http://${NetAddress.formatInet(address)}`
```
tim-smart: Generated by AI review: This returns a string that `new URL()` rejects for scoped IPv6 addresses, for example `http://[fe80::1%2]:3000`. `formatAddress` is also used as a URL in the server tests. Can we avoid returning an invalid URL here?

packages/effect/src/unstable/http/HttpServer.ts
```diff
@@ -283,11 +253,15 @@ export const makeTestClient: Effect.Effect<
...
-    return yield* Effect.die(new Error("HttpServer.layerTestClient: UnixAddress not supported"))
+  if (address._tag === "UnixPathAddress") {
+    return yield* Effect.die(new Error("HttpServer.layerTestClient: UnixPathAddress not supported"))
   }
-  const host = address.hostname === "0.0.0.0" ? "127.0.0.1" : address.hostname
-  const url = `http://${host}:${address.port}`
+  const host = NetAddress.isUnspecified(address.address)
+    ? address._tag === "InetAddressV4"
+      ? NetAddress.ipv4Loopback
+      : NetAddress.ipv6Loopback
```
tim-smart: Generated by AI review: This now maps a server bound to `::` to `::1` in `makeTestClient`; it previously used `127.0.0.1`. That can fail on systems without IPv6 loopback support. Is that change intentional?

packages/effect/src/unstable/net/IpInterface.ts
```diff
@@ -0,0 +1,284 @@
...
+  }
+}
+
+const interfaceError = (
+  kind: IpInterfaceError["kind"],
+  input: unknown,
+  reason: string
+): Result.Result<never, IpInterfaceError> => Result.fail(new IpInterfaceError({ kind, input, reason }))
+
+const width = (address: NetAddress.IpAddress): 32 | 128 => NetAddress.isIpv4Address(address) ? 32 : 128
```
tim-smart: Generated by AI review: There is a lot of duplicated parsing and construction code between `IpInterface` and `IpNetwork`. Can we move the shared parts into an internal helper so they don't drift?

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1156 @@
...
+  isIpv6Address(self) ? Option.getOrElse(fromIpv4Mapped(self), () => self) : self
+
+const InetV4Proto = {
+  _tag: "InetAddressV4",
+  [TypeId]: TypeId,
+  [Equal.symbol](this: InetAddressV4, that: Equal.Equal): boolean {
+    return isInetAddressV4(that) && this.port === that.port && Equal.equals(this.address, that.address)
+  },
+  [Hash.symbol](this: InetAddressV4): number {
+    return Hash.combine(Hash.hash(this.address))(Hash.number(this.port))
```
tim-smart: Generated by AI review: Can we use the data-first `Hash.combine(a, b)` form here too? The curried form reverses the argument order, and mixing both styles between the IPv4 and IPv6 protos looks accidental.

packages/sql/pg/src/PgTypes.ts
```diff
@@ -790,11 +694,19 @@ const decodeInet = (bytes: Uint8Array, offset: number, size: number): string =>
...
+    : NetAddress.ipv6FromBytesUnsafe(addressBytes)
+  const interfaceAddress = IpInterface.make(address, bits)
+  if (Result.isFailure(interfaceAddress)) return fail(interfaceAddress.failure.message)
+  if (isCidr) {
+    const network = IpNetwork.make(interfaceAddress.success.address, interfaceAddress.success.prefixLength)
+    if (Result.isFailure(network)) return fail(network.failure.message)
+    return IpNetwork.format(network.success)
+  }
+  const text = IpInterface.format(interfaceAddress.success)
+  return bits === addressSize * 8 ? text.slice(0, text.lastIndexOf("/")) : text
```
tim-smart: Generated by AI review: Can we format the address directly instead of formatting the interface and slicing off the prefix?

```suggestion
return bits === addressSize * 8
  ? NetAddress.formatIp(address)
  : IpInterface.format(interfaceAddress.success)
```

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1156 @@
...
+      scopeId = Number(scopeText)
+      if (!Number.isInteger(scopeId) || scopeId > 0xffffffff) {
+        return addressError("InetAddress", input, "scope identifier must be an unsigned 32-bit integer")
+      }
+      host = host.slice(0, scopeSeparator)
+    }
+  } else {
+    const separator = input.lastIndexOf(":")
+    if (separator < 0 || input.indexOf(":") !== separator) {
+      return addressError("InetAddress", input, "IPv6 addresses must be bracketed")
```
tim-smart: Generated by AI review: This branch only allows one `:`, so `ipFromString(host)` can never return IPv6 here. The later bracket check is only useful for bracketed IPv4. Also, inputs with no colon get `IPv6 addresses must be bracketed`; `expected host:port or [IPv6]:port` would be clearer.

.changeset/net-address-values.md
```diff
@@ -0,0 +1,14 @@
...
+Add platform-neutral network values under `effect/unstable/net`: `NetAddress` for MAC, IP, internet, and Unix socket addresses; `IpInterface` for host addresses that preserve their prefix and host bits; and `IpNetwork` for canonical IPv4 and IPv6 CIDR values. These modules provide strict parsing, canonical formatting, checked and unsafe construction, equality, hashing, matching unstable `effect/Schema` codecs, and network operations including containment, overlap, exact bounds, and address counts.
+
+Add `NetAddress.SocketAddress.Input` and checked and unsafe constructors for normalizing concrete socket address configuration. HTTP and socket servers now accept these inputs and expose canonical `NetAddress.SocketAddress` values. Replace their TCP address types with `NetAddress.InetAddress`, use `address` instead of `hostname`, and replace Unix address types with `NetAddress.UnixPathAddress`. Bun listener hostnames continue to resolve before binding; native conversion failures remain typed server-open errors, and IPv6 authorities are formatted with brackets. Servers bound to the IPv6 unspecified address now log addresses such as `http://[::]:3000` instead of substituting `0.0.0.0`.
```
tim-smart: Generated by AI review: Can we mention that the Bun and Deno layers now have a `ServeError` error channel? This is a compile-time breaking change for code expecting `never`.

tim-smart: Should probably just rewrite the changeset so it isn't a giant wall of text

packages/effect/src/Schema.ts
```diff
@@ -11265,6 +11268,333 @@ export const DateTimeZonedFromString: DateTimeZonedFromString = DateTimeZonedStr
...
+      },
+      encode: (value) => Effect.succeed(encode(value))
+    })
+  )).annotate({ identifier })
+
+/**
+ * Schema for already-constructed MAC address values.
+ *
+ * @category schemas
+ * @since 4.0.0
```
tim-smart, line 11294: Should these have the `@unstable` tag?

tim-smart (changes requested):

packages/platform/bun/src/BunHttpServer.ts
```diff
@@ -154,10 +172,14 @@ export const make = Effect.fnUntraced(
 
     yield* Scope.addFinalizer(scope, shutdown)
 
+    const address = "unix" in options
```
tim-smart, line 175: Keep the `options.unix !== undefined` check in both places. `{ unix: undefined, port: 3000 }` now reports `unix://undefined` for a TCP server and breaks `makeTestClient`.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1141 @@
...
+
+/**
+ * Returns `true` when a value is a portable concrete socket address.
+ *
+ * @category guards
+ * @since 4.0.0
+ */
+export const isSocketAddress = (u: unknown): u is SocketAddress => isInetAddress(u) || isUnixPathAddress(u)
+
+const Ipv4Proto = {
```
tim-smart, line 238: Can we add `toJSON` to these protos, plus `IpInterface` and `IpNetwork`? `JSON.stringify` currently dumps the private bytes. Have `NodeInspectSymbol` use it too.

packages/effect/src/unstable/http/HttpServer.ts
```diff
@@ -283,11 +255,13 @@ export const makeTestClient: Effect.Effect<
...
   const client = yield* HttpClient.HttpClient
   const address = server.address
-  if (address._tag === "UnixAddress") {
-    return yield* Effect.die(new Error("HttpServer.layerTestClient: UnixAddress not supported"))
+  if (address._tag === "UnixPathAddress") {
+    return yield* Effect.die(new Error("HttpServer.layerTestClient: UnixPathAddress not supported"))
   }
-  const host = address.hostname === "0.0.0.0" ? "127.0.0.1" : address.hostname
-  const url = `http://${host}:${address.port}`
+  const host = NetAddress.isUnspecified(address.address)
```
tim-smart, line 261: This changes the test client host from `127.0.0.1` to `::1` for a server bound to `::`. Can we keep the IPv4 fallback for dual-stack listeners? Some environments don't have IPv6 loopback support.

packages/effect/src/unstable/http/HttpServer.ts
```diff
@@ -283,11 +255,13 @@ export const makeTestClient: Effect.Effect<
...
+    ? NetAddress.isIpv4Address(address.address) ? NetAddress.ipv4Loopback : NetAddress.ipv6Loopback
+    : address.address
+  const url = `http://${NetAddress.formatUrlHost(host)}:${address.port}`
```
tim-smart, line 264: Can we reject scoped IPv6 addresses here? `formatUrlHost` silently drops the scope, so the client loses the interface it needs. `formatAddress` has the same issue.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1141 @@
...
+  return Object.freeze(self)
+}
+
+const makeIpv6 = (bytes: Uint8Array): Ipv6Address => {
+  const self = Object.assign(Object.create(Ipv6Proto), { bytes: new Uint8Array(bytes) })
+  return Object.freeze(self)
+}
+
+const makeMac = (bytes: Uint8Array): MacAddress => {
+  const self = Object.assign(Object.create(MacProto), { bytes: bytes.slice() })
```
tim-smart, line 312: Use `new Uint8Array(bytes)` here too. `Buffer.slice()` shares the underlying memory, which is what we just fixed for IPv4 and IPv6. No public byte constructor reaches this yet, but it should be consistent.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -0,0 +1,1141 @@
...
+    | { readonly path: string }
+}
+
+/**
+ * A checked network-address operation failure.
+ *
+ * @category errors
+ * @since 4.0.0
+ */
+export class NetAddressError extends Data.TaggedError("NetAddressError")<{
```
tim-smart, line 152: Can we keep the original input on `NetAddressError`? `PgTypes` now forwards the message directly, so SQL errors have lost the value and column type that failed.

packages/sql/pg/src/PgTypes.ts
```diff
@@ -790,11 +694,20 @@ const decodeInet = (bytes: Uint8Array, offset: number, size: number): string =>
...
-  if (isCidr && hasHostBits(bytes, offset + 4, addressSize, bits)) {
-    return fail("CIDR address has host bits set")
-  }
-  const text = family === PGSQL_AF_INET ? formatIPv4(bytes, offset + 4) : formatIPv6(bytes, offset + 4)
-  return isCidr || bits !== addressSize * 8 ? `${text}/${bits}` : text
+  const addressBytes = bytes.slice(offset + 4, offset + 4 + addressSize)
+  const address = family === PGSQL_AF_INET
+    ? NetAddress.ipv4FromBytesUnsafe(addressBytes)
+    : NetAddress.ipv6FromBytesUnsafe(addressBytes)
+  const interfaceAddress = IpInterface.make(address, bits)
```
tim-smart, line 701: The prefix is already validated above. Just use `IpNetwork.make` for `cidr` and `IpInterface.makeUnsafe` for `inet`, so we don't validate it three times.

packages/effect/src/Schema.ts
```diff
@@ -11265,6 +11268,333 @@ export const DateTimeZonedFromString: DateTimeZonedFromString = DateTimeZonedStr
...
+  identifier: "MacAddress"
+})
+
+/**
+ * Schema for MAC addresses encoded as canonical colon-separated hexadecimal strings.
+ *
+ * @category schemas
+ * @since 4.0.0
+ */
+export const MacAddressFromString = netAddressFromString(
```
tim-smart, line 11306: Add named interfaces and explicit annotations to the new `*FromString` schemas, consistent with the other codecs in this module.

packages/effect/src/unstable/net/IpNetwork.ts
```diff
@@ -0,0 +1,315 @@
...
+export const isIpNetwork = (u: unknown): u is IpNetwork => hasProperty(u, TypeId)
+
+const IpNetworkProto = {
+  _tag: "IpNetwork",
+  [TypeId]: TypeId,
+  [Equal.symbol](this: IpNetwork, that: Equal.Equal): boolean {
+    return isIpNetwork(that) && this.prefixLength === that.prefixLength && Equal.equals(this.address, that.address)
+  },
+  [Hash.symbol](this: IpNetwork): number {
+    return Hash.combine(Hash.hash(this.address))(Hash.number(this.prefixLength))
```
tim-smart, line 78: These two `Hash.combine` calls should use the data-first form too.

Also in `packages/effect/src/unstable/net/NetAddress.ts:1070`.

tim-smart (approved):

PR #7525, opened 2026-08-30
Add exact ByteSize value module

fubhy:
## Summary

- add an exact, non-negative, bigint-backed `ByteSize` value module with SI and IEC units, parsing, formatting, checked arithmetic, ordering, and safe number conversion
- add `Schema.ByteSize`, string/bigint/number transformations, JSON revival support, and `Config.ByteSize`
- add runtime and type tests, the implementation plan, and a patch changeset

## Testing

- `pnpm lint-fix`
- `pnpm test --run packages/effect/test/ByteSize.test.ts packages/effect/test/Config.test.ts packages/effect/test/schema/Schema.test.ts packages/effect/test/schema/representation/builtInRevivers.test.ts`
- `pnpm test-types ByteSize.tst.ts`
- `pnpm check`

### DIFF
--- a/packages/effect/src/ByteSize.ts
+++ b/packages/effect/src/ByteSize.ts
@@ -46,12 +46,12 @@ export interface ByteSize extends Equal.Equal, Pipeable, Inspectable.Inspectable
 export type Input = ByteSize | bigint | number | string
 
 /**
- * Canonical decimal and binary byte unit symbols.
+ * Canonical decimal byte unit symbols.
  *
  * @category models
  * @since 4.0.0
  */
-export type Unit =
+export type DecimalUnit =
   | "B"
   | "kB"
   | "MB"
@@ -63,6 +63,15 @@ export type Unit =
   | "YB"
   | "RB"
   | "QB"
+
+/**
+ * Canonical binary byte unit symbols.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export type BinaryUnit =
+  | "B"
   | "KiB"
   | "MiB"
   | "GiB"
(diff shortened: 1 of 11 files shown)


packages/effect/src/ByteSize.ts
```diff
@@ -0,0 +1,677 @@
...
+ */
+export type Input = ByteSize | bigint | number | string
+
+/**
+ * Canonical decimal and binary byte unit symbols.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export type Unit =
```
IMax153: We should separate out the binary and decimal types into their own types in case a user want to restrict input to one or the other. We can still expose `Unit` as the union of `DecimalUnit` and `BinaryUnit`.

packages/effect/src/ByteSize.ts
```diff
@@ -0,0 +1,677 @@
...
+
+/**
+ * Options controlling compact byte-size formatting.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export interface FormatOptions {
+  readonly system?: "decimal" | "binary" | undefined
+  readonly unit?: Unit | undefined
```
IMax153: It seems like these two fields could be at odds with one another when configuring formatting.

For example, you could specify `system: "decimal"` but then use a binary `format`?

packages/effect/src/ByteSize.ts
```diff
@@ -0,0 +1,677 @@
...
+
+let zeroValue: ByteSize | undefined
+
+/**
+ * The byte size containing zero bytes.
+ *
+ * @category constants
+ * @since 4.0.0
+ */
+export const zero: ByteSize = zeroValue = make(bigint0)
```
IMax153: The lazy initialization here seems unnecessary, I would remove it.

packages/effect/src/ByteSize.ts
```diff
@@ -0,0 +1,677 @@
...
+      if (input < bigint0) return invalid(`expected a non-negative bigint, received ${input}`)
+      return make(input)
+    case "number":
+      return fromNumber(input)
+    case "string":
+      return parse(input)
+    case "object":
+      if (isByteSize(input)) return input
+  }
+  return invalid(`unsupported input ${globalThis.String(input)}`)
```
IMax153:
```suggestion
  return invalid(`unsupported input ${input}`)
```

Should probably audit the module for more improper usage of `globalThis.String`

packages/effect/src/ByteSize.ts
```diff
@@ -0,0 +1,677 @@
...
+const unitConstructor = (symbol: Unit) => (value: number | bigint): ByteSize =>
+  fromQuantity(value, unitsBySymbol.get(symbol)!)
+
+/**
+ * Creates a decimal kilobyte value.
+ *
+ * @category constructors
+ * @since 4.0.0
+ */
+export const kilobytes = unitConstructor("kB")
```
IMax153, line 270: Should these be camel-cased? I'm not sure. 

`ByteSize.Kilobytes` or `ByteSize.kilobytes`

packages/effect/src/ByteSize.ts
```diff
@@ -0,0 +1,677 @@
...
+ */
+export const toBytes = (self: ByteSize): bigint => self.value
+
+/**
+ * Converts a byte size to a safe integer, returning `None` when it is too large.
+ *
+ * @category converting
+ * @since 4.0.0
+ */
+export const toNumber = (self: ByteSize): Option.Option<number> => {
```
IMax153, line 421: Why `toNumber` instead of `toBigint`?

IMax153 (approved):

PR #7528, opened 2026-08-30
Defer Undici loading in platform-node

### LINKED ISSUE #6648: platform-node: eager undici import hijacks Node's built-in fetch via legacy global-dispatcher symbol (FetchHttpClient POSTs fail with "invalid content-length header")
## What version of Effect is running?

`effect@4.0.0-beta.65`, `@effect/platform-node@4.0.0-beta.65`

## What steps can reproduce the bug?

Importing `@effect/platform-node/NodeHttpClient` (statically importing `undici`) breaks Node's **built-in** `globalThis.fetch` for the whole process: any POST carrying an explicit `content-length` header afterwards rejects before any network I/O.

```js
// node v22.21.1
const body = new TextEncoder().encode(JSON.stringify({ a: 1 }))
const init = {
  method: 'POST',
  headers: { 'content-type': 'application/json', 'content-length': String(body.byteLength) },
  body
}

await fetch('https://example.com/post', init)          // works (HTTP response)
await import('@effect/platform-node/NodeHttpClient')   // or any module that pulls it in
await fetch('https://example.com/post', init)
// TypeError: fetch failed
//   cause: InvalidArgumentError: invalid content-length header [UND_ERR_INVALID_ARG]
```

Root cause is an undici v8 import-time side effect (reported upstream in nodejs/undici#5500): `undici/lib/global.js` runs `setGlobalDispatcher(new Agent())` at module load when no global dispatcher exists, and also writes the legacy `Symbol.for('undici.globalDispatcher.1')` — the internal symbol Node's bundled fetch reads. From then on Node's fetch dispatches into npm undici v8's core, whose stricter `Request.processHeader` rejects request-set `content-length` that the bundled core accepts. Reproduced with undici 8.2.0 through 8.9.0 (latest).

Two Effect-specific aggravations:

1. `@effect/platform-node/NodeHttpClient` imports `undici` **eagerly** (top-level `import Undici from "undici"` in `Undici.ts`), so even consumers using only `layerNodeHttp` (the `node:http` backend) trigger the side effect.
2. `FetchHttpClient` always sets an explicit `content-length` header from `HttpBody`, which is exactly the shape the hijacked dispatcher rejects — so in a process that has loaded `platform-node` anywhere in its module graph, **every `FetchHttpClient` POST with a body fails** with an opaque `Transport error` (`RequestError` wrapping `TypeError: fetch failed`), while GETs keep working.

## What is the expected behavior?

Importing `@effect/platform-node` (especially when only the `node:http`-backed client is used) should not alter the behavior of Node's built-in `fetch`, and `FetchHttpClient` POSTs should not fail depending on whether an unrelated module was imported earlier in the process.

## What do you see instead?

`HttpClientError.RequestError: Transport error` for every `FetchHttpClient` request with a body, with the nested cause `InvalidArgumentError: invalid content-length header`. Bodyless requests succeed, which makes the failure look nondeterministic/context-dependent in real apps.

We hit this in production (Next.js on Vercel and locally, Node 22): a PR that added one file importing `NodeHttpClient` into a server module graph broke unrelated `FetchHttpClient`-backed OAuth and model POSTs process-wide.

## Possible mitigations on the Effect side

- Lazy-load `undici` in `platform-node` so only the undici-backed layers (`layerUndici`, `makeUndici`, dispatcher layers) trigger the import, not `layerNodeHttp`.
- Consider having `FetchHttpClient` omit explicit `content-length` (it is a forbidden header per the fetch spec; compliant implementations compute it from the body), which would also sidestep the hijacked-dispatcher failure mode.

## Additional information

- Upstream report: nodejs/undici#5500 (same mechanism, hit via a different package).
- The failure mode is especially confusing because the poisoning is process-wide and load-order dependent: fresh processes work until some route/module graph loads `platform-node`, after which unrelated fetch POSTs fail.

Kashkovsky:
Importing `@effect/platform-node` or `@effect/platform-node/NodeHttpClient` no longer evaluates `undici`, so these imports do not replace the dispatcher used by Node global fetch.

**Detail:**

• Load the Undici facade only when `makeDispatcher`, `dispatcherLayerGlobal`, or the Kubernetes TLS dispatcher is acquired.
• Remove the `Undici` namespace from the package root; use `@effect/platform-node/Undici` for direct Undici APIs.
• Cover side-effect-free package imports and dispatcher-triggered loading.

Closes #6648


Kashkovsky:
@tim-smart could you review this? It changes the Undici loading path from #2410 and removes the eager root re-export while keeping `@effect/platform-node/Undici`.

packages/platform/node/src/NodeHttpClient.ts
```diff
@@ -90,6 +90,8 @@ export class Dispatcher extends Context.Service<Dispatcher, Undici.Dispatcher>()
   "@effect/platform-node/NodeHttpClient/Dispatcher"
 ) {}
 
+const loadUndici = Effect.promise(() => import("./Undici.ts"))
```
tim-smart: This should use Effect.cached too

packages/platform/node/test/NodeHttpClientImport.test.ts
```diff
@@ -0,0 +1,40 @@
+import { assert, describe, it } from "@effect/vitest"
```
tim-smart: You can remove this file

tim-smart (changes requested):

packages/platform/node/src/NodeHttpClient.ts
Kashkovsky: Thanks, done

packages/platform/node/test/NodeHttpClientImport.test.ts
Kashkovsky: Removed

packages/platform/node/src/NodeHttpClient.ts
```diff
@@ -90,6 +90,8 @@ export class Dispatcher extends Context.Service<Dispatcher, Undici.Dispatcher>()
   "@effect/platform-node/NodeHttpClient/Dispatcher"
 ) {}
 
+const loadUndici = Effect.promise(() => import("./Undici.ts")).pipe(Effect.cached, Effect.runSync)
```
tim-smart:
```suggestion
const loadUndici = Effect.promise(() => import("./Undici.ts"))
```

Actually they are already cached

tim-smart (approved):

Issue #7529, opened 2026-08-30
NodeTerminal.readLine does not echo typed input and ignores Ctrl+C when stdin is a TTY


dragove:
### What version of Effect is running?

4.0.0-rc.112

### What steps can reproduce the bug?

[example](https://www.effect.website/docs/v4/platform/terminal#reading-from-standard-input) on the website can reproduce this bug.

### What is the expected behavior?

Characters typed on a TTY are echoed while reading stdin.

### What do you see instead?

Nothing respond before hit `enter` key.

### Additional information

Node.js: 26.8.1
Arch Linux
Kitty Terminal

shahidbeig-a11y:
Taking this. I'll fix NodeTerminal.readLine so a TTY echoes input and Ctrl+C interrupts, plus tests and a changeset. Not changing the Terminal interface.

PR #7530, opened 2026-08-30
Fix NodeTerminal.readLine not echo input

dragove:
## Type

- [x] Bug Fix

## Description

`NodeTerminal` created its `readline` interface without an `output` stream, so `terminal` resolved to `input.isTTY && output.isTTY` = `false` even on a real TTY.

## Related

- Related Issue https://github.com/Effect-TS/effect/issues/7529

### DIFF
--- a/packages/platform/node-shared/src/NodeTerminal.ts
+++ b/packages/platform/node-shared/src/NodeTerminal.ts
@@ -53,7 +53,7 @@ export const make: (
     const rlRef = yield* RcRef.make({
       acquire: Effect.acquireRelease(
         Effect.sync(() => {
-          const rl = readline.createInterface({ input: stdin, escapeCodeTimeout: 50 })
+          const rl = readline.createInterface({ input: stdin, output: stdout, escapeCodeTimeout: 50 })
           const onLine = (line: string) => Queue.offerUnsafe(lines, line)
           const onClose = () => {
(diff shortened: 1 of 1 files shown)


packages/platform/node-shared/src/NodeTerminal.ts
```diff
@@ -53,7 +53,7 @@ export const make: (
     const rlRef = yield* RcRef.make({
       acquire: Effect.acquireRelease(
         Effect.sync(() => {
-          const rl = readline.createInterface({ input: stdin, escapeCodeTimeout: 50 })
+          const rl = readline.createInterface({ input: stdin, output: stdout, escapeCodeTimeout: 50 })
```
tim-smart, line 56: This will cause readline to echo input while prompts are running. For password and hidden prompts, the raw input is printed before the prompt redraws, so passwords can end up in the terminal scrollback. Normal text prompts are also echoed twice and readline's line editing conflicts with the prompt redraw.

tim-smart (changes requested):

packages/platform/node-shared/src/NodeTerminal.ts
dragove: Sorry, I don't have the knowledge to fix this issue. I'll close this pr.

Issue #7526, opened 2026-08-30
`Atom`: stream-backed atom's finalizer never runs when consumed via a derived atom and `idleTTL` exceeds the emission interval


jonathn:
### What version of Effect is running?

effect@4.0.0-rc.112

### What steps can reproduce the bug?

_I was using Claude Opus 5 to evaluate Effect v4, and it uncovered what it believes to be a genuine, reproducible bug.  
The following is the bug report prepared by Claude._

--------------

An `Atom` backed by a `Stream` that holds an external resource (via `Effect.acquireRelease`) is
never disposed — and its finalizer never runs — when **both** of the following hold:

1. it is consumed through a **derived** atom rather than being subscribed to directly, and
2. the registry's `idleTTL` is **greater than the stream's emission interval**.

The leak is self-sustaining: the node is scheduled for removal, but the still-running stream's own
writes cancel the pending removal, so it is rescheduled forever.

Standalone repro — **no UI framework involved**, `effect` core only:
```js
// npm i effect@4.0.0-rc.112 && node repro.mjs <idleTTL> <emitIntervalMs>
import * as Atom from "effect/unstable/reactivity/Atom"
import * as AtomRegistry from "effect/unstable/reactivity/AtomRegistry"
import * as AsyncResult from "effect/unstable/reactivity/AsyncResult"
import * as Effect from "effect/Effect"
import * as Stream from "effect/Stream"
import * as Queue from "effect/Queue"

const ttl     = process.argv[2] === undefined ? 400 : Number(process.argv[2])
const EMIT_MS = Number(process.argv[3] ?? 100)
let released  = false

const source = Stream.callback((queue) =>
  Effect.acquireRelease(
    Effect.sync(() => {
      console.log("[ACQUIRE]  resource opened")
      return setInterval(() => Queue.offerUnsafe(queue, Date.now()), EMIT_MS)
    }),
    (handle) => Effect.sync(() => {
      console.log("[RELEASE]  resource closed")
      released = true
      clearInterval(handle)
    }),
  ).pipe(Effect.asVoid),
)

const feedAtom    = Atom.withLabel(Atom.make(() => source), "FEED")
const derivedAtom = Atom.withLabel(
  Atom.make((get) => AsyncResult.isSuccess(get(feedAtom)) ? "ready" : "init"), "DERIVED")

const registry = AtomRegistry.make({ defaultIdleTTL: ttl })

const dump = (tag) => {
  const rows = []
[...]

PR #7533, opened 2026-08-30
fix(atom-solid): forward default idle TTL

### LINKED ISSUE #7527: `@effect/atom-solid`: RegistryProvider hardcodes `defaultIdleTTL ?? 400`, diverging from `@effect/atom-react` and making the TTL impossible to disable

effect@4.0.0-rc.112  and  @effect/atom-solid@4.0.0-rc.112

_I was using Claude Opus 5 to evaluate Effect v4, and it uncovered this SolidJS adapter issue.
The following is the bug report prepared by Claude._

---------

`@effect/atom-solid`'s `RegistryProvider` substitutes a default idle TTL that the caller cannot
turn off, whereas `@effect/atom-react`'s equivalent passes the option straight through:

```ts
// @effect/atom-solid/src/RegistryContext.ts
defaultIdleTTL: options.defaultIdleTTL ?? 400   // <- hardcoded fallback

// @effect/atom-react/src/RegistryContext.ts
defaultIdleTTL: options.defaultIdleTTL          // <- undefined unless supplied
```

Because the fallback is applied with `??`, the only way to reach "no TTL" is `defaultIdleTTL={0}`,
which is a different code path from `undefined` (and currently triggers
`TimeoutNaNWarning: NaN is not a number` from core). There is no way to express "behave like the
React binding's default".

The two bindings should have the same default registry behaviour, and `defaultIdleTTL` should be
fully caller-controlled — including being left `undefined`.

### What do you see instead?

Identical atom graphs behave differently under the two bindings purely because of this line.

With a stream-backed atom consumed via a derived atom (see companion core issue #7526), the hardcoded
400ms TTL means the same code leaks under Solid but not under React's `RegistryProvider`:

```
SOLID  via derived atom, stock            LEAK  (finalizer never runs)
SOLID  via derived atom, defaultIdleTTL=0 CLEAN
REACT  via derived atom, RegistryProvider CLEAN
REACT  via derived atom, default registry LEAK  (that registry also sets 400)
```

Removing the fallback makes the Solid binding clean at every stream rate tested (50/100/300/2000ms):

```diff
- defaultIdleTTL: options.defaultIdleTTL ?? 400
+ defaultIdleTTL: options.defaultIdleTTL
```

### Additional information

The underlying disposal bug is in core and is reported separately (#7526) — this issue is only about the
unexpected divergence between the two bindings, which turns a core edge case into a
framework-dependent behavioral difference and removes the user's ability to configure around it.

apoorvdarshan:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`@effect/atom-solid`'s `RegistryProvider` previously replaced an omitted `defaultIdleTTL` with 400 ms, unlike the React binding and contrary to its forwarding documentation. This change forwards `undefined`, allowing unused atom nodes to be removed immediately unless a TTL is explicitly configured.

Includes a patch changeset for `@effect/atom-solid`.

## Related

- Related Issue #7526
- Closes #7527
- Closes EFF-1285

## Testing

- `nix develop -c pnpm test --run packages/atom/solid/test/index.test.tsx` (10 passed)
- `nix develop -c pnpm check`

### DIFF
--- a/packages/atom/solid/test/index.test.tsx
+++ b/packages/atom/solid/test/index.test.tsx
@@ -1,5 +1,6 @@
 import {
   RegistryContext,
+  RegistryProvider,
   useAtom,
   useAtomInitialValues,
   useAtomRef,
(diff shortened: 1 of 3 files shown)


packages/atom/solid/test/index.test.tsx
```diff
@@ -10,9 +11,52 @@ import {
 } from "@effect/atom-solid"
 import { assert, describe, it } from "@effect/vitest"
 import { AsyncResult, Atom, AtomRef, AtomRegistry } from "effect/unstable/reactivity"
-import { type Accessor, createComponent, createEffect, createRoot, type Resource } from "solid-js"
+import { type Accessor, createComponent, createEffect, createRoot, type Resource, useContext } from "solid-js"
 
 describe("atom-solid", () => {
+  describe("RegistryProvider", () => {
+    it("does not set a default idle TTL", () => {
```
tim-smart: You can revert the test addition

tim-smart (changes requested):

tim-smart (approved):

PR #7540, opened 2026-08-31
fix(platform-node-shared): keep cooked mode for NodeTerminal.readLine on TTY

### LINKED ISSUE #7529: NodeTerminal.readLine does not echo typed input and ignores Ctrl+C when stdin is a TTY

shahidbeig-a11y:
## Type

- [x] Bug Fix

## Description

`NodeTerminal` enabled stdin raw mode whenever its shared readline interface was acquired. That is required for `readInput` keypress handling, but it breaks `readLine` on a TTY: typed characters are not echoed and Ctrl+C no longer interrupts the process.

This change keeps stdin in cooked mode for `readLine`, enables raw mode only while `readInput` is active, and passes `stdout` to `readline.createInterface` so line editing works on a TTY.

Fixes #7529

## Testing

- `pnpm test --run packages/platform/node-shared/test/NodeTerminal.test.ts`
- Added TTY regression coverage for cooked-mode `readLine`, raw-mode `readInput`, and Ctrl+C interruption

### DIFF
--- a/.changeset/node-terminal-readline-tty.md
+++ b/.changeset/node-terminal-readline-tty.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform-node-shared": patch
+---
+
+NodeTerminal: keep stdin in cooked mode for `readLine` on a TTY so typed input is echoed and Ctrl+C interrupts as expected. Raw mode is now enabled only while reading key input.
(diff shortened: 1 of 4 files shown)


shahidbeig-a11y:
Closing for now.

PR #7547, opened 2026-08-31
PersistedQueue hardening and optimization pass

tim-smart:
Hardening and optimization pass over `PersistedQueue`, covering the SQL, Redis, and memory stores plus the factory. SQL schema changes are applied by `0002_upgrade_schema`, which upgrades existing `0001_create_table` data in place.

Closes EFF-980

## Semantics

- **Attempts count on claim, not on outcome.** The claim bumps `attempts`, interrupt-release decrements it back, and the failure path no longer bumps. Handler metadata is now 1-based. A handler that crashes the process on its final attempt is flipped to `failed` by a periodic pass over exhausted, lock-expired pending rows (mirrored in the Redis reset script).
- **Retry backoff.** New `visible_at` column set from a retry schedule. The poll filters `visible_at <= now` and orders by `visible_at, sequence`. Redis uses a delayed zset swept by the take script; memory delays re-eligibility.
- **Failed state.** `completed BOOLEAN` becomes `state` (`pending` / `completed` / `failed`). Exhausted rows are marked failed instead of becoming invisible, matching the Redis `:failed` destination. Inspect/requeue APIs are a later pass.
- **Decode failures dead-letter immediately.** A schema decode (or JSON parse) failure marks the row failed with the raw element preserved and the cause in `last_failure`, then takes the next item. Decode errors leave `take`'s error channel.
- **Config moves to `make()`.** `maxAttempts` and `retrySchedule` are queue-definition options; `take()` loses its options. This also removes `maxAttempts` from the mailbox key.
- **Patient acks.** `complete`/`retry` retry with backoff up to the lock expiration window (the refresh fiber holds ownership meanwhile) instead of dying after ~1s.

## Retention and cleanup

Completed rows are kept for de-duplication (DurableQueue replays re-offer the same id) and pruned by a new `PersistedQueueStore.cleanup` plus `PersistedQueue.layerCleanup({ interval, timeToLive, failedTimeToLive })`. Defaults: 30 days for completed; failed rows are the dead-letter record and exempt unless `failedTimeToLive` is set. SQL batch-deletes with small LIMITs, Redis trims the ids zset and failed lists, memory prunes its entries.

## Bug fixes

- `id`/`queue_name` widened to 255 (DurableQueue ids like `${queueName}/${idempotencyKey}` overflowed `VARCHAR(36)`).
- `sequence` is 64-bit on all dialects (was int32 on pg and mssql).
- MySQL `element`/`last_failure` are `MEDIUMTEXT` (64KB cap before).
- Memory store take race: two takers passing the latch together no longer crash on an empty set.
- Redis `:ids` dedup set no longer grows forever (timestamp-scored zset trimmed by cleanup).
- mssql: `GETDATE()` → `SYSDATETIME()` (1/300s rounding could put fresh `visible_at` values in the future, hiding them from the poll), and two mssql migration statements had invalid identifier interpolation.

## Optimizations

- `offer` nudges the local poll latch, so same-process work skips the up-to-`pollInterval` wait. No LISTEN/NOTIFY this pass.
- Partial index on pending rows (pg/sqlite/mssql) so pollers stop scanning completed rows.

## Docs

At-least-once delivery (idempotent handlers) is now stated on the module and `take`, and DurableQueue documents that a dead-lettered item parks its workflow until the future requeue API, plus the single-cleanup-instance recommendation.

## Testing

Shared queue behavior runs against memory, SQLite, PostgreSQL, MySQL, MSSQL, and Redis. Migration coverage exercises fresh installs and upgrades from the original `0001` on every SQL backend, including preservation of pending and completed rows, de-duplication, widened identifiers, and payloads over 64 KB. `pnpm check` and `pnpm lint` pass.


packages/effect/src/unstable/persistence/PersistedQueue.ts
```diff
@@ -170,15 +209,39 @@ export const makeFactory = Effect.gen(function*() {
...
+                      ? f(exit.value as S["Type"], { id: item.id, attempts: item.attempts })
+                      : Cause.hasInterruptsOnly(exit.cause)
+                      ? Effect.failCause(exit.cause as Cause.Cause<never>)
+                      : Effect.fail(new DeadLetter(exit.cause))
+                )
+              )
+            )
+          ).pipe(
+            Effect.catchIf(
+              (e): e is DeadLetter => e instanceof DeadLetter,
```
tim-smart: Instead of instanceof use Predicate.isTagged or add a TypeId to the error and use Predicate.hasProperty

packages/effect/src/unstable/persistence/PersistedQueue.ts
```diff
@@ -147,10 +174,20 @@ export const makeFactory = Effect.gen(function*() {
     make<S extends Schema.Constraint>(options: {
       readonly name: string
       readonly schema: S
+      readonly maxAttempts?: number | undefined
+      readonly backoff?: ((attempts: number) => Duration.Input) | undefined
```
tim-smart: Could we use a Schedule here instead?

packages/effect/src/unstable/persistence/PersistedQueue.ts
```diff
@@ -315,38 +495,97 @@ export const layerStoreMemory: Layer.Layer<
...
-          if (exit._tag === "Success") {
-            return Effect.void
-          } else if (!Exit.hasInterrupts(exit)) {
-            item.attempts += 1
-          }
-          if (item.attempts >= options.maxAttempts) {
-            return Effect.void
+        // close before scanning so a mutation after the scan reopens the latch
+        queue.latch.closeUnsafe()
+        const now = yield* Clock.currentTimeMillis
```
tim-smart: I think we can access the clock in the constructor, and then use the unsafe methods in the service members.

Same applies to the redis layer

packages/effect/test/unstable/persistence/PersistedQueueTest.ts
```diff
@@ -141,81 +185,360 @@ export const suiteWith = <R>(
...
+        const fiber = yield* queue.take(Effect.succeed).pipe(Effect.forkScoped)
+        yield* assertNotDelivered(fiber)
+      }), testOptions)
+  })
+}
 
 export const suite = (name: string, layer: Layer.Layer<PersistedQueue.PersistedQueueStore, unknown>) =>
   suiteWith(name, layer, it)
+
+const originalSqlMigration = (tableName: string) =>
```
tim-smart: Lets remove the migration tests

packages/effect/src/unstable/persistence/PersistedQueue.ts
```diff
@@ -281,83 +491,156 @@ export class PersistedQueueStore extends Context.Service<
...
           }
-          queue.items.add(item)
+          const entry = item
+          entry.state = "processing"
+          entry.attempts += 1
+          queue.pending.delete(entry)
           queue.latch.openUnsafe()
-          return Effect.void
+          yield* Effect.addFinalizer((exit) =>
+            Effect.gen(function*() {
```
tim-smart: Effect.fnUntraced

PR #7548, opened 2026-09-01
Add Alchemy integration for Cloudflare cluster

IMax153:
## Summary
- add an Alchemy-native Worker resource with fixed Durable Object bindings, SQLite migrations, and singleton Cron Trigger configuration
- add a shared isolate runtime for class-based Cloudflare Workers
- document Alchemy deployment and Wrangler adoption
- add Miniflare, unit, and type coverage

## Validation
- `pnpm --filter @effect/platform-cloudflare check`
- `pnpm test --run packages/platform/cloudflare/test/CloudflareAlchemy.test.ts packages/platform/cloudflare/test/CloudflareWorker.test.ts packages/platform/cloudflare/test/CloudflareCluster.test.ts packages/platform/cloudflare/test/CloudflareDurableObjects.test.ts`
- `pnpm test-types CloudflareAlchemy.tst.ts`
- `pnpm lint`

Repository-wide `pnpm check` still reports pre-existing Buffer type errors in platform-node and sql/pg tests. Alchemy beta.76 requires Effect RC.112, so the resource is typechecked here while its deployment-time execution awaits that version alignment.

PR #7550, opened 2026-09-01
Add Alchemy deployment for the Cloudflare cluster

tim-smart:
Adds Alchemy v2 deployment for the Cloudflare cluster, stacked on #7322. Closes EFF-988.

`AlchemyCloudflareCluster.make({ entities, layer })` registers four library-owned Durable Object classes, builds the cluster and handler layers into the isolate-lifetime scope, and returns `provide`, `wake`, native namespace bindings, and `context`. Users own the Worker and cron declaration; they do not declare or re-export Alchemy Durable Object classes.

- Wrangler and Alchemy now use the same `CloudflareDurableObjectPrograms`. Wrangler builds each program inside its existing initialization barrier, then stores the completed program for RPCs and alarms. Transport types are shared with the internals.
- Plan-time namespace and context access throws a field-specific diagnostic. `provide` and `wake` remain inert during planning. Native namespace RPC methods retain a `never` error channel to avoid Alchemy typed-error envelopes. Alchemy's default fetch 404 and Wrangler's fetch rejection are deliberate.
- Alchemy is an optional peer (`>=2.0.0-beta <3`), pinned to `2.0.0-beta.76` for development. The Wrangler entry stays Alchemy-free. Release-age exclusions use package-name patterns; the existing Node type override remains necessary for Alchemy's dependency graph.
- The example and typetests have a dedicated Alchemy typecheck step in normal PR CI and the manual `Alchemy Canary` workflow. The example project is no longer referenced by the root typecheck. Only the Alchemy typetest is excluded from the shared tests project.

`wake(name)` returns the handler function required by `Cloudflare.Workers.cron`. The example provides `CronEventSourceLive` on Worker initialization; the cluster layer itself uses the isolate scope.

Validation: all 170 Cloudflare tests and 18 Alchemy typetests pass; root typecheck, dedicated Alchemy typecheck, lint, package build, stripped declaration emission, and a fresh dependency resolution pass. Test changes are confined to the supplied regression-test commit. Repository JSDoc validation reports unrelated existing errors in `packages/effect/src/Logger.ts`.

The Alchemy glue remains typecheck-only against workspace Effect because the pinned Alchemy runtime targets the published Effect API. The README live `alchemy deploy` smoke is explicitly **not verified**; it still requires Cloudflare credentials. Supersedes #7548.

Singleton alarm state reads are deferred with `Effect.suspend`, and workflow, queue, and singleton programs capture Effect's Clock service while reading its current timestamp on each operation. The public cron example now provides `CronEventSourceLive`. All six alarm and Clock regressions pass; the live smoke remains unverified.

`Cluster` has a single definition in the Alchemy-free `internal/cluster.ts` module and remains exported by `AlchemyCloudflareCluster`. A clean shared-tests compilation resolves zero Alchemy declarations; stripped declaration emission preserves the public type and its imports.

### DIFF
--- a/packages/platform/cloudflare/src/internal/entityRuntime.ts
+++ b/packages/platform/cloudflare/src/internal/entityRuntime.ts
@@ -25,7 +25,7 @@ import * as Reply from "effect/unstable/cluster/Reply"
 import * as RunnerAddress from "effect/unstable/cluster/RunnerAddress"
 import * as Rpc from "effect/unstable/rpc/Rpc"
 import * as RpcSchema from "effect/unstable/rpc/RpcSchema"
-import type { EntityDeliveryOptions } from "../CloudflareDurableObjectPrograms.ts"
+import type { EntityDeliveryOptions, EntityInvokeResult } from "../CloudflareDurableObjectPrograms.ts"
 import { encodeName } from "./clusterName.ts"
 import type { EntityKeepAlive } from "./entityKeepAlive.ts"
 import {
@@ -44,7 +44,7 @@ import {
 import { type EntityRegistration, getEntityRegistration } from "./entityRegistry.ts"
 import { CurrentEntityName, CurrentReplyRegistry, type EntityReplyRegistry, makeReplyRegistry } from "./entityReply.ts"
 import { armAlarm, earliestDeliverAt, withTransaction } from "./entityStorage.ts"
-import { decodeReplyFor, decodeRequest, encodeReplyFor, type InvokeResult, peekEnvelopeTag } from "./entityWire.ts"
+import { decodeReplyFor, decodeRequest, encodeReplyFor, peekEnvelopeTag } from "./entityWire.ts"
 
 interface CachedHandlers {
   readonly handlers: Record<string, (request: any) => any>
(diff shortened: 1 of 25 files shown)


packages/platform/cloudflare/src/internal/entityRuntime.ts
```diff
@@ -243,11 +244,7 @@ interface RunOptions {
 }
 
 /** @internal */
-export interface DeliveryOptions {
-  readonly deliverAt?: number | undefined
-  readonly primaryKey?: string | null | undefined
-  readonly replyTo?: string | undefined
-}
+export type DeliveryOptions = EntityDeliveryOptions
```
tim-smart: Instead of re-exporting use the original types

packages/platform/cloudflare/src/internal/entityWire.ts
```diff
@@ -31,10 +32,10 @@ export const InvokeResult = Schema.Union([
   Schema.Struct({ _tag: Schema.Literal("MailboxFull") }),
   Schema.Struct({ _tag: Schema.Literal("EncodedMessageTooLarge") }),
   Schema.Struct({ _tag: Schema.Literal("AskDeduplicatedToTell") })
-])
+]) satisfies Schema.Codec<EntityInvokeResult>
 
 /** @internal */
-export type InvokeResult = typeof InvokeResult.Type
+export type InvokeResult = EntityInvokeResult
```
tim-smart: remove re-export

PR #7561, opened 2026-09-01
fix(httpapi): transform middleware-declared errors like endpoint errors

jensdev:
## Problem

Attaching an `HttpApiMiddleware` that declares an `error` schema rewrote the response for that status on **every** endpoint, even when the middleware declared the exact same schema object the endpoint already declared:

```jsonc
"500": { "schema": { "anyOf": [
  { "$ref": "#/components/schemas/Err" },
  { "$ref": "#/components/schemas/Err" }   // duplicate
] } }
```

Endpoints store declared errors as response codecs (transformResponseSchema → Schema.toCodecJson); middleware stored them raw. Everything downstream merges the two by object identity, so they could never match.

This also broke encoding, not just the document: getResponseEncode runs JSON.stringify over the schema's encoded form, and a raw schema's encoded form isn't JSON-safe. A middleware error with a Schema.BigInt field fails at response time with Expected a JSON-serializable response body. (It looked fine for Schema.Date only because JSON.stringify happens to reproduce the ISO string.)

Fix

- Middleware errors now go through the same response transform as endpoint errors, so middleware.error and endpoint.error hold the same kind of thing.
- transformResponseSchema is memoized on its input schema, so one declaration yields one instance and the existing Set in getErrorSchemas collapses it.

Behavior preserved

A middleware error that is genuinely new for a status is still documented; two different schemas on one status still emit an anyOf.

Note on the snapshot change

The updated platform/node snapshot is the same bug via a different door: the fixture declares error: [UserError, UserError]. The transform ran per element, so the Set couldn't collapse those either. Status 400 is now a single $ref.

Tests

Five cases across OpenApi.test.ts and HttpApiBuilder.test.ts. Two fail on main (same schema collapses; middleware error encodes through the response codec); three pass before and after, guarding against over-correction.

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### DIFF
--- a/.changeset/httpapi-middleware-error-dedupe.md
+++ b/.changeset/httpapi-middleware-error-dedupe.md
@@ -1,0 +1,23 @@
+---
+"effect": patch
+---
+
+Fix `HttpApiMiddleware` errors duplicating an endpoint's response schema
+
+An endpoint stored its declared errors as response codecs (`transformResponseSchema`)
+while a middleware stored the schemas it was given, raw. Everything downstream merges
+the two sets by identity, so a middleware error could never match an endpoint error —
+even when both sides declared the exact same schema.
+
+Two consequences, both fixed:
+
+- `OpenApi.fromApi` emitted `anyOf: [{ $ref: "#/components/schemas/Err" }, { $ref: "#/components/schemas/Err" }]`
+  for the shared status on every endpoint the middleware was attached to.
+- A middleware error whose encoded form is not JSON-serializable (a `Schema.BigInt`
+  field, for example) failed to encode at runtime with
+  `Expected a JSON-serializable response body`.
+
+Middleware errors now go through the same response transform as endpoint errors, and
+that transform is memoized on its input schema, so declaring one schema in both places
+yields one instance. Middleware errors that are genuinely new for a status are still
+documented, and two different schemas on one status still produce an `anyOf`.
--- a/.changeset/httpapi-middleware-error-dedupe.md
+++ b/.changeset/httpapi-middleware-error-dedupe.md
@@ -17,7 +17,11 @@ Two consequences, both fixed:
   field, for example) failed to encode at runtime with
   `Expected a JSON-serializable response body`.
 
-Middleware errors now go through the same response transform as endpoint errors, and
-that transform is memoized on its input schema, so declaring one schema in both places
-yields one instance. Middleware errors that are genuinely new for a status are still
-documented, and two different schemas on one status still produce an `anyOf`.
+`HttpApiEndpoint.getErrorSchemas` now gives a middleware's declared errors the same
+treatment the endpoint gave its own, so a middleware attached to endpoints with
+different `disableCodecs` settings is resolved correctly for each. The transform is
+memoized on its input schema, so declaring one schema in both places yields one
+instance.
+
+Middleware errors that are genuinely new for a status are still documented, and two
+different schemas on one status still produce an `anyOf`.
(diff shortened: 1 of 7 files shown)


packages/effect/src/unstable/httpapi/HttpApiMiddleware.ts
```diff
@@ -380,7 +380,8 @@ export const Service = <
 
 function getError(error: ErrorConstraint | undefined): ReadonlySet<Schema.Top> {
   if (error === undefined) return new Set()
-  return new Set(Array.isArray(error) ? error : [error])
+  const schemas = Array.isArray(error) ? error : [error]
```
tim-smart: We can't transform here because middleware can be shared by endpoints with different `disableCodecs` values. This breaks deduplication and encoding for `disableCodecs: true`. Can we transform in `getErrorSchemas` based on the endpoint instead? We'll also need to store `disableCodecs` on the endpoint and preserve it in `optionsFromEndpoint`.

packages/effect/src/unstable/httpapi/HttpApiEndpoint.ts
```diff
@@ -1157,7 +1157,18 @@ function getSuccessResponse(
   return new Set(disableCodecs ? schemas : schemas.map(transformResponseSchema))
 }
 
-function transformResponseSchema(schema: Schema.Top): Schema.Top {
+const transformedResponseSchemas = new WeakMap<Schema.Top, Schema.Top>()
+
+/** @internal */
```
tim-smart: Can we use `Function.memoize` here? It has the same `WeakMap` behavior and avoids the extra uncached function. This should be `memoize`, not `memoizeIdempotent`, since the transform isn't idempotent.

packages/effect/test/unstable/httpapi/OpenApi.test.ts
```diff
@@ -316,6 +316,111 @@ describe("OpenApi", () => {
     })
   })
 
+  it("collapses a middleware error that repeats an endpoint error", () => {
```
tim-smart, line 319: Can we add a test where a `disableCodecs: true` endpoint and its middleware use the same error schema? That catches the regression above.

.changeset/httpapi-middleware-error-dedupe.md
```diff
@@ -0,0 +1,23 @@
+---
+"effect": patch
+---
+
+Fix `HttpApiMiddleware` errors duplicating an endpoint's response schema
```
tim-smart: Can we shorten this to the summary and bullets? The before/after detail can stay in the PR description.

.changeset/httpapi-middleware-error-dedupe.md
```diff
@@ -0,0 +1,8 @@
+---
+"effect": patch
+---
+
+Fix `HttpApiMiddleware`-declared errors being duplicated and mis-encoded:
+
+- `OpenApi.fromApi` emitted a duplicated `anyOf` for the middleware's status on every endpoint it was attached to, even when the middleware declared the same schema the endpoint already declared.
+- A middleware error whose encoded form is not JSON-serializable (a `Schema.BigInt` field, for example) failed to encode with `Expected a JSON-serializable response body`.
```
tim-smart:
```suggestion
Fix `HttpApiMiddleware`-declared errors being duplicated and mis-encoded.
```

jensdev:
@tim-smart  Thanks for the review. I added another test and interrupted the auto-merge

PR #7570, opened 2026-09-01
Fix Worker.run hanging when a worker dies before the ready handshake

### LINKED ISSUE #7566: Worker `run` parks uninterruptibly when a worker dies before the handshake
If a worker dies before sending its ready message, `WorkerPlatform.spawn(...).run(...)` parks forever in an uninterruptible region. The owning scope never closes and interrupting the fiber does nothing.

In `Worker.makePlatform` (`packages/effect/src/unstable/workers/Worker.ts`, `run` at L180, the wait at L219):

```ts
const run = (handler, opts) =>
  Effect.uninterruptibleMask((restore) =>
    Effect.scopedWith(Effect.fnUntraced(function*(scope) {
      ...
      yield* options.listen({ port, scope, emit, deferred: fiberSet.deferred })
      yield* ready.await                                // never opens
      ...
      return yield* restore(FiberSet.join(fiberSet))    // never reached
    })))
```

`ready` is opened only from `emit`, on the platform ready message (`data[0] === 0`). The platform layers report worker death by failing `fiberSet.deferred` — `NodeWorker`'s `listen` does this on `error`, `messageerror` and `exit` — but that deferred is only ever observed by the `FiberSet.join` *after* `ready.await`. So a pre-handshake death is never seen. And because `ready.await` is inside the uninterruptible region rather than under `restore`, the fiber can't be interrupted out of it either.

`RpcClient.layerProtocolWorker` inherits this. Its `Effect.retry(Schedule.spaced(1000))` around `backing.run(...)` never fires, because `run` hangs rather than failing.

### Reproduction

`dying-worker.mjs` — a worker that dies before the handshake:

```js
process.exit(1)
```

```js
import { Worker as NodeWorkerThreads } from "node:worker_threads"
import { NodeWorker } from "@effect/platform-node"
import { Effect, Fiber } from "effect"
import { Worker } from "effect/unstable/workers"

  const platform = yield* Worker.WorkerPlatform
  const backing = yield* platform.spawn(0)
  yield* backing.run(() => Effect.void)
}).pipe(Effect.scoped, Effect.provide(NodeWorker.layer(() => new NodeWorkerThreads("./dying-worker.mjs"))))

const fiber = Effect.runFork(program)
await new Promise((r) => setTimeout(r, 1000))

console.log(await Promise.race([
  Effect.runPromise(Fiber.interrupt(fiber)).then(() => "interrupt completed"),
  new Promise((r) => setTimeout(() => r("STILL PARKED after 5s"), 5000))
]))
// => STILL PARKED after 5s
```

Control: the identical program against a worker that posts `[0]` and *then* dies prints

```
run() failed as expected: Cause([Fail(WorkerError: The worker has exited with code: 1)])
interrupt completed -- scope closed
```

so the hang is specific to a death before the handshake.

Reproduced on `effect@4.0.0-rc.112` with `@effect/platform-node@4.0.0-rc.112`, Node 24.

### Possible fix

Have the wait observe the deferred as well, so a pre-handshake death surfaces the same `WorkerError` a later one does:

```ts
yield* Effect.raceFirst(ready.await, Deferred.await(fiberSet.deferred))
```

Two things I couldn't settle from outside, so I'd rather ask than guess:

1. Should a pre-handshake death **fail** `run` — which is what makes `RpcClient.layerProtocolWorker`'s retry respawn the worker — or just become interruptible?
2. Is there a reason `ready.await` needs to be uninterruptible? Moving it under `restore` fixes the teardown half on its own but still leaves a dead worker unnoticed.

Happy to send a PR with a test once you've picked a direction.

tim-smart:
`Worker.run` wraps its body in `Effect.uninterruptibleMask` and waited on the ready latch with a bare `yield* ready.await`. When a worker dies before sending the platform ready message, the platform `listen` hook fails `fiberSet.deferred`, but nothing observes that failure until `FiberSet.join` after the latch wait, so the fiber parks forever in an uninterruptible region. Neither `Fiber.interrupt` nor `Effect.timeout` can unwind it, and `RpcClient.layerProtocolWorker`'s retry never fires because the fiber hangs instead of failing.

The ready wait now races the latch against `FiberSet.join(fiberSet)` under `restore`:

- A worker exiting (or erroring) before the handshake fails `run` with the `WorkerError` from the platform listen hook, which also lets the `RpcClient` retry schedule kick in.
- The handshake wait is interruptible, so callers can time out or interrupt while waiting for readiness. Scope finalizers still run on interruption, terminating the worker.

Added tests in `packages/effect/test/unstable/workers/Worker.test.ts` using an in-memory platform covering: death before `run` starts waiting, death while waiting, interruption while waiting, and death after the handshake. The first three hang on the previous code.

Validation: `pnpm lint-fix`, `pnpm check`, targeted runs of the new Worker tests, `WorkerError.test.ts`, `rpc/RpcClient.test.ts`, and platform-node `RpcServer.test.ts` (95 tests, exercises NodeWorker end to end).

Closes #7566
Closes EFF-999

### DIFF
--- a/.changeset/worker-run-early-exit.md
+++ b/.changeset/worker-run-early-exit.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-Fix `Worker.run` hanging uninterruptibly when a worker dies before the ready handshake. The ready wait now races the worker's failure signal, so an early exit fails `run` with a `WorkerError`, and the wait is interruptible.
+Fix `Worker.run` hanging uninterruptibly when a worker dies before the ready handshake.
--- a/.changeset/worker-run-early-exit.md
+++ b/.changeset/worker-run-early-exit.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `Worker.run` hanging uninterruptibly when a worker dies before the ready handshake. The ready wait now races the worker's failure signal, so an early exit fails `run` with a `WorkerError`, and the wait is interruptible.
--- a/packages/effect/src/unstable/workers/Worker.ts
+++ b/packages/effect/src/unstable/workers/Worker.ts
@@ -216,7 +216,9 @@ export const makePlatform = <W>() =>
                   },
                   deferred: fiberSet.deferred as any
                 })
-                yield* ready.await
+                // Fail fast if the worker dies before signalling readiness,
+                // and allow interruption while waiting for the handshake.
+                yield* restore(Effect.raceFirst(ready.await, FiberSet.join(fiberSet)))
                 currentPort = port
                 if (buffer.length > 0) {
                   for (const [message, transfers] of buffer) {
--- a/packages/effect/test/unstable/workers/Worker.test.ts
+++ b/packages/effect/test/unstable/workers/Worker.test.ts
@@ -1,0 +1,91 @@
+import { assert, describe, it } from "@effect/vitest"
+import { Deferred, Effect, Exit, Fiber } from "effect"
+import * as Worker from "effect/unstable/workers/Worker"
+import { isWorkerError, WorkerError, WorkerReceiveError } from "effect/unstable/workers/WorkerError"
+
+interface TestPort {
+  readonly postMessage: (message: any, transfers?: any) => void
+}
+
+const testPort: TestPort = { postMessage: () => {} }
+
+const makePlatform = (
+  onListen: (options: {
+    readonly emit: (data: any) => void
+    readonly deferred: Deferred.Deferred<never, WorkerError>
+  }) => void
+) =>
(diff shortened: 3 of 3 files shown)


.changeset/worker-run-early-exit.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `Worker.run` hanging uninterruptibly when a worker dies before the ready handshake. The ready wait now races the worker's failure signal, so an early exit fails `run` with a `WorkerError`, and the wait is interruptible.
```
tim-smart:
```suggestion
Fix `Worker.run` hanging uninterruptibly when a worker dies before the ready handshake.
```

PR #7574, opened 2026-09-01
unstable/ai: recover unknown tool calls

candrewlee14:
Follow-up to #7452, rewritten on top of #7588. Supersedes #6432.

## Problem

If the model calls a tool that isn't in the toolkit, the whole `generateText` / `streamText` call fails. `Response.Part(toolkit)` builds its union from the toolkit's tools, so there's no schema member for the unknown name and the decode fails. Even if it decoded, `resolveToolCalls` skips it, so it'd be left without a result and the next request would be rejected.

For agents whose toolkit changes between turns this is a normal thing to happen, not a hallucination. The model saw the tool a turn ago and now it's gone. Telling the model "that tool doesn't exist, try again" is a better outcome than failing the turn.

## Change

`generateText` and `streamText` take a new option, `unknownToolCalls: "error" | "return"`. Default is `"error"`, which is today's behaviour. With `"return"`, an unknown tool call comes back as a `tool-call-error` part with the original call, its params, and a `ToolNotFoundError`. `Prompt.fromResponseParts` turns that into the assistant tool call plus a failed tool result, so the model can correct itself.

```ts
const response = yield* LanguageModel.generateText({ prompt, toolkit, unknownToolCalls: "return" })
response.toolCallErrors // [{ name, params, error }]
```

Calls to tools that *are* in the toolkit aren't affected. #7588 already routes their parameter failures through `failureMode`, and I left that alone. With `disableToolCallResolution: true` those calls still fail the operation like they do on main. So the split is: the tool decides for its own calls, the operation decides for calls that don't belong to any tool. If there's no toolkit, or it's empty, every call is unknown and the option applies the same way.

On @IMax153's concern from #7452 about callers silently missing invalid calls: with the default, `toolCallErrors` is `Array<never>` and `"tool-call-error"` isn't in the response part type, so nothing changes for existing code unless you opt in.

## Notes

- Recovery happens in the same decode that validates the response. Parts are decoded one at a time, and only a `tool-call` naming an unknown tool is recovered. Anything else that fails to decode still fails the operation. Params are never validated twice.
- The `tool-call-error` part is produced by decoding, not constructed by hand, so a call that's malformed in some other way (non-string `id`, bad metadata) keeps failing rather than coming back under a misleading error.
- Provider-executed calls are never recovered.
- Params on the returned part go through `Formatter` so they're JSON-safe in history (the `1n` case from the last review).
- `HttpRequestDetails` / `HttpResponseDetails` moved to `internal/http-details.ts` so `Response` can import `AiError` without a cycle. Both modules still re-export them.
- `Response.AllParts` includes the new part unconditionally.

Tests cover `generateText`, `streamText` and `generateObject`, with a toolkit, without one, and with an empty one, resolution on and off, a non-literal option value, and a property test that a returned part always satisfies the response schema. Also asserted: known-tool parameter failures still fail with resolution disabled, an unrecoverable response runs no handlers, and history never ends up with a call without a result or a result without a call.

If you'd rather this lived on the toolkit (a second type parameter, which would flow to `Chat` and `generateObject` for free) I'm happy to rework it that way.

https://claude.ai/code/session_019wFhhXHNc4Fnkye1orWQqr

### DIFF
--- a/packages/effect/src/unstable/ai/LanguageModel.ts
+++ b/packages/effect/src/unstable/ai/LanguageModel.ts
@@ -20,6 +20,7 @@ import type * as JsonSchema from "../../JsonSchema.ts"
 import * as Option from "../../Option.ts"
 import * as Predicate from "../../Predicate.ts"
 import * as Queue from "../../Queue.ts"
+import * as Result from "../../Result.ts"
 import * as Schema from "../../Schema.ts"
 import * as SchemaAST from "../../SchemaAST.ts"
 import * as Semaphore from "../../Semaphore.ts"
@@ -29,7 +30,9 @@ import type { Span } from "../../Tracer.ts"
 import type { Concurrency, Mutable, NoExcessProperties } from "../../Types.ts"
 import * as AiError from "./AiError.ts"
 import { defaultIdGenerator, IdGenerator } from "./IdGenerator.ts"
+import { encodeAiError } from "./internal/ai-error.ts"
 import * as InternalCodecTransformer from "./internal/codec-transformer.ts"
+import { toJson } from "./internal/json.ts"
 import * as Prompt from "./Prompt.ts"
 import * as Response from "./Response.ts"
 import * as ResponseIdTracker from "./ResponseIdTracker.ts"
@@ -101,7 +104,7 @@ export interface Service {
     <Options extends NoExcessProperties<GenerateTextOptionsWithoutToolkit, Options>>(
       options: Options & GenerateTextOptionsWithoutToolkit
     ): Effect.Effect<
-      GenerateTextResponse<{}>,
+      GenerateTextResponse<{}, "decoded", ExtractUnknownToolCalls<Options>>,
       ExtractError<Options>,
       ExtractServices<Options>
     >
@@ -115,7 +118,7 @@ export interface Service {
     >(
       options: Options & GenerateTextOptions<Tools> & { readonly toolkit: ToolkitInput<Tools> }
     ): Effect.Effect<
-      GenerateTextResponse<Tools, ExtractToolParametersMode<Options>>,
+      GenerateTextResponse<Tools, ExtractToolParametersMode<Options>, ExtractUnknownToolCalls<Options>>,
       ExtractError<Options>,
       ExtractServices<Options>
     >
@@ -127,7 +130,11 @@ export interface Service {
     >(
       options: Options & GenerateTextOptions<ExtractTools<Options>> & { readonly toolkit: Options["toolkit"] }
     ): Effect.Effect<
-      GenerateTextResponse<ExtractTools<Options>, ExtractToolParametersMode<Options>>,
+      GenerateTextResponse<
+        ExtractTools<Options>,
+        ExtractToolParametersMode<Options>,
+        ExtractUnknownToolCalls<Options>
+      >,
       ExtractError<Options>,
       ExtractServices<Options>
     >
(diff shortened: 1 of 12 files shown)


tim-smart:
Hey! It will be a while until this gets a proper review as it is a large change and we have higher priorities atm :)

candrewlee14:
Thanks for letting me know! I'm also totally understanding if you throw this out and solve it in a different way, of course.
I think the underlying problem is worth solving though :)

packages/effect/src/unstable/ai/LanguageModel.ts
```diff
@@ -569,6 +625,20 @@ export type ExtractToolParametersMode<Options> = Options extends {
...
+ * keeps both modes, so the response type admits the part the operation may
+ * produce.
+ *
+ * @category utility types
+ * @since 4.0.0
+ */
+export type ExtractUnknownToolCalls<Options> = Options extends {
+  readonly unknownToolCalls: infer Mode
+} ? [Exclude<Mode, undefined>] extends [never] ? "error" : Extract<Mode, "error" | "return">
+  : "error"
```
danieljvdm, line 640: optional `unknownToolCalls` gets typed as `"error"` even when set to `"return"`, so `toolCallErrors` becomes `never[]`. same issue in streaming and `UnknownToolCallsOf`.

packages/effect/src/unstable/ai/LanguageModel.ts
```diff
@@ -2378,6 +2475,95 @@ const makeToolkitWithOpaqueParameters = <Tools extends Record<string, Tool.Any>>
...
+      // reasons validate their own fields. A call which does not carry both as
+      // strings is malformed rather than unresolvable, and keeps failing.
+      if (!Predicate.isString(part.name) || !Predicate.isString(part.id)) {
+        return Effect.fail(decoded.failure)
+      }
+      // A call which names a tool in the toolkit belongs to that tool, and
+      // `Toolkit` routes its parameter failure through the tool's own
+      // `failureMode` when it resolves the call. When resolution is disabled
+      // the caller owns the call, so it fails here exactly as it does today.
+      if (Object.hasOwn(toolkit.tools, part.name)) {
```
danieljvdm, line 2536: `toString`, `constructor` and `__proto__` still fail in return mode. the remaining lookups need `Object.hasOwn` too.

`toString` with finish reason `"length"` also produces two results for one call.

packages/effect/src/unstable/ai/LanguageModel.ts
```diff
@@ -2378,6 +2475,95 @@ const makeToolkitWithOpaqueParameters = <Tools extends Record<string, Tool.Any>>
...
+      // must keep failing the operation instead of being returned to the model
+      // under a misleading error.
+      return Effect.catchCause(
+        decodePart({
+          type: "tool-call-error",
+          id: part.id,
+          name: part.name,
+          params: toJson(part.params),
+          error: encodeAiError(AiError.make({ module: "LanguageModel", method, reason })),
+          metadata: part.metadata
```
danieljvdm, line 2560: `providerExecuted: "true"` gets dropped here and escapes validation. same for `1` and `null`. can we validate before converting?

packages/effect/test/unstable/ai/LanguageModel.test.ts
```diff
@@ -2001,6 +2002,515 @@ describe("LanguageModel", () => {
...
+          for (const part of result.success.content) {
+            if (part.type === "tool-call-error") {
+              strictEqual(typeof part.id, "string")
+              strictEqual(typeof part.name, "string")
+              assertTrue(AiError.isAiError(part.error))
+            }
+            if (part.type === "tool-call") {
+              strictEqual(typeof part.id, "string")
+              strictEqual(typeof part.name, "string")
+            }
```
danieljvdm, line 2411: can we assert the full schema here and generate invalid `providerExecuted` values too?

danieljvdm (changes requested):
the new response part and generic make sense for keeping the tool types intact. my concern is recovery rewriting calls while execution still reads the originals.

could we classify each call once and use that for recovery, execution and history? we'd still keep the raw params for `Toolkit` to decode.

PR #7577, opened 2026-09-02
fix(Queue): prevent duplicate messages from resumed producers

### LINKED ISSUE #7575: Queue duplicates a message when a resumed producer takes from the same queue

Effect v4, reproduced from `main` at `d812596a1` (package version `4.0.0-rc.112`), using Node.js `26.5.0` on macOS.

Fill a capacity-one queue, start a producer that offers another message and then takes from the same queue, and consume the first message to unblock the producer.

```ts
import { Effect, Fiber, Queue } from "effect"

const result = await Effect.runPromise(Effect.gen(function*() {
  const queue = yield* Queue.bounded<number>(1)
  yield* Queue.offer(queue, 0)

  const producer = yield* Effect.forkChild(
    Effect.gen(function*() {
      yield* Queue.offer(queue, 1)
      return yield* Queue.take(queue)
    }),
    { startImmediately: true }
  )

  const first = yield* Queue.take(queue)
  const second = yield* Fiber.join(producer)
  const remaining = yield* Queue.clear(queue)
  yield* Queue.shutdown(queue)
  return { first, second, remaining }
}))

console.log(result)
```

Both offered messages have been consumed, so the queue should be empty:

```ts
{ first: 0, second: 1, remaining: [] }
```

### What do you see instead?

The queue contains an extra copy of `1`:

```ts
{ first: 0, second: 1, remaining: [1] }
```

### Additional information

The reproduction uses only public APIs, without mocks, sleeps, or a custom scheduler. `startImmediately` ensures the producer has reached its blocked offer before the first take.

In [`releaseCapacity`](https://github.com/Effect-TS/effect/blob/d812596a1/packages/effect/src/Queue.ts#L1939-L1946), the pending offer is removed after `entry.resume(exitTrue)`. Resuming the producer can synchronously run its `Queue.take`, reentering `releaseCapacity` while the same offer is still registered and enqueueing it a second time.

This is the same state-before-callback ordering concern addressed for Deferred in #7364.

tim-smart:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### Why

A bounded queue can retain an extra copy of a message after both offered messages have been consumed. `releaseCapacity` resumed a blocked producer before removing its pending offer from `state.offers`. When the resumed producer synchronously takes from the same queue, the take reenters `releaseCapacity` while the offer is still registered, so the same message is admitted twice.

### What Changes

In `releaseCapacity`:

- Remove the offer entry from `state.offers` before calling `entry.resume`, so a reentrant call cannot see it.
- Recompute available capacity on each loop iteration instead of caching it once, since a reentrant call can add or remove messages while the loop is suspended in `resume`.
- Bail out if a reentrant call transitioned the queue to `Done` mid-loop.

Includes the regression test from #7576 (thanks for the deterministic repro), which now passes.

### Verification

```bash
pnpm install
pnpm lint-fix
pnpm vitest --run packages/effect
pnpm check
git diff --check
```

All 285 test files pass (9003 tests), including the new regression test. Typecheck and lint are clean.

## Related

- Closes #7575
- Supersedes the reproduction-only draft #7576

Closes EFF-1006

### DIFF
--- a/.changeset/queue-release-capacity-reentry.md
+++ b/.changeset/queue-release-capacity-reentry.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix duplicate Queue messages when a resumed producer synchronously takes from the same queue. `releaseCapacity` now removes a pending offer before resuming its producer and recomputes available capacity on every iteration, so reentrant calls can no longer admit the same offer twice or exceed the queue's capacity.
--- a/packages/effect/src/Queue.ts
+++ b/packages/effect/src/Queue.ts
@@ -1936,22 +1936,28 @@ const releaseCapacity = <A, E>(self: Dequeue<A, E>): boolean => {
     }
     return false
   }
-  let n = self.capacity - self.messages.length
   for (const entry of self.state.offers) {
-    if (n === 0) break
-    else if (entry._tag === "Single") {
+    // `entry.resume` can run the producer fiber synchronously, which may
+    // reenter this function and mutate `self.state` and `self.messages`, so
+    // remove the entry before resuming and recompute capacity every iteration
+    const state = self.state as Queue.State<A, E>
+    if (state._tag === "Done") {
+      return Pull.isDoneCause(state.exit.cause)
+    }
+    let n = self.capacity - self.messages.length
+    if (n <= 0) break
+    if (entry._tag === "Single") {
       MutableList.append(self.messages, entry.message)
-      n--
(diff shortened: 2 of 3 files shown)


tim-smart:
Superseded by #7576, which includes the same releaseCapacity fix plus the zero-capacity handoff fix and broader regression coverage.

PR #7583, opened 2026-09-02
fix(RequestResolver): evict cancelled pending cache entries

### LINKED ISSUE #7582: RequestResolver.withCache retains cancelled pending entries and hangs equal retries

1. Create a resolver that completes requests with `"Alice"`.
2. Hold its batching delay with `RequestResolver.setDelayEffect(Deferred.await(gate))`.
3. Wrap it with `RequestResolver.withCache({ capacity: 10 })`.
4. Start a request and interrupt its caller before opening the gate.
5. Start an equal request, then open the gate.

[Executable regression with an uncached control](https://github.com/kitlangton/effect/blob/repro/request-cache-cancellation/packages/effect/test/Request.test.ts).

The retry invokes the resolver and completes with `"Alice"`, as the uncached control does.

### What do you see instead?

The resolver does not execute for the retry, which stays suspended even after the batching gate opens.

### Additional information

The cache registers the first pending entry before execution. Cancellation removes that entry from the batch but does not complete or evict its cached entry. A subsequent equal request finds an unfinished cached entry, attaches its completion to the abandoned request, and is excluded from execution by `preCheck`.

The regression controls the ordering with a Deferred rather than relying on a timeout. It also covers cancellation while other batch entries remain and retries started by cancellation finalizers.

kitlangton:
## Type

- [x] Bug Fix
- [x] Documentation Update

### Why

Cancelling a request before its batch runs removes it from the batch but leaves an unfinished entry in `RequestResolver.withCache`. An equal retry attaches to that abandoned entry and never reaches the resolver.

### What Changes

Complete removed pending entries internally with interruption and evict their cache entries. Suppress notifications to cancelled callers, preserve entry-identity guards, and finish cache cleanup before interrupting the delay fiber, whose finalizers may immediately retry.

| Scenario | Before | After |
| --- | --- | --- |
| Cancel pending request, retry equal request | Retry stays suspended | Retry executes successfully |
| Delay cancellation finalizer immediately retries | Retry attaches to abandoned work | Retry sees a fresh cache entry |
| One of two coalesced callers cancels | Other caller remains active | Unchanged |
| Success or ordinary failure completes | Result cached | Unchanged |

### Scope

Pending-request cancellation and interruption-result caching, with an `effect` patch changeset and matching JSDoc. Shared work retains its existing cancellation protection; no blanket uninterruptibility is added.

### Verification

```bash
pnpm test --run packages/effect/test/Request.test.ts
pnpm lint
pnpm check
pnpm jsdocs --check
git diff --check origin/main
```

All 23 tests pass. The original cached-retry regression fails without the fix, while the uncached control passes. Coverage includes surviving batches, either coalesced caller cancelling, ordinary failure caching, interruption results, replacement entries, and silent/idempotent cancellation with finalizer-triggered retry. Lint, typecheck, JSDoc, and whitespace checks pass.

## Related

Closes #7582.

## Audit

### DIFF
--- a/.changeset/request-cache-cancellation.md
+++ b/.changeset/request-cache-cancellation.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `RequestResolver.withCache` retaining abandoned entries when a pending request is cancelled, which could cause later equal requests to hang. Results containing interruptions are no longer cached; completed successes and other failures remain cached, and cancelling one caller does not interrupt requests shared with other callers.
--- a/.changeset/request-cache-cancellation.md
+++ b/.changeset/request-cache-cancellation.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-Fix `RequestResolver.withCache` retaining abandoned entries when a pending request is cancelled, which could cause later equal requests to hang. Results containing interruptions are no longer cached; completed successes and other failures remain cached, and cancelling one caller does not interrupt requests shared with other callers.
+Fix `RequestResolver.withCache` retaining abandoned entries when a pending request is cancelled
--- a/packages/effect/src/RequestResolver.ts
+++ b/packages/effect/src/RequestResolver.ts
@@ -1118,8 +1118,9 @@ export const asCache: {
  *
  * **Gotchas**
  *
- * Entries do not expire by time, and completed failures are cached the same as
- * successes. Request equality controls cache hits.
+ * Entries do not expire by time, and completed failures without interruptions
+ * are cached the same as successes. Results containing interruptions are not
+ * cached. Request equality controls cache hits.
  *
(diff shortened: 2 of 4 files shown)


.changeset/request-cache-cancellation.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `RequestResolver.withCache` retaining abandoned entries when a pending request is cancelled, which could cause later equal requests to hang. Results containing interruptions are no longer cached; completed successes and other failures remain cached, and cancelling one caller does not interrupt requests shared with other callers.
```
tim-smart:
```suggestion
Fix `RequestResolver.withCache` retaining abandoned entries when a pending request is cancelled
```

PR #7584, opened 2026-09-02
fix(PartitionedSemaphore): preserve new waiters during stale cleanup

### LINKED ISSUE #7580: PartitionedSemaphore late cancellation can delete a replacement partition and strand its waiter

1. Hold the only permit of a `PartitionedSemaphore`.
2. Start waiter A under key `"same"`.
3. Release the permit, pausing A after selection but before its take finishes.
4. Register waiter B under the same key, creating a new waiter partition.
5. Interrupt A so its acquired permit is refunded.

[Executable regression with a different-key control](https://github.com/kitlangton/effect/blob/repro/partitioned-semaphore/packages/effect/test/PartitionedSemaphore.test.ts).

The test uses a queued public `Scheduler` to expose the interruption window deterministically, following the existing semaphore test pattern. It does not mutate private semaphore state.

B acquires the refunded permit and its take completes successfully.

### What do you see instead?

B remains suspended. The otherwise identical different-key control completes.

### Additional information

Selection removes A from its waiter set and removes the now-empty partition. Interruption runs cleanup again; the captured old set is still empty, so cleanup removes the map entry by key, deleting B's replacement partition. Cleanup must only remove a partition when it actually removes that partition's last waiter, while still refunding A's acquired permit.

kitlangton:
## Type

- [x] Bug Fix

### Why

Interrupting a selected waiter before its acquisition finishes can run cleanup twice. If another waiter has registered under the same key, the second cleanup deletes that replacement partition and leaves its waiter suspended despite the permit refund.

### What Changes

Remove a partition only when cleanup actually deletes its last waiter. Keep permit refunds outside that guard.

**Before:** the same-key waiter remains suspended; the different-key control succeeds.

**After:** both waiters acquire their refunded permit successfully.

### Scope

One conditional change, a deterministic same-key regression and different-key control, and an `effect` patch changeset. The scheduler fixture follows the existing semaphore tests and does not fabricate private semaphore state.

### Verification

```bash
pnpm test --run packages/effect/test/PartitionedSemaphore.test.ts packages/effect/test/Semaphore.test.ts
pnpm lint
pnpm check
git diff --check origin/main
```

All 26 tests pass. The same-key regression fails on the original runtime while its control passes. Lint, typecheck, and whitespace checks pass.

## Related

Closes #7580.

## Audit

### DIFF
--- a/packages/effect/test/PartitionedSemaphore.test.ts
+++ b/packages/effect/test/PartitionedSemaphore.test.ts
@@ -1,5 +1,5 @@
 import { assert, describe, it } from "@effect/vitest"
-import { Effect, Fiber, Option, PartitionedSemaphore } from "effect"
+import { Effect, Exit, Fiber, Option, PartitionedSemaphore, Scheduler } from "effect"
 
 describe("PartitionedSemaphore", () => {
   it.effect("module-level combinators delegate to the instance api", () =>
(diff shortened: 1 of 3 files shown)


packages/effect/test/PartitionedSemaphore.test.ts
```diff
@@ -89,4 +89,47 @@ describe("PartitionedSemaphore", () => {
       assert.strictEqual(yield* PartitionedSemaphore.available(sem), 4)
       yield* PartitionedSemaphore.take(sem, "c", 4)
     }))
+
+  for (const key of ["other", "same"]) {
+    it.effect(`repro: interrupting a resumed waiter preserves the next ${key}-key waiter`, () =>
```
tim-smart:
```suggestion
    it.effect(`interrupting a resumed waiter preserves the next ${key}-key waiter`, () =>
```

PR #7587, opened 2026-09-02
fix(Cache): preserve replacements during conditional invalidation

### LINKED ISSUE #7581: Cache.invalidateWhen and ScopedCache.invalidateWhen can delete replacement entries

1. Start a lookup for a key and pause it on a `Deferred`.
2. Start `Cache.invalidateWhen(cache, key, value => value === 1)`; it waits for that original lookup.
3. Replace the key with `Cache.set(cache, key, 99)`.
4. Complete the original lookup with `1` and await conditional invalidation.

[Executable Cache regression](https://github.com/kitlangton/effect/blob/repro/cache-invalidate-when/packages/effect/test/Cache.test.ts).

The same race affects `ScopedCache`: invalidate the pending entry and acquire resource `2` before letting resource `1`'s lookup complete. [Executable ScopedCache regression](https://github.com/kitlangton/effect/blob/repro/cache-invalidate-when/packages/effect/test/ScopedCache.test.ts).

Conditional invalidation returns `false` because its original entry is no longer cached, preserving the replacement. In the scoped case, resource `2` remains alive until owner shutdown and is then released.

### What do you see instead?

Conditional invalidation returns `true` and removes replacement `99`, although `99` does not match the predicate. `ScopedCache` also loses the replacement's scope: the release log after owner shutdown contains only `[1]`, rather than `[1, 2]`.

### Additional information

Both implementations await an entry, evaluate the predicate against its value, then remove the current map entry by key without verifying entry identity. The scoped version closes the captured old scope instead of the replacement's scope. Tests cover original lookup success and failure; failed lookups already skip the predicate and preserve replacements.

kitlangton:
## Type

- [x] Bug Fix

### Why

`Cache.invalidateWhen` and `ScopedCache.invalidateWhen` can remove a replacement entry after awaiting an older lookup. The predicate matches the old value, not the replacement; ScopedCache additionally loses ownership of the replacement's resources.

### What Changes

Recheck entry identity after predicate evaluation. Return `false` if the original entry is no longer cached, leaving the replacement intact.

| Observation | Before | After |
| --- | --- | --- |
| Old value 1 matches; key was replaced with 99 | Returns true, deletes 99 | Returns false, keeps 99 |
| Scoped resource 1 replaced with 2 | Release log after shutdown omits 2 | Resource 2 stays live until shutdown, then releases |
| Old lookup fails | Predicate skipped | Unchanged |

### Scope

Identity guards in both conditional-invalidation implementations, success/failure regressions and resource-lifetime assertions, and an `effect` patch changeset. No new abstractions or public API changes.

### Verification

```bash
pnpm test --run packages/effect/test/Cache.test.ts packages/effect/test/ScopedCache.test.ts
pnpm lint
pnpm check
git diff --check origin/main
```

All 203 tests pass. Both successful-old-lookup regressions fail on the original runtime; failed-old-lookup controls continue to pass. Lint, typecheck, and whitespace checks pass.

## Related

Closes #7581.

## Audit

### DIFF
--- a/.changeset/cache-invalidate-when-replacement.md
+++ b/.changeset/cache-invalidate-when-replacement.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-Fix `Cache.invalidateWhen` and `ScopedCache.invalidateWhen` deleting a replacement entry while waiting for an earlier lookup. They now return `false` when the original entry is no longer cached, preserving the replacement value and ensuring scoped replacement resources are released when the cache's owning scope closes.
+Fix `Cache.invalidateWhen` and `ScopedCache.invalidateWhen` deleting a replacement entry while waiting for an earlier lookup.
--- a/.changeset/cache-invalidate-when-replacement.md
+++ b/.changeset/cache-invalidate-when-replacement.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `Cache.invalidateWhen` and `ScopedCache.invalidateWhen` deleting a replacement entry while waiting for an earlier lookup. They now return `false` when the original entry is no longer cached, preserving the replacement value and ensuring scoped replacement resources are released when the cache's owning scope closes.
--- a/packages/effect/src/Cache.ts
+++ b/packages/effect/src/Cache.ts
@@ -1041,6 +1041,10 @@ export const invalidateWhen: {
       return oentry.await().pipe(
         effect.map((value) => {
           if (f(value)) {
+            const current = MutableHashMap.get(self.map, key)
+            if (Option.isNone(current) || current.value !== oentry) {
+              return false
+            }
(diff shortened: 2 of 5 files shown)


.changeset/cache-invalidate-when-replacement.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `Cache.invalidateWhen` and `ScopedCache.invalidateWhen` deleting a replacement entry while waiting for an earlier lookup. They now return `false` when the original entry is no longer cached, preserving the replacement value and ensuring scoped replacement resources are released when the cache's owning scope closes.
```
tim-smart:
```suggestion
Fix `Cache.invalidateWhen` and `ScopedCache.invalidateWhen` deleting a replacement entry while waiting for an earlier lookup.
```

PR #7588, opened 2026-09-02
unstable/ai: route tool param validation through failureMode and drop ToolParameterValidationError.toolParams

### LINKED ISSUE #6335: unstable/ai: tool-call parameter decode failure bypasses failureMode: "return" and kills the whole streamText stream

### LINKED ISSUE #7416: unstable/ai: constructing ToolParameterValidationError throws when toolParams contain NaN/Infinity, turning a typed tool error into a defect

tim-smart:
## Summary

Fixes two related `unstable/ai` bugs around tool call parameter validation:

- Constructing `ToolParameterValidationError` threw when the invalid params contained non-JSON values such as `NaN` or `Infinity`, turning a typed tool error into a defect (#7416).
- A tool-call parameter decode failure bypassed `failureMode: "return"` and killed the whole `streamText` stream (#6335) — both in `LanguageModel` transport decode and in the provider packages' own parameter validation.

## Changes

- **`AiError.ToolParameterValidationError`**: removed the `toolParams` field (typed `Schema.Json`, it crashed construction on non-JSON values and could not encode safely). `message` is built from `toolName` and `description` only. All construction sites were updated.
- **`Toolkit.handle`**: parameter decode failures now route through the tool's `failureMode`. With `"return"`, the handle stream emits a `tool-result` with the normalized `AiError` as `result` and `isFailure: true`, running through the existing `encodeResult` path exactly like handler failures. The default `"error"` mode still fails with `ToolParameterValidationError`.
- **Provider packages (OpenAI, OpenAI-compat, Anthropic)**: `transformToolCallParams` no longer fails the response on schema drift. It decodes wire parameters with the provider codec (handling provider-specific rewrites such as strict-mode nullable optionals) and re-encodes them into the tool's standard encoded form; when that validation fails, the raw parameters pass through unchanged so `Toolkit` — the single authority for parameter validation — routes the failure through `failureMode`. This also fixes a latent bug where transforming parameter schemas (e.g. `Schema.FiniteFromString`) were decoded twice on the provider path. OpenRouter already passed parsed params through untouched.
- **`LanguageModel.streamText` / `generateText`**: with tool call resolution enabled, response parts are decoded with opaque tool parameters, so schema-drifted params can no longer produce an `InvalidOutputError` that kills the stream/effect before `Toolkit` runs. `generateText`'s previous prevalidation/final double decode collapsed into one decode.
- **Provider-executed tool calls**: since they never reach `Toolkit`, their parameters are validated explicitly as part of response validation and fail with `InvalidOutputError` on mismatch, preserving `main`'s behavior.
- **Emitted `params` typing**: with resolution enabled, tool-call parts carry raw (encoded) params at runtime and are now typed `params: unknown` via the new `Response.ToolParametersMode` (`false` = decoded, `true` = encoded, `"opaque"` = unknown), threaded through `ExtractEncodedToolParameters`, `GenerateTextResponse`, and `Response.Part`/`StreamPart`. This replaces the previous internal cast that typed raw params as decoded.

## Testing

- Toolkit/LanguageModel: `failureMode: "return"` with invalid and `NaN` params yields an `isFailure` tool result for `generateText` and `streamText` (stream stays alive, handler never runs); `failureMode: "error"` still fails without a defect; transforming parameter schemas with resolution enabled preserve encoded params on emitted parts while handlers receive decoded values; provider-executed tool calls fail with `InvalidOutputError` on invalid params and pass through on valid ones.
- Providers: mocked OpenAI response/stream tests prove schema-drifted params route through `failureMode: "return"` instead of failing, and that transformed params reach handlers decoded; an equivalent Anthropic streaming test covers the same path.
- All `unstable/ai` and `packages/ai/*` suites pass (1269 tests), plus `pnpm check`, `pnpm lint`, `pnpm jsdocs --check`, and the AiError doctests.

Closes EFF-1008
Closes #7416
Closes #6335

### DIFF
--- a/.changeset/eff-1008-tool-param-failure-mode.md
+++ b/.changeset/eff-1008-tool-param-failure-mode.md
@@ -7,9 +7,3 @@
 ---
 
 Route tool call parameter validation failures through the tool's `failureMode` and drop `ToolParameterValidationError.toolParams`.
-
-With `failureMode: "return"`, invalid tool call parameters now produce a failed tool result (`isFailure: true`) instead of failing the effect or killing the `streamText` stream, matching how tool handler failures are treated. The default `failureMode: "error"` still fails with `ToolParameterValidationError`. This applies to the provider packages as well: the OpenAI, OpenAI-compat, and Anthropic packages no longer fail the response when tool call parameters do not match the tool's schema. Instead they convert valid parameters into the tool's standard encoded form and pass invalid parameters through unchanged, leaving `Toolkit` as the single authority for parameter validation.
-
-`ToolParameterValidationError` no longer carries a `toolParams` field, which previously made error construction throw for non-JSON parameter values such as `NaN` or `Infinity`.
-
-**Breaking runtime/type change**: when tool call resolution is enabled, tool call parts emitted by `generateText` / `streamText` now carry the raw (encoded) parameters instead of decoded ones, and their `params` field is typed as `unknown` (see the new `Response.ToolParametersMode`). Code that read decoded `params` off returned tool calls must decode them itself or read handler inputs instead. Provider-executed tool calls are still validated against the tool's parameter schema and fail with `InvalidOutputError` on mismatch, as before.
--- a/.changeset/eff-1008-tool-param-failure-mode.md
+++ b/.changeset/eff-1008-tool-param-failure-mode.md
@@ -1,0 +1,13 @@
+---
+"effect": patch
+"@effect/ai-anthropic": patch
+"@effect/ai-openai": patch
+"@effect/ai-openai-compat": patch
+"@effect/ai-openrouter": patch
+---
+
+Route tool call parameter validation failures through the tool's `failureMode` and drop `ToolParameterValidationError.toolParams`.
+
+With `failureMode: "return"`, invalid tool call parameters now produce a failed tool result (`isFailure: true`) instead of failing the effect or killing the `streamText` stream, matching how tool handler failures are treated. The default `failureMode: "error"` still fails with `ToolParameterValidationError`.
+
+`ToolParameterValidationError` no longer carries a `toolParams` field, which previously made error construction throw for non-JSON parameter values such as `NaN` or `Infinity`.
--- a/.changeset/eff-1008-tool-param-failure-mode.md
+++ b/.changeset/eff-1008-tool-param-failure-mode.md
@@ -8,6 +8,8 @@
 
 Route tool call parameter validation failures through the tool's `failureMode` and drop `ToolParameterValidationError.toolParams`.
 
-With `failureMode: "return"`, invalid tool call parameters now produce a failed tool result (`isFailure: true`) instead of failing the effect or killing the `streamText` stream, matching how tool handler failures are treated. The default `failureMode: "error"` still fails with `ToolParameterValidationError`.
+With `failureMode: "return"`, invalid tool call parameters now produce a failed tool result (`isFailure: true`) instead of failing the effect or killing the `streamText` stream, matching how tool handler failures are treated. The default `failureMode: "error"` still fails with `ToolParameterValidationError`. This applies to the provider packages as well: the OpenAI, OpenAI-compat, and Anthropic packages no longer fail the response when tool call parameters do not match the tool's schema. Instead they convert valid parameters into the tool's standard encoded form and pass invalid parameters through unchanged, leaving `Toolkit` as the single authority for parameter validation.
 
 `ToolParameterValidationError` no longer carries a `toolParams` field, which previously made error construction throw for non-JSON parameter values such as `NaN` or `Infinity`.
+
+**Breaking runtime/type change**: when tool call resolution is enabled, tool call parts emitted by `generateText` / `streamText` now carry the raw (encoded) parameters instead of decoded ones, and their `params` field is typed as `unknown` (see the new `Response.ToolParametersMode`). Code that read decoded `params` off returned tool calls must decode them itself or read handler inputs instead. Provider-executed tool calls are still validated against the tool's parameter schema and fail with `InvalidOutputError` on mismatch, as before.
--- a/packages/effect/src/unstable/ai/Response.ts
+++ b/packages/effect/src/unstable/ai/Response.ts
@@ -226,11 +226,11 @@ export const AllParts = <T extends Toolkit.Any | Toolkit.WithHandler<any>>(
  */
 export type Part<
   Tools extends Record<string, Tool.Any>,
-  EncodedToolParameters extends ToolParametersMode = false
+  ParametersMode extends ToolParametersMode = "decoded"
 > =
   | TextPart
   | ReasoningPart
-  | ToolCallParts<Tools, EncodedToolParameters>
+  | ToolCallParts<Tools, ParametersMode>
   | ToolResultParts<Tools>
   | ToolApprovalRequestPart
   | FilePart
@@ -307,7 +307,7 @@ export const Part = <T extends Toolkit.Any | Toolkit.WithHandler<any>>(
  */
 export type StreamPart<
   Tools extends Record<string, Tool.Any>,
-  EncodedToolParameters extends ToolParametersMode = false
+  ParametersMode extends ToolParametersMode = "decoded"
 > =
   | TextStartPart
   | TextDeltaPart
@@ -318,7 +318,7 @@ export type StreamPart<
   | ToolParamsStartPart
   | ToolParamsDeltaPart
   | ToolParamsEndPart
-  | ToolCallParts<Tools, EncodedToolParameters>
+  | ToolCallParts<Tools, ParametersMode>
   | ToolResultParts<Tools>
   | ToolApprovalRequestPart
   | FilePart
@@ -410,26 +410,25 @@ export const StreamPart = <T extends Toolkit.Any | Toolkit.WithHandler<any>>(
  */
 export type ToolCallParts<
   Tools extends Record<string, Tool.Any>,
-  EncodedParameters extends ToolParametersMode = false
-> = ToolCallPartForName<Tools, EncodedParameters, keyof Tools>
+  ParametersMode extends ToolParametersMode = "decoded"
+> = ToolCallPartForName<Tools, ParametersMode, keyof Tools>
 
 /**
- * Controls tool parameter types: decoded (`false`), encoded (`true`), or
- * `unknown` (`"opaque"`).
+ * Controls whether tool parameters are decoded, encoded, or opaque.
  *
  * @category utility types
  * @since 4.0.0
  */
-export type ToolParametersMode = boolean | "opaque"
+export type ToolParametersMode = "decoded" | "encoded" | "opaque"
 
 type ToolCallPartForName<
   Tools extends Record<string, Tool.Any>,
-  EncodedParameters extends ToolParametersMode,
+  ParametersMode extends ToolParametersMode,
   Name extends keyof Tools
 > = Name extends string ? ToolCallPart<
     Name,
-    EncodedParameters extends true ? Tool.ParametersEncoded<Tools[Name]>
-      : EncodedParameters extends "opaque" ? unknown
+    ParametersMode extends "encoded" ? Tool.ParametersEncoded<Tools[Name]>
+      : ParametersMode extends "opaque" ? unknown
       : Tool.Parameters<Tools[Name]>
   >
   : never
--- a/packages/effect/src/unstable/ai/Response.ts
+++ b/packages/effect/src/unstable/ai/Response.ts
@@ -414,15 +414,8 @@ export type ToolCallParts<
 > = ToolCallPartForName<Tools, EncodedParameters, keyof Tools>
 
 /**
- * Utility type that determines how tool call parameters are typed on response
- * parts.
- *
- * **Details**
- *
- * - `false`: parameters are fully decoded via the tool's parameter schema
- * - `true`: parameters are typed in their encoded form
- * - `"opaque"`: parameters are typed as `unknown` because transport decode
- *   leaves them untouched and validation happens in `Toolkit`
+ * Controls tool parameter types: decoded (`false`), encoded (`true`), or
+ * `unknown` (`"opaque"`).
  *
  * @category utility types
  * @since 4.0.0
--- a/packages/effect/src/unstable/ai/Response.ts
+++ b/packages/effect/src/unstable/ai/Response.ts
@@ -226,7 +226,7 @@ export const AllParts = <T extends Toolkit.Any | Toolkit.WithHandler<any>>(
  */
 export type Part<
   Tools extends Record<string, Tool.Any>,
-  EncodedToolParameters extends boolean = false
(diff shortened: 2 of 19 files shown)


.changeset/eff-1008-tool-param-failure-mode.md
```diff
@@ -0,0 +1,15 @@
...
+
+`ToolParameterValidationError` no longer carries a `toolParams` field, which previously made error construction throw for non-JSON parameter values such as `NaN` or `Infinity`.
+
+**Breaking runtime/type change**: when tool call resolution is enabled, tool call parts emitted by `generateText` / `streamText` now carry the raw (encoded) parameters instead of decoded ones, and their `params` field is typed as `unknown` (see the new `Response.ToolParametersMode`). Code that read decoded `params` off returned tool calls must decode them itself or read handler inputs instead. Provider-executed tool calls are still validated against the tool's parameter schema and fail with `InvalidOutputError` on mismatch, as before.
```
tim-smart:
```suggestion
Route tool call parameter validation failures through the tool's `failureMode` and drop `ToolParameterValidationError.toolParams`.
```

packages/effect/src/unstable/ai/Response.ts
```diff
@@ -410,16 +410,27 @@ export const StreamPart = <T extends Toolkit.Any | Toolkit.WithHandler<any>>(
...
 > = ToolCallPartForName<Tools, EncodedParameters, keyof Tools>
 
+/**
+ * Controls tool parameter types: decoded (`false`), encoded (`true`), or
+ * `unknown` (`"opaque"`).
+ *
+ * @category utility types
+ * @since 4.0.0
+ */
+export type ToolParametersMode = boolean | "opaque"
```
tim-smart: I think make this an explicit string union, as the boolean is ambiguous.

tim-smart: Addressed in `2f702793b6`: `ToolParametersMode` is now `"decoded" | "encoded" | "opaque"`, with the new modes propagated through the implementation and tests.

PR #7604, opened 2026-09-02
fix(Pool): preserve reservations during borrower release

### LINKED ISSUE #7599: Pool.reserve loses exclusivity when another borrower returns

Effect `4.0.0-rc.112`; reproduced on `main` at `1a86166f1`, Node.js `26.5.0`.

Borrow a shared item twice, reserve it, then return the other borrow:

```ts
import { Effect, Exit, Fiber, Pool, Scope } from "effect"

const result = await Effect.runPromise(Effect.scoped(Effect.gen(function*() {
  const pool = yield* Pool.make({ acquire: Effect.succeed("resource"), size: 1, concurrency: 2 })
  const owner = yield* Scope.make()
  const borrower = yield* Scope.make()
  const reservation = yield* Scope.make()
  yield* Pool.get(pool).pipe(Scope.provide(owner))
  yield* Pool.get(pool).pipe(Scope.provide(borrower))
  yield* Pool.reserve(pool, "resource").pipe(Scope.provide(reservation))
  yield* Scope.close(borrower, Exit.void)

  const next = yield* Pool.use(pool, Effect.succeed).pipe(Effect.forkChild({ startImmediately: true }))
  const beforeRelease = next.pollUnsafe()
  yield* Scope.close(reservation, Exit.void)
  yield* Fiber.join(next)
  yield* Scope.close(owner, Exit.void)
  return beforeRelease?._tag ?? "Waiting"
})))
console.log(result)
```

`"Waiting"`: the next checkout cannot use the item until its reservation closes.

### What do you see instead?

`"Success"`: returning the other borrow puts the reserved item back into circulation.

### Additional information

Usage-TTL reclaim and closing one of two overlapping reservations can also reopen the item prematurely. PostgreSQL multiplexing uses this API to pin connections.

[Regression tests and lifecycle controls](https://github.com/kitlangton/effect/blob/fix/pool-reservation-lifecycle/packages/effect/test/Pool.test.ts#L311-L466).

kitlangton:
## Type

- [x] Bug Fix

### Why
Returning another shared borrow can put a reserved item back into circulation before its reservation closes. Usage-TTL reclaim and overlapping reservations have the same problem. PostgreSQL multiplexing uses this API to pin connections.

### What Changes
Track active reservations separately from resource-owning leases and check them on both availability insertion paths.

| Event while reserved | Before | After |
| --- | --- | --- |
| Peer returns, TTL reclaims, or one overlapping reservation closes | Item becomes borrowable | Item stays reserved |
| Last reservation closes | Waiter count assumes one borrow remains | Wake all available slots |

### Scope
Preserves exported types and existing resource lifetime: reservations prevent reuse, not invalidation or shutdown cleanup. Includes controls for those boundaries and an `effect` patch changeset.

### Verification
```bash
pnpm test --run packages/effect/test/Pool.test.ts
EFFECT_INTEGRATION_TESTS=1 pnpm test --run packages/sql/pg/test/PgPool.integration.test.ts
pnpm lint
pnpm check
git diff --check
```
37 Pool tests and 15 PostgreSQL integration tests pass; lint and typecheck pass. Against the original Pool implementation, four regressions fail and 33 tests pass.

## Related

Closes #7599.

## Audit

### DIFF
--- a/.changeset/pool-preserve-reservations.md
+++ b/.changeset/pool-preserve-reservations.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Keep `Pool.reserve` items out of shared circulation when other borrowers return or overlapping reservations close. Restore available slots only after the last reservation closes.
--- a/.changeset/pool-preserve-reservations.md
+++ b/.changeset/pool-preserve-reservations.md
@@ -3,3 +3,5 @@
 ---
 
 Keep `Pool.reserve` items out of shared circulation when other borrowers return or overlapping reservations close. Restore available slots only after the last reservation closes.
+
+Overlapping reservations on the same item now count `usage` once instead of once per reservation, and the usage TTL strategy no longer reclaims reserved items, acquiring a fresh item for waiters instead. `Pool.reserve` is now a full no-op when per-item concurrency is `1`; previously it removed the item from availability while leaving its capacity unchanged.
--- a/.changeset/pool-preserve-reservations.md
+++ b/.changeset/pool-preserve-reservations.md
@@ -3,5 +3,3 @@
 ---
 
 Keep `Pool.reserve` items out of shared circulation when other borrowers return or overlapping reservations close. Restore available slots only after the last reservation closes.
-
-Overlapping reservations on the same item now count `usage` once instead of once per reservation, and the usage TTL strategy no longer reclaims reserved items, acquiring a fresh item for waiters instead. `Pool.reserve` is now a full no-op when per-item concurrency is `1`; previously it removed the item from availability while leaving its capacity unchanged.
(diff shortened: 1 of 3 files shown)


.changeset/pool-preserve-reservations.md
```diff
@@ -0,0 +1,7 @@
+---
+"effect": patch
+---
+
+Keep `Pool.reserve` items out of shared circulation when other borrowers return or overlapping reservations close. Restore available slots only after the last reservation closes.
+
+Overlapping reservations on the same item now count `usage` once instead of once per reservation, and the usage TTL strategy no longer reclaims reserved items, acquiring a fresh item for waiters instead. `Pool.reserve` is now a full no-op when per-item concurrency is `1`; previously it removed the item from availability while leaving its capacity unchanged.
```
tim-smart:
```suggestion
Keep `Pool.reserve` items out of shared circulation when other borrowers return or overlapping reservations close. Restore available slots only after the last reservation closes.
```

PR #7643, opened 2026-09-02
Reduce allocations in Effect primitives and fibers

tim-smart:
## Summary

- size Effect, Exit, Option, and Result primitives with constructor functions instead of `Object.create`
- replace retained `arguments` objects in Async, Iterator, and OnExit with named fields
- add dedicated Map and As primitives to remove closure and intermediate primitive allocations
- share context-derived fiber state per `cacheRoot`
- replace the supervised-child observer closure with a parent back-reference

The five commits are independent to make review and selective landing easier. This is an early-review PR, with the runtime and context changes called out for extra scrutiny.

## Measurements

On Node 26.7 without pointer compression:

- Effect and data primitives retain 30-65% less memory (`Effect.map`: 208 B to 72 B; exits and `Option.some` / `Result.succeed`: 56 B to 32 B)
- live supervised fibers retain 48% less memory (731 B to 379 B)
- fork/join is 15-25% faster in interleaved A/B runs
- the sequential flatMap run loop is unchanged within measurement noise

## Follow-ups worth considering

- make `_observers` lazy now that supervised children no longer allocate one; most fibers never gain an observer
- apply dedicated primitives selectively to hot combinators that still create closures through `flatMap`, especially `tap`, `mapError`, and `andThen`
- remove the now-unused `single: false` branch from `makePrimitive`
- keep the memory benchmark as a repeatable runtime benchmark so future representation changes can be checked against both retained bytes and CPU

The context cache and parent cleanup are the highest-risk parts of this PR. In particular, review should verify cache invalidation for every `fiberCached` context key and cleanup ordering for interrupted supervised children.

## Verification

- full Effect suite: 9,104 tests passed on the originating branch
- targeted Effect, Fiber, Context, Option, and Result tests: 368 passed locally
- `pnpm check`
- `pnpm lint`

Closes EFF-1036

### DIFF
--- a/packages/effect/src/internal/effect.ts
+++ b/packages/effect/src/internal/effect.ts
@@ -520,6 +520,7 @@ export class FiberImpl<A = any, E = any> implements Fiber.Fiber<A, E> {
     this._yielded = undefined
     this._running = false
     this._deferredInterrupt = false
+    this._parent = undefined
     this.runtimeMetrics?.recordFiberStart(this.context)
   }
 
@@ -536,6 +537,7 @@ export class FiberImpl<A = any, E = any> implements Fiber.Fiber<A, E> {
   _yielded: Exit.Exit<any, any> | (() => void) | undefined
   _running: boolean
   _deferredInterrupt: boolean
+  _parent: FiberImpl<any, any> | undefined
 
   // set in setContext
   context!: Context.Context<never>
@@ -638,6 +640,10 @@ export class FiberImpl<A = any, E = any> implements Fiber.Fiber<A, E> {
 
     this._exit = exit
     this.runtimeMetrics?.recordFiberEnd(this.context, this._exit)
+    if (this._parent) {
+      this._parent._children?.delete(this)
+      this._parent = undefined
+    }
     for (let i = 0; i < this._observers.length; i++) {
       this._observers[i](exit)
     }
@@ -1453,11 +1459,24 @@ export const as: {
   <A, E, R, B>(
     self: Effect.Effect<A, E, R>,
     value: B
-  ): Effect.Effect<B, E, R> => {
-    const b = succeed(value)
-    return flatMap(self, (_) => b)
-  }
+  ): Effect.Effect<B, E, R> => new (AsImpl as any)(self, succeed(value))
 )
+
+const AsProto = makePrimitiveProto({
+  op: "As",
+  [evaluate](this: any, fiber: FiberImpl): Primitive {
+    fiber._stack.push(this)
+    return this[args]
+  },
+  [contA](this: any, _value, _fiber) {
+    return this.exit
+  }
+})
+function AsImpl(this: any, self: any, exit: any) {
+  this[args] = self
+  this.exit = exit
+}
+AsImpl.prototype = AsProto
 
 /** @internal */
 export const asSome = <A, E, R>(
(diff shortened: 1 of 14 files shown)


packages/effect/src/internal/effect.ts
```diff
@@ -536,18 +537,39 @@ export class FiberImpl<A = any, E = any> implements Fiber.Fiber<A, E> {
...
-  currentSpan: Tracer.AnySpan | undefined
-  currentLogLevel!: LogLevel.LogLevel
-  minimumLogLevel!: LogLevel.LogLevel
-  currentStackFrame: StackFrame | undefined
-  runtimeMetrics: Metric.FiberRuntimeMetricsService | undefined
-  maxOpsBeforeYield!: number
-  currentPreventYield!: boolean
+  _ctxCache!: FiberContextCache
+
+  get currentScheduler(): Scheduler.Scheduler {
```
tim-smart: This can be a breaking change by removing these getters and having a fiber.cache property on the public interface.

tim-smart: Done in 858095791f: removed the getters and exposed the shared cache as `fiber.cache` (`Fiber.Cache` interface: `scheduler`, `tracerContext`, `span`, `logLevel`, `minimumLogLevel`, `stackFrame`, `runtimeMetrics`, `maxOpsBeforeYield`, `preventYield`). Updated all in-repo readers (Scheduler, ManagedRuntime, request batching, Otlp logger/tracer, opentelemetry), the Logger migration annotation, and added a changeset noting the breaking interface change. Same commit also dedupes the identical continuation `evaluate` bodies to claw back some of the bundle-size increase.

tim-smart:
## Memory results (branch @ 7f66db7 vs main @ de7774f)

Node v26.7 official build (no pointer compression). Retained bytes per object: `heapUsed` after 3 forced GCs, 0.5-1M objects held live in a preallocated array. Benchmark scripts are attached to EFF-1036.

| allocation | main | this PR | change |
|---|---|---|---|
| `Effect.succeed` / exit | 56 B | 32 B | -43% |
| `Effect.flatMap` | 112 B | 72 B | -36% |
| `Effect.map` | 208 B | 80 B | -62% |
| `Effect.tap` / `Effect.andThen` | 152 B | 48 B | -68% |
| `Effect.sync` | 152 B | 128 B | -16% |
| `Effect.gen` (unexecuted) | 312 B | 280 B | -10% |
| `Effect.fail` | 208 B | 184 B | -12% |
| `Option.some` / `Result.succeed` | 56 B | 32 B | -43% |
| `Deferred.makeUnsafe` | 56 B | 40 B | -29% |
| live supervised fiber (`forkChild` + `never`) | 731 B | 347 B | -53% |

For scale: 100k suspended supervised fibers retain ~35 MB instead of ~73 MB.

## GC impact

Mixed request workload (`gen` + `forkChild` + `Deferred` + `tap`/`map`/`andThen` + `scoped`), 360k requests per run, no forced GC, GC activity recorded via `perf_hooks`, main and PR interleaved for 3 rounds:

| metric | main | this PR |
|---|---|---|
| minor GCs (scavenges) | 108-109 | 84-86 (-21%) |
| major GCs | 5 | 5 |
| incremental steps | 5 | 5 |
| total GC pause | 49-51 ms | 45-48 ms |
| wall time | 1316-1332 ms | 1206-1216 ms (-8%) |

The scavenge count is the direct signal: with the same new-space size, ~21% fewer minor collections means ~21% fewer transient bytes allocated per request. Old-generation behavior is unchanged (same major/incremental counts); the retained-size wins above show up as a smaller old-gen live set instead. GC pause stays ~3.8% of wall time on both sides, but the wall itself is 8% shorter.

PR #7647, opened 2026-09-02
fix(schema): render finite numbers as Schema.Finite and reference generated aliases directly

### LINKED ISSUE #7646: @effect/openapi-generator: generated output triggers the language service `schemaNumber` and `unnecessaryTypeofType` diagnostics
The `httpclient` output of `@effect/openapi-generator` 4.0.0-rc.112 is flagged by `@effect/language-service` (via `@effect/tsgo` 0.38.0) with two suggestion diagnostics that the generator could avoid by construction.

**Reproduction** (spec in the collapsed block below, `format: 'httpclient'`):

```ts
const generator = await Effect.runPromise(OpenApiGenerator.make);
const out = await Effect.runPromise(
  generator.generate(spec, { name: 'ReproClient', format: 'httpclient' }),
);
```

Generated (abridged):

```ts
export const Addon = Schema.Struct({
  "id": Schema.Number.check(Schema.isInt().annotate({ "expected": "an integer" })),
  "downloads": Schema.Number.check(Schema.isFinite().annotate({ "expected": "a finite number" })),
})
// ...
readonly "getAddon": <Config extends OperationConfig>(id: string, options: ...) =>
  Effect.Effect<WithOptionalResponse<typeof GetAddon200.Type, Config>, HttpClientError.HttpClientError | SchemaError>
```

`tsc` with the language-service plugin reports:

```
repro.ts(11,51): suggestion TS377098: This Schema number API accepts `NaN`, `Infinity`, and `-Infinity`. Use `Schema.Finite` for finite domain numbers. ... effect(schemaNumber)
repro.ts(11,140): suggestion TS377098: ... effect(schemaNumber)
repro.ts(98,170): suggestion TS377090: This `typeof ... .Type` query can be replaced with `GetAddon200`. effect(unnecessaryTypeofType)
```

**Expected**

1. Numbers from JSON are always finite, so `Schema.Finite` (and `Schema.Int` for `type: integer`) would satisfy `schemaNumber`. The generator already appends `Schema.isFinite()` for `number`, so the intent is there; only the base schema differs.
2. Operation signatures reference `typeof X.Type` even though the generator emits `export type X = ...` for the same name in the same file. Emitting `X` would satisfy `unnecessaryTypeofType`.

Since both tools live in this monorepo, it would be good for the generator's default output to pass the language service's defaults. We currently post-process the output (replace `Schema.Number` with `Schema.Finite`, and `typeof X.Type` with `X` when the alias exists), which works but should not be necessary.

**Side observation:** for a spec with no error responses, the generated client declares `decodeError` without using it, which fails under `noUnusedLocals` (`TS6133`).

<details><summary>Minimal spec</summary>

```json
{
  "openapi": "3.0.0",
  "info": { "title": "Repro", "version": "1.0.0" },
  "paths": {
    "/addons/{id}": {
      "get": {
        "operationId": "getAddon",
        "parameters": [{ "name": "id", "in": "path", "required": true, "schema": { "type": "integer" } }],
        "responses": {
          "200": { "description": "OK", "content": { "application/json": { "schema": { "$ref": "#/components/schemas/Addon" } } } }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Addon": {
        "type": "object",
        "required": ["id", "downloads"],
        "properties": { "id": { "type": "integer" }, "downloads": { "type": "number" } }
      }
    }
  }
}
```

</details>

Versions: `@effect/openapi-generator` 4.0.0-rc.112, `effect` 4.0.0-rc.112, `@effect/tsgo` 0.38.0.

Stanzilla:
Closes #7646.

`@effect/openapi-generator` output triggered two `@effect/language-service` diagnostics that the generator can avoid by construction.

## `effect`: `SchemaRepresentation.toCodeDocument`

The JSON Schema importer already attaches `effect/schema/isFinite` (for `type: number`) and `effect/schema/isInt` (for `type: integer`) checks to number nodes. The renderer now emits such nodes from `Schema.Finite` instead of `Schema.Number`, and drops an `isFinite` check when it carries nothing beyond the defaults of `Schema.isFinite()` (a custom-annotated, aborted, or branded check is kept on top of the `Schema.Finite` base). Plain `Schema.Number` without those checks renders unchanged.

Before / after for `{ "type": "number" }` and `{ "type": "integer" }`:

```ts
Schema.Number.check(Schema.isFinite().annotate({ "expected": "a finite number" }))  // before
Schema.Finite                                                                       // after

Schema.Number.check(Schema.isInt().annotate({ "expected": "an integer" }))          // before
Schema.Finite.check(Schema.isInt().annotate({ "expected": "an integer" }))          // after
```

## `@effect/openapi-generator`: client signatures

Operation signatures referenced `typeof X.Type` for success, error, and SSE schemas even though the generator emits `export type X = ...` for the same name in the same file. They now reference `X` directly (`typeof X.Encoded` and `typeof X.DecodingServices` are unchanged, as no alias exists for those).

```ts
// before
Effect.Effect<WithOptionalResponse<typeof GetAddon200.Type, Config>, ...>
// after
Effect.Effect<WithOptionalResponse<GetAddon200, Config>, ...>
```

## Verification

- `pnpm vitest run packages/effect/test/schema/representation packages/tools/openapi-generator/test`: 25 files, 807 tests passed (new cases cover unannotated, default-annotated, custom-annotated `isFinite`, `isInt`, and plain `Number`).
- `pnpm lint`, and `pnpm check` in both packages: clean.
- The minimal spec from the issue now renders `"downloads": Schema.Finite`, `"id": Schema.Finite.check(Schema.isInt()...)` and `WithOptionalResponse<GetAddon200, Config>`, with no `schemaNumber` or `unnecessaryTypeofType` diagnostics.

Two changesets are included (`effect` patch, `@effect/openapi-generator` patch).

### DIFF
--- a/.changeset/direct-lanterns.md
+++ b/.changeset/direct-lanterns.md
@@ -1,0 +1,5 @@
+---
+"@effect/openapi-generator": patch
+---
+
+Reference generated type aliases directly instead of `typeof X.Type` in client signatures.
--- a/.changeset/gentle-orbits.md
+++ b/.changeset/gentle-orbits.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Render JSON-derived finite numbers as `Schema.Finite` in `SchemaRepresentation.toCodeDocument`.
--- a/packages/effect/src/internal/schema/toCodeDocument.ts
+++ b/packages/effect/src/internal/schema/toCodeDocument.ts
@@ -1,6 +1,7 @@
 import * as Arr from "../../Array.ts"
 import { format, formatPropertyKey } from "../../Formatter.ts"
 import type * as Schema from "../../Schema.ts"
+import * as SchemaAST from "../../SchemaAST.ts"
 import type * as SchemaRepresentation from "../../SchemaRepresentation.ts"
 import { errorWithPath } from "../errors.ts"
 import * as InternalRecord from "../record.ts"
@@ -10,6 +11,14 @@ type Path = ReadonlyArray<string | number>
 type CheckRepresentationAnnotation = SchemaRepresentation.CheckRepresentationAnnotation<
   SchemaRepresentation.Representation
 >
+
+function hasFiniteNumberCheck(checks: ReadonlyArray<SchemaRepresentation.Check>): boolean {
+  return checks.some((check) =>
+    check._tag === "Filter"
+      ? check.representation?.id === "effect/schema/isFinite" || check.representation?.id === "effect/schema/isInt"
+      : hasFiniteNumberCheck(check.checks)
+  )
+}
 
 /** @internal */
 export function makeCode(runtime: string, Type: string): SchemaRepresentation.Code {
@@ -330,6 +339,10 @@ export function toCodeDocument(
     return rendered === undefined ? "" : `.${method}(${rendered})`
   }
 
+  function defaultFiniteAnnotations(): string {
+    return runtimeAnnotate(SchemaAST.isFinite().annotations)
+  }
+
   function compileCheck(
     check: SchemaRepresentation.Check,
     path: Path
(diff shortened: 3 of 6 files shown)


gcanti (commented):
The `schemaNumber` false positive has been fixed upstream. For the rest, one option is to add comments only to generated files to disable LSP/oxlint, something like `// @effect-diagnostics *:off` and the oxlint equivalent.

cc @mattiamanzati

PR #7652, opened 2026-09-02
Honor bufferSize in NodeStream.fromReadable and fromDuplex

### LINKED ISSUE #7645: NodeStream.fromReadable, fromDuplex, and pipeThroughDuplex expose a bufferSize option that is silently ignored

_No response_

bufferSize is a dead option. It is exposed in fromReadable, fromDuplex, and pipeThroughDuplex, but it is never forwarded or read anywhere. fromReadableChannel doesn’t even accept it, and readableToPullUnsafe has no bufferSize parameter. So a caller can write bufferSize: 100, but it does absolutely nothing.

[see](https://github.com/Effect-TS/effect/blob/main/packages/platform/node-shared/src/NodeStream.ts)

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

xia-chao:
## Summary

`NodeStream.fromReadable`, `fromDuplex`, and `pipeThroughDuplex` accepted `bufferSize` but never forwarded it. `fromReadableChannel` did not take the option, and `readableToPullUnsafe` drained every available chunk in one pull.

A positive finite `bufferSize` now caps how many already-buffered readable chunks are taken per pull. Omitting it keeps the previous unbounded drain.

Fixes #7645

## Test plan

- [x] `pnpm test --run packages/platform/node-shared/test/NodeStream.test.ts`
- [x] `fromReadable({ bufferSize: 1 })` on a readable with `a`, `b`, `c` already buffered emits chunks `[["a"], ["b"], ["c"]]`

### DIFF
--- a/.changeset/nodestream-buffersize.md
+++ b/.changeset/nodestream-buffersize.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform-node-shared": patch
+---
+
+Honor `bufferSize` in `NodeStream.fromReadable`, `fromDuplex`, and `pipeThroughDuplex` so it caps how many readable chunks are taken per pull.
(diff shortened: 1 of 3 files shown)


tim-smart:
Thanks, addressed in another PR

PR #7655, opened 2026-09-02
fix(Stream): preserve large source chunks when rechunking

### LINKED ISSUE #7654: Stream.rechunk throws when buffering a large source chunk

`effect` v4 source. Discovery reproduced at `0af0985d5c132b33e14a2df59e0567d7ba2a1710`; publication base: `243c72f001dd87385839c057af48f346abcf480b`.

```ts
import { Effect, Stream } from "effect"

const values = Array.from({ length: 200000 }, (_, i) => i)
const result = await Effect.runPromise(
  Stream.fromArray(values).pipe(
    Stream.rechunk(200001),
    Stream.chunks,
    Stream.runCollect
  )
)
console.log(result.map((chunk) => chunk.length))
```

[Executable regressions and controls](https://github.com/kitlangton/effect/blob/audit/i2-stream-rechunk/packages/effect/test/Stream.test.ts): `rechunk preserves a large source chunk with target $size`.

`[200000]`: one final partial chunk containing every input value in order.

### What do you see instead?

The effect fails with a `RangeError: Maximum call stack size exceeded` defect while buffering the source chunk. Equal/smaller target sizes and the split-source control pass.

### Additional information

The buffering branch uses `chunk.push(...arr)`, passing the entire source chunk as function arguments.

335 tests pass on `243c72f00`; lint, typecheck and whitespace checks pass. Red suites ran on `0af0985d5`. The affected implementation is byte-identical between those bases.

## Audit

kitlangton:
## Type

- [x] Bug Fix

## Description

`Stream.rechunk` used an array spread when buffering a source chunk. Large chunks can exceed the JavaScript engine's call-argument limit, causing a `RangeError` instead of preserving the stream elements.

The buffering branch now appends values individually. Regression coverage includes larger, equal, and smaller target sizes, plus input split across source chunks. The split case buffers one value before appending a 200,000-element chunk, so it exercises the same known failing size as the primary regression.

This PR also includes a patch changeset for `effect`.

## Verification

- `nix develop -c pnpm test --run --project effect packages/effect/test/Stream.test.ts -t rechunk` (6 passed)
- `nix develop -c pnpm check`
- `git diff --check`

Closes #7654
Closes EFF-1044

### DIFF
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -6905,8 +6905,8 @@ export const rechunk: {
             if (chunk.length === 0 && arr.length === target) {
               return Effect.succeed(arr)
             } else if (chunk.length + arr.length < target) {
-              for (const value of arr) {
-                chunk.push(value)
+              for (let i = 0; i < arr.length; i++) {
+                chunk.push(arr[i])
               }
               return loop()
             }
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -6905,7 +6905,9 @@ export const rechunk: {
             if (chunk.length === 0 && arr.length === target) {
               return Effect.succeed(arr)
             } else if (chunk.length + arr.length < target) {
-              chunk.push(...arr)
+              for (const value of arr) {
+                chunk.push(value)
+              }
               return loop()
             }
             current = arr
--- a/.changeset/stream-rechunk-large-source.md
+++ b/.changeset/stream-rechunk-large-source.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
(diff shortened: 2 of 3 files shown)


packages/effect/src/Stream.ts
```diff
@@ -6905,7 +6905,9 @@ export const rechunk: {
             if (chunk.length === 0 && arr.length === target) {
               return Effect.succeed(arr)
             } else if (chunk.length + arr.length < target) {
-              chunk.push(...arr)
+              for (const value of arr) {
```
tim-smart: use a for loop with an index instead

tim-smart: Changed this to an indexed loop in 86484819fd.

PR #7657, opened 2026-09-02
refactor(Channel): remove runDone

### LINKED ISSUE #7656: Channel.runDone returns the first output instead of the completion value

`effect` v4 source. Discovery reproduced at `0af0985d5c132b33e14a2df59e0567d7ba2a1710`; publication base: `243c72f001dd87385839c057af48f346abcf480b`.

```ts
import { Channel, Effect } from "effect"

const result = await Effect.runPromise(
  Channel.fromArray([1, 2]).pipe(
    Channel.concat(Channel.end("finished")),
    Channel.runDone
  )
)
console.log(result)
```

[Executable regressions and controls](https://github.com/kitlangton/effect/blob/audit/i2-channel-rundone/packages/effect/test/Channel.test.ts) are under `destructors`, in the `runDone` and `runDrain` groups.

Return `"finished"` after consuming both outputs. Effects and failures after an emitted element must still be observed.

### What do you see instead?

The result is `1`. Later effects and failures are skipped. The equivalent `Channel.runDrain` call returns the completion value correctly.

### Additional information

`runDone` currently uses a single-pull consumer rather than draining the channel.

54 tests pass on `243c72f00`; lint, typecheck and whitespace checks pass. Red suites ran on `0af0985d5`. The affected implementation is byte-identical between those bases.

## Audit

kitlangton:
## Type

- [x] Refactor

## Description

Remove `Channel.runDone`, which duplicates `Channel.runDrain`. Callers should use `runDrain` to consume emitted elements and return the channel's done value.

The `Sink.toChannel` example and v3-to-v4 migration guidance now point to `runDrain`. The Channel test covers that replacement behavior with an emitted value followed by a distinct done value.

## Verification

```sh
pnpm lint-fix
pnpm test --run packages/effect/test/Channel.test.ts
pnpm doctest --run packages/effect/src/Sink.ts
pnpm jsdocs --check
pnpm check
```

## Related

Closes #7656
Closes EFF-1045

### DIFF
--- a/packages/effect/src/Channel.ts
+++ b/packages/effect/src/Channel.ts
@@ -8076,7 +8076,7 @@ export const runCollect = <OutElem, OutErr, OutDone, Env>(
  */
 export const runDone = <OutElem, OutErr, OutDone, Env>(
   self: Channel<OutElem, OutErr, OutDone, unknown, unknown, unknown, Env>
-): Effect.Effect<OutDone, OutErr, Env> => runWith(self, identity_, Effect.succeed)
+): Effect.Effect<OutDone, OutErr, Env> => runDrain(self)
 
 /**
  * Runs a channel until the first output element is available, returning it in
--- a/packages/effect/src/Channel.ts
+++ b/packages/effect/src/Channel.ts
@@ -8069,16 +8069,6 @@ export const runCollect = <OutElem, OutErr, OutDone, Env>(
   })
 
 /**
- * Runs a channel and outputs the done value.
- *
- * @category running
- * @since 4.0.0
- */
-export const runDone = <OutElem, OutErr, OutDone, Env>(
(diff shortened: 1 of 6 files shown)


packages/effect/src/Channel.ts
```diff
@@ -8076,7 +8076,7 @@ export const runCollect = <OutElem, OutErr, OutDone, Env>(
  */
 export const runDone = <OutElem, OutErr, OutDone, Env>(
```
tim-smart: Lets just remove it as it does the same thing as runDrain?

tim-smart: Removed `Channel.runDone` in b3228e1915 and updated the affected test, docs, migration guidance, and changeset to use `runDrain`.

PR #7687, opened 2026-09-02
fix(cli): allow optional wrappers around alternative flags

### LINKED ISSUE #7686: Flag.optional defects when wrapping Flag.orElse alternatives

`effect` v4 source at discovery base `243c72f001dd87385839c057af48f346abcf480b`; publication base: `243c72f001dd87385839c057af48f346abcf480b`.

Define `Flag.string("config").pipe(Flag.orElse(() => Flag.string("config-url")), Flag.optional)`. Parse primary-only input, alternate-only input, or no flags.

[Runnable regression and controls](https://github.com/kitlangton/effect/blob/audit/i3-optional-alternatives/packages/effect/test/unstable/cli/Param.test.ts): the `optional alternatives` suite provides the required CLI test services. Run from the linked branch's repository root:

```sh
pnpm test --run packages/effect/test/unstable/cli/Param.test.ts -t 'optional alternatives'
```

Supplied alternatives return `Some(value)`; omitting both returns `None`. Unrelated positional operands remain available.

### What do you see instead?

All three cases defect at a single-parameter assertion before the alternatives are parsed. Removing `Flag.optional` lets the primary and alternate flags parse normally.

### Additional information

`Param.optional` calls `getUnderlyingSingleOrThrow` but does not use its result. Alternative flags have multiple leaves, yet optional wrapping needs only their parse result. This differs from alternative-name registration: the defect happens inside the optional parser after registration. `withDefault` also delegates to this wrapper.

Refreshed validation: Isolated suite: 34 passed, 0 skipped. Root `pnpm lint-fix`, `pnpm check`, and committed/worktree whitespace checks passed. Matching test identities in the fresh base run: 3 failed / 31 passed. Fifteen-fix combined run: 372 passed, versus 65 failed / 307 passed on base; no skips.

## Audit

kitlangton:
## Type

- [x] Bug Fix

## Description

`Flag.optional` no longer rejects `Flag.orElse` parameter trees before parsing. `Flag.withDefault`, which delegates to `optional`, now supports alternative flags as well.

The fix removes a stale single-leaf assertion whose result was unused.

## Verification

- `pnpm test --run packages/effect/test/unstable/cli/Param.test.ts`
- `pnpm lint-fix`
- `pnpm check`

## Related

Closes #7686
Closes EFF-1060

### DIFF
--- a/packages/effect/src/unstable/cli/Param.ts
+++ b/packages/effect/src/unstable/cli/Param.ts
@@ -1273,8 +1273,6 @@ export const optional = <Kind extends ParamKind, A>(
   param: Param<Kind, A>
 ): Param<Kind, Option.Option<A>> => {
   const parse: Parse<Option.Option<A>> = Effect.fnUntraced(function*(args) {
-    getUnderlyingSingleOrThrow(param)
-
     return yield* param.parse(args).pipe(
       Effect.map(([leftover, value]) => [leftover, Option.some(value)] as const),
       // Catch both MissingOption (for flags) and MissingArgument (for positional arguments)
--- a/packages/effect/src/unstable/cli/Param.ts
+++ b/packages/effect/src/unstable/cli/Param.ts
@@ -1272,8 +1272,8 @@ export const mapTryCatch: {
 export const optional = <Kind extends ParamKind, A>(
   param: Param<Kind, A>
 ): Param<Kind, Option.Option<A>> => {
-  const parse: Parse<Option.Option<A>> = Effect.fnUntraced(function*(args) {
-    return yield* param.parse(args).pipe(
+  const parse: Parse<Option.Option<A>> = (args) =>
+    param.parse(args).pipe(
       Effect.map(([leftover, value]) => [leftover, Option.some(value)] as const),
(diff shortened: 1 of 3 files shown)


kitlangton:
The only failing check is the Node shard, where `@effect/sql-mysql2`'s `PersistedCache (sql-mysql2-single) > smoke test` hit `ER_LOCK_DEADLOCK`. The other 12 checks passed. This PR changes only the CLI optional-parameter assertion, its tests, and a changeset, so this appears unrelated.

I cannot rerun Actions with my repository permissions. Could a maintainer retry the [failed Node job](https://github.com/Effect-TS/effect/actions/runs/33676338806/job/100401848364)?

packages/effect/src/unstable/cli/Param.ts
```diff
@@ -1273,8 +1273,6 @@ export const optional = <Kind extends ParamKind, A>(
   param: Param<Kind, A>
 ): Param<Kind, Option.Option<A>> => {
   const parse: Parse<Option.Option<A>> = Effect.fnUntraced(function*(args) {
```
tim-smart: This doesn't need fnUntraced any more

PR #7714, opened 2026-09-02
fix(NodeSink): invalidate drain callbacks on interruption

### LINKED ISSUE #7713: Interrupted Node sinks continue submitting writes when backpressure clears

`@effect/platform-node-shared` v4 source at discovery base `243c72f001dd87385839c057af48f346abcf480b`; publication base: `243c72f001dd87385839c057af48f346abcf480b`. Reproduced on Node v26.5.0.

Run `Stream.make("a", "b")` through `NodeSink.fromWritable` into a real `Writable` with `highWaterMark: 1`. Hold the first write callback until backpressure is established, interrupt the sink, then release the callback and allow Node's drain event to run.

[Runnable regression and controls](https://github.com/kitlangton/effect/blob/audit/i4-node-sink-cancel-drain/packages/platform/node-shared/test/NodeSink.test.ts): the `backpressure: endOnDone=$endOnDone, interrupted=$interrupted` matrix, `cancellation during drain dispatch: endOnDone=%s`, and `cancellation during resumed write: endOnDone=$endOnDone, backpressure=$backpressure`. After branch push, run from its repository root:

```sh
pnpm test --run packages/platform/node-shared/test/NodeSink.test.ts
```

The already-submitted `a` may finish, but the interrupted sink must not submit `b`. This also holds when an earlier drain listener cancels the sink during the same dispatch. Cancellation inside a resumed write must prevent further submissions and listener registration.

### What do you see instead?

The drain callback continues submitting data after the sink has exited with interruption. A callback selected before listener removal can still execute, and cancellation during a resumed write can leave another drain listener installed.

### Additional information

The fixture uses native drain dispatch and public `Fiber.interrupt` or `AbortController` cancellation, not manually emitted events or a custom scheduler. The correction must invalidate pending callback work, not cancel bytes already owned by Node. Finalization errors are a separate issue.

Validation: Post-commit validation on `3a5be2dc6644`: 16 distinct tests passed, 0 failed, 0 skipped. All listed PR Verification test commands, root `pnpm lint-fix`, root `pnpm check`, and committed/worktree whitespace checks passed; the worktree remained clean. The frozen paired base run at `243c72f001dd` has 9 failing and 7 passing matched cases for this fix. Only the selected test files were run; integration and cluster suites were disabled.

## Audit

kitlangton:
## Type

- [x] Bug Fix

## Description

Interrupting a Node sink while it waits for writable backpressure could still submit the next buffered value. Node may already have selected the sink's `drain` callback when interruption removes that listener.

The write loop now keeps a local cancellation flag. Cleanup sets the flag before removing the pending `drain` listener, so an already-selected callback returns before writing, registering another listener, or completing.

The regression uses `Stream.run(NodeSink.fromWritable(...))`, a real writable `drain` event, and public cancellation through an `AbortSignal`.

## Verification

```sh
nix develop -c pnpm lint-fix
nix develop -c pnpm test --run packages/platform/node-shared/test/NodeSink.test.ts
nix develop -c pnpm check
```

## Related

Closes EFF-1073
Closes #7713

### DIFF
--- a/packages/platform/node-shared/src/NodeSink.ts
+++ b/packages/platform/node-shared/src/NodeSink.ts
@@ -83,29 +83,26 @@ export const pullIntoWritable = <A, IE, E>(options: {
   options.pull.pipe(
     Effect.flatMap((chunk) => {
       let i = 0
-      return Effect.callback<void, E>((resume) => {
-        let cancelled = false
+      return Effect.callback<void, E>((resume, signal) => {
         const loop = () => {
           for (; i < chunk.length;) {
-            if (cancelled) {
+            if (signal.aborted) {
               return
             }
             const success = options.writable.write(chunk[i++], options.encoding as any)
             if (!success) {
-              if (!cancelled) {
+              if (!signal.aborted) {
                 options.writable.once("drain", loop)
               }
               return
             }
           }
-          if (!cancelled) {
+          if (!signal.aborted) {
             resume(Effect.void)
           }
         }
         loop()
         return Effect.sync(() => {
-          // Removing a listener does not invalidate an in-progress drain dispatch.
-          cancelled = true
           options.writable.off("drain", loop)
         })
       })
--- a/packages/platform/node-shared/src/NodeSink.ts
+++ b/packages/platform/node-shared/src/NodeSink.ts
@@ -83,15 +83,31 @@ export const pullIntoWritable = <A, IE, E>(options: {
   options.pull.pipe(
     Effect.flatMap((chunk) => {
       let i = 0
-      return Effect.callback<void, E>(function loop(resume) {
-        for (; i < chunk.length;) {
-          const success = options.writable.write(chunk[i++], options.encoding as any)
-          if (!success) {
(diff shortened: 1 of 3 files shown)


packages/platform/node-shared/src/NodeSink.ts
```diff
@@ -83,15 +83,28 @@ export const pullIntoWritable = <A, IE, E>(options: {
   options.pull.pipe(
     Effect.flatMap((chunk) => {
       let i = 0
-      return Effect.callback<void, E>(function loop(resume) {
-        for (; i < chunk.length;) {
-          const success = options.writable.write(chunk[i++], options.encoding as any)
-          if (!success) {
-            options.writable.once("drain", () => (loop as any)(resume))
-            return
+      return Effect.callback<void, E>((resume, signal) => {
```
tim-smart: Avoid using the signal here, and instead a boolean flag

tim-smart: Updated in ee50cf0201 to use a local boolean cancellation flag. The focused regression and required local checks pass.

PR #7772, opened 2026-09-02
Improve HTTP server and HttpRouter request throughput

tim-smart:
Improves `HttpRouter` and HTTP server throughput, based on CPU and heap profiles against Hono on `@hono/node-server`.

Closes EFF-1042

## Benchmark snapshot

`autocannon -c 100`, Node 26, logger disabled, interleaved fresh-process rounds at revision `9a4c30b`:

| route | effect before | effect at `9a4c30b` | hono |
|---|---:|---:|---:|
| `GET /` | 75.0k rps | 89.7k rps | 94.4k rps |
| `GET /user/:id` | 62.0k rps | 85.6k rps | 89.9k rps |

The review fixes after that revision restore request and tracing semantics and have not been re-benchmarked.

### FindMyWay router

- Resolve static routes with a map lookup after URL normalization.
- Resolve single trailing-parameter routes such as `/users/:id` with a prefix-key map lookup. Disable this fast path per method when regex, wildcard-adjacent, mid-segment, or suffixed parameters can shadow it.
- Skip URL normalization for already-canonical paths.
- Compile safe, unique parameter names into stable null-prototype objects. Fall back to assignment for other names and runtimes that block dynamic code generation.
- Reuse backtracking and parameter scratch arrays per router instance, and skip duplicate-slash replacement when no duplicate exists.

### HTTP request handling

- Set route context directly and enter interruptibility without an extra effect wrapper.
- Resolve `HttpRouter.params` with `Effect.withFiberSucceed`.
- Fuse request scope setup with uninterruptibility in `HttpEffect.toHandled`.
- Use one boundary request throughout response handling, including middleware and failure paths.
- Skip server tracing only when tracing is disabled. The default native tracer still provides `Tracer.ParentSpan` to handlers.
- Keep the single-continuation response path for requests where tracing is disabled and no middleware is present.

### Fiber runtime

- Read the fiber cache once per interpreter operation.
- Avoid a duplicate ensure-continuation lookup.
- Use a property read in `effectIsExit`.

### Bodies and responses

- Retain text bodies and encode them lazily. Node writes strings directly, while Bun and Web responses also reuse the text source.
- Build default response headers without spreading and resetting the prototype.
- Copy the exact byte range when exposing an `ArrayBuffer` from a pooled or offset byte view.
- Handle RPC notification-only responses explicitly while preserving legacy coercion for untyped `HttpBody.text` callers.

## Validation

- `packages/effect/test`: 9,114 passed, 47 skipped.
- `packages/platform/node/test`: 404 passed.
- Focused HTTP and FindMyWay suites: 96 passed.
- `pnpm check`, `pnpm lint-fix`, `pnpm circular`, and `pnpm jsdocs --check`: clean.
- `pnpm test-types`: 2,226 passed across TypeScript 5.9 and 6.0.
- `pnpm doctest --run`: 3,388 passed.


packages/effect/src/internal/core.ts
```diff
@@ -571,6 +571,24 @@ export const withFiber: <A, E = never, R = never>(
...
 })
 
+/**
+ * A variant of `withFiber` for synchronously computed success values, which
+ * continues directly with the result instead of evaluating a `succeed` exit
+ * as a separate operation.
+ *
+ * @internal
+ */
+export const succeedWithFiber: <A, R = never>(
```
tim-smart: This could be exported as Effect.withFiberSucceed

packages/effect/src/unstable/http/HttpMiddleware.ts
```diff
@@ -171,6 +171,15 @@ export const logger: <E, R>(
...
 )
 
+/**
+ * Returns whether server tracing is unobservable for the current fiber.
+ *
+ * @internal
+ */
+export const isTracerDisabledFastUnsafe = (fiber: {
+  getRef: <A>(ref: Context.Reference<A>) => A
+}): boolean => fiber.getRef(Tracer) === nativeTracer || !fiber.getRef(TracerEnabled)
```
tim-smart: Can this use the fiber cache?

packages/effect/src/unstable/http/internal/httpMiddleware.ts
```diff
@@ -0,0 +1,18 @@
+import * as Context from "../../../Context.ts"
+import type * as Fiber from "../../../Fiber.ts"
+import { constFalse } from "../../../Function.ts"
+import type { Predicate } from "../../../Predicate.ts"
+import { TracerEnabled } from "../../../References.ts"
+import type { HttpServerRequest } from "../HttpServerRequest.ts"
+
+/** @internal */
+export const TracerDisabledWhen = Context.Reference<Predicate<HttpServerRequest>>(
```
tim-smart: Remove this file and keep it in the public module. Also it is no longer using fiber.cache for the TracerEnabled ref

PR #7829, opened 2026-09-03
fix(sql-libsql): isolate transaction contexts per client

### LINKED ISSUE #7828: libSQL transactions can redirect another client's queries to the wrong database

Effect v4 source at `f74282cb08de808b678ce67687acbcd61939d2f6`, using `@effect/sql-libsql`.

1. Construct clients A and B for two different local database files.
2. Seed each database with a distinct marker and confirm each client reads its own marker.
3. Execute a query through B inside `A.withTransaction(...)`.
4. Compare with B's query outside A's transaction. Repeat with A and B swapped.

[Native regression coverage](https://github.com/kitlangton/effect/blob/audit/c2-r1-libsql-transaction-client-isolation/packages/sql/libsql/test/TransactionIsolation.test.ts) uses two real local databases, not a mocked query result.

Queries issued through B keep targeting B's configured database. A transaction belongs to its specific client; constructing a second client does not opt it into a cross-database transaction.

### What do you see instead?

B's query uses A's transaction connection. Reads return A's marker, writes can target A instead of B, and a nested B transaction can be mistaken for a savepoint in A.

### Additional information

The adapter installs and retrieves transaction connections under one module-global service key. Allocating a unique private transaction key per constructed client preserves database identity while retaining same-client nesting and transformed-view behavior. Moving the declaration without changing its string key does not fix the collision.

## Audit

kitlangton:
## Why

Two libSQL clients shared one module-level transaction context key. A query through client B inside client A's transaction could therefore run against A's database.

## What changed

- Allocate a private transaction context key for each constructed libSQL client.
- Add one public-API regression to the existing libSQL client integration suite.
- Add a patch changeset for `@effect/sql-libsql`.

## Verification

The focused regression fails against the unfixed source because client B returns A's marker row. It passes on this branch.

```bash
nix develop -c pnpm lint-fix
nix develop -c env EFFECT_INTEGRATION_TESTS=1 pnpm test --run --project @effect/sql-libsql packages/sql/libsql/test/Client.integration.test.ts --maxWorkers=1 --no-file-parallelism --sequence.concurrent=false
nix develop -c pnpm check
git diff --check
```

Closes #7828
Closes EFF-1141


packages/sql/libsql/test/TransactionIsolation.test.ts
```diff
@@ -0,0 +1,23 @@
+import { LibsqlClient } from "@effect/sql-libsql"
```
tim-smart: This is the wrong test file

tim-smart: Moved the regression into `packages/sql/libsql/test/Client.integration.test.ts` and deleted the standalone test file in 0c8f8ed744. The gated integration suite passes 8/8 with `EFFECT_INTEGRATION_TESTS=1`.

tim-smart (approved):

PR #7831, opened 2026-09-03
fix(sql-sqlite-bun): allow database export within transactions

### LINKED ISSUE #7830: Bun SQLite export waits on its own transaction lock

Effect v4 source at `f74282cb08de808b678ce67687acbcd61939d2f6`, using `@effect/sql-sqlite-bun` on Bun 1.4.0.

1. Create an in-memory `SqliteClient` and seed a table.
2. Confirm `sql.export` returns a readable snapshot outside a transaction.
3. Execute `sql.withTransaction(sql.export)`.
4. Compare with an ordinary query inside a transaction and with direct native `Database.serialize()` inside `BEGIN IMMEDIATE`.

[Native regression coverage](https://github.com/kitlangton/effect/blob/audit/c2-r1-sqlite-bun-export-in-transaction/packages/sql/sqlite-bun/test/TransactionExport.test.ts) includes the driver control and resource/serialization checks.

Export uses the current client's transaction connection and returns its native SQLite snapshot. It does not commit the transaction. Calls outside that transaction must still respect the client's serialization lock.

### What do you see instead?

Export waits for the same permit already held by its enclosing transaction, so the transaction cannot complete. Outside exports and ordinary transactional queries work. Native Bun SQLite supports serialization while the transaction is active.

### Additional information

The export accessor always uses the nontransactional acquirer. It should bypass that acquisition only when the matching client's transaction context is present—not unconditionally and not merely because another client's transaction is active.

## Audit

kitlangton:
## Why

`sql.withTransaction(sql.export)` waits for its own transaction's serialization permit. The transaction cannot release that permit until export completes, even though native Bun SQLite can serialize the active database connection.

## What changes

Check the current client's transaction service when export executes. Reuse its connection when that context is present; otherwise keep the existing acquirer and locking behavior.

| Export context | Before | After |
| --- | --- | --- |
| Inside this client's transaction | Waits on its own permit | Returns a native snapshot |
| Same-client nested transaction or transformed view | Same self-wait | Preserves owning transaction context |
| Outside this client's transaction | Waits for its permit | Unchanged |
| Unrelated client's transaction context | Must still acquire this client's permit | Preserved |

## Scope

The Bun SQLite export path, a regression test in the existing client test file, and an `@effect/sql-sqlite-bun` patch changeset. The fix does not make the semaphore reentrant or change extension loading, transaction commits or snapshot format.

## Verification

```bash
pnpm lint-fix
bun node_modules/vitest/vitest.mjs run --project @effect/sql-sqlite-bun packages/sql/sqlite-bun/test/Client.test.ts --maxWorkers=1 --no-file-parallelism --sequence.concurrent=false
bun node_modules/vitest/vitest.mjs run --project @effect/sql-sqlite-bun --maxWorkers=1 --no-file-parallelism --sequence.concurrent=false
pnpm check
git diff --check
```

The focused client test and full package suite both pass 5/5. The regression verifies that an export inside an active transaction completes and includes that transaction's uncommitted write.

Closes #7830
Closes EFF-1142

## Audit


packages/sql/sqlite-bun/test/TransactionExport.test.ts
```diff
@@ -0,0 +1,215 @@
+import type { SqliteClient } from "@effect/sql-sqlite-bun"
```
tim-smart: This is the wrong test file

tim-smart: Moved this into `Client.test.ts` and cut the coverage to one regression case. It verifies that an export inside the active transaction completes and contains the row written by that transaction. The focused and package suites both pass.

PR #7833, opened 2026-09-03
fix(platform-deno): decode Redis URL authority credentials once

### LINKED ISSUE #7832: Deno Redis URLs send percent-encoded authority credentials literally

Effect v4 source at `a3f2b319343f28edb028270f7074066cea0ec429`, using `@effect/platform-deno` and Deno 2.9.4.

1. Supply a Redis URL whose authority contains percent-encoded username/password characters.
2. Compare the emitted AUTH command with the equivalent explicit connection settings.
3. For example, fictional URL credentials `app%3Auser:p%40ss%25` should identify username `app:user` and password `p@ss%`.

[Regression cases](https://github.com/kitlangton/effect/blob/audit/c2-r1-deno-redis-url-credential-decoding/packages/platform/deno/test/DenoRedis.credentials.test.ts) run the real installed driver against an in-memory RESP peer; no real credentials or Redis server are needed.

URL authority credentials are decoded once. Explicit options retain precedence and are not URI-decoded. The driver's already-decoded query-password fallback remains unchanged.

### What do you see instead?

The driver sends the percent-encoded authority strings literally in AUTH, so the URL form fails against an identity that succeeds with equivalent explicit settings.

### Additional information

Effect already normalizes the dependency's URL representation and merges explicit options. Normalizing raw authority fields at that boundary avoids both the omission and the double-decoding caused by blindly decoding every parsed password.

## Audit

kitlangton:
## Why

Deno Redis URLs with percent-encoded authority credentials send those escape sequences literally in AUTH. The same identity works when supplied through explicit username and password options.

## What changes

Decode raw URL authority usernames and passwords once before applying explicit connection options. Query-string passwords keep their existing decoding behavior, and explicit options remain authoritative.

The Redis integration suite uses a local TCP listener to inspect the real AUTH frame. The new cases cover one-time authority decoding and unchanged query-password behavior; the suite already covers explicit-option precedence.

## Malformed percent escapes

Malformed authority escape sequences fail during layer construction and are wrapped in `RedisError`. This is deliberate. Falling back to the raw value would silently send invalid URL encoding as a credential and hide the configuration error.

## Verification

```bash
pnpm lint-fix
EFFECT_INTEGRATION_TESTS=1 deno run -A --cached-only --no-prompt --no-lock node_modules/vitest/vitest.mjs run --project @effect/platform-deno packages/platform/deno/test/DenoRedis.integration.test.ts --maxWorkers=1 --no-file-parallelism -t 'decodes URL authority credentials once|does not decode a query-string password twice'
deno check --no-remote --deny-import --no-lock --node-modules-dir=manual packages/platform/deno/src/DenoRedis.ts packages/platform/deno/test/DenoRedis.integration.test.ts
pnpm check
git diff --check
```

Both added integration cases pass. Removing the three runtime lines makes the authority-decoding case fail while the query-password case continues to pass.

Closes EFF-1143

Closes #7832

## Audit


packages/platform/deno/test/DenoRedis.credentials.test.ts
```diff
@@ -0,0 +1,62 @@
+import * as DenoRedis from "@effect/platform-deno/DenoRedis"
```
tim-smart: This is the wrong test file

tim-smart: Moved the two non-duplicate cases into DenoRedis.integration.test.ts and removed the standalone file in bcb4bfb5f6. The existing integration suite already covers explicit-option precedence.

PR #7837, opened 2026-09-03
fix(SqlMessageStorage): preserve reply IDs in by-ID reads

### LINKED ISSUE #7836: SQL message by-ID reads lose joined replies and mask acknowledgement reply IDs

Effect v4 source at `8364ddda233d35ee472cc2356312317a60a148a2`, using cluster `SqlMessageStorage`.

1. Save a streaming request, a reply chunk, and its acknowledgement.
2. Read that now-eligible request through `unprocessedMessagesById`.
3. Compare `lastSentReply` with the persisted reply and the equivalent shard-reader result.
4. Read an AckChunk envelope by its own message ID and check its `replyId`.

[Native SQLite regressions](https://github.com/kitlangton/effect/blob/audit/c2-r2-sql-message-storage-by-id-reply-alias/packages/platform/node/test/cluster/SqlMessageStorage.byId.test.ts) use the real adapter and ordinary public persistence operations.

The request retains the saved reply's ID, request ID, sequence and encoded values. The acknowledgement retains its own acknowledged `replyId`.

### What do you see instead?

The by-ID request loses `lastSentReply`, returning `None` even though the shard reader returns the saved chunk. In the native SQLite control, the acknowledgement's `replyId` also becomes `"null"`.

### Additional information

The by-ID SELECT aliases the joined reply's ID as `reply_id`, but the common decoder expects `reply_reply_id`. The shorter alias also collides with the message row's own `reply_id`. Matching the existing shard-query alias repairs both observations without changing eligibility or adding a decoder fallback.

## Audit

kitlangton:
## Why

`unprocessedMessagesById` aliases the joined reply ID as `reply_id`, which collides with an `AckChunk` message's own reply ID. The decoded acknowledgement therefore receives `"null"` instead of the saved chunk ID.

## What changed

- Alias the joined value as `reply_reply_id`, matching the shared row decoder and shard-based query.
- Cover the collision in the existing SQL message storage integration test.
- Add an `effect` patch changeset.

## Verification

```bash
nix develop -c pnpm lint-fix
nix develop -c env EFFECT_INTEGRATION_TESTS=1 pnpm test --run packages/platform/node/test/cluster/SqlMessageStorage.integration.test.ts
nix develop -c pnpm check
git diff --check
```

Closes #7836
Closes EFF-1145


packages/platform/node/test/cluster/SqlMessageStorage.byId.test.ts
```diff
@@ -0,0 +1,41 @@
+import { NodeCrypto } from "@effect/platform-node"
```
tim-smart: This test is in the wrong file

tim-smart: Moved the regression into the existing SqlMessageStorage.integration.test.ts coverage and removed the standalone file. The integration suite passes all 56 tests.

PR #7840, opened 2026-09-03
fix(DurableClock): preserve explicit zero in-memory thresholds

### LINKED ISSUE #7839: DurableClock replaces numeric zero thresholds with the default

Effect v4 source at `b628bb1240a09df2244c3ff6c587f978feb25225`, using `DurableClock.sleep`.

1. Request a positive duration, such as 10 milliseconds.
2. Set `inMemoryThreshold` to numeric `0` or bigint `0n`.
3. Compare routing with the equivalent `Duration.zero` or `"0 millis"` threshold.

[Regression coverage](https://github.com/kitlangton/effect/blob/audit/c2-r2-durable-clock-zero-threshold-defaulted/packages/effect/test/unstable/workflow/DurableClock.boundary.test.ts) records the public engine calls with finite, already-completed responses.

All zero representations select durable scheduling for a positive duration. Only an omitted/undefined threshold uses the documented 60-second default; durations at or below a supplied threshold use the in-memory activity path.

### What do you see instead?

Numeric and bigint zero are treated as absent, selecting the default threshold and the in-memory path. Equivalent object/string zero values select the durable path.

### Additional information

The threshold is tested for truthiness before conversion to `Duration`. An explicit undefined check preserves valid zero values without changing the duration comparison or scheduling implementation.

## Audit

kitlangton:
## Why

`DurableClock.sleep` treated numeric `0` and bigint `0n` thresholds as missing and substituted the 60-second default. Positive durations could therefore use the in-memory activity path instead of durable scheduling.

## What changed

- Default `inMemoryThreshold` only when it is `undefined`.
- Cover explicit numeric and bigint zero in the existing cluster workflow engine tests.
- Add a patch changeset for `effect`.

## Verification

Against the unfixed conditional, both regression cases fail because they call `activityExecute`. With the fix, both call `scheduleClock` followed by `deferredResult`.

```bash
nix develop -c pnpm lint-fix
nix develop -c pnpm test --run packages/effect/test/cluster/ClusterWorkflowEngine.test.ts --maxWorkers=1 --no-file-parallelism
nix develop -c pnpm check
git diff --check
```

Closes #7839
Closes EFF-1146


packages/effect/test/unstable/workflow/DurableClock.boundary.test.ts
```diff
@@ -0,0 +1,50 @@
+import { assert, describe, it } from "@effect/vitest"
```
tim-smart: This test is in the wrong file

tim-smart: Moved the two regression cases into `packages/effect/test/cluster/ClusterWorkflowEngine.test.ts` and removed the standalone test file in 6e296d9c45.

PR #7854, opened 2026-09-03
fix(utils): preserve absolute codegen discovery matches

### LINKED ISSUE #7853: Codegen duplicates cwd when discovering absolute patterns

Repository `@effect/utils` codegen tool at `2122e07b8688344a1d49c334a10223a9e00c60da`.

1. Create a small source directory containing an annotated barrel and a module with a valid `@since` annotation.
2. Run codegen with that directory as `--cwd` and a relative barrel `--pattern`.
3. Repeat using the absolute spelling of the same pattern, or call `BarrelGenerator.discoverFiles` with those two spellings.

[Registered regression coverage](https://github.com/kitlangton/effect/blob/audit/c2-r2-codegen-absolute-discovery-pattern-rebased-twice/packages/tools/utils/test/Codegen.test.ts) uses real Glob, discovery and CLI calls against owned temporary fixtures.

Both spellings find the same barrel. Absolute matches already identify their files; relative matches remain relative to the supplied working directory.

### What do you see instead?

The absolute spelling fails with a filesystem `ENOENT` error. Discovery joins `cwd` onto a filename that Glob has already made absolute, producing a duplicated path before reading the annotation.

### Additional information

Only relative matches should be joined to `cwd`. Unconditionally resolving every result would also change the helper's existing relative-cwd output paths, which is unnecessary for this fix.

## Audit

kitlangton:
## Why

`glob@13` preserves an absolute input pattern as an absolute match. `BarrelGenerator.discoverFiles` always joined matches with `cwd`, so absolute matches became duplicated paths and failed with `FileSystem.readFile` `NotFound`.

## What changed

Keep absolute discovery matches unchanged and prefix `cwd` only for relative matches. Register the Node-only `@effect/utils` Vitest project, cover the behavior through the public `BarrelGenerator.discoverFiles` interface, and add a patch changeset.

## Verification

```bash
pnpm lint-fix
pnpm test --run --project @effect/utils packages/tools/utils/test/Codegen.test.ts --maxWorkers=1 --no-file-parallelism --maxConcurrency=1
pnpm check
git diff --check
```

Closes #7853
Closes EFF-1152

## Audit


packages/tools/utils/src/Codegen.ts
```diff
@@ -182,7 +182,10 @@ export const layer: Layer.Layer<BarrelGenerator, never, FileSystem.FileSystem |
         nodir: true
       })
 
-      const results = yield* Effect.forEach(indexFiles, (file) => discoverFile(path.join(cwd, file)))
+      const results = yield* Effect.forEach(
+        indexFiles,
+        (file) => discoverFile(path.isAbsolute(file) ? file : path.join(cwd, file))
+      )
```
fubhy: This can never include absoute files because `glob` returns relative paths unless explicitly configured to `absolute: true`

fubhy (dismissed):

packages/tools/utils/src/Codegen.ts
tim-smart: Confirmed through the public interface with the real `Glob.layer` and filesystem: on the baseline, `discoverFiles(absoluteBarrelPath, cwd)` tries to read a path with `cwd` duplicated and fails with `FileSystem.readFile` `NotFound`.

For `glob@13`, an absolute input pattern produces an absolute match without `absolute: true`. Setting `absolute: true` here would also change path spelling for relative patterns. I kept the narrow `path.isAbsolute(file)` guard so relative behavior stays unchanged. The focused regression is in `packages/tools/utils/test/Codegen.test.ts`.

PR #7860, opened 2026-09-03
fix(SqlRunnerStorage): bound PostgreSQL acquisition results to requested shards

### LINKED ISSUE #7859: PostgreSQL runner storage returns held shards outside the requested set

Effect v4 source at `a62f4838f007ed74fefc9253c2836261943624e2`, using PostgreSQL advisory locking in `SqlRunnerStorage`.

1. Create a healthy PostgreSQL runner storage and acquire shard `default:1`.
2. Using the same storage, acquire only `default:2`.
3. Inspect the returned shard list. Also check a separate storage instance using the same prefix cannot acquire the still-held first shard.

[Integration regressions](https://github.com/kitlangton/effect/blob/audit/c2-r2-sql-runner-pg-acquire-unrequested-held-shards/packages/platform/node/test/cluster/SqlRunnerStorage.acquire.integration.test.ts) exercise normal public acquire/refresh/release calls.

The second call returns only `default:2`, the shard it was asked to acquire. The first shard remains locked until it is explicitly released.

### What do you see instead?

It returns both shards. The PostgreSQL advisory path adds every recognized held lock to the result, even when that lock is absent from the current request. Nonempty refresh reuses the same path.

### Additional information

The input-bounded contract already holds for the native PostgreSQL row-backed path. Filtering the held-lock results must not release unrelated locks or reacquire requested held locks and increase their advisory-lock counts.

## Audit

kitlangton:
## Why

PostgreSQL advisory acquisition included shards already held by the reserved connection even when they were absent from the current request.

## What changed

- Filter held advisory locks through the current request before returning shard IDs.
- Extend the existing `SqlRunnerStorage` integration coverage with a disjoint request.
- Add a patch changeset.

## Verification

- Without the guard, the PostgreSQL regression received shards 1-4 when only shard 4 was requested.
- `pnpm lint-fix`
- `EFFECT_INTEGRATION_TESTS=1 pnpm test --run packages/platform/node/test/cluster/SqlRunnerStorage.integration.test.ts -t acquireShards`
- `pnpm check`
- `git diff --check`

Closes #7859
Closes EFF-1153


packages/platform/node/test/cluster/SqlRunnerStorage.acquire.integration.test.ts
```diff
@@ -0,0 +1,20 @@
+import { describe, expect, it } from "@effect/vitest"
```
tim-smart: This is the wrong test file

PR #7862, opened 2026-09-03
fix(EventLogRemote): retry wrapped authentication Forbidden errors

### LINKED ISSUE #7861: EventLogRemote fails to retry authentication Forbidden responses

Effect v4 source at `4372c79a32fcecc7b6b284d49a215b2c4fef9d77`, using `EventLogRemote`.

1. Create an encrypted remote against the in-process event-log server and memory storage.
2. Write an entry for a newly generated identity A.
3. Write an entry for identity B through the same remote.
4. Compare B's first result with an immediate manual repeat of the same entry.

[Regression tests](https://github.com/kitlangton/effect/blob/audit/c2-r2-eventlog-remote-wrapped-forbidden-not-retried/packages/effect/test/unstable/eventlog/EventLogRemote.retry.test.ts) use new local fixture identities, real cryptography and in-process RPC; no existing keys or remote service are required.

The remote uses its documented bounded Forbidden recovery: obtain a fresh challenge, authenticate B and complete its first write. A's cached authentication remains usable.

### What do you see instead?

B's first write fails through `write → authenticate → Forbidden`, while the manual repeat succeeds. Authentication wraps the protocol failure before the retry predicate sees it, so the predicate misses the Forbidden code.

### Additional information

Raw write-level Forbidden responses already retry. Classification needs to inspect only the immediate known authentication wrapper, leaving original errors, non-Forbidden behavior and the existing retry bound intact. No server authorization or signature policy change is needed.

## Audit

kitlangton:
## Summary

- Retry `EventLogRemote` writes and change streams when authentication returns a wrapped `Forbidden` error.
- Limit unwrapping to the immediate `authenticate` wrapper so other errors keep their existing behavior.
- Add a focused regression test to the existing `EventLogRemote` test file.

## Verification

- `pnpm lint-fix`
- `pnpm test --run packages/effect/test/unstable/eventlog/EventLogRemote.test.ts --maxWorkers=1 --no-file-parallelism`
- `pnpm check`
- `git diff --check origin/main...HEAD`

Closes #7861
Closes EFF-1154


packages/effect/src/unstable/eventlog/EventLogRemote.ts
```diff
@@ -206,8 +206,11 @@ export const makeWith = Effect.fnUntraced(function*({ encodeWrite, decodeChanges
     Effect.retry(effect, {
       while(e) {
         hello = null
-        const isForbidden = Predicate.isTagged(e, "EventLogProtocolError") &&
-          (e as any as EventLogProtocolError).code === "Forbidden"
+        const error = e instanceof EventLogRemoteError && e.method === "authenticate"
```
tim-smart: Use Predicate.isTagged - instanceof is banned in this repo.

tim-smart: Updated in `1eb5503c21`: the classifier now uses `Predicate.isTagged` while retaining the `authenticate` method guard. The focused test, lint, and typecheck pass.

packages/effect/src/unstable/eventlog/EventLogRemote.ts
```diff
@@ -206,8 +206,12 @@ export const makeWith = Effect.fnUntraced(function*({ encodeWrite, decodeChanges
     Effect.retry(effect, {
       while(e) {
         hello = null
-        const isForbidden = Predicate.isTagged(e, "EventLogProtocolError") &&
-          (e as any as EventLogProtocolError).code === "Forbidden"
+        const error = Predicate.isTagged(e, "EventLogRemoteError") &&
```
tim-smart: I think add a static method `EventLogRemoteError.is` to the errors which have a `u is EventLogRemoteError` to remove the types casts.

tim-smart: Added `EventLogRemoteError.is` in `c0efa5dc47` and used it to remove the `EventLogRemoteError` casts from the classifier. The focused test, lint, and typecheck pass.

packages/effect/src/unstable/eventlog/EventLogRemote.ts
```diff
@@ -206,8 +215,11 @@ export const makeWith = Effect.fnUntraced(function*({ encodeWrite, decodeChanges
     Effect.retry(effect, {
       while(e) {
         hello = null
-        const isForbidden = Predicate.isTagged(e, "EventLogProtocolError") &&
-          (e as any as EventLogProtocolError).code === "Forbidden"
+        const error = EventLogRemoteError.is(e) && e.method === "authenticate"
+          ? e.cause
+          : e
+        const isForbidden = Predicate.isTagged(error, "EventLogProtocolError") &&
```
tim-smart: Add a .is method to EventLogProtocolError too

tim-smart: Added `EventLogProtocolError.is` in `b3e65310e8` and used it to remove the remaining classifier cast. The focused test, lint, and typecheck pass.

PR #7875, opened 2026-09-03
Wait for inherited stdio close in NodeChildProcessSpawner release and kill

### LINKED ISSUE #7821: platform-node-shared: scoped release and `kill` wait on the leader's `exit`, abandoning descendants when the leader is a non-exec'ing shell (Linux `sh -c`)
## Summary

`NodeChildProcessSpawner` signals the **process group** on scoped release and on `kill`, but completion is measured on the **leader's `exit`** (`Deferred.await(exitSignal)`). When the leader exits before its descendants, the release returns immediately and `forceKillAfter` never escalates, so the descendants keep running unobserved.

This is deterministic on Linux for any `shell: true` command: `/bin/sh` is dash (0.5.12 on Debian/Ubuntu), and dash does not `exec` the `-c` command, so the spawned leader is the shell itself. SIGTERM kills the shell instantly, `exit` fires, and the real process is still shutting down. On macOS `/bin/sh` is bash, which execs, so the leader _is_ the real process and the problem does not reproduce.

Version: `@effect/platform-node-shared@4.0.0-rc.112` (newest rc; `main` unchanged at `packages/platform-node-shared/src/NodeChildProcessSpawner.ts`: release awaits `exitSignal` at the "Process already exited" and "Process is still running" branches, and `kill` at the `killProcessGroup … andThen(Deferred.await(exitSignal))` site).

## Topology

`shell: true` on the two platforms. Effect signals and waits on the **leader** (the pid `spawn()` returned):

```mermaid
flowchart LR
  subgraph mac["macOS · /bin/sh = bash (execs)"]
    E1["Effect spawner"] -->|"leader = node wrapper"| W1["node wrapper.mjs"]
    W1 -.->|"detached (own pgid)"| C1["node child"]
  end
  subgraph linux["Linux · /bin/sh = dash (no exec)"]
    E2["Effect spawner"] -->|"leader = sh"| S2["sh -c"]
    S2 --> W2["node wrapper.mjs"]
    W2 -.->|"detached (own pgid)"| C2["node child"]
  end
```

```text
macOS (bash execs)                     Linux (dash keeps the shell)
Effect ──► node wrapper  ◄─ leader     Effect ──► sh -c        ◄─ leader
              └─(detached) child                   └─► node wrapper
                                                         └─(detached) child
```

## Timeline of `Scope.close` on Linux

```mermaid
sequenceDiagram
  participant E as Effect release
  participant S as sh (leader)
  participant W as node wrapper
  participant C as child (detached)
  E->>S: SIGTERM to process group
  E->>W: SIGTERM (same group)
  S-->>E: exit (t ≈ 1 ms)
  Note over E: exitSignal resolved, release returns,<br/>forceKillAfter timer dropped
  Note over W: still alive, forwards at t = 150 ms
  W->>C: SIGTERM (t = 150 ms)
  C-->>W: exit
  W-->>W: exit
  Note over E,C: caller already moved on. With the close-wait<br/>patch the release returns here instead (≈ 159 ms)
```

```text
t=0     Effect: kill(-pgid, SIGTERM)  ── sh dies ── 'exit' fires ── release returns (1 ms)
t=150ms wrapper forwards SIGTERM ─► child prints 'child:SIGTERM' ─► wrapper exits
        ▲ nobody is waiting any more; forceKillAfter never escalates
```

## Reproduction

Three files in an empty directory:

```json
// package.json
{ "private": true, "type": "module",
  "dependencies": { "effect": "4.0.0-rc.112", "@effect/platform-node": "4.0.0-rc.112" } }
```

```js
// wrapper.mjs
// Models a supervisor such as portless: forwards SIGTERM to a detached child after 150 ms,
// exits only after the child exits.
import { spawn } from 'node:child_process'
const child = spawn(process.execPath, ['-e', `
  process.once('SIGTERM', () => { console.log('child:SIGTERM'); process.exit(0) })
  setTimeout(() => process.exit(0), 4000)
`], { detached: true, stdio: ['ignore', 'inherit', 'inherit'] })
process.once('SIGTERM', () => {
  console.log('wrapper:SIGTERM')
  setTimeout(() => process.kill(-child.pid, 'SIGTERM'), 150)
})
child.once('exit', () => process.exit(0))
console.log('READY')
```

```js
// run.mjs
import * as ChildProcess from 'effect/unstable/process/ChildProcess'
import * as ChildProcessSpawner from 'effect/unstable/process/ChildProcessSpawner'
import * as NodeServices from '@effect/platform-node/NodeServices'
import { Effect, Scope, Exit, Stream } from 'effect'

  const spawner = yield* ChildProcessSpawner.ChildProcessSpawner
  const scope = yield* Scope.make()
  const handle = yield* Scope.provide(scope)(spawner.spawn(
    ChildProcess.make(`"${process.execPath}" wrapper.mjs`, [], {
      shell: true, stdin: 'ignore', stdout: 'pipe', stderr: 'pipe',
      killSignal: 'SIGTERM', forceKillAfter: '15 seconds',
    }),
  ))
  const lines = []
  yield* handle.stdout.pipe(Stream.decodeText, Stream.splitLines,
    Stream.runForEach((l) => Effect.sync(() => lines.push(l))), Effect.forkScoped)
  while (!lines.includes('READY')) yield* Effect.sleep('10 millis')
  const started = Date.now()
  yield* Scope.close(scope, Exit.void)
  const releaseMs = Date.now() - started
  const atRelease = lines.filter((l) => l.endsWith(':SIGTERM'))
  yield* Effect.sleep('1 second')
  console.log(JSON.stringify({ platform: process.platform, releaseMs, atRelease,
    oneSecondLater: lines.filter((l) => l.endsWith(':SIGTERM')) }))
}).pipe(Effect.scoped)

Effect.runPromise(program.pipe(Effect.provide(NodeServices.layer))).then(() => process.exit(0))
```

Run on Linux without leaving macOS/Windows (Docker):

```sh
docker run --rm -v "$PWD":/w -w /w node:22 sh -c 'npm install --silent --no-audit --no-fund && node run.mjs'
```

Or natively: `npm install && node run.mjs`.

Measured output (node 22, published `@effect/platform-node-shared@4.0.0-rc.112`; `releaseMs` = time for `Scope.close` to return, `atRelease` = signal lines already printed then):

| Platform              | `/bin/sh`                   | spawner                   | releaseMs | atRelease              |
| --------------------- | --------------------------- | ------------------------- | --------- | ---------------------- |
| Linux (Docker recipe above) | dash 0.5.12 (does not exec) | rc.112 as published | **1–2** | none yet (both appear ~150 ms later, unobserved) |
| Linux | dash | rc.112 + our close-wait patch | 159 | both |
| macOS | bash (execs) | rc.112 as published | 163 | both |

`ps` inside the container confirms the topology: `spawn().pid` is the `sh` process and the node process is its child; on macOS they are the same pid.

## Expected

Scoped release and `kill` should not complete until the process group (or at least everything holding the leader's inherited stdio) has stopped, and `forceKillAfter` should keep escalating until then. Public `exitCode` / `isRunning` can keep Node's `exit` semantics.

## What we run with today

A pnpm patch that keeps `exitSignal` for the public API but makes the acquireRelease finalizer and `kill` await the `close` event (inherited stdio released) instead of `exit`. That covers the shell-leader case (the shell's stdout pipe is inherited by the real process). It is a proxy, though: it only works when stdio is piped, and a surviving descendant in another process group that still holds the pipe makes `close` never fire. Polling the process group (`kill(-pgid, 0)`) within the `forceKillAfter` budget, then SIGKILL-ing the group, would not have that coupling. Happy to open a PR either way if you tell me which shape you want.

<details>
<summary>Our patch (the <code>src/</code> hunks; <code>dist/</code> mirrors them)</summary>

```diff
diff --git a/src/NodeChildProcessSpawner.ts b/src/NodeChildProcessSpawner.ts
--- a/src/NodeChildProcessSpawner.ts
+++ b/src/NodeChildProcessSpawner.ts
@@ -65,6 +65,7 @@ const toPlatformError = (
 
 type ExitCodeWithSignal = readonly [code: number | null, signal: NodeJS.Signals | null]
 type ExitSignal = Deferred.Deferred<ExitCodeWithSignal>
+type CloseSignal = Deferred.Deferred<ExitCodeWithSignal>
 
 const taskkill = (
   childProcess: NodeChildProcess.ChildProcess,
@@ -336,10 +337,11 @@ const make = Effect.gen(function*() {
     spawnOptions: NodeChildProcess.SpawnOptions
   ) =>
     Effect.callback<
-      readonly [NodeChildProcess.ChildProcess, ExitSignal],
+      readonly [NodeChildProcess.ChildProcess, ExitSignal, CloseSignal],
       PlatformError.PlatformError
     >((resume) => {
-      const deferred = Deferred.makeUnsafe<ExitCodeWithSignal>()
+      const exitSignal = Deferred.makeUnsafe<ExitCodeWithSignal>()
+      const closeSignal = Deferred.makeUnsafe<ExitCodeWithSignal>()
       const handle = NodeChildProcess.spawn(
         command.command,
         command.args,
@@ -349,10 +351,13 @@ const make = Effect.gen(function*() {
         resume(Effect.fail(toPlatformError("spawn", error, command)))
       })
       handle.on("exit", (...args) => {
-        Deferred.doneUnsafe(deferred, Exit.succeed(args))
+        Deferred.doneUnsafe(exitSignal, Exit.succeed(args))
+      })
+      handle.on("close", (...args) => {
+        Deferred.doneUnsafe(closeSignal, Exit.succeed(args))
       })
       handle.on("spawn", () => {
-        resume(Effect.succeed([handle, deferred]))
+        resume(Effect.succeed([handle, exitSignal, closeSignal]))
       })
       return Effect.sync(() => {
         handle.kill("SIGTERM")
@@ -481,12 +486,12 @@ const make = Effect.gen(function*() {
         const env = resolveEnvironment(cmd.options)
         const stdio = buildStdioArray(stdinConfig, stdoutConfig, stderrConfig, resolvedAdditionalFds)
 
-        const [childProcess, exitSignal] = yield* Effect.acquireRelease(
+        const [childProcess, exitSignal, closeSignal] = yield* Effect.acquireRelease(
           spawn(cmd, buildSpawnOptions(cmd.options, { cwd, env, stdio }, process.platform)),
-          Effect.fnUntraced(function*([childProcess, exitSignal]) {
-            const exited = yield* Deferred.isDone(exitSignal)
+          Effect.fnUntraced(function*([childProcess, exitSignal, closeSignal]) {
+            const closed = yield* Deferred.isDone(closeSignal)
             const killWithTimeout = withTimeout(childProcess, cmd, cmd.options)
-            if (exited) {
+            if (closed) {
-              // Process already exited, check if children need cleanup
-              const [code] = yield* Deferred.await(exitSignal)
+              // Process and inherited stdio already closed; check whether the process group needs cleanup
+              const [code] = yield* Deferred.await(closeSignal)
               if (code !== 0 && Predicate.isNotNull(code)) {
@@ -501,8 +506,8 @@ const make = Effect.gen(function*() {
-            // Process is still running, kill it
+            // Process or inherited stdio is still open; terminate the process group
             return yield* killWithTimeout((command, childProcess, signal) =>
               killProcessGroup(command, childProcess, signal).pipe(
                 Effect.catch(() => killProcess(command, childProcess, signal)),
-                Effect.andThen(Deferred.await(exitSignal))
+                Effect.andThen(Deferred.await(closeSignal))
               )
             ).pipe(
               Effect.ignore
@@ -548,7 +553,7 @@ const make = Effect.gen(function*() {
           return killWithTimeout((command, childProcess, signal) =>
             killProcessGroup(command, childProcess, signal).pipe(
               Effect.catch(() => killProcess(command, childProcess, signal)),
-              Effect.andThen(Deferred.await(exitSignal))
+              Effect.andThen(Deferred.await(closeSignal))
             )
           ).pipe(
             Effect.asVoid
```

</details>

Related: #6411 (force-kill escalation) fixed the timeout scope but is still leader-bound; #5592 (children surviving parent SIGINT) is the same family.

xia-chao:
## Summary

`NodeChildProcessSpawner` already signals the process group on scoped release and `kill`, but it waited on the leader's `exit`. On Linux, `shell: true` uses dash (`sh -c`), which does not `exec` the command. SIGTERM kills the shell immediately, `exit` fires, `forceKillAfter` is dropped, and descendants keep running.

Release and `kill` now wait for the Node `close` event (inherited stdio released). Public `exitCode` / `isRunning` still use `exit`.

Fixes #7821

## Test plan

- [x] `vitest --run packages/platform/node-shared/test/NodeChildProcessSpawner.test.ts`
- [x] Linux `shell: true` scope close does not return until the wrapper and its inherited-stdio child have printed `*:SIGTERM`
- [x] macOS (bash execs) still completes the same wait
- [x] `exitCode` / `isRunning` remain bound to the leader `exit` event

### DIFF
--- a/.changeset/childprocess-close-wait.md
+++ b/.changeset/childprocess-close-wait.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform-node-shared": patch
+---
+
+Wait for the spawned process `close` event (inherited stdio released) in scoped release and `kill`, so a non-exec'ing shell leader cannot abandon descendants.
--- a/packages/platform/node-shared/src/NodeChildProcessSpawner.ts
+++ b/packages/platform/node-shared/src/NodeChildProcessSpawner.ts
@@ -65,6 +65,7 @@ const toPlatformError = (
 
 type ExitCodeWithSignal = readonly [code: number | null, signal: NodeJS.Signals | null]
 type ExitSignal = Deferred.Deferred<ExitCodeWithSignal>
+type CloseSignal = Deferred.Deferred<ExitCodeWithSignal>
 
 const taskkill = (
   childProcess: NodeChildProcess.ChildProcess,
@@ -336,10 +337,11 @@ const make = Effect.gen(function*() {
     spawnOptions: NodeChildProcess.SpawnOptions
   ) =>
     Effect.callback<
-      readonly [NodeChildProcess.ChildProcess, ExitSignal],
+      readonly [NodeChildProcess.ChildProcess, ExitSignal, CloseSignal],
       PlatformError.PlatformError
     >((resume) => {
-      const deferred = Deferred.makeUnsafe<ExitCodeWithSignal>()
+      const exitSignal = Deferred.makeUnsafe<ExitCodeWithSignal>()
+      const closeSignal = Deferred.makeUnsafe<ExitCodeWithSignal>()
       const handle = NodeChildProcess.spawn(
         command.command,
         command.args,
@@ -349,10 +351,13 @@ const make = Effect.gen(function*() {
         resume(Effect.fail(toPlatformError("spawn", error, command)))
       })
       handle.on("exit", (...args) => {
-        Deferred.doneUnsafe(deferred, Exit.succeed(args))
+        Deferred.doneUnsafe(exitSignal, Exit.succeed(args))
+      })
+      handle.on("close", (...args) => {
+        Deferred.doneUnsafe(closeSignal, Exit.succeed(args))
       })
       handle.on("spawn", () => {
-        resume(Effect.succeed([handle, deferred]))
+        resume(Effect.succeed([handle, exitSignal, closeSignal]))
       })
       return Effect.sync(() => {
         handle.kill("SIGTERM")
@@ -481,14 +486,14 @@ const make = Effect.gen(function*() {
         const env = resolveEnvironment(cmd.options)
         const stdio = buildStdioArray(stdinConfig, stdoutConfig, stderrConfig, resolvedAdditionalFds)
 
-        const [childProcess, exitSignal] = yield* Effect.acquireRelease(
+        const [childProcess, exitSignal, closeSignal] = yield* Effect.acquireRelease(
           spawn(cmd, buildSpawnOptions(cmd.options, { cwd, env, stdio }, process.platform)),
-          Effect.fnUntraced(function*([childProcess, exitSignal]) {
-            const exited = yield* Deferred.isDone(exitSignal)
+          Effect.fnUntraced(function*([childProcess, , closeSignal]) {
+            const closed = yield* Deferred.isDone(closeSignal)
             const killWithTimeout = withTimeout(childProcess, cmd, cmd.options)
-            if (exited) {
-              // Process already exited, check if children need cleanup
-              const [code] = yield* Deferred.await(exitSignal)
+            if (closed) {
+              // Process and inherited stdio already closed; check whether the process group needs cleanup
+              const [code] = yield* Deferred.await(closeSignal)
               if (code !== 0 && Predicate.isNotNull(code)) {
                 // Non-zero exit code ,attempt to clean up process group
                 return yield* Effect.ignore(killWithTimeout(killProcessGroup))
@@ -498,11 +503,11 @@ const make = Effect.gen(function*() {
             if (!isReferenced) {
               return yield* Effect.void
             }
-            // Process is still running, kill it
+            // Process or inherited stdio is still open; terminate the process group
             return yield* killWithTimeout((command, childProcess, signal) =>
               killProcessGroup(command, childProcess, signal).pipe(
                 Effect.catch(() => killProcess(command, childProcess, signal)),
-                Effect.andThen(Deferred.await(exitSignal))
+                Effect.andThen(Deferred.await(closeSignal))
               )
             ).pipe(
               Effect.ignore
@@ -548,7 +553,7 @@ const make = Effect.gen(function*() {
           return killWithTimeout((command, childProcess, signal) =>
             killProcessGroup(command, childProcess, signal).pipe(
               Effect.catch(() => killProcess(command, childProcess, signal)),
-              Effect.andThen(Deferred.await(exitSignal))
+              Effect.andThen(Deferred.await(closeSignal))
             )
           ).pipe(
             Effect.asVoid
--- a/packages/platform/node-shared/test/NodeChildProcessSpawner.test.ts
+++ b/packages/platform/node-shared/test/NodeChildProcessSpawner.test.ts
@@ -5,8 +5,12 @@ import * as NodePath from "@effect/platform-node-shared/NodePath"
 import { assert, describe, it } from "@effect/vitest"
 import * as ChildProcessSpawnerTest from "effect-test/unstable/process/ChildProcessSpawnerTest"
 import * as Effect from "effect/Effect"
+import * as Exit from "effect/Exit"
 import * as FileSystem from "effect/FileSystem"
 import * as Layer from "effect/Layer"
+import * as Schedule from "effect/Schedule"
+import * as Scope from "effect/Scope"
+import * as Stream from "effect/Stream"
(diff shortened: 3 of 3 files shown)


tim-smart:
Already fixed on main

PR #7877, opened 2026-09-03
fix(Arbitrary): encode transformed template literal parts

### LINKED ISSUE #7876: Arbitrary template generation stringifies decoded transformed parts

Effect v4 source at `a2c1ce62f04214d98d848f4cfe33d677e1a87d39`, using `Arbitrary.schema`.

Sample `Schema.TemplateLiteral([Schema.BooleanFromBit])` with count 1, size 0, maxDiscards 0 and seed 0. Validate the returned string with `Schema.is` for that same template. Compare with a template containing plain literal bits.

[Regression tests](https://github.com/kitlangton/effect/blob/audit/c2-r3-arbitrary-template-transformed-part-stringification/packages/effect/test/unstable/arbitrary/R3Template.test.ts) exercise public sampling and shrinking.

The generated string satisfies the template: `"0"` or `"1"`, not a decoded boolean word.

### What do you see instead?

The sample is `["true"]`, which the schema rejects. Generation converts decoded parts directly to strings rather than encoding their template spelling.

### Additional information

The correction must retain decoded-part checks and validate both root and shrink spellings. Simply generating unchecked encoded parts would lose those constraints.

## Audit

kitlangton:
## Why

Template arbitrary generation stringified decoded transformed values. For `BooleanFromBit`, it produced invalid `"true"` or `"false"` strings instead of the template's encoded bit spelling.

## What changed

Encode each generated tuple through the template parser's reverse direction, then validate the resulting string. The same validation applies to generated roots and automatic shrinks. Focused regressions now live in the existing Arbitrary suite.

## Scope

Transformation callbacks now run during template generation. Failed encodings or rejected spellings consume the existing discard budget, while defects still propagate.

## Verification

```bash
pnpm lint-fix
pnpm test --run packages/effect/test/unstable/arbitrary/Arbitrary.test.ts --maxWorkers=1 --no-file-parallelism
pnpm check
git diff --check origin/main...HEAD
```

An independent probe on `main` generated `["true"]` for `Schema.TemplateLiteral([Schema.BooleanFromBit])`, and `Schema.is` rejected it. The original two-file selection also reproduced the claimed 148 passed / 7 failed on publication base `a2c1ce62f04214d98d848f4cfe33d677e1a87d39` and 155 passed with the initial patch. After consolidating the regression coverage, the final Arbitrary suite has 149 passing tests.

Closes #7876
Closes EFF-1178

## Audit


gcanti:
Closing in favor of https://github.com/Effect-TS/effect/pull/8094

PR #7879, opened 2026-09-03
fix(Arbitrary): validate indexed object generation and shrinking

### LINKED ISSUE #7878: Arbitrary object generation can violate applicable index constraints

Effect v4 source at `a2c1ce62f04214d98d848f4cfe33d677e1a87d39`, using schema-derived Arbitrary generation.

1. Create a `StructWithRest` with fixed field `value: Schema.String` and a string-keyed `Schema.NonEmptyString` record constraint.
2. Confirm the schema rejects `{ value: "" }` and accepts `{ value: "x" }`.
3. Sample with count 1, size 0, maxDiscards 0 and seed 0.

[Regression tests](https://github.com/kitlangton/effect/blob/audit/c2-r3-arbitrary-struct-rest-skips-fixed-value-index-validation/packages/effect/test/unstable/arbitrary/R3StructRest.test.ts) distinguish output validity from generation productivity.

Every emitted value satisfies all applicable field and index constraints. If no valid candidate is produced within the discard budget, report the existing typed exhaustion result instead of publishing an invalid value.

### What do you see instead?

Sampling succeeds with `[{ value: "" }]`, despite the same schema rejecting that value.

### Additional information

The schema is inhabited, but this report does not require constructive intersection generation. A validity guard can exhaust at size zero; nonzero direct samples and an explicit fixed-field NonEmptyString control distinguish that limit from blanket rejection.

## Audit

kitlangton:
## Why

Schema-derived arbitraries can emit objects whose fixed fields violate an applicable index signature. For example, a `String` field combined with a `NonEmptyString` record emits an empty value that `Schema.is` rejects.

## What changes

Validate generated objects and automatic shrink candidates against an index-signature-only object AST. This checks fixed fields and overlapping indexes without reparsing named properties, preserves generated object identity and prototypes, and leaves objects without index signatures on the existing path. Plain records with one index signature also stay on the existing path because their keys and values already come from that signature's generators.

## Scope

This prevents invalid output but does not construct productive intersections. A size-zero `String` and `NonEmptyString` intersection can exhaust its discard budget. That matches the existing behavior for inhabited residual filters and avoids changing generation distributions or replay semantics.

Regression coverage now lives in the existing `Arbitrary.test.ts` suite and checks generation, automatic shrinking, exact size-zero exhaustion, and overlapping indexes.

## Verification

```bash
pnpm lint-fix
pnpm test --run packages/effect/test/unstable/arbitrary/Arbitrary.test.ts --maxWorkers=1 --no-file-parallelism
pnpm check
git diff --check origin/main...HEAD
```

The independent regression failed on `main` with 147 passing tests and only the new validity assertion failing. The updated branch passes all 148 tests in the focused suite.

Closes #7878
Closes EFF-1179

## Audit


gcanti:
Closing in favor of https://github.com/Effect-TS/effect/pull/8095

PR #7906, opened 2026-09-03
fix(Sharding): preserve explicit registration service overrides

### LINKED ISSUE #7905: Sharding construction context shadows explicit entity registration services

Effect v4 source at `7e74c58bed55f3d90a3a6fc1c863fd2bec87e981`, using entity registration in Sharding.

1. Construct a real in-memory Sharding layer with an application service whose value is `"construction"`.
2. Register an entity builder under an explicit provider for the same service with value `"registration"`.
3. Invoke a handler that returns the builder's captured value. Repeat using `Entity.toLayer`.

[Registration regressions](https://github.com/kitlangton/effect/blob/audit/c2-r4-sharding-registration-services-shadowed-by-construction-context/packages/effect/test/cluster/ShardingRegistrationContext.test.ts) use ordinary public registration and clients.

The explicit registration provider wins for application services. Construction-only services can remain fallbacks; runner-owned internals retain their existing ownership.

### What do you see instead?

The handler returns `"construction"`. The broad construction context is overlaid onto the registration context before EntityManager captures it, shadowing the explicit registration value.

### Additional information

This is registration-time precedence, not first-request context capture. A correction must preserve constructor-only application/logger/tracer fallback, first-caller isolation and the runner's owned config, clock, reaper, generator and registration lifetime.

## Audit

kitlangton:
## Why

Sharding's construction context could shadow application services explicitly supplied while registering an entity. Handler builders then captured the construction-time service instead of the registration override.

## What changed

- Capture the context at entity registration and merge it over construction-time fallbacks.
- Reapply runner-owned configuration, clock, reaper, scope, and Snowflake generator after the merge.
- Add a focused regression to the existing Sharding cluster suite.
- Add an `effect` patch changeset.

## Verification

- `pnpm lint-fix`
- `packages/effect/test/cluster/Sharding.test.ts`: 54 tests passed
- `pnpm check`
- `git diff --check`

Closes #7905
Closes EFF-1169

## Audit


packages/effect/test/cluster/ShardingRegistrationContext.test.ts
```diff
@@ -0,0 +1,177 @@
+import { assert, describe, it } from "@effect/vitest"
```
tim-smart: All the tests files are wrong

tim-smart: Fixed in 983485f179. I removed the three standalone files and moved a focused regression into Sharding.test.ts using the shared TestSharding layer and it.effect convention.

PR #7907, opened 2026-09-03
add Effectable.Mixin

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds `Effectable.Mixin` to extend an existing concrete or abstract class with Effect behavior. Implement `asEffect()` on the subclass; its return type determines the success, error, and service types.

```ts
import { Effect, Effectable } from "effect"

class Box {
  constructor(readonly value: number) {}
  double() {
    return this.value * 2
  }
}

class EffectBox extends Effectable.Mixin(Box) {
  override asEffect() {
    return Effect.succeed(this.value)
  }
}

  const effectValue = yield* new EffectBox(1)
  //    ^? number
  return effectValue
})
```

Constructor parameters and base members are preserved. Effect's prototype methods shadow matching base prototype methods, including `pipe`, `toString`, `toJSON`, the iterator, and Node's inspect method.

## Validation

Effectable runtime tests, type tests (TypeScript 5.9 and 6.0), doctest, full typecheck, and lint pass. The JSDoc check reports existing errors in the unchanged `Multipart.ts` and `ChildProcess.ts` files.

## Related

[Discord question](https://discord.com/channels/795981131316985866/1544760617255313438)

Closes EFF-1299

### DIFF
--- a/packages/effect/src/Effectable.ts
+++ b/packages/effect/src/Effectable.ts
@@ -43,10 +43,10 @@ export const Prototype = <A extends Effect.Effect<any, any, any>>(options: {
     [evaluate]: options.evaluate
   }) as any
 
-const proto = Prototype({
+const proto = Prototype<Class<any, any, any>>({
   label: "Effectable",
   evaluate(_) {
-    return (this as any).override
+    return this.override
   }
 })
 
--- a/packages/effect/src/Effectable.ts
+++ b/packages/effect/src/Effectable.ts
@@ -46,7 +46,7 @@ export const Prototype = <A extends Effect.Effect<any, any, any>>(options: {
 const proto = Prototype<Class<any, any, any>>({
   label: "Effectable",
   evaluate(_) {
-    return this.override
+    return this.asEffect()
   }
 })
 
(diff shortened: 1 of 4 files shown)


packages/effect/src/Effectable.ts
```diff
@@ -60,9 +65,90 @@ const Base: new<A, E, R>() => Effect.Effect<A, E, R> = (() => {
...
+ * Effect.isEffect(box) // => true
+ * await Effect.runPromise(box) // => 2
+ * ```
+ *
+ * @see {@link Prototype} for a lower-level primitive approach to creating custom Effect-like values without a class
+ * @see {@link Class} for a base constructor to extend
+ * @category constructors
+ * @since 4.0.0
+ */
+export const Mixin = <TBase extends new(...args: ReadonlyArray<any>) => any>(
```
tim-smart:
```suggestion
export const Mixin = <TBase extends abstract new(...args: ReadonlyArray<any>) => object>(
```

PR #7915, opened 2026-09-03
fix(Effectable): evaluate Class with asEffect

### LINKED ISSUE #7914: Effectable.Class never evaluates its required override

Effect v4 source at `7e74c58bed55f3d90a3a6fc1c863fd2bec87e981`, using `Effectable.Class`.

Extend the public class with a finite successful `override` effect. Run the instance and compare it with running that override directly or using a finite `Effectable.Prototype` evaluator. A getter-backed override also makes it observable whether delegation occurs.

[Regression tests](https://github.com/kitlangton/effect/blob/audit/c2-r5-effectable-class-ignores-override-effect/packages/effect/test/Effectable.r5.test.ts) use finite fork/snapshot/interruption cleanup, not an unbounded synchronous runner on the broken class.

The custom Effect evaluates its required override and produces the delegate's success or failure.

### What do you see instead?

The base evaluator returns the wrapper itself repeatedly and never reads the override. Direct delegate and prototype controls terminate normally.

## Audit

kitlangton:
## Why

`Effectable.Class` currently evaluates by returning its own instance to the interpreter. Subclasses loop instead of producing their success or failure result.

## What changes

Define `asEffect()` as the class interface and call it whenever an `Effectable.Class` instance is evaluated. The method runs on the actual subclass receiver for each execution.

## Scope

One evaluator call, the abstract class method, two focused cases in the existing Effect test suite and an `effect` patch changeset. Public class generics and the lower-level `Prototype` API remain unchanged. The returned effect is not cached, so `asEffect()` can use current instance state and provided services.

## Verification

```bash
pnpm test --run packages/effect/test/Effect.test.ts --maxWorkers=1 --no-file-parallelism --sequence.concurrent=false
pnpm check
pnpm exec oxlint packages/effect/src/Effectable.ts packages/effect/test/Effect.test.ts
pnpm exec dprint check packages/effect/src/Effectable.ts packages/effect/test/Effect.test.ts .changeset/effectable-class-override-delegation.md
git diff --check
```

The regression cases cover direct class evaluation and lazy, receiver-bound `asEffect()` calls across repeated runs. All 245 tests in `Effect.test.ts` pass on `56446afbe2`, along with the root typecheck, targeted lint and formatting checks, and whitespace checks.

Closes #7914
Closes EFF-1170

## Audit


packages/effect/src/Effectable.ts
tim-smart:
```suggestion
  abstract asEffect(): Effect.Effect<A, E, R>
```

I think this is a better interface.

tim-smart: Implemented in `183cf48b66`: `Effectable.Class` now declares `asEffect()`, and evaluation calls `this.asEffect()` on every run. I also updated the tests, changeset, and PR text. The focused suite and checks pass.

PR #7927, opened 2026-09-03
feat(http): reduce web handler cold start on Cloudflare

tim-smart:
Two of the cold-start improvements from the Cloudflare investigation.

## Changes

- `HttpServerRespondable` still calls `Schema.isSchemaError`. The predicate and type id now live in a small internal leaf shared with the public `Schema.isSchemaError` API, so the HTTP path does not retain the main Schema module graph. A minimal `HttpRouter` worker bundled from the published output stays at 283 KB (41 KB min+gzip), with no `Schema*` or `internal/schema/*` modules retained, and keeps the roughly 3 ms workerd cold-start improvement.
- `HttpRouter.toWebHandler`, `HttpEffect.toWebHandlerLayer` and `HttpEffect.toWebHandlerLayerWith` build the layer when the handler is created instead of on the first request. On Cloudflare the module is evaluated before the first request, so the layer and router construction (about 2.5 ms) move off the request path. A failed build never surfaces as an unhandled rejection at startup; every request rejects with the build error, which Workers turns into a 500 and logs. A layer doing async work while building may still be in progress on the first request, which then waits; the docs say so.
- Guard for the Schema removal: `packages/tools/bundle/test/HttpRouterEsbuild.test.ts` bundles `fixtures/http-router.ts` with esbuild using worker conditions and asserts that no `Schema*` or `internal/schema/*` module is retained. The rollup based Bundle job cannot see this regression because rollup lifts single exports out of `Schema.js` (21 bytes of movement), while wrangler's esbuild drags the whole graph in. The test reproduces the `annotate-pure-calls` build step per Effect module so it shakes like the published package. Proven on demand: importing the full `Schema.ts` module for the same `Schema.isSchemaError` call makes the guard fail with `Schema.ts`, `SchemaAST.ts`, `SchemaParser.ts`, `SchemaIssue.ts` and six more retained modules; using the shared leaf passes.
- `packages/tools/bundle/fixtures/http-router.ts` stays as a size trend for the router and catches Schema regressions that reach the parser (about +10 % in the Bundle job).
- Tests cover a `SchemaError` mapping to 400 through `toResponseOrElse`, eager layer construction without rebuilding on the first request, and repeated request failures without an unhandled rejection when layer construction fails.

Layers that do I/O during construction (fetch, sleep, random values) still cannot be built in the global scope of a Worker; they need to be turned into a handler inside `fetch`, which is also where bindings become available.

## Validation

- `pnpm lint-fix`
- `pnpm check`
- `pnpm jsdocs --check`
- `pnpm test --run packages/tools/bundle/test/HttpRouterEsbuild.test.ts packages/effect/test/unstable/http/HttpServerRespondable.test.ts packages/effect/test/schema/Schema.test.ts` (590 tests passed)
- Rebuilt `dist` and ran the router in `workerd`: the layer is observed to be built before the first request, and a layer that fails on purpose gives a 500 on every request with the error logged, with no unhandled rejection at startup under `node --unhandled-rejections=strict`

Closes EFF-1156

### DIFF
--- a/packages/effect/src/unstable/http/HttpServerRespondable.ts
+++ b/packages/effect/src/unstable/http/HttpServerRespondable.ts
@@ -11,7 +11,6 @@
 import * as Cause from "../../Cause.ts"
 import * as Effect from "../../Effect.ts"
 import { hasProperty } from "../../Predicate.ts"
-import * as Schema from "../../Schema.ts"
 import type { HttpServerResponse } from "./HttpServerResponse.ts"
 import * as Response from "./HttpServerResponse.ts"
 
@@ -50,6 +49,12 @@ export const isRespondable = (u: unknown): u is Respondable => hasProperty(u, sy
 const badRequest = Response.empty({ status: 400 })
 const notFound = Response.empty({ status: 404 })
 
+// Checked by type id instead of `Schema.isSchemaError` so that this module does
+// not pull the Schema modules into bundles that never use them.
+const SchemaErrorTypeId = "~effect/Schema/SchemaError"
+const isSchemaError = (u: unknown): boolean =>
+  hasProperty(u, SchemaErrorTypeId) && u[SchemaErrorTypeId] === SchemaErrorTypeId
+
 /**
  * Converts a `Respondable` value into an `HttpServerResponse`.
  *
@@ -86,7 +91,7 @@ export const toResponseOrElse = (u: unknown, orElse: HttpServerResponse): Effect
   } else if (isRespondable(u)) {
     return Effect.catchCause(u[symbol](), () => Effect.succeed(orElse))
     // add support for some commmon types
-  } else if (Schema.isSchemaError(u)) {
+  } else if (isSchemaError(u)) {
     return Effect.succeed(badRequest)
   } else if (Cause.isNoSuchElementError(u)) {
     return Effect.succeed(notFound)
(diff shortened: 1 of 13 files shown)


packages/effect/src/unstable/http/HttpServerRespondable.ts
```diff
@@ -86,7 +91,7 @@ export const toResponseOrElse = (u: unknown, orElse: HttpServerResponse): Effect
   } else if (isRespondable(u)) {
     return Effect.catchCause(u[symbol](), () => Effect.succeed(orElse))
     // add support for some commmon types
-  } else if (Schema.isSchemaError(u)) {
```
tim-smart, line 89: I think keep using Schema.isSchemaError

tim-smart: Addressed in 92422b0332. `HttpServerRespondable` still calls `Schema.isSchemaError`; the predicate and type id now live in a small internal leaf shared with the public `Schema.isSchemaError` API, so the router path does not retain the main Schema graph.

I used the existing esbuild guard as the failing test. Pointing the call at the full `Schema.ts` module failed with ten retained Schema modules; pointing it at the shared leaf passes the unchanged guard. `pnpm check`, `pnpm jsdocs --check`, and the targeted bundle/HTTP/Schema tests all pass (590 tests).

PR #7929, opened 2026-09-03
Optimize OtlpTracer span creation and export

tim-smart:
Closes EFF-1157

## Changes

- `OtlpTracer` spans are now a class with lazy `traceId`, `spanId`, `attributes`, and `events`, matching `Tracer.NativeSpan`. Unsampled spans no longer generate identifiers or allocate a `Map` and event array.
- `makeOtlpSpan` avoids the `Option.match` closure allocation and the `.map` calls on empty event and link arrays.
- `Encoding.randomHex` builds 16 and 32 character identifiers with a single `String.fromCharCode` call. This roughly halves generation cost and produces flat strings, so `JSON.stringify` no longer flattens rope strings for every span.

## Measurements

`pnpm runtimeperf-compare otlp-tracer` with the benchmark guard from #7918 applied locally (base = `main`, head = this branch):

| fixture | base | head | delta |
| --- | --- | --- | --- |
| otlp-tracer/unsampled-span | 232 ns | 6.5 ns | -97% |
| otlp-tracer/sampled-span-batch-100 | 81.65 µs | 46.91 µs | -43% |

Remaining cost in the sampled batch is roughly 20 µs `JSON.stringify`, 16 µs span creation and conversion, and 8 µs per-flush HTTP client overhead.

## Validation

- `pnpm lint-fix`
- `pnpm test --run` for `Tracer`, `OtlpExporter`, `OtlpSerialization`, and `Encoding` tests
- `pnpm check`

### DIFF
--- a/.changeset/otlp-tracer-span-performance.md
+++ b/.changeset/otlp-tracer-span-performance.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Speed up `OtlpTracer` span creation and export. Spans now allocate identifiers, attributes, and events lazily, and `Encoding.randomHex` produces flat strings for 16 and 32 character identifiers so serialization no longer flattens ropes.
--- a/packages/effect/src/Encoding.ts
+++ b/packages/effect/src/Encoding.ts
@@ -424,13 +424,107 @@ export const encodeHex: (input: Uint8Array | string) => string = (input) =>
  * @since 4.0.0
  */
 export const randomHex = (length: number): string => {
-  let result = ""
-  for (let i = length >>> 3; i > 0; i--) {
-    const word = (Math.random() * 0x100000000) >>> 0
-    result += byteToHex[word >>> 24] + byteToHex[(word >>> 16) & 0xff] + byteToHex[(word >>> 8) & 0xff] +
-      byteToHex[word & 0xff]
+  switch (length) {
+    case 16:
+      return randomHex16()
+    case 32:
+      return randomHex32()
+    default: {
+      let result = ""
+      for (let i = length >>> 3; i > 0; i--) {
+        result += randomHex8()
+      }
+      return result
+    }
   }
-  return result
+}
+
+const hexCharCodes = new Uint8Array(16)
+for (let i = 0; i < 16; i++) {
+  hexCharCodes[i] = "0123456789abcdef".charCodeAt(i)
+}
+
+const randomWord = (): number => (Math.random() * 0x100000000) >>> 0
+
+// Trace and span identifiers are the common lengths. A single
+// String.fromCharCode call produces a flat string, which avoids rope
+// flattening when the identifier is later serialized.
+const randomHex8 = (): string => {
+  const a = randomWord()
+  return String.fromCharCode(
+    hexCharCodes[a >>> 28],
+    hexCharCodes[(a >>> 24) & 15],
+    hexCharCodes[(a >>> 20) & 15],
+    hexCharCodes[(a >>> 16) & 15],
+    hexCharCodes[(a >>> 12) & 15],
+    hexCharCodes[(a >>> 8) & 15],
+    hexCharCodes[(a >>> 4) & 15],
+    hexCharCodes[a & 15]
+  )
+}
+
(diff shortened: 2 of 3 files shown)


tim-smart:
## Equivalence audit (Code Auditor)

I audited this PR for behavioural equivalence against `main`, per the request. No correctness, security, or concurrency defects found. Details below.

### `Encoding.randomHex`

I verified equivalence empirically, not just by reading: I stubbed `Math.random` with a seeded PRNG and ran the old and new implementations on the same random stream. Outputs are byte-identical across 24 length inputs (0, non-multiples of 8, fractional, `NaN`, `±Infinity`, `2^32` boundary) times 4 seeds. Since both consume one word per 8 characters in the same order and map nibbles MSB-first, any transcription or ordering mistake would have shown up as a mismatch.

- Character set unchanged: old `byteToHex` (`i.toString(16).padStart(2, "0")`) and new `hexCharCodes` (`"0123456789abcdef"`) are both lowercase `[0-9a-f]`.
- Distribution: chi-square per nibble position on 200k samples for both 16- and 32-char outputs is uniform (worst 22.6, df=15, critical value at p<1e-6 is 54.7).
- Length semantics preserved: `length >>> 3` round-down behaviour is identical, including `NaN`/`Infinity`/`2^32` coercion. The existing tests in `packages/effect/test/Encoding.test.ts:12-30` cover this and pass.

### Lazy `traceId`/`spanId`/`attributes`/`events`

The lazy getters mirror the pre-existing `Tracer.NativeSpan` pattern (`packages/effect/src/Tracer.ts:707`), so this aligns OtlpTracer with established span semantics rather than inventing a new one. I verified with tests against the PR branch:

- Unsampled spans generate nothing and are never exported (`packages/effect/src/unstable/observability/OtlpTracer.ts:92` checks `sampled` first). `_traceId`/`_spanId`/`_attributes`/`_events` stay `undefined` after end.
- Reading an unsampled span yields stable, valid ids (memoized via `??=`), so consumers like `OtlpLogger` (`packages/effect/src/unstable/observability/OtlpLogger.ts:231`) see the same values as before, materialized on first read.
- Ids read before export match the exported payload exactly; re-reads after export return the same value.
- Child of unsampled parent: parent ids materialize when the sampled child is exported, traceId is shared, `parentSpanId` equals the parent's memoized spanId. Sampled parent + child: exported `child.parentSpanId === parent.spanId` regardless of end order.
- Attributes/events: writes go through the getters, so the map/array materializes on first write and captures everything, including writes after an early read.
- Runtime contract confirmed: the tracer always receives a fresh `links` array per span (`packages/effect/src/internal/effect.ts:5915`), so `addLinks` mutation is per-span, and initial `attributes` are applied via `span.attribute()` after creation.

### `TraceData` shape

Differential test: I ran the same seven span scenarios (rich root with links/events/attributes, bare root, sampled child of unsampled parent, sampled parent+child pair, interrupted, failed, plus an unsampled span that must not appear) against `main` and this branch, with a capturing `HttpClient`, then compared the exported JSON payloads with random ids and clock timestamps normalized. The payloads are identical. All 14 `OtlpSpan` fields are still present, empty `attributes`/`events`/`links` are still `[]` rather than omitted, `dropped*Count` still 0, and root `parentSpanId` is `undefined` in both (omitted by `JSON.stringify` either way). The protobuf path consumes the same structure, so it is unaffected.

`{ code: StatusCode.Ok }` replacing the shared `constOtelStatusSuccess` is output-identical (covered by the differential) and avoids sharing a mutable object, which is arguably safer.

### Verification run

- Targeted suites: Encoding, Tracer, OtlpSerialization, OtlpExporter, OtlpConfig — 51 tests pass.
- `pnpm check` clean for the changed files.
- The two hardening notes I have are pre-existing and out of scope here: with `maxBatchSize: "disabled"` the exporter's `runExport` skips the empty-buffer guard and never clears the buffer (`packages/effect/src/unstable/observability/OtlpExporter.ts:205`), so idle flushes post empty payloads and re-send prior items; and `randomHex` with negative lengths attempts a multi-GB string in both old and new code (documented coercion behaviour).

From an equivalence standpoint this is safe to merge.

PR #7945, opened 2026-09-03
fix(HttpClient): constrain catch recovery to HTTP responses

### LINKED ISSUE #7944: Data-first HttpClient.catch accepts non-response recovery values

Effect v4 source at `42fd969b45025eed6d60cb87b97c5ff88d187671`, using data-first `HttpClient.catch`.

Use a failing in-process client and recover with an effect producing a number or undefined. Compare data-first and data-last catch calls, and inspect the advertised `execute` result. No network request or runtime execution of an invalid recovery is required for the type proof.

[Type regressions](https://github.com/kitlangton/effect/blob/audit/c2-r5-httpclient-catch-data-first-response-constraint/packages/effect/typetest/unstable/http/HttpClient.catch.tst.ts) check genuine producer inference, public callability and valid explicit generic calls.

Both call forms require response-producing recovery effects, since the returned HttpClient still promises HTTP responses. Never-success recovery effects remain valid.

### What do you see instead?

The data-first overload accepts arbitrary success values through unconstrained `A2`, while its returned client still advertises response success. The data-last overload already rejects those callbacks.

## Audit

kitlangton:
## Why

The data-first `HttpClient.catch` overload permits recovery values that are not HTTP responses, even though the returned client's `execute` method still promises a response. The data-last form already enforces that contract.

## What changes

Match the data-first recovery result with the data-last overload by requiring `HttpClientResponse`. Callers that need to recover to arbitrary application values can use `Effect.catch` on `client.execute(request)`.

The existing `HttpClient` typetest suite now has two focused checks: it rejects a non-response data-first recovery and accepts a response recovery. This is a type-only fix, so no runtime test was added.

## Verification

On the actual merge base, `a9d1ee3d4d`, the new non-response rejection assertion fails. It passes with the one-line signature fix in this branch.

```bash
nix develop -c pnpm exec tstyche --target '*' packages/effect/typetest/unstable/http/HttpClient.tst.ts
nix develop -c pnpm check
nix develop -c pnpm lint
git diff --check
```

The focused typetest passes 12 tests and 10 assertions, with 2 expected diagnostics matched. The root typecheck, lint, and whitespace checks also pass.

Closes EFF-1194
Closes #7944

## Audit


packages/effect/src/unstable/http/HttpClient.ts
```diff
@@ -316,7 +316,7 @@ const catch_: {
   <E, E2, R2>(
     f: (e: E) => Effect.Effect<HttpClientResponse.HttpClientResponse, E2, R2>
   ): <R>(self: HttpClient.With<E, R>) => HttpClient.With<E2, R2 | R>
-  <E, R, A2, E2, R2>(
+  <E, R, A2 extends HttpClientResponse.HttpClientResponse, E2, R2>(
     self: HttpClient.With<E, R>,
     f: (e: E) => Effect.Effect<A2, E2, R2>
```
tim-smart: This should just match the data-last overload above

packages/effect/typetest/unstable/http/HttpClient.tst.ts
```diff
@@ -145,4 +175,162 @@ describe("HttpClient", () => {
       >()
     })
   })
+
+  describe("catch", () => {
+    it("rejects a number recovery in data-first form", () => {
```
tim-smart: Can we just 1-2 tests that prevent regression

PR #7954, opened 2026-09-03
Expose SSE decode options per HttpApiClient call

fitchmultz:
Generated `HttpApiClient` calls now accept native SSE decode options in their request object:

```ts
const stream = yield* client.users.stream({
  sseOptions: { maxEventSize: Infinity }
})
```

`sseOptions` is an unconditional optional field on `HttpApiEndpoint.ClientRequest`, alongside `responseMode`. All endpoint requests accept it; it affects SSE decoding only. `AtomHttpApi` queries and mutation requests accept the same field, and query cache keys include it so different decoding limits remain independent.

Omitted options retain native defaults and reuse the precompiled decoder. The API schema is unchanged.

Tests were updated in separate commits before implementation. Coverage includes concurrent calls, mixed buffered/SSE responses with `WithHeaders`, atom query cache isolation, per-mutation option forwarding, and request types with and without required inputs.

Local validation passed (all commands run through `nix develop -c`):

- `pnpm test --run packages/effect/test/unstable/httpapi/HttpApiClient.test.ts packages/effect/test/reactivity/AtomHttpApi.test.ts` (38 tests)
- `pnpm test-types HttpApiClient AtomHttpApi` (276 assertions across TypeScript 5.9.3 and 6.0.3)
- `pnpm check`
- `pnpm lint-fix` and `pnpm lint`

`pnpm jsdocs --check` reports six existing formatting diagnostics in `Multipart.ts` and `ChildProcess.ts`. Both files are unchanged from upstream; independent review confirmed the same diagnostics on clean upstream `main`.

Closes EFF-1272

### DIFF
--- a/packages/effect/src/unstable/httpapi/HttpApiClient.ts
+++ b/packages/effect/src/unstable/httpapi/HttpApiClient.ts
@@ -182,7 +182,8 @@ export declare namespace Client {
         Endpoint["~Query"],
         Endpoint["~Payload"],
         Endpoint["~Headers"],
-        Mode
+        Mode,
+        Endpoint["~Success"]
       >
     >
   ) => MethodReturn<Endpoint, E, R, Mode>
--- a/packages/effect/src/unstable/httpapi/HttpApiClient.ts
+++ b/packages/effect/src/unstable/httpapi/HttpApiClient.ts
@@ -333,8 +333,6 @@ export const makeClient = <ApiId extends string, Groups extends HttpApiGroup.Con
         const { group, endpoint, errors, successes } = onEndpointOptions
         const makeUrl = compilePath(endpoint.path)
         const decodeMap: Record<number | "orElse", ResponseDecoder> = { orElse: statusOrElse }
-        const decodeResponse: ResponseDecoder = (response, sseOptions) =>
-          (decodeMap[response.status] ?? decodeMap.orElse)(response, sseOptions)
         const errorAlternatives = new Map<number, Array<ResponseAlternative>>()
         for (const [status, schemas] of errors.entries()) {
           const grouped = groupSchemasByContentType(schemas)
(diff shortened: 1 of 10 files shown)


packages/effect/src/unstable/httpapi/HttpApiSchema.ts
```diff
@@ -274,6 +275,7 @@ export interface StreamSse<
   readonly mode: "sse"
   readonly sseMode: StreamSseMode
   readonly contentType: string
+  readonly decodeOptions: Sse.DecodeOptions | undefined
```
tim-smart: I think decode options should live on the client, not in the schema.

tim-smart (changes requested):

packages/effect/src/unstable/httpapi/HttpApiSchema.ts
fitchmultz: Agreed—moved this to `sseDecodeOptions` on `HttpApiClient.make`, `makeWith`, `group`, and `endpoint`, and removed the schema changes. The options pass unchanged to the native decoder; omitting them keeps its defaults. The regression test uses the same API declaration with separate limited and default clients, and the type test covers all four constructors.

tim-smart: What if you want to set them per endpoint? `client.users.stream({}, { sseOptions: ... })`

fitchmultz: Moved this to the individual call as you suggested: `client.users.stream({}, { sseOptions: { maxEventSize: Infinity } })`. The constructor options are removed, and the existing request argument and `responseMode` stay unchanged. Omitted options keep the native defaults and reuse the precompiled decoder. Tests cover concurrent call isolation and option forwarding through mixed/`WithHeaders` responses.

packages/effect/src/unstable/httpapi/HttpApiClient.ts
tim-smart, line 180: They could be added to the ClientRequest options instead.

tim-smart: Also the atom integration will need to be updated

tim-smart (changes requested):

tim-smart (approved):

PR #7987, opened 2026-09-04
fix(Multipart): add isStreamPart guard

### LINKED ISSUE #7986: Multipart.isPart accepts PersistedFile outside the Part union
## Problem

Multipart.isPart returns true for an authentic PersistedFile even though its predicate narrows to Field | File. A cast-free consumer can consequently enter a File branch whose content property is missing.

**Before:** A library-produced persisted file passes isPart and can be mistaken for a streamed File.

**After:** isPart accepts Field and File only; isPersistedFile still recognizes persisted files.

## Reproduction evidence

Base: `53eb88875dace386188435884037e4bc25109c37`. The identical selection goes from 2 failed / 21 passed to 23 passed, repeated, including all 16 existing Multipart tests. Tests use actual parsed parts and actual toPersisted output, with a no-op filesystem and a custom writer that drains owned bytes; no disk persistence is claimed. Both baseline and fixed fixtures compile under the existing public predicate signature.

The accompanying correction includes the frozen regression and passing-control fixtures:

```sh
pnpm test --run packages/effect/test/unstable/http/Multipart.test.ts packages/effect/test/unstable/http/Multipart.isPart.test.ts
```

## Audit

kitlangton:
## Why

`Multipart.isPart` is the broad brand guard shared by parsed and persisted multipart values. Consumers that need only parser output also need a guard that narrows to the streamed `Part` union (`Field | File`).

## What changes

- Restore `isPart` to its existing behavior for every branded multipart part, including `PersistedFile`.
- Add `isStreamPart`, which checks the brand and directly accepts only the `Field` and `File` tags.
- Cover both guards with real parsed and persisted values in the existing `Multipart.test.ts` suite.

The PR includes an `effect` patch changeset for the new public guard.

## Verification

```sh
pnpm lint-fix
pnpm test --run packages/effect/test/unstable/http/Multipart.test.ts
pnpm check
git diff --check
```

All 16 focused tests pass.

Closes EFF-1218
Closes #7986

## Audit


packages/effect/src/unstable/http/Multipart.ts
```diff
@@ -103,7 +103,9 @@ export interface Field extends Part.Proto {
  * @category guards
  * @since 4.0.0
  */
-export const isPart = (u: unknown): u is Part => Predicate.hasProperty(u, TypeId)
```
tim-smart, line 106: I think we keep `isPart` as it was, then add a `isStreamPart` that also checks the _tag directly.

tim-smart: Implemented in ac68c92b72: `isPart` is restored to the brand-only check, and the new `isStreamPart` guard checks the brand plus the `Field` / `File` tags directly. Tests, docs, the changeset, and PR wording now reflect the split; the focused suite, lint, and typecheck pass.

PR #7998, opened 2026-09-04
fix(platform): preserve multipart limit failures

### LINKED ISSUE #7996: @effect/platform 0.x: multipart limit fix from #6395 was never backported — maxFileSize/maxParts still silently ignored

`@effect/platform` 0.97.1 (latest, released 2026-07-30) with `effect` 3.22.1.

This is the same defect as #6392, which was fixed for Effect v4 by #6395 — that PR only touched `packages/effect/src/unstable/http/Multipart.ts`. The Effect 3 line (`packages/platform`, published as `@effect/platform` 0.x) never received the fix, and it is still being released: 0.97.1 shipped two weeks *after* #6395 was merged.

`packages/platform/src/Multipart.ts:335-340` is unchanged:

```ts
onError(error_) {
  exit = Option.some(Exit.fail(convertError(error_)))
},
onDone() {
  exit = Option.some(Exit.void)   // overwrites a captured failure
}
```

Runnable repro against `@effect/platform@0.97.1` — a 100-byte part under a 10-byte `maxFileSize`:

```js
import { Multipart } from "@effect/platform"
import { Chunk, Effect, Option, Stream } from "effect"

const boundary = "X"
const body =
    `--${boundary}\r\n` +
    `Content-Disposition: form-data; name="file"; filename="big.bin"\r\n` +
    `Content-Type: application/octet-stream\r\n\r\n` +
    "A".repeat(100) +
    `\r\n--${boundary}--\r\n`

const program = Stream.fromChunk(Chunk.of(new TextEncoder().encode(body))).pipe(
    Stream.pipeThroughChannel(
        Multipart.makeChannel({
            "content-type": `multipart/form-data; boundary=${boundary}`,
        }),
    ),
    Stream.runCollect,
    Multipart.withLimits({ maxFileSize: Option.some(10) }),
)

console.log((await Effect.runPromiseExit(program))._tag)
```

A `MultipartError` failure, as on v4 after #6395.

### What do you see instead?

`Success` — the part is delivered as if no limit had been declared. Splitting the body so the violation and the terminator arrive in separate chunks does not change it:

```
single chunk               -> Success (limit ignored)
violation in own chunk     -> Success (limit ignored)
```

### Additional information

`maxTotalSize` still trips, because the body-size check aborts in multipasta rather than only reporting; `maxParts` and `maxFileSize` do not. The complementary upstream half — multipasta reporting `maxPartSize`/`maxParts` via `onError` without returning, so bytes keep flowing — is filed as tim-smart/multipasta#44.

Practical impact for a consumer persisting parts as they arrive: a file over `maxFileSize` is written to disk and answered `200`. We measured a request 10 bytes over a 20 MB `maxFileSize` being stored successfully, which is what led us here; we now enforce the size in the handler instead of relying on the annotation.

Happy to open the backport PR if you would take one against the `@effect/platform` 0.x line — otherwise it would be good to know that these annotations are not enforced there, since they read as if they are.

xia-chao:
Fixes #7996.

`multipasta` invokes `onDone` after reporting a limit violation through `onError`. In the v3 multipart channel, `onDone` unconditionally replaced the captured `MultipartError` with successful completion, so `maxFileSize` and `maxParts` could be silently ignored.

Record normal completion only when no parser failure has already been captured.

Tests cover:
- `maxFileSize` with both a single input chunk and a boundary split after the violation
- `maxParts` with a three-part body and a limit of two

Verified with:
- `pnpm vitest run packages/platform/test/Multipart.test.ts`
- `pnpm --filter @effect/platform check`

### DIFF
--- a/.changeset/old-trees-worry.md
+++ b/.changeset/old-trees-worry.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform": patch
+---
+
+Fix multipart limit failures being silently ignored.
--- a/packages/platform/src/Multipart.ts
+++ b/packages/platform/src/Multipart.ts
@@ -336,7 +336,9 @@ export const makeChannel = <IE>(
           exit = Option.some(Exit.fail(convertError(error_)))
         },
         onDone() {
-          exit = Option.some(Exit.void)
+          if (Option.isNone(exit)) {
+            exit = Option.some(Exit.void)
+          }
         }
       })
 
--- a/packages/platform/test/Multipart.test.ts
+++ b/packages/platform/test/Multipart.test.ts
@@ -1,9 +1,63 @@
 import * as Multipart from "@effect/platform/Multipart"
 import { describe, test } from "@effect/vitest"
-import { deepStrictEqual } from "@effect/vitest/utils"
-import { Chunk, Effect, identity, pipe, Stream, Unify } from "effect"
+import { deepStrictEqual, strictEqual } from "@effect/vitest/utils"
+import { Chunk, Effect, identity, Option, pipe, Stream, Unify } from "effect"
 
 describe("Multipart", () => {
+  test("fails when maxFileSize is exceeded, including when the terminator follows in the same chunk", () =>
+    Effect.gen(function*() {
+      const boundary = "X"
+      const body = `--${boundary}\r\n` +
+        `Content-Disposition: form-data; name="file"; filename="big.bin"\r\n` +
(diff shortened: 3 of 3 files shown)


tim-smart (changes requested):
Missing a changeset

javascript-unsafe:
Added a patch changeset for `@effect/platform` in 2a59e169e.

tim-smart (approved):

PR #8014, opened 2026-09-04
fix(Effect): retain union-record errors and services

### LINKED ISSUE #8013: Effect.all omits errors and services from non-shared record keys
## Problem

Effect.all can infer a union of differently shaped records as infallible or service-free because it collects channels only from keys common to every branch. At runtime, the selected branch still executes its other entries, including their failures and service requirements.

**Before:** Branch-specific errors and required services disappear from the inferred result when the record keys differ.

**After:** One outer Effect retains every branch's possible errors and requirements, along with the original branch-specific success shapes.

## Reproduction evidence

Pinned base: `88593e4abca7065512214a0affd15fa401b0d64e`. Each exact TypeScript version goes from 13 failed / 6 passed canonical assertions to 19 passed. Controls cover shared, literal, index and empty records, result/discard modes, public utility types, generic/explicit calls, piping and service provision. Runtime preservation passes 3 new and 7 selected existing tests on both sides, with 240 existing cases excluded. Whole-module JavaScript is byte-identical under both exact TypeScript transforms and actual configured native emission. This is a type correction, not runtime red/green.

The accompanying correction includes frozen type assertions and runtime preservation controls:

```sh
pnpm test-types --target 5.9.3 packages/effect/typetest/Effect.allRecords.tst.ts
pnpm test-types --target 6.0.3 packages/effect/typetest/Effect.allRecords.tst.ts
pnpm test --run packages/effect/test/Effect.allRecords.test.ts
pnpm test --run packages/effect/test/Effect.test.ts -t '^Effect all '
```

## Audit

kitlangton:
## Why

`Effect.all` inferred unions of differently shaped records from keys shared by every branch. Errors and required services attached only to branch-specific keys were therefore omitted even though those effects still run.

## What changed

Record-value extraction now distributes across the input union before deriving error and service channels. The outer `Effect` and branch-specific success shapes remain unchanged. The regression assertion lives in the existing `Effect.tst.ts` suite, and runtime behavior is unchanged.

## Verification

```sh
pnpm lint-fix
pnpm test-types --target 5.9.3 packages/effect/typetest/Effect.tst.ts
pnpm test-types --target 6.0.3 packages/effect/typetest/Effect.tst.ts
pnpm check
```

Closes #8013
Closes EFF-1237

## Audit

Runtime correctness audit. Intended label: `audit`.


packages/effect/src/Effect.ts
```diff
@@ -315,6 +315,8 @@ export declare namespace All {
   > extends infer X ? X
     : never
 
+  type ObjectValues<T> = T extends unknown ? T[keyof T] : never
```
tim-smart: You can move this into the namespace at the bottom

PR #8016, opened 2026-09-04
fix(Effect): retain direct constructor error types

### LINKED ISSUE #8015: Direct try and tryPromise calls can infer an incorrect error channel from context
## Problem

A return annotation can make direct Effect.try or Effect.tryPromise calls claim an application error type, or no error at all. Their direct-function runtime branches do not perform that mapping: thrown or rejected values are wrapped in Cause.UnknownError.

**Before:** A direct constructor can typecheck under an Effect<number, LoadError> return annotation without any error mapper.

**After:** Direct forms retain UnknownError; mapped options infer the actual catch result; union inputs include the direct branch's UnknownError.

## Reproduction evidence

Pinned base: `88593e4abca7065512214a0affd15fa401b0d64e`. Each exact TypeScript version goes from 12 failed / 35 passed tests to 47 passed, with 45 assertions and six matched negative directives. Baseline evaluates 41 assertions and has six intentional unused directives. Controls preserve mapped, partial/explicit generic, generic-wrapper, union and signal typing. Runtime preservation passes 250 tests before and after, with a passing fixed repeat. Whole-module JavaScript is byte-identical; original same-base baseline/parity evidence was retained, not described as a maintained rerun.

The accompanying correction includes frozen regression and passing-control fixtures:

```sh
pnpm test-types packages/effect/typetest/Effect.tryDirect.tst.ts --target 5.9.3
pnpm test-types packages/effect/typetest/Effect.tryDirect.tst.ts --target 6.0.3
pnpm test --run packages/effect/test/Effect.tryDirect.test.ts packages/effect/test/Effect.test.ts
```

## Audit

kitlangton:
Direct `Effect.try` and `Effect.tryPromise` calls could claim an application error through a return annotation even though their runtime failures are `Cause.UnknownError`.

Separate direct and mapped overloads now preserve `UnknownError` for direct calls and infer `E` from `catch` for mapped calls. The mapped overloads have no error-type default. Explicit two-generic direct calls, union-valued arguments, and generic aliases that erase `UnknownError` are rejected. Runtime behavior is unchanged.

Regression coverage is in the existing `Effect.tst.ts` suite. Each constructor retains its original direct/mapped inference checks plus focused checks for the return-annotation regression and generic-alias hole. The expanded call-shape matrix and standalone test files were removed.

## Verification

- `nix develop -c pnpm test-types typetest/Effect.tst.ts`: 256 tests passed across TypeScript 5.9.3 and 6.0.3.
- `nix develop -c pnpm test --run packages/effect/test/Effect.test.ts`: 247 tests passed.
- `nix develop -c pnpm check`: passed.
- `nix develop -c pnpm lint-fix`: passed.

Closes #8015
Closes EFF-1238

## Audit


packages/effect/src/Effect.ts
```diff
@@ -966,11 +966,12 @@ export const promise: <A>(
...
  * @since 2.0.0
  */
-export const tryPromise: <A, E = Cause.UnknownError>(
-  options:
-    | { readonly try: (signal: AbortSignal) => PromiseLike<A>; readonly catch: (error: unknown) => E }
-    | ((signal: AbortSignal) => PromiseLike<A>)
-) => Effect<A, E> = internal.tryPromise
+export const tryPromise: {
+  <A>(options: (signal: AbortSignal) => PromiseLike<A>): Effect<A, Cause.UnknownError>
+  <A, E = Cause.UnknownError>(
```
tim-smart: I don't think this default is needed?

tim-smart: Removed the mapped-overload error defaults from both `Effect.tryPromise` and `Effect.try` in 1cd580f2bf. `E` is inferred from `catch`. The existing type suite passes on TypeScript 5.9 and 6.0; the full typecheck, lint, and 247 Effect runtime tests also pass. Tests are unchanged in this commit.

PR #8022, opened 2026-09-04
fix(Effect): type repeatOrElse fallback metadata correctly

### LINKED ISSUE #8021: fix(Effect): type repeatOrElse fallback metadata correctly
## Problem

`Effect.repeatOrElse` exposes its fallback's second argument as `Option<B>`, but the v4 runtime passes `Option<Schedule.Metadata<B, A>>`. Both public call forms consequently reject valid metadata access and can accept output-only operations on a metadata object.

| Fallback observation | Public declaration | Actual v4 value |
| --- | --- | --- |
| Previous completed step is present | `Some<B>` | `Some<Schedule.Metadata<B, A>>` |
| Read the previous schedule output | Treat `previous.value` as `B` | After narrowing to `Some`, read `previous.value.output` |

Expected: type the existing metadata envelope accurately, rather than change runtime behavior back to an output-only contract.

## Evidence

The standalone comparison is pinned to `b2cc832e8e50c9d4f87d0d706e81b5c32a7b5ef9`. On **each of TypeScript 5.9.3 and 6.0.3**, the same type fixture goes from **7 failed / 1 passed tests** to **8 passed tests with 24 passed assertions** after correcting the declarations.

Runtime preservation checks pass **285 tests in each state: 7 new and 278 existing Effect/Schedule tests**. This is a type correction, not runtime red/green evidence. Explicit old output-only fallback annotations need to change to the metadata type.

## Audit

kitlangton:
## Why

`Effect.repeatOrElse` already passes the previous step's `Schedule.Metadata` to its fallback, but both public overloads type that argument as `Option<B>`. The stale declarations hide metadata fields and allow callers to treat the metadata object as the schedule output.

Closes #8021
Closes EFF-1240

## What changed

- Type the fallback argument as `Option<Schedule.Metadata<B, A>>` in both overloads.
- Update the JSDoc example to narrow `Some` and read `previous.value.attempt`.
- Add two type assertions in `Effect.tst.ts`, one per overload, for the fallback's metadata argument.
- Add migration guidance for callers with explicit `Option.Option<Output>` annotations.

Runtime behavior is unchanged.

## Verification

Earlier validation against unmodified `main` confirmed both overloads inferred `Option<number>` instead of `Option<Schedule.Metadata<number, string>>`. The two remaining assertions check this corrected contract directly.

- `pnpm test-types Effect.tst.ts`: 252 tests and 342 assertions passed across TypeScript 5.9.3 and 6.0.3
- `pnpm test --run packages/effect/test/Effect.test.ts`: 249 existing tests passed; this PR adds no runtime tests
- `pnpm check`: passed
- `pnpm lint-fix`: passed

The unchanged JSDoc example passed all 240 `Effect.ts` doctests in the preceding pass. `pnpm jsdocs --check` reported three existing diagnostics in `packages/effect/src/unstable/http/Multipart.ts`, which this PR does not modify.


.changeset/repeat-or-else-metadata-argument.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+Correct the `Effect.repeatOrElse` fallback type to expose the previous step's `Schedule.Metadata`, matching the existing runtime value. Callers with an explicit `Option.Option<Output>` annotation must use `Option.Option<Schedule.Metadata<Output, Input>>` and read fields such as `previous.value.output` after narrowing to `Some`. Runtime behavior is unchanged.
```
tim-smart:
```suggestion
Correct the `Effect.repeatOrElse` fallback type to expose the previous step's `Schedule.Metadata`, matching the existing runtime value.
```

PR #8032, opened 2026-09-04
fix(Effect): complete source cleanup before timeout fallback

### LINKED ISSUE #8031: fix(Effect): complete source cleanup before timeout fallback
## Problem

`Effect.timeoutOrElse` runs the fallback inside the timer side of the race. The fallback factory and body can start before the source's interruption cleanup, and a source that completes while the fallback is blocked can still win after the timeout has fired.

| Observation after the deadline is reached | Observed | Expected |
| --- | --- | --- |
| Fallback factory / body | Can run before source cleanup | Run after source interruption cleanup completes |
| Source completes while fallback is blocked | Can defeat the fallback | No longer competes once timeout is selected |

## Evidence

The canonical v2 standalone comparison is pinned to `b2cc832e8e50c9d4f87d0d706e81b5c32a7b5ef9`, not a later commit or publication tip. The same selection gives **11 failed / 40 passed tests → 51 passed**, followed by another unchanged 51-test pass. It includes **44 new and 7 existing tests; 242 tests are excluded**.

A Deferred/Ref handshake keeps the fallback blocked while observing whether the source can still complete, rather than racing two newly ready results or asserting elapsed time. Ordering checks independently observe source cleanup before both fallback construction and execution. Successful source values, including nested Options, `0` and `undefined`, remain distinct from the timeout decision.

## Audit

kitlangton:
## Why

The existing `Effect.timeoutOrElse` JSDoc says that when the timeout wins, the source is interrupted before the fallback runs. The current implementation contradicts that contract: it races the source against the timer and fallback together, so fallback work can start before source cleanup finishes, and a slow fallback can let the source win after the deadline.

Closes #8031
Closes EFF-1245

## What changes

Use `timeoutOption` to settle the timeout decision and finish source interruption cleanup, then return the source value or lazily invoke the fallback outside the race.

| Case | Before | After |
| --- | --- | --- |
| Deadline reached | Fallback can start before source cleanup | Source cleanup finishes before the fallback factory and body run |
| Source completes while fallback is blocked | Source can still win | The selected fallback wins |
| Ready source returns `Option.none()` at zero duration | Source value is returned | Source value remains distinct from timeout |

## Compatibility

The fallback now runs in the caller continuation rather than the timer child. `Effect.forkChild` inside the fallback follows the caller's supervision and lifetime, and the fallback inherits the caller's ambient interruptibility. Code relying on the timer child's lifetime or interruptibility should choose an explicit child lifetime or interruption boundary. The provided `Scope` identity and Scope-owned cleanup remain unchanged.

## Verification

Run the focused cases on the fixed implementation:

```sh
pnpm test --run packages/effect/test/Effect.test.ts -t "timeoutOrElse"
```

To compare the merge-base runtime while keeping the current tests:

```sh
git restore --source=b2cc832e8e50c9d4f87d0d706e81b5c32a7b5ef9 packages/effect/src/internal/effect.ts
pnpm test --run packages/effect/test/Effect.test.ts -t "timeoutOrElse"
git restore packages/effect/src/internal/effect.ts
```

- Merge-base runtime: 4 failed, 2 passed. The failures cover cleanup ordering for both call forms, the slow-fallback race, and caller-fiber execution.
- Fixed runtime: 6 passed. The full `Effect.test.ts` file passes 255/255.
- `pnpm lint-fix` and `pnpm check` pass.

## Audit


packages/effect/src/internal/effect.ts
```diff
@@ -3790,9 +3790,9 @@ export const timeoutOrElse: {
       readonly orElse: LazyArg<Effect.Effect<A2, E2, R2>>
     }
   ): Effect.Effect<A | A2, E | E2, R | R2> =>
-    raceFirst(
-      self,
-      flatMap(sleep(options.duration), options.orElse)
+    flatMap(
+      timeoutOption(self, options.duration),
+      (option): Effect.Effect<A | A2, E2, R2> => Option.isNone(option) ? suspend(options.orElse) : succeed(option.value)
```
tim-smart:
```suggestion
      (option): Effect.Effect<A | A2, E2, R2> => Option.isNone(option) ? options.orElse() : succeed(option.value)
```

PR #8065, opened 2026-09-04
fix(ai): bound Toolkit preliminary result buffering

danieljvdm:
`Toolkit.handle` puts preliminary results in a private, unbounded queue. A handler can keep filling it while its consumer is paused, even when it yields every `preliminary` call. Downstream buffer limits cannot constrain that queue.

Bound it to 16 entries so producers wait for consumers. Use `Stream.ensuring` instead of `onEnd` to interrupt handlers when consumers stop early or cancel, including producers waiting on a full queue.

```mermaid
sequenceDiagram
  participant Handler
  participant Queue as Native Toolkit queue
  participant Consumer as Returned Stream
  Handler->>Queue: preliminary(result)
  Note over Handler,Queue: At most 16 queued results<br/>Further offers suspend
  Consumer->>Queue: consume result
  Queue-->>Handler: resume pending offer
  Consumer->>Handler: interrupt on stream close
```

With consumption paused, the regression accepts 64 updates before the fix and 16 after. Tests also verify ordered delivery, success/failure settlement, and cleanup while the parent remains alive. All 51 Tool tests, type checks, and lint pass. The docs check reports six existing diagnostics in `Multipart.ts` and `ChildProcess.ts`.

### DIFF
--- a/.changeset/bounded-toolkit-progress.md
+++ b/.changeset/bounded-toolkit-progress.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Bound each Toolkit handler's result buffer to 16 entries so preliminary results backpressure slow consumers. Interrupt handlers when their result stream closes, including early termination and cancellation.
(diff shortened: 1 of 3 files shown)


kvnloo:
Exact-head comment on Effect-TS/effect#8065 head 132d312b.

KEEP bound Toolkit preliminary Queue to capacity 16 (backpressure while full) and Stream.ensuring(Fiber.interrupt(fiber)) so handlers stop when the result stream closes.

CHECK new Tool.test.ts backpressure case pins accepted==16 while latch open, then drains to 64 for both success and failure settlement.

Author danieljvdm not kvnloo. No competing PR from me.

PR #8070, opened 2026-09-05
Fix durable suspension for child workflow fan-out inside activities

tim-smart:
Parallel child workflows inside an activity could interrupt sibling dispatch, repeatedly retry the activity, and eventually exhaust its interrupt retry budget. Register children before execution-ID calculation and let intentional suspension return a durable `Suspended` result, releasing activity resources while children wait.

Preserve child wakeups that arrive during parent cleanup in both engines. Cluster resume messages use a separate key per child, and concurrent wakes share one pending wait for the parent's persisted reply. Coalescing stops before the parent is reset so later completions can wake the next replay. Empty resume envelopes persisted by older runners remain valid. Resume handlers do not occupy the slot reserved for deferred completions.

Pending child registrations are released on failure or cancellation and adopted once by the matching workflow instance. Attempting to consume a registration twice fails with a defect instead of silently adding a slot that could deadlock suspension.

Activities retain normal replay semantics: their bodies can execute again when the parent resumes. Bounded concurrency dispatches the next batch on replay. No child-wait polling loop is added.

Validation:

- Regression pass reproduced activity suspension failures and incomplete sibling dispatch before implementation.
- 164 workflow and cluster tests pass, including bounded replay in both engines, long suspension, resource release, cancellation during ID calculation, ID calculation failure, and child completion during cleanup. Existing durable race and clock-in-activity tests also pass.
- The 64-child cleanup regression reduces peak parked reply handlers from 64 to 1 while retaining each durable wake. Fresh-entity tests cover legacy resume envelopes against both suspended and queued parent runs.
- Repository-wide `pnpm check` and `pnpm lint` pass (all toolchain commands run through `nix develop -c`).
- `pnpm jsdocs --check` reports existing errors in unchanged `Multipart.ts` and `ChildProcess.ts`; the updated Activity documentation has no diagnostics.
- Includes an `effect` patch changeset.

Closes EFF-1268

### DIFF
--- a/packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts
+++ b/packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts
@@ -88,7 +88,7 @@ export const make = Effect.gen(function*() {
         >,
         Schema.declare<Workflow.Result<any, any>>
       >
-      | Rpc.Rpc<"resume", Schema.Struct<{}>>
+      | typeof ResumeRpc
     >
   >()
   const partialEntities = new Map<
@@ -101,7 +101,7 @@ export const make = Effect.gen(function*() {
         Schema.Struct<{ name: typeof Schema.String; attempt: typeof Schema.Int }>,
         Schema.declare<Workflow.Result<any, any>>
       >
-      | Rpc.Rpc<"resume">
+      | typeof ResumeRpc
     >
   >()
   const ensureEntity = (workflow: Workflow.Any) => {
@@ -293,17 +293,17 @@ export const make = Effect.gen(function*() {
   const sendResumeParent = Effect.fnUntraced(function*(options: {
     readonly workflowName: string
     readonly executionId: string
-  }) {
+  }, childExecutionId: string) {
     const requestId = yield* requestIdFor({
       workflow: workflows.get(options.workflowName)!,
       entityType: `Workflow/${options.workflowName}`,
       executionId: options.executionId,
       tag: "resume",
-      id: ""
+      id: childExecutionId
     })
     if (Option.isNone(requestId)) {
       const client = (yield* RcMap.get(clientsPartial, options.workflowName))(options.executionId)
-      return yield* client.resume({} as any, { discard: true })
+      return yield* client.resume({ childExecutionId }, { discard: true })
     }
     const reply = yield* replyForRequestId(requestId.value)
     if (Option.isNone(reply)) return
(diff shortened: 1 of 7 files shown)


packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts
```diff
@@ -293,23 +294,45 @@ export const make = Effect.gen(function*() {
...
+    const waiter = yield* storage.registerReplyHandler(
+      new Message.OutgoingRequest<any>({
+        envelope: request,
+        context: Context.empty() as Context.Context<any>,
+        rpc: ensureEntity(workflow).protocol.requests.get("run")!,
+        lastReceivedReply: Option.none(),
+        respond: () => Effect.void,
+        annotations: Context.empty()
+      })
+    ).pipe(Effect.interruptible, Effect.forkScoped({ startImmediately: true }))
```
tim-smart: Forked fibers are already interruptible

packages/effect/src/unstable/workflow/WorkflowEngine.ts
```diff
@@ -708,6 +709,18 @@ export const layerMemory: Layer.Layer<WorkflowEngine> = Layer.effect(WorkflowEng
       if (exit && exit._tag === "Success" && exit.value._tag === "Complete") {
         return
       } else if (state.fiber && !exit) {
+        // A child can finish while the parent is still releasing its activity.
+        // Preserve the wake until that run has published its suspended result.
+        if (!state.resumeRequested) {
+          state.resumeRequested = true
+          yield* Fiber.await(state.fiber).pipe(
+            Effect.andThen(() => {
```
tim-smart: Can this be Effect.flatMap?

packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts
tim-smart: Removed the redundant `Effect.interruptible` in 8c00afb7cc. All 51 targeted workflow/cluster tests, repository-wide typecheck, and lint pass.

packages/effect/src/unstable/workflow/WorkflowEngine.ts
tim-smart: Changed this to `Effect.flatMap` in 8c00afb7cc. All 51 targeted workflow/cluster tests, repository-wide typecheck, and lint pass.

PR #8085, opened 2026-09-05
fix(Schema): preserve Class JSON Schema annotations

### LINKED ISSUE #8084: toJsonSchemaDocument drops a Schema.Class's class-level annotations; the same annotations on a hoisted Schema.Struct survive
**Version:** `effect@4.0.0-rc.112` · Node v26.8.1

## Summary

A `Schema.Class`'s class-level annotations — `title` and `description`, and equally any custom key admitted through `includeAnnotationKey` — never reach the generated JSON Schema document. The root becomes a bare `$ref` and the definition it points at carries no annotations.

The same annotations on a `Schema.Struct` hoisted into `$defs` by an `identifier` **do** survive, on the pool entry. So this is not a limitation of hoisting or of the `$defs` representation; the generator already does the right thing one code path over.

## Reproduction

```ts
import { JsonSchema, Schema } from "effect"

class Foo extends Schema.Class<Foo>("Foo")({ a: Schema.String }, { title: "T", description: "D" }) {}

console.log(JSON.stringify((Foo.ast as any).annotations))
const doc = Schema.toJsonSchemaDocument(Foo)
console.log(JSON.stringify(doc))
console.log(JSON.stringify(JsonSchema.toDocumentDraft07(doc)))
```

Output:

```text
{"identifier":"Foo","~sentinels":[],"title":"T","description":"D"}

{"dialect":"draft-2020-12","schema":{"$ref":"#/$defs/FooEncoded"},
 "definitions":{"FooEncoded":{"type":"object","properties":{"a":{"type":"string"}},
 "required":["a"],"additionalProperties":false}}}

{"dialect":"draft-07","schema":{"$ref":"#/definitions/FooEncoded"},
 "definitions":{"FooEncoded":{"type":"object","properties":{"a":{"type":"string"}},
 "required":["a"],"additionalProperties":false}}}
```

The annotations are present on the AST and absent from the document, at both dialects. `Foo.annotate({ title, description })` after the fact behaves identically.

## The contrast — a hoisted Struct keeps them

```ts
const S = Schema.Struct({ a: Schema.String }).annotate({ identifier: "S", title: "T", description: "D" })
console.log(JSON.stringify(Schema.toJsonSchemaDocument(S)))
```

```text
{"dialect":"draft-2020-12","schema":{"$ref":"#/$defs/S"},
 "definitions":{"S":{"type":"object","properties":{"a":{"type":"string"}},
 "required":["a"],"additionalProperties":false,"title":"T","description":"D"}}}
```

Same shape — bare `$ref` root, annotations on the pool entry — and here they are carried.

## Mechanism

The class AST carries the annotations (first line of the reproduction's output) plus an `encoding` link to a **separate** `Objects` AST holding just the property signatures:

```text
CLASS AST KEYS: ["~effect/Schema","annotations","checks","encoding","context",
                 "_tag","typeParameters","run","encodingChecks","encodingRun"]
LINK: [{"to":{"~effect/Schema":"~effect/Schema","_tag":"Objects",
        "propertySignatures":[{"name":"a","type":{...,"_tag":"String"}}],
        "indexSignatures":[]}, "transformation":{...}}]
```

The generator emits the definition from that encoded AST — hence the `FooEncoded` pool name — and the encoded AST has no annotations. Nothing copies them across the `encoding` link.

## Not limited to `title` / `description`

Any class-level annotation is lost, including custom keys:

```ts
class Q extends Schema.Class<Q>("Q")({ a: Schema.String }, { "x-taplo": { hidden: true } }) {}
Schema.toJsonSchemaDocument(Q, { includeAnnotationKey: () => true })
// => schema {"$ref":"#/$defs/QEncoded"}; definitions.QEncoded has no x-taplo
```

This is what makes it more than cosmetic for us: we publish SchemaStore documents, where the editor-facing keywords (`markdownDescription`, `x-taplo`, `x-tombi-*`) are the payload. A `Schema.Class` root cannot carry any of them.

## Ask

Propagate a class's own annotations onto the definition generated from its encoded AST, matching what an identifier-hoisted `Schema.Struct` already does.

## Workaround

Use an annotated `Schema.Struct` as the root rather than a `Schema.Class`.

xia-chao:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Schema.toJsonSchemaDocument` follows a class’s encoding link to the underlying object schema. The class annotations stayed on the outer class node, so the generated `$defs` entry lost `title`, `description`, and custom JSON annotations.

This change carries serializable class annotations to the encoded schema used for JSON Schema generation. If the encoded side already has the same annotation, the encoded-side value still wins. Internal class metadata and callbacks are not copied.

The regression test covers an annotation passed to `Schema.Class`, annotations added later with `.annotate()`, and a custom key enabled through `includeAnnotationKey`.

Validation:

- `pnpm lint-fix`
- `pnpm test --run packages/effect/test/schema/toJsonSchemaDocument.test.ts packages/effect/test/schema/representation/toRepresentation.test.ts packages/effect/test/schema/jsonSchemaRoundTrip.test.ts packages/effect/test/schema/toCodec.test.ts packages/effect/test/schema/Schema.test.ts` (1038 tests)
- `pnpm check`

## Related

- Closes #8084

### DIFF
--- a/.changeset/calm-owls-share.md
+++ b/.changeset/calm-owls-share.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Preserve `Schema.Class` annotations on generated JSON Schema definitions.
--- a/packages/effect/src/internal/schema/toCodec.ts
+++ b/packages/effect/src/internal/schema/toCodec.ts
@@ -4,6 +4,7 @@ import type * as PublicSchema from "../../Schema.ts"
 import * as SchemaAST from "../../SchemaAST.ts"
 import * as SchemaGetter from "../../SchemaGetter.ts"
 import * as InternalTransformation from "../../SchemaTransformation.ts"
+import * as InternalAnnotations from "./annotations.ts"
 import * as InternalMake from "./make.ts"
 
 /** @internal */
@@ -12,7 +13,17 @@ export function toCodecJson<S extends PublicSchema.Constraint>(schema: S): Publi
 }
 
 /** @internal */
-export const toCodecJsonAST = SchemaAST.applyToSelfOrLastLinkEncodingIdempotent((ast) => {
+export const toCodecJsonAST = memoize((ast: SchemaAST.AST): SchemaAST.AST => {
+  const out = toCodecJsonASTBase(ast)
+  const annotations = SchemaAST.isDeclaration(ast) ? getClassJsonAnnotations(ast) : undefined
+  return annotations === undefined
+    ? out
+    : SchemaAST.applyToSelfOrLastLinkEncoding((encoded) =>
+      SchemaAST.annotate(encoded, { ...annotations, ...InternalAnnotations.resolve(encoded) })
+    )(out)
+})
+
+const toCodecJsonASTBase = SchemaAST.applyToSelfOrLastLinkEncodingIdempotent((ast) => {
   const out = toCodecJsonASTStep(ast, toCodecJsonAST)
   const context = ast.context
   if (out === ast || context === undefined) return out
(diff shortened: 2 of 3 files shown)


gcanti:
The current APIs already handle the case raised in the related issue without requiring a special case.

PR #8092, opened 2026-09-06
feat(atom-angular): add Angular bindings for Effect Atom

tabernerojerry:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

**Suggested Labels:** `4.0`

Adds `@effect/atom-angular`, Angular bindings for the Effect Atom modules. It joins
`@effect/atom-react`, `@effect/atom-solid`, and `@effect/atom-vue` under `packages/atom`.

The package shares one `AtomRegistry` through Angular's dependency injection and surfaces atoms as
Angular signals, so a component reads atom state the same way it reads any other signal and writes
it through functions that resolve the registry at injection time.

```ts
import { Component } from "@angular/core"
import { injectMakeAtom, provideAtomRegistry } from "@effect/atom-angular"

@Component({
  selector: "app-users",
  providers: [provideAtomRegistry()],
  template: `
    @switch (result().status) {
      @case ("initial") { <p>Loading…</p> }
      @case ("success") { <ul>@for (u of result().value; track u.id) { <li>{{ u.name }}</li> }</ul> }
      @case ("failure") { <p>{{ result().error }}</p> }
    }
    <button (click)="handleRefresh()">Reload</button>
  `
})
export class UsersComponent {
  readonly #store = injectMakeAtom(usersAtom)
  readonly result = this.#store.matchResult()

  handleRefresh() {
    this.#store.refresh()
  }
}
```

### API surface

`src/Providers.ts` — dependency-injection wiring:

- `provideAtomRegistry(options?)` — the provider array an application or route supplies
- `ATOM_REGISTRY` / `injectAtomRegistry()` — the registry token and its accessor
- `REGISTRY_OPTIONS` / `injectRegistryOptions()` — configures the registry a scope creates

`src/Injects.ts` — signal-based injection functions:

- Read: `injectAtomValue`, `injectAtomRef`, `injectAtomComputed`, `injectAtomMount`,
  `injectAtomSubscribe`
- Write: `injectAtomSet`, `injectAtomUpdate`, `injectAtomSetExit`, `injectAtomSetPromise`,
  `injectAtomRefresh`
- Composition: `injectMakeAtom` bundles the above into a single store whose surface is selected from
  the atom's type — reads for every atom, writes for writable ones, `matchResult` for
  `Result`-shaped ones; `atomMatchResult` flattens a `Result`-shaped store into one template-ready
  view tagged by `ATOM_RESULT_STATUS`
- Lifecycle: `injectDestroyRef`

`src/Constants.ts` — `ATOM_RESULT_STATUS`, the `initial` / `success` / `failure` tag set.

`src/Types.ts` — the type-level surface: option bags, the helpers that read an atom's value/write/
error types back out of its type, `AtomResultView`, and the layered store interfaces
(`ReadableAtomStore`, `WritableAtomStore`, `ResultAtomStore`, `AsyncAtomStore`).

Each branch of the flattened view also carries `waiting`, which is orthogonal to the tag: a success
branch can be waiting on the next run. Success and failure payloads default to the atom's own
success type and to a `Cause.pretty` string; `transformSuccess` / `transformFailure` fix them by
inference, and `transformFailure` receives the whole `Cause`, not the bare error.

### Design notes

**DI instead of a provider component.** React and Solid ship a `RegistryProvider` component.
Angular's equivalent is a provider array, so this package exposes `provideAtomRegistry()` plus
injection tokens rather than a component. `ATOM_REGISTRY` is `providedIn: "root"`, so a registry
exists without any explicit wiring; calling `provideAtomRegistry()` at application or route level
creates a scoped one.

**`inject*` instead of `use*`.** The siblings name their entry points `useAtom*`. Angular's
convention for functions that must run inside an injection context is the `inject` prefix, and
following it makes the constraint legible at the call site — so the equivalents here are
`injectAtomValue`, `injectAtomSet`, and so on.

**Injection-context contract.** Every `inject*` function must be called from an injection context.
Subscriptions and keepalives are created synchronously and released when the surrounding injector is
destroyed. What those functions _return_ — setters, refresh callbacks, and the store's methods —
captures the registry at injection time and stays callable outside an injection context, so a store
can be handed to a callback or a service method without losing its binding.

**Signal identity.** `value()` builds its signal on first call and shares it from then on, so
repeated calls share one subscription. `computed()` builds a new derived atom and subscription per
call and is meant to be stored on the component, not called from a template expression.

**Divergence from the sibling packages, stated up front.** The API is not a one-to-one port, and
I would rather surface that here than have a reviewer discover it:

- No equivalent of React's `useAtom` read+write tuple. The closest is `injectMakeAtom`, which
  returns a store object instead of a tuple — a tuple destructure reads poorly in an Angular
  component field, and the store lets the type system add or withhold `set` / `matchResult` based on
  the atom.
- No equivalent of `useAtomInitialValues`, `useAtomRefProp`, `useAtomRefPropValue`,
  `useAtomSuspense`, `ScopedAtom`, or `HydrationBoundary` (React), or `useAtomResource` (Solid).
  Suspense and hydration have no Angular counterpart; the others were left out until there is a
  concrete use for them rather than ported speculatively.
- Adds `injectAtomComputed`, `injectAtomUpdate`, `injectAtomSetExit`, `injectAtomSetPromise`,
  `atomMatchResult`, and `injectDestroyRef`, which have no sibling equivalent.

**Dedicated `Types.ts` / `Constants.ts`.** The siblings colocate types with implementations. The
store interfaces here are layered so that a store exposes only what its atom supports, and that
conditional surface is large enough that keeping it beside the injection functions made both harder
to read. `ATOM_RESULT_STATUS` is split out for the same reason: it is the single source of truth for
the tag set, with `AtomResultStatus` derived from it and `AtomResultView` discriminated by it.

### Testing

`packages/atom/angular/test/index.test.ts` — (53) cases covering the registry providers, each
injection function, the store surfaces, and the `Result` view. `vitest.setup.ts` bootstraps `TestBed`
with `platformBrowserTesting` and resets it per test; the vitest project runs under `jsdom`, matching
`atom-react` and `atom-solid`. Type-level expectations are asserted inline with `expectTypeOf`, as in
the sibling bindings.

Validated with `pnpm lint-fix`, `pnpm test --run packages/atom/angular/test/index.test.ts`,
`pnpm check`, `pnpm circular`, and `pnpm ai-docgen` (tree clean), plus a build and dry `pnpm pack` to
confirm the packed surface resolves `publishConfig.exports` to `dist` and keeps the internal and
legacy paths blocked. `pnpm jsdocs --check` reports no findings for this package.

### Monorepo registration

Following `.agents/skills/package-development/registration.md`:

- `.changeset/config.json` — added to the `fixed` version group alongside the other atom packages
- `tsconfig.packages.json` — project reference
- `tsconfig.tests.json` — `@effect/atom-angular` and `@effect/atom-angular/*` path mappings
- `vitest.config.ts` — project entry with the `jsdom` environment and setup file
- `README.md` — catalog table row
- `ai-docs/package.json` — workspace dependency, matching the other three atom packages
- `exports` / `publishConfig.exports` / `files` mirror the sibling packages
- Covered without change: `pnpm-workspace.yaml` (`packages/atom/*` glob), `snapshot.yml`
  (`./packages/atom/*` glob), `jsdocs.config.json` and lint configs (glob-based), `deno.json`
  (`packages/atom` is excluded), `tstyche.json` (no `typetest` directory)

### Changeset

`.changeset/angular-atom-bindings.md`, `patch` for `@effect/atom-angular`. The repository is in
`rc` pre-mode, where new surface is recorded as `patch`.

### Peer range

`@angular/core: ">=20.0.0 <23.0.0"`, developed and tested against 22.1.x. Please confirm that floor
matches the Angular support window you want to commit to.

### Follow-up

A sample application demonstrating the bindings is ready and will follow in a separate PR once this
lands, so the toolchain and lockfile cost of an Angular CLI app can be reviewed on its own.

Sample Demo: `https://github.com/Effect-TS/effect/commit/55fa911c40f9e513f8421e72fa76ca5f1a5d0709`

### DIFF
--- a/.changeset/angular-atom-bindings.md
+++ b/.changeset/angular-atom-bindings.md
@@ -1,0 +1,22 @@
+---
+"@effect/atom-angular": patch
+---
+
+Add `@effect/atom-angular`, Angular bindings for the Effect Atom modules.
+
+The package exposes the Atom registry through Angular dependency injection with `provideAtomRegistry`,
+`ATOM_REGISTRY`, `injectAtomRegistry`, and `REGISTRY_OPTIONS` / `injectRegistryOptions` for configuring the
+registry a scope creates.
+
+Injection functions surface atoms as Angular signals and operations on them: `injectAtomValue`,
+`injectAtomRef`, `injectAtomComputed`, `injectAtomMount`, `injectAtomSubscribe`, `injectAtomRefresh`,
+`injectAtomSet`, `injectAtomUpdate`, `injectAtomSetExit`, `injectAtomSetPromise`, and `injectDestroyRef`.
+`injectMakeAtom` bundles those operations into a single store whose surface is selected from the atom's
+type — reads for every atom, writes for writable ones, and `matchResult` for `Result`-shaped ones —
+and `atomMatchResult` flattens a `Result`-shaped store into one template-ready view tagged by
+`ATOM_RESULT_STATUS`.
+
+Every `inject*` function must be called from an injection context. Subscriptions and keepalives are created
+synchronously and released when the surrounding injector is destroyed; what those functions return — setters,
+refresh callbacks, and the store's methods — captures the registry at injection time and stays callable
+outside an injection context.
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -19,6 +19,7 @@
       "@effect/ai-openai",
       "@effect/ai-openai-compat",
       "@effect/ai-openrouter",
+      "@effect/atom-angular",
       "@effect/atom-react",
       "@effect/atom-solid",
       "@effect/atom-vue",
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ This monorepo contains the core `effect` package alongside integration packages 
 | [`@effect/ai-openai`](packages/ai/openai)                             | OpenAI provider for the Effect AI modules                | [docs](https://effect.website/docs/v4/api/ai-openai)               |
 | [`@effect/ai-openai-compat`](packages/ai/openai-compat)               | OpenAI-compatible API provider for the Effect AI modules | [docs](https://effect.website/docs/v4/api/ai-openai-compat)        |
 | [`@effect/ai-openrouter`](packages/ai/openrouter)                     | OpenRouter provider for the Effect AI modules            | [docs](https://effect.website/docs/v4/api/ai-openrouter)           |
+| [`@effect/atom-angular`](packages/atom/angular)                       | Angular bindings for Effect Atom                         | [docs](https://effect.website/docs/v4/api/atom-angular)            |
 | [`@effect/atom-react`](packages/atom/react)                           | React bindings for Effect Atom                           | [docs](https://effect.website/docs/v4/api/atom-react)              |
 | [`@effect/atom-solid`](packages/atom/solid)                           | SolidJS bindings for Effect Atom                         | [docs](https://effect.website/docs/v4/api/atom-solid)              |
 | [`@effect/atom-vue`](packages/atom/vue)                               | Vue bindings for Effect Atom                             | [docs](https://effect.website/docs/v4/api/atom-vue)                |
--- a/ai-docs/package.json
+++ b/ai-docs/package.json
@@ -8,6 +8,7 @@
     "@effect/ai-openai": "workspace:*",
     "@effect/ai-openai-compat": "workspace:*",
     "@effect/ai-openrouter": "workspace:*",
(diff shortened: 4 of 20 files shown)


tim-smart:
For now this can be a community package :) We might consider an official package later.

tabernerojerry:
> For now this can be a community package :) We might consider an official package later.

Awesome, thank you for the feedback.

PR #8093, opened 2026-09-06
fix(persistence)!: correct token-bucket timing and expiration

Vicentesan:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

Correct token-bucket `retryAfter`, `onExceeded: "delay"`, and `resetAfter` in both memory and Redis stores.

With five tokens per five minutes, exhausting the bucket and retrying after 59 seconds now reports **1 second** instead of 60 seconds. Reserving one token at that point reports a **1-second delay** and a **301-second reset time**.

Both stores calculate timing within the atomic consumption operation, accounting for elapsed refill time and existing reservations. Refill boundaries use the full window and a stable timestamp to avoid fractional rounding drift at epoch timestamps. Total waits round up to whole milliseconds, and negative elapsed refill time is clamped to prevent backward clock skew from inflating waits.

This also aligns refill and expiration behavior:

- Full buckets discard partial refill progress, preventing an extra token immediately after exhausting a full bucket.
- Configuration changes settle refills using the previous settings. Fully recovered buckets start fresh with the new configuration, consistently across retained memory state and expired Redis state.
- Redis preserves fractional token counts instead of truncating replies.
- Redis expiration rounds up to whole-token refills and whole milliseconds, preventing fractional consumption from expiring state before recovery.

Redis maintains an additional refill-state hash that expires alongside the existing keys. Existing numeric refill timestamps remain readable, and no separate Redis round trip is needed.

### Breaking change

`RateLimiterStore.tokenBucket` now:

- Accepts `window: Duration.Duration` instead of `refillRate`.
- Returns `readonly [remaining, retryAfterMillis, resetAfterMillis]` instead of a number.

Custom stores must calculate all three values atomically. Reset timing reflects the persisted token count, so rejected attempts do not extend it. Direct callers must destructure the tuple.

JSDoc and the changeset document the updated contract, behavior, and migration.

## Validation

- 45 limiter unit tests and 16 RateLimiter tests against real Redis pass.
- Coverage includes rejection timing, exact refill boundaries, reset timing, delayed reservations, fractional intervals and token counts, epoch timestamps, backward clocks, concurrency, full-bucket recovery, configuration changes, and legacy Redis state.
- Expiration regressions exercise actual Redis key expiration and verify that fractional token state survives until recovery.
- Redis timing tests use independent `TestClock` instances.
- Type-contract tests pass on TypeScript 5.9 and 6.0.
- `pnpm check`, `pnpm lint-fix`, and `pnpm lint` pass.
- `pnpm jsdocs --check` reports existing errors in unrelated `Multipart.ts` and `ChildProcess.ts` documentation.


Vicentesan:
Closing in favor of #8103, which fixes the original timing issue and now also covers the fractional-token and premature-expiration cases I raised, with regression tests.

Thanks for incorporating those cases and working through this with me!

PR #8103, opened 2026-09-06
Fix token bucket refill timing and fractional costs

tim-smart:
After exhausting five tokens in a five-minute window, retrying 59 seconds later now reports `retryAfter: 1s` instead of `60s`. Memory and Redis use whole-token refill boundaries for retry, delay, and reset timing, including fractional token costs.

Redis preserves signed fractional balances, so a remaining count of `-0.5` rejects instead of becoming `0`. Both tuple values use round-trip precision, keeping accumulated fractional balances and timing consistent with memory. Both keys expire when capacity actually refills, with the TTL rounded up to a millisecond. A 0.5-token cost with a 4s refill interval therefore keeps the bucket until the 4s boundary.

`RateLimiterStore.tokenBucket` now returns `readonly [remaining: number, elapsedMillis: number]` atomically. Custom stores must retain numeric precision in both values and provide elapsed time in milliseconds; returning `[remaining, 0]` retains the timing bug. Restart the interval only at full capacity after refill and before consumption. Otherwise, advance the boundary by whole refill intervals without restarting it on reads.

`resetAfter` measures time until full capacity with no further consumption, including reserved debt. Failed requests do not reserve debt.

Validation:

- Memory suite: 50 passed.
- Real Redis timing/expiry coverage: 30 passed; 21 unrelated tests excluded. Includes actual wall-clock expiry and accumulated fractional balances compared with persisted values.
- `pnpm check`, `pnpm lint-fix`, `pnpm lint`, and `git diff --check` passed.
- JSDoc checks report existing errors in unchanged `Multipart.ts` and `ChildProcess.ts`.
- Tests are unchanged from regression commit `eddbf84`.

Closes EFF-1297

### DIFF
--- a/packages/effect/test/unstable/persistence/RateLimiterTest.ts
+++ b/packages/effect/test/unstable/persistence/RateLimiterTest.ts
@@ -3,18 +3,18 @@ import { Duration, Effect, type Layer } from "effect"
 import { TestClock } from "effect/testing"
 import { RateLimiter } from "effect/unstable/persistence"
 
-export const tokenBucketTimingSuite = (
+export const suite = (
   name: string,
   layer: Layer.Layer<RateLimiter.RateLimiterStore, unknown>
 ) => {
   it.layer(layer, { timeout: "30 seconds" })(`RateLimiter token-bucket timing (${name})`, (it) => {
     for (const onExceeded of ["fail", "delay"] as const) {
       for (
-        const [elapsed, tokens, expected] of [
-          [59_000, 1, 1_000],
-          [59_500, 1, 500],
-          [59_000, 2, 61_000],
-          [119_000, 2, 1_000]
+        const [elapsed, tokens, expected, resetAfter] of [
+          [59_000, 1, 1_000, 301_000],
+          [59_500, 1, 500, 300_500],
+          [59_000, 2, 61_000, 361_000],
+          [119_000, 2, 1_000, 301_000]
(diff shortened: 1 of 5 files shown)


packages/effect/test/unstable/persistence/RateLimiterTest.ts
```diff
@@ -0,0 +1,90 @@
+import { assert, it } from "@effect/vitest"
+import { Duration, Effect, type Layer } from "effect"
+import { TestClock } from "effect/testing"
+import { RateLimiter } from "effect/unstable/persistence"
+
+export const tokenBucketTimingSuite = (
```
tim-smart: This can just be exported as `suite`, in case it needs to be extended later.

packages/effect/test/unstable/persistence/RateLimiterTest.ts
```diff
@@ -0,0 +1,183 @@
...
+ * Consumes `tokens`, runs `idle`, and asserts the bucket is back at capacity
+ * with a fresh refill interval.
+ */
+export const restartsInterval = <E, R>(
+  key: string,
+  onExceeded: OnExceeded,
+  tokens: number,
+  idle: Effect.Effect<void, E, R>
+) =>
+  Effect.gen(function*() {
```
tim-smart: Use fnUntraced

PR #8106, opened 2026-09-07
Fix tool result serialization to respect failure status

tim-smart:
`Response.ToolResultPart` used the first matching result schema even when `isFailure` selected the other branch. With `Schema.Number` for success and `Schema.NumberFromString` for failure, a failed `404` encoded as a number, and an `AllParts` JSON round trip lost the original string `encodedResult`.

Select between complete success and failure codecs using literal `isFailure` tags. Validate `encodedResult` against the selected branch as well. Preserve metadata and flag defaults and propagate the supplied schemas' decoding and encoding services. The factory now declares a `Schema.Codec` return type over the existing `ToolResultPart` and `ToolResultPartEncoded` models.

Branch selection exposed a second gap: failed results produced by the framework (`AiError`, and the `execution-denied` / `execution-interrupted` results minted for denied or interrupted calls) were never part of the tool's failure schema, so they could not round trip through `AllParts` once the failure branch was enforced. `Tool.failureResultSchema(tool)` now defines the complete failure result once, as `AiError | failureSchema | Tool.ExecutionFailure`, and both `Toolkit` result encoding and the `Response` part schemas use it. `AiError` comes first so it is always reconstructed; the tool's own schema comes before `ExecutionFailure` so user failures keep their codec and fields. `Tool.FailureResult`, `FailureResultEncoded`, `Result`, and `ResultEncoded` include `ExecutionFailure` regardless of `failureMode`, matching what the codec produces.

`HttpRequestDetails` and `HttpResponseDetails` move to `AiError` and are re-exported from `Response`, so `Response` can import `Tool` without an import cycle.

The `effect` patch changeset documents migration of persisted results that do not match their selected branch, the public `Schema.Codec` return type, the new `Tool.failureResultSchema` / `Tool.ExecutionFailure` exports, and the widened failure result types.

Regression coverage checks direct encoding, `AllParts` JSON round trips, rejection of cross-branch `encodedResult` values, provider-defined denial and interruption results, `AiError` reconstruction with an unknown failure schema, and preservation of user failure fields for struct, unknown, and record failure schemas.

Validation:

- `nix develop -c pnpm vitest run packages/effect/test/unstable/ai/ packages/ai/anthropic/test packages/ai/openai/test`: 1237 passed, 44 skipped.
- `nix develop -c pnpm test-types packages/effect/typetest/unstable/ai/`: 124 assertions passed on TypeScript 5.9.3 and 6.0.3.
- `nix develop -c pnpm check`, `pnpm lint-fix`, `pnpm circular`, `pnpm ai-docgen` (no changes), and `git diff --check`: passed.

Closes EFF-1128


packages/effect/src/unstable/ai/Tool.ts
```diff
@@ -255,6 +255,21 @@ export interface Tool<
...
+   *
+   * **Details**
+   *
+   * A failed result is either a value of the tool's `failureSchema`, an
+   * `AiError` raised while handling the call, or an {@link ExecutionFailure}
+   * synthesized when the call was denied or interrupted. The toolkit result
+   * encoder and the `Response` part schemas both use this schema, so failed
+   * results encode and decode consistently.
+   */
+  failureResultSchema(): Schema.Union<
```
tim-smart: This can be a module level api instead

PR #8110, opened 2026-09-07
Fix filesystem and HTTP file range precision loss

tim-smart:
Filesystem writes and stats could silently round `9007199254740993` to `9007199254740992`. Writes now reject unsafe positions with `PlatformError` / `BadArgument`. Stats request bigint values and preserve `size` and `blksize` exactly.

Number-typed metadata (`dev`, `ino`, `rdev`, `mode`, `nlink`, `uid`, `gid`, and `blocks`) must be safe integers. An unsafe value fails the whole stat operation, including optional fields. The public `File.Info` documentation explains that large inode values can therefore prevent stat and HTTP file serving even for small files.

`HttpPlatform` file responses keep content lengths as bigint and write exact decimal headers, so whole files above the safe integer limit remain supported. EOF clamping and static-server range arithmetic retain exact sizes, including suffix, open-ended, and 416 totals. Invalid numeric inputs and unrepresentable runtime range bounds fail with `BadArgument` instead of defects or rounding.

`HttpPlatform.make` now supplies bigint `contentLength` to its platform callback. Node and Deno compare empty lengths as bigint; Deno derives bounded stream lengths from checked numeric start/end bounds. The portable platform retains numeric body-length metadata only when safe and always preserves the exact header. Bun also writes the exact length header.

Writes retain checked numeric positions because callback-based `fs.write` ignores bigint positions. Small-file probes confirmed this on Node 18.0.0, 20.0.0, 22.0.0, 24.0.0, 26.4.0, and 26.7.0, matching [Node's documented non-number position behavior](https://nodejs.org/api/fs.html#fswritefd-buffer-offset-length-position-callback). Oversized regression cases use doubles.

Validation at `3abe687`:

- 131 targeted Node/core tests, 20 Bun tests, and 75 Deno tests pass; one existing Deno test is skipped.
- `nix develop -c pnpm lint-fix`, `nix develop -c pnpm check`, and Deno type checking of `DenoHttpPlatform.ts` pass.
- `nix develop -c pnpm test-types FileSystem.tst.ts` passes on both selected TypeScript targets (16 assertions).
- `nix develop -c pnpm jsdocs --check` reports only two errors in unchanged `Logger.ts`: an out-of-order `@see` and an unresolved `consolePretty` link.

Review cleanup at `f2fd26b`: precision coverage is consolidated into the existing test files and reduced from 63 cases to 20. Shared range validation runs in core; Node and Bun retain adapter-specific end-bound and exact-header checks. The changeset uses the suggested single sentence. Implementation is unchanged.

Current validation: all 85 tests across the five affected suites pass (71 Node/core, 14 Bun), down from 128 (108 Node/core, 20 Bun). Both runs used shuffled execution with seed 1309. `pnpm lint-fix`, `pnpm check`, and `git diff --check` pass.

Separate backlog follow-up: [EFF-1310](mention://issue/01a079b7-2eaa-7e90-8e18-2187ff29d1b4) covers `HttpBody.file` / `fileFromInfo` content-length precision loss and invalid range inputs causing defects, including `HttpClientRequest.bodyFile`. Assigned to The Boys; this PR does not implement that follow-up.

Closes EFF-1309


packages/effect/test/unstable/http/HttpStaticServer.precision.test.ts
```diff
@@ -0,0 +1,85 @@
+import { assert, describe, it } from "@effect/vitest"
```
tim-smart: Please don't create one-off tests files for this

.changeset/fix-filesystem-integer-precision.md
```diff
@@ -0,0 +1,18 @@
...
+Prevent precision loss in Node/Bun filesystem operations and `HttpPlatform` file responses.
+
+Writes reject unsafe positions with `BadArgument`. Stats preserve `size` and
+`blksize` exactly and reject unsafe numeric metadata, including optional fields.
+Large inode values can therefore prevent stat and file serving even for small files.
+
+`HttpPlatform.make` now passes bigint `contentLength` to its callback, preserving
+exact headers for whole files above `Number.MAX_SAFE_INTEGER`. Invalid range
+inputs and unsafe runtime range bounds fail with `BadArgument`.
+`HttpStaticServer` also preserves exact range calculations and `Content-Range` totals.
```
tim-smart:
```suggestion
Prevent precision loss in Node/Bun filesystem operations and `HttpPlatform` file responses.
```

PR #8112, opened 2026-09-07
Fix invalid File.readAlloc sizes on Node and Deno

tim-smart:
`File.readAlloc` now fails with `PlatformError` (`BadArgument`, `FileSystem.readAlloc`) for negative, fractional, non-finite, and unallocatable sizes on Node and Deno. Previously, `1.5` consumed one byte, Deno accepted `NaN` as an empty read, and other invalid allocations surfaced as defects.

Validate and allocate before reading so rejected sizes preserve the cursor. `readAlloc(0)` continues to return `Option.none()` without moving the cursor. Includes shared regression coverage and a patch changeset for both platform packages.

Validation passed:

- Node file-system suite: 38 tests.
- Deno file-system suite: 36 tests, 1 existing skip.
- `nix develop -c pnpm check`
- `nix develop -c deno check packages/platform/deno/src/DenoFileSystem.ts`

Closes EFF-1308

### DIFF
--- a/packages/platform/deno/src/DenoFileSystem.ts
+++ b/packages/platform/deno/src/DenoFileSystem.ts
@@ -247,31 +247,27 @@ class FileImpl implements FileSystem.File {
   }
 
   readAlloc(size: number) {
-    return Effect.flatMap(
-      Effect.try({
-        try: () => {
-          if (!Number.isInteger(size) || size < 0) {
-            throw new RangeError("size must be a non-negative integer")
-          }
-          return new Uint8Array(size)
-        },
-        catch: (cause) =>
-          PlatformError.badArgument({
-            module: "FileSystem",
-            method: "readAlloc",
-            description: `Could not allocate a buffer of size ${size}`,
-            cause
-          })
-      }),
-      (buffer) => {
+    return Effect.suspend(() => {
+      try {
+        if (!Number.isInteger(size) || size < 0) {
+          throw new RangeError("size must be a non-negative integer")
+        }
+        const buffer = new Uint8Array(size)
         return Effect.map(this.readChunk("readAlloc", buffer), (bytesRead) => {
           if (bytesRead === 0) {
             return Option.none()
           }
           return Option.some(bytesRead === size ? buffer : buffer.subarray(0, bytesRead))
         })
+      } catch (cause) {
+        return Effect.fail(PlatformError.badArgument({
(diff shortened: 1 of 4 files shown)


packages/platform/deno/src/DenoFileSystem.ts
```diff
@@ -247,15 +247,31 @@ class FileImpl implements FileSystem.File {
...
-      const buffer = new Uint8Array(size)
-      return Effect.map(this.readChunk("readAlloc", buffer), (bytesRead) => {
-        if (bytesRead === 0) {
-          return Option.none()
-        }
-        return Option.some(bytesRead === size ? buffer : buffer.subarray(0, bytesRead))
-      })
-    })
+    return Effect.flatMap(
+      Effect.try({
```
tim-smart: I think just use a try/catch in a Effect.suspend block.

tim-smart: Same for the other one.

packages/platform/deno/src/DenoFileSystem.ts
```diff
@@ -248,13 +248,25 @@ class FileImpl implements FileSystem.File {
...
-      return Effect.map(this.readChunk("readAlloc", buffer), (bytesRead) => {
-        if (bytesRead === 0) {
-          return Option.none()
+      try {
+        if (!Number.isInteger(size) || size < 0) {
+          throw new RangeError("size must be a non-negative integer")
         }
-        return Option.some(bytesRead === size ? buffer : buffer.subarray(0, bytesRead))
-      })
+        const buffer = new Uint8Array(size)
```
tim-smart, line 255: Can this throw? If not we could drop the try / catch

tim-smart: Yes, the allocation can throw even after the guard passes. Reproduced on Deno 2.9.4:

```ts
const size = Number.MAX_VALUE
console.log(Number.isInteger(size) && size >= 0) // true
new Uint8Array(size)
// RangeError: Invalid typed array length: 1.7976931348623157e+308
```

The existing shared `Number.MAX_VALUE` regression covers this case and verifies `BadArgument` plus cursor preservation. All six Deno `readAlloc` cases pass at `d6f624fa4e`. Removing the catch would turn this native allocation failure back into a defect, so I kept it.

PR #8114, opened 2026-09-07
Skip optional stack capture when stackTraceLimit is zero

### LINKED ISSUE #8038: Effect.fn constructs an Error at every definition site, which costs most of a Cloudflare Workers startup budget
`Effect.fn` constructs an `Error` at module scope for every definition, to remember the
definition site for stack frames. On Cloudflare Workers that construction costs between 19 and
256 microseconds each, depending on the size of the module the definition sits in, so an
application with about 1,600 definitions spends more than its entire 400 ms startup CPU budget
there.

## Reproduction

Repository: https://github.com/agcty/effect-fn-startup-repro

```
bun install
bun run all
```

Each variant emits 1,600 definitions and ends with

```js
globalThis.__evaluated = defs.length
export default { fetch() { return new Response(String(globalThis.__evaluated)) } }
```

Every number below comes from a run whose worker answered a request with the expected
definition count. This matters because `wrangler check startup` reports a fast startup for a
bundle that throws during evaluation, so a timing number on its own does not prove the code
ran.

Two gates are reported. The first is the scripted CPU total from the profile
`wrangler check startup` writes. The second uses no profiler: the wall time from spawning
`wrangler dev` to the first correct response. Both are the median of five runs, each against a
freshly started server. About 800 ms of every cold start is wrangler and workerd starting up,
so that gate confirms effects above roughly 100 ms and cannot resolve smaller ones.

## Measurements

Apple M2 Max, macOS 26.6.2, wrangler 4.127.1, workerd 1.20260828.1, Node 24.15.0
(V8 13.6.233.17), esbuild 0.25.10, bun 1.4.0.

Padded means the module is filled out to 60,000 trivial `p = (p ^ N) >>> 0` statements, so the
enclosing top-level module function is large. The 1,600 definitions are spread evenly through
it. Those statements are cheap to run and cannot be eliminated, and on their own they cost
0.0 ms of startup CPU.

| variant | 1,600 definitions built with | workerd startup CPU | workerd cold start | Node eval |
| --- | --- | --- | --- | --- |
| A | `Effect.fn`, small module | 31.5 ms | 821 ms | 8.4 ms |
| A0 | `Effect.fnUntraced`, small module | 1.7 ms | 825 ms | 6.0 ms |
| B | `Effect.fn`, padded module | 410.7 ms | 1282 ms | 38.5 ms |
| D | `Effect.fnUntraced`, padded module | 1.9 ms | 877 ms | 36.1 ms |
| C | `Effect.fn`, same code split across 12 modules | 62.3 ms | 900 ms | 43.8 ms |
| C0 | `Effect.fnUntraced`, split across 12 modules | 1.9 ms | 821 ms | 41.6 ms |

Subtracting each untraced baseline from its `Effect.fn` counterpart, the definition-site
capture costs:

| enclosing module | 1,600 definitions | per definition |
| --- | --- | --- |
| small, 1,600 statements | 29.8 ms | 19 us |
| split into 12 modules of 5,000 statements | 60.4 ms | 38 us |
| single module, 15,000 statements | 134.0 ms | 84 us |
| single module, 30,000 statements | 227.9 ms | 142 us |
| single module, 60,000 statements | 408.8 ms | 256 us |

The cost is linear in the number of definitions. In the 60,000 statement module, 1, 16, 160
and 1,600 definitions cost 1.6, 4.6, 34.6 and 410.7 ms of startup CPU.

## Where the cost is

`Effect.fn` constructs the definition Error before it looks at anything else:

```js
export const fn = function () {
  const nameFirst = typeof arguments[0] === "string"
  const name = nameFirst ? arguments[0] : "Effect.fn"
  const spanOptions = nameFirst ? arguments[1] : undefined
  const prevLimit = getStackTraceLimit()
  setStackTraceLimit(2)
  const defError = new globalThis.Error()
  setStackTraceLimit(prevLimit)
  ...
```

`defError` is kept for the lifetime of the function and read lazily. When the built function
runs, `makeFn` puts it into `CurrentStackFrame` as the parent frame named
`${name} (definition)`, alongside a second `Error` captured per call. `CurrentStackFrame` is
what annotates a `Cause` with the stack trace on failure and on interruption, so the definition
Error is a real feature and not dead weight. Nothing reads it at definition time.

Four controls locate the cost precisely. All of them use the padded 60,000 statement module and
report workerd startup CPU.

| control | result |
| --- | --- |
| 1,600 plain `new Error()`, no Effect, `stackTraceLimit` 2 | 372.4 ms |
| 1,600 plain `new Error()`, no Effect, `stackTraceLimit` 0 | 373.5 ms |
| 1,600 `Error.stackTraceLimit` write pairs, no `Error` constructed | 0.0 ms |
| the same padding, no capture at all | 0.0 ms |

The cost is therefore the `Error` construction itself. Effect adds nothing measurable on top of
it, and the `Error.stackTraceLimit` writes are free. Two consequences matter for any fix:

- The stack trace limit is irrelevant. A capture that keeps zero frames costs the same as one
  that keeps two. Capturing a lighter value, or lowering the limit to 1, would not help.
- Setting `Error.stackTraceLimit = 0` in application code does not help either, because `fn`
  overrides the limit for the duration of the capture. Measured at 409.9 ms, unchanged from
  variant B.

Moving the construction into a small helper called from the large module, which is the shape
`Effect.fn` already has, reproduces the cost at 343.5 ms. The size that matters is the calling
function's, not the constructing function's.

## Mechanism

Under Node 24 the same 1,600 captures cost 2.4 ms in both the small and the padded module,
about 1.5 microseconds each, and that figure does not change with module size. Under workerd
the same captures cost 19 us each in a small module and 256 us each in a 60,000 statement one.
Constructing an `Error` in workerd costs time in proportion to the size of the JavaScript
function on the stack, whether or not any frames are retained.

I have not identified what workerd or its V8 embedding does differently here. The behaviour is
observed and reproducible, not explained. The workerd side is filed separately as https://github.com/cloudflare/workerd/issues/7245. The practical burden still falls on Effect, because Effect is what places an
`Error` construction at every definition site in an application's module graph.

The behaviour has not improved with newer runtimes. On workerd 1.20251001.0 variant B cost
365.5 ms; on 1.20260828.1 it costs 410.7 ms.

## Real-world impact

The application this came from evaluates about 1,600 `Effect.fn` definitions at Worker startup,
spread across six packages. In a CPU profile of its startup, 232 ms of 364 ms of scripted CPU
were in these captures. Cloudflare enforces a 400 ms startup CPU limit, so deploys began
failing on the first attempt and succeeding only on a retry.

There is no good application-side fix. Rewriting every definition to `Effect.fnUntraced` gives
up spans and definition frames everywhere. Splitting modules until each one is small helps, and
the repro measures a nearly seven-fold improvement from splitting one module into twelve, but
that is a strange constraint to place on a codebase and it does not remove the cost.

## Proposed remedies, in order of preference

1. **Honour `captureStackTrace: false` in `fn`.** `Effect.fn(name, spanOptions)` already accepts
   span options, and `Tracer.SpanOptions.captureStackTrace` already means "do not capture".
   Today `fn` constructs `defError` before it ever inspects `spanOptions`. Reading the option
   first and skipping the construction would give a per-site opt-out that reuses an existing
   convention and changes no defaults.

2. **A global opt-out.** An environment variable or a module-level flag that switches the
   definition capture off for a whole bundle. An application with 1,600 definitions cannot
   annotate them one at a time, and a Worker build knows at build time that it wants this. A
   build-time constant a bundler can fold would also let the capture be removed outright rather
   than branched around.

3. **Avoid the eager construction.** If the definition frame can be recorded with something
   cheaper than a full `Error` at definition time, that would fix the problem for everyone
   without an opt-out. I do not know whether that is possible, since the definition site is only
   on the stack while the module is evaluating.

`Effect.fnUntraced` is the only escape hatch today. It avoids the cost completely, 1.9 ms
against 410.7 ms, but it also drops the span, so it is not a substitute for `fn` where tracing
is wanted.

## Versions

- effect 4.0.0-rc.112
- workerd 1.20260828.1 via wrangler 4.127.1
- Node 24.15.0, V8 13.6.233.17
- macOS 26.6.2 on Apple M2 Max
- esbuild 0.25.10, bun 1.4.0

tim-smart:
`Effect.fn` overrides `Error.stackTraceLimit = 0` and constructs errors to capture definition and invocation sites. These captures can be expensive during Cloudflare Workers startup even with a zero-length stack.

Skip optional error construction when the limit is zero in functions, spans, layer and middleware service definitions, atom labels, and the internal stack probe. Leave a zero limit unchanged when formatting causes. Named spans, function frames, and enabled stack locations are preserved. Explicit `captureStackTrace: true` still opts into span capture at limit zero, and supplied stack callbacks remain supported.

Service definitions expose an undefined `stack` at limit zero and retain their definition location at positive limits. Includes an Effect patch changeset documenting these behaviors.

Adds four focused regressions for zero-limit function definition/invocation capture, default span capture, and explicit span opt-in. Existing tests cover the surrounding tracing behavior.

Validation:

- All 783 tests in 21 related files pass under Node, including the four focused regressions.
- `nix develop -c bun run --bun vitest run packages/effect/test/StackCapture.test.ts`: all four pass.
- `nix develop -c pnpm lint-fix`: passed.
- `nix develop -c pnpm check`: passed.

Closes EFF-1312
Closes #8038

### DIFF
--- a/.changeset/skip-disabled-stack-capture.md
+++ b/.changeset/skip-disabled-stack-capture.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Skip optional stack capture when `Error.stackTraceLimit` is zero in `Effect.fn`, spans, layer and middleware service definitions, and atom labels. Preserve named spans and explicit stack callbacks, and avoid raising a zero limit when formatting causes.
--- a/.changeset/skip-disabled-stack-capture.md
+++ b/.changeset/skip-disabled-stack-capture.md
@@ -2,4 +2,6 @@
 "effect": patch
 ---
 
-Skip optional stack capture when `Error.stackTraceLimit` is zero in `Effect.fn`, spans, layer and middleware service definitions, and atom labels. Preserve named spans and explicit stack callbacks, and avoid raising a zero limit when formatting causes.
+Skip optional stack capture when `Error.stackTraceLimit` is zero in `Effect.fn`, spans, layer and middleware service definitions, and atom labels. Preserve named spans and enabled stack locations, and leave a zero limit unchanged when formatting causes.
+
+Explicit span options remain supported at a zero global limit: `captureStackTrace: true` opts into capture, and a supplied stack callback is preserved. The `stack` property on `LayerMap.Service`, `LayerRef.Service`, `HttpApiMiddleware.Service`, and `RpcMiddleware.Service` definitions is `undefined` when defined at limit zero; positive limits retain the definition location.
--- a/.changeset/skip-disabled-stack-capture.md
+++ b/.changeset/skip-disabled-stack-capture.md
@@ -2,6 +2,4 @@
 "effect": patch
 ---
 
-Skip optional stack capture when `Error.stackTraceLimit` is zero in `Effect.fn`, spans, layer and middleware service definitions, and atom labels. Preserve named spans and enabled stack locations, and leave a zero limit unchanged when formatting causes.
-
-Explicit span options remain supported at a zero global limit: `captureStackTrace: true` opts into capture, and a supplied stack callback is preserved. The `stack` property on `LayerMap.Service`, `LayerRef.Service`, `HttpApiMiddleware.Service`, and `RpcMiddleware.Service` definitions is `undefined` when defined at limit zero; positive limits retain the definition location.
+Skip optional stack capture when `Error.stackTraceLimit` is zero.
--- a/packages/effect/src/LayerMap.ts
+++ b/packages/effect/src/LayerMap.ts
@@ -431,11 +431,13 @@ export const Service = <Self>() =>
   Options extends { readonly dependencies: ReadonlyArray<Layer.Layer<any, any, any>> } ? Options["dependencies"][number]
     : never
 > => {
-  const Err = globalThis.Error as any
   const limit = getStackTraceLimit()
-  setStackTraceLimit(2)
-  const creationError = new Err()
-  setStackTraceLimit(limit)
+  let creationError: Error | undefined
+  if (limit !== 0) {
+    setStackTraceLimit(2)
+    creationError = new globalThis.Error()
+    setStackTraceLimit(limit)
+  }
 
   function TagClass() {}
   const TagClass_ = TagClass as any as Mutable<TagClass<Self, Id, string, any, any, any, any, any>>
@@ -443,7 +445,7 @@ export const Service = <Self>() =>
   TagClass.key = id
   Object.defineProperty(TagClass, "stack", {
     get() {
-      return creationError.stack
+      return creationError?.stack
     }
   })
 
--- a/packages/effect/src/LayerRef.ts
+++ b/packages/effect/src/LayerRef.ts
@@ -349,11 +349,13 @@ export const Service = <Self>() =>
   [Preload] extends [true] ? E : never,
   Deps[number]
 > => {
-  const Err = globalThis.Error as any
   const limit = getStackTraceLimit()
-  setStackTraceLimit(2)
-  const creationError = new Err()
-  setStackTraceLimit(limit)
+  let creationError: Error | undefined
+  if (limit !== 0) {
+    setStackTraceLimit(2)
+    creationError = new globalThis.Error()
+    setStackTraceLimit(limit)
+  }
 
   function TagClass() {}
   const TagClass_ = TagClass as any as Mutable<TagClass<Self, Id, any, any, any, any, any>>
@@ -361,7 +363,7 @@ export const Service = <Self>() =>
   TagClass.key = id
   Object.defineProperty(TagClass, "stack", {
     get() {
-      return creationError.stack
+      return creationError?.stack
     }
   })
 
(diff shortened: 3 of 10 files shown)


.changeset/skip-disabled-stack-capture.md
```diff
@@ -0,0 +1,7 @@
+---
+"effect": patch
+---
+
+Skip optional stack capture when `Error.stackTraceLimit` is zero in `Effect.fn`, spans, layer and middleware service definitions, and atom labels. Preserve named spans and enabled stack locations, and leave a zero limit unchanged when formatting causes.
+
+Explicit span options remain supported at a zero global limit: `captureStackTrace: true` opts into capture, and a supplied stack callback is preserved. The `stack` property on `LayerMap.Service`, `LayerRef.Service`, `HttpApiMiddleware.Service`, and `RpcMiddleware.Service` definitions is `undefined` when defined at limit zero; positive limits retain the definition location.
```
tim-smart:
```suggestion
Skip optional stack capture when `Error.stackTraceLimit` is zero.
```

PR #8115, opened 2026-09-07
Fix file body content-length precision and range errors

tim-smart:
File body constructors can round Content-Length when sizes or offsets exceed the safe integer range, and invalid ranges currently cause defects. Calculate ranges with exact bigint arithmetic, clamp to EOF, and convert only the final safe length to a number. Invalid inputs and final lengths above `Number.MAX_SAFE_INTEGER` now fail with typed `PlatformError` / `BadArgument`; larger sizes and ranges remain valid when their resulting length is representable.

The change covers `HttpBody.file`, `HttpBody.fileFromInfo`, and propagation through `HttpClientRequest.bodyFile`, with public policy documentation and a changeset. Regression tests use stat/stream doubles and cover lazy evaluation, small-file bytes, EOF clamping, oversized ranges, invalid inputs, and outgoing headers.

Validation:

- All 43 tests pass in `HttpBody.test.ts` and `HttpClientRequest.test.ts` (previously 147). File-body cases were reduced from 126 to 22 by removing redundant permutations and repeated entry-point coverage; the 21 unrelated request tests are unchanged.
- `nix develop -c pnpm lint-fix`, `nix develop -c pnpm lint`, and `nix develop -c pnpm check` pass.
- `nix develop -c pnpm jsdocs --check` reports two existing errors in unchanged `packages/effect/src/Logger.ts`: an out-of-order `@see` and an unresolved `consolePretty` link. No diagnostics concern the modified HTTP docs.

Closes EFF-1310

### DIFF
--- a/.changeset/exact-http-file-body-length.md
+++ b/.changeset/exact-http-file-body-length.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-Calculate `HttpBody.file`, `HttpBody.fileFromInfo`, and `HttpClientRequest.bodyFile` content lengths with exact bigint arithmetic and EOF clamping. Fail with typed `PlatformError` / `BadArgument` when the final length exceeds `Number.MAX_SAFE_INTEGER`, while allowing larger sizes and range inputs when the resulting length is representable. Invalid range inputs now fail through the typed error channel during effect evaluation instead of causing defects.
+Calculate `HttpBody.file`, `HttpBody.fileFromInfo`, and `HttpClientRequest.bodyFile` content lengths with exact bigint arithmetic and EOF clamping.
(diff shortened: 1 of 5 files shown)


.changeset/exact-http-file-body-length.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+Calculate `HttpBody.file`, `HttpBody.fileFromInfo`, and `HttpClientRequest.bodyFile` content lengths with exact bigint arithmetic and EOF clamping. Fail with typed `PlatformError` / `BadArgument` when the final length exceeds `Number.MAX_SAFE_INTEGER`, while allowing larger sizes and range inputs when the resulting length is representable. Invalid range inputs now fail through the typed error channel during effect evaluation instead of causing defects.
```
tim-smart:
```suggestion
Calculate `HttpBody.file`, `HttpBody.fileFromInfo`, and `HttpClientRequest.bodyFile` content lengths with exact bigint arithmetic and EOF clamping.
```

PR #8121, opened 2026-09-07
Align CLI constructor names with Schema and Config

tim-smart:
Use PascalCase for CLI value and control constructors, with scalar value names aligned to Schema and Config. This is a breaking API cleanup for the Effect 4 release candidate.

The value constructor renames apply per module:

| Modules | New constructor names |
| --- | --- |
| Primitive, Param, Flag, Argument | `String`, `Int`, `Finite`, `Date`, `Redacted`, `Path`, `FileText`, `FileParse`, `FileSchema` |
| Primitive, Param, Flag | `Boolean`, `KeyValuePair` |
| Param, Flag, Argument | `Literals`, `ChoiceWithValue`, `File`, `Directory` |
| Primitive | `Choice` |
| Primitive, Param, Flag, Argument | `Never` |

`Argument` has no `Boolean` or `KeyValuePair`. `Primitive` uses `Path` with a path type and `Choice` with key/value pairs; it has no `File`, `Directory`, or `ChoiceWithValue`.

Prompt constructors are `String`, `Int`, `Number`, `Date`, `File`, `Confirm`, `Toggle`, `Select`, `MultiSelect`, `AutoComplete`, `List`, `Password`, `Hidden`, and `Custom`. `Prompt.Number` retains its numeric parser and does not enforce `Schema.Finite`. Built-in String and Number references are qualified to avoid shadowing.

Plain string choices use `Literals` in Flag, Argument, and Param. `Primitive.Choice` and `ChoiceWithValue` retain their pair-mapping semantics. `Primitive.Never`, `Param.Never`, `Flag.Never`, and `Argument.Never` consistently name the always-failing sentinels. Their output type remains `never`. `KeyValuePair`, `File`, and `Directory` are unchanged by this final alignment.

`GlobalFlag.Action` and `GlobalFlag.Setting` coexist with the types of those names. Factories keep their lowercase names: `Command.make`, `CliConfig.make`, `Param.makeSingle`, `Prompt.makeTheme`, `Completions.generate`, and `CliOutput.defaultFormatter`. `Prompt.succeed` also stays lowercase, following `Effect.succeed`. Combinators, guards, runners, and layers keep their names.

Prompt’s public option types also change: `IntegerOptions` becomes `IntOptions`, and `FloatOptions` becomes `NumberOptions`. `NumberOptions` still extends `IntOptions`; fields and behavior are preserved. Shared `TextOptions` keeps its name for String, Hidden, Password, and List.

The public `Primitive.Never._tag` changes from `"None"` to `"Never"`; consumers matching this tag must update. The internal sentinel parameter name changes from `"__none__"` to `"__never__"`.

The other public tags change too: `Primitive.Int._tag` changes from `"Integer"` to `"Int"`, and `Primitive.Finite._tag` changes from `"Float"` to `"Finite"`. Both `Completions.FlagType` and `Completions.ArgumentType` change their `_tag` alternatives from `"Integer"`/`"Float"` to `"Int"`/`"Finite"`. Custom completion descriptors and switches or pattern matches on these unions must be updated. Parsing behavior, help labels, and generated completion scripts are unchanged.

Consumers, examples, migration annotations, and the migration reference are updated. Regeneration also surfaced missing guidance for `SortedSet.toggle`, `@effect/platform/Url.fromString`, and `@effect/sql/Model.Date`; those annotations describe existing behavior and make no changes to those implementations.

The migration reference uses implementation snapshot `b4773df70e`, including the unified `Never` constructors and Prompt option types. Its header records the branch used during generation and the full snapshot SHA.

Validation after the sentinel and option-type correction: all 413 CLI runtime tests, 232 CLI type assertions, 84 affected JSDoc examples, `pnpm check`, `pnpm lint-fix`, `pnpm lint`, and `git diff --check` passed. Tests are unchanged from tests-first commit `78826f4839`. The migration reference was regenerated. `pnpm jsdocs --check` reports only the existing `Logger.ts` diagnostics; no CLI JSDoc diagnostics were reported.

Closes EFF-1317


packages/effect/src/unstable/cli/Prompt.ts
```diff
@@ -1104,7 +1104,7 @@ export const hidden = (
  * @category constructors
  * @since 4.0.0
  */
-export const integer = (options: IntegerOptions): Prompt<number> => {
+export const Integer = (options: IntegerOptions): Prompt<number> => {
```
tim-smart: This can be `Int` to match the other modules

packages/effect/typetest/unstable/cli/Param.tst.ts
```diff
@@ -1,16 +1,42 @@
-import { Effect } from "effect"
-import { Argument, Flag, Prompt } from "effect/unstable/cli"
+import { Effect, type Redacted } from "effect"
+import { Argument, Flag, Param, Primitive, Prompt } from "effect/unstable/cli"
 import { describe, expect, it } from "tstyche"
 
 describe("Param", () => {
+  it("uses Schema and Config names for scalar constructors", () => {
```
tim-smart: Remove this test

packages/effect/typetest/unstable/cli/Constructors.tst.ts
```diff
@@ -0,0 +1,120 @@
+import { Effect, type Redacted, Schema } from "effect"
```
tim-smart: Remove this file

packages/effect/src/unstable/cli/Flag.ts
```diff
@@ -153,33 +153,23 @@ export const date = (name: string): Flag<Date> => Param.date(Param.flagKind, nam
...
- * @see {@link choiceWithValue} for mapping accepted strings to different typed values
+ * @see {@link ChoiceWithValue} for mapping accepted strings to different typed values
  *
  * @category constructors
  * @since 4.0.0
  */
-export const choice = <const Choices extends ReadonlyArray<string>>(
+export const Literals = <const Choices extends ReadonlyArray<string>>(
   name: string,
   choices: Choices
```
tim-smart: Should rename the argument name too.

PR #8139, opened 2026-09-09
Fix OpenTelemetry propagation through skipped spans

tim-smart:
Raw OpenTelemetry spans created inside an unnamed `Effect.fn` or a disabled span received a fake `noop` parent and started a separate trace. The bridge now skips spans marked `DisablePropagation` and uses the nearest propagated ancestor. `Tracer.currentOtelSpan` uses `Effect.currentPropagatedSpan`, including its missing-span failure when no propagated span exists.

Adds two focused regressions to the existing tracer suite: a raw OTel child inside an unnamed `Effect.fn` retains its enclosing trace and parent, and `currentOtelSpan` resolves to the enclosing span. Both reuse the existing tracing fixture; the suite's context manager is registered and cleaned up with suite hooks. Includes a patch changeset for `@effect/opentelemetry`.

When no propagated ancestor exists, the SDK bridge preserves the ambient OpenTelemetry context. `currentOtelSpan` fails with `NoSuchElementException` rather than returning a fake span. The separate `OtlpTracer` user-supplied `context` callback path still receives the unfiltered span and is deferred to a follow-up; this PR does not fix propagation through that callback.

Validation:

- `nix develop -c pnpm test run packages/opentelemetry/test packages/effect/test/Tracer.test.ts`: 83 passed, including all 19 tests in the existing OpenTelemetry tracer suite.
- `nix develop -c pnpm check`
- `nix develop -c pnpm build`
- `nix develop -c pnpm docgen`

Closes EFF-1325


.changeset/fix-otel-skipped-span-propagation.md
```diff
@@ -0,0 +1,7 @@
+---
+"@effect/opentelemetry": patch
+---
+
+Preserve OpenTelemetry trace propagation through unnamed `Effect.fn` and disabled spans by using the nearest propagated parent. Make `Tracer.currentOtelSpan` skip these spans as well.
+
+When no propagated ancestor exists, the bridge preserves the ambient OpenTelemetry context, allowing child spans to inherit an active OpenTelemetry parent. `Tracer.currentOtelSpan` now fails with `NoSuchElementException` in this case instead of returning a span backed by the fake `noop` context.
```
tim-smart:
```suggestion
Preserve OpenTelemetry trace propagation through unnamed `Effect.fn` and disabled spans by using the nearest propagated parent. Make `Tracer.currentOtelSpan` skip these spans as well.
```

PR #8145, opened 2026-09-09
fix(sql-pglite): support parameterless SQL scripts

k70suK3-k06a7ash1:
Parameterless SQL scripts currently fail through the PGlite row-query interface with `cannot insert multiple commands into a prepared statement` (SQLSTATE `42601`). For example:

```ts
yield* sql`
  CREATE TABLE example (id INTEGER, value TEXT);
  CREATE INDEX example_id ON example (id);
  CREATE INDEX example_value ON example (value);
`
```

The adapter always calls PGlite's single-statement `query` API. This change separates execution into an internal `Script | Parameterized` tagged union: parameterless SQL uses [`exec`](https://pglite.dev/docs/api#exec), while statements with a non-empty parameter array keep using `query` with bound values. SQL is passed intact, including semicolons inside literals, comments and dollar-quoted blocks.

Scripts return the final statement's rows, or an empty array when there are no rows. This applies to ordinary row queries, `.unprepared` and `.stream`, preserving result-name transformations and SQL error classification. The `.raw`, `.values` and `.valuesUnprepared` paths retain their existing single-statement behavior.

This was originally encountered with the combined index creation in `SqlMessageStorage` on `4.0.0-rc.112`. Current `main` already executes those indexes separately; the regression tests here reproduce the remaining adapter limitation directly.

Validation:

- Before the fix, the new regression file had 7 failures and 1 pass, reproducing SQLSTATE `42601`.
- `pnpm test --run packages/sql/pglite/test`: 7 files, 41 tests passed, covering DDL, final/empty results, execution modes, parameter binding, SQL quoting, failure atomicity, transaction rollback and result transformations.
- `pnpm check`, `pnpm lint-fix` and `git diff --check` passed.
- Includes a patch changeset for `@effect/sql-pglite`.

### DIFF
--- a/.changeset/pglite-parameterless-sql-scripts.md
+++ b/.changeset/pglite-parameterless-sql-scripts.md
@@ -1,0 +1,5 @@
+---
+"@effect/sql-pglite": patch
+---
+
+Support parameterless SQL scripts in PGlite row execution, including unprepared and streaming queries. Execute scripts with PGlite's `exec` API and return the final statement's rows, while retaining parameter binding for parameterized statements.
--- a/packages/sql/pglite/src/PgliteClient.ts
+++ b/packages/sql/pglite/src/PgliteClient.ts
@@ -12,8 +12,10 @@
  * @since 4.0.0
  */
 import { PGlite, type PGliteInterface, type PGliteOptions } from "@electric-sql/pglite"
+import * as Arr from "effect/Array"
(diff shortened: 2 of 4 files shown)


kyosuke-lvgs:
When will this PR be merged?

kyosuke-lvgs:
@tim-smart
Please review

PR #8150, opened 2026-09-09
fix(Formatter): include Error message in formatJson

### LINKED ISSUE #8144: Logger.consoleJson does not format Error correctly

4.0.0-rc.112

The Logger.consoleJson logger does not log correctly Errors that are not Data.TaggedError. The error message is not included in the logs.
Minimal reproducible example: https://effect.website/play#a737a4f170c5

OR the code:
```typescript
import { NodeRuntime } from "@effect/platform-node"
import { Effect, Logger } from "effect"

  yield* Effect.fail(new Error("This message is currently not parsed"))
}).pipe(
  Effect.tapError(Effect.logError),
  Effect.catch(() => Effect.succeed(""))
)

NodeRuntime.runMain(program.pipe(
  Effect.provide(Logger.layer([Logger.consoleJson]))
  //Effect.provide(Logger.layer([Logger.consoleStructured]))
))

```

`{"message":{},"level":"ERROR","timestamp":"2026-09-09T11:07:49.837Z","annotations":{},"spans":{},"fiberId":"#2"}`

To have something similar to Logger.formatStructured

 `{"message":"Error: This message is currently not parsed","level":"ERROR","timestamp":"2026-09-09T11:00:20.270Z","annotations":{},"spans":{},"fiberId":"#2"}`

### What do you see instead?

`{"message":{},"level":"ERROR","timestamp":"2026-09-09T11:00:20.270Z","annotations":{},"spans":{},"fiberId":"#2"}`

### Additional information

I think the problem is in Formatter.formatJson, maybe we need a epecial case for isntaceof Error.

Tyagiquamar:
Fixes #8144

## What

`Formatter.formatJson` now serializes `Error` instances without `toJSON` as objects containing their enumerable properties plus `name` and `message`. This makes `Logger.formatJson` and `Logger.consoleJson` retain the message of a plain `Error`:

```ts
import { Formatter } from "effect"

Formatter.formatJson(new Error("boom"))
// {"name":"Error","message":"boom"}
```

For the reported logging case:

- Before: `{"message":{},"level":"ERROR",...}`
- After: `{"message":{"name":"Error","message":"boom"},"level":"ERROR",...}`

Existing enumerable properties are preserved. For example, Node system error fields such as `errno`, `code`, `syscall`, and `path` remain available as structured JSON.

## Why

`Error#name` and `Error#message` are non-enumerable, so `JSON.stringify(new Error("boom"))` produces `{}`. Adding those fields explicitly preserves the useful message while retaining properties that `JSON.stringify` already emitted.

Errors with `toJSON`, including `Data.TaggedError`, keep their custom structured serialization.

## How

The `formatJson` replacer copies enumerable Error properties and adds `name` and `message`. It tracks both the original Error and its serialized representation so circular references remain omitted. Redaction still runs before Error serialization.

## Tests

- `packages/effect/test/Formatter.test.ts` covers root and nested plain Errors, Node-style enumerable properties, custom `toJSON`, and circular references.
- `packages/effect/test/Logger.test.ts` covers `Effect.tapError(Effect.logError)` through `Logger.formatJson`.

Validated with:

- `pnpm test --run packages/effect/test/Formatter.test.ts packages/effect/test/Logger.test.ts`
- `pnpm jsdocs --check`
- `pnpm lint`
- `pnpm check`

### DIFF
--- a/.changeset/fix-formatjson-error-message.md
+++ b/.changeset/fix-formatjson-error-message.md
@@ -1,0 +1,11 @@
+---
+"effect": patch
+---
+
+Fix `Formatter.formatJson` to include the message when stringifying plain `Error` values, matching the output of `Logger.formatStructured`.
+
+```ts
+import { Formatter } from "effect"
+
+Formatter.formatJson(new Error("boom")) // now `"Error: boom"`, previously `{}`
+```
--- a/packages/effect/src/Formatter.ts
+++ b/packages/effect/src/Formatter.ts
@@ -316,6 +316,9 @@ export function formatJson(input: unknown, options?: {
       if (typeof redacted !== "object" || redacted === null) {
         return redacted
       }
+      if (redacted instanceof Error && !Predicate.hasProperty(redacted, "toJSON")) {
+        return safeToString(redacted)
+      }
       while (ancestors.length > 0 && ancestors[ancestors.length - 1] !== this) {
         ancestors.pop()
       }
--- a/packages/effect/test/Formatter.test.ts
+++ b/packages/effect/test/Formatter.test.ts
@@ -344,6 +344,20 @@ describe("Formatter", () => {
       strictEqual(formatJson([1n, 2n]), `["1n","2n"]`)
     })
 
+    it("should stringify Error messages", () => {
+      strictEqual(formatJson(new Error("boom")), `"Error: boom"`)
+      strictEqual(formatJson({ error: new Error("boom") }), `{"error":"Error: boom"}`)
+    })
+
+    it("should keep structured serialization for Errors that define toJSON", () => {
+      class Tagged extends Error {
+        toJSON() {
+          return { _tag: "Tagged", message: "boom" }
+        }
+      }
+      strictEqual(formatJson(new Tagged("boom")), `{"_tag":"Tagged","message":"boom"}`)
+    })
+
     it("should redact sensitive data", () => {
       const date = Object.assign(new Date(0), {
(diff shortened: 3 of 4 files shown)


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

gcanti (changes requested):
This introduces a regression for Node system errors. `Formatter.formatJson` currently preserves their enumerable fields, such as `errno`, `code`, `syscall`, and `path`. Returning `safeToString(error)` replaces that structured information with a single runtime-dependent string.

We could instead preserve the existing enumerable properties and add `name` and `message` for errors without `toJSON`

Conceptually:

```ts
{ ...error, name: error.name, message: error.message }
```

This would fix plain errors serializing to `{}` while preserving the structured fields already emitted for Node errors. The implementation should retain the existing redaction and circular-reference handling.

Tyagiquamar:
Hi @gcanti, updated as suggested.

The implementation now preserves all enumerable properties (e.g. `errno`, `code`, `syscall`, `path` on Node system errors) while explicitly including `name` and `message` for errors without `toJSON`:
```ts
let current = redacted
if (current instanceof Error && !Predicate.hasProperty(current, "toJSON")) {
  current = { ...current, name: current.name, message: current.message }
}
```

Circular reference detection and redaction handling remain preserved, and tests have been updated covering plain errors, Node system errors, custom `toJSON` errors, and circular references.

gcanti (approved):

Issue #7187, opened 2026-08-11
Schema.d.ts references SchemaAST.Sentinel, which @internal-stripping removes from the published SchemaAST.d.ts — consumers without skipLibCheck fail to compile


spencerbeggs:
> **Status: work in progress.** Filing early to track the defect; I will follow up with a minimal reproduction repository.

## What happens

The published `effect` package's `dist/Schema.d.ts` references `SchemaAST.Sentinel` in a public type:

```ts
// dist/Schema.d.ts (4.0.0-beta.107, line ~10815)
readonly "~sentinels"?: ReadonlyArray<SchemaAST.Sentinel> | undefined;
```

but `Sentinel` is declared `/** @internal */` in `src/SchemaAST.ts` (~line 2515), so it is stripped from the published `dist/SchemaAST.d.ts` (zero occurrences). Any consumer whose tsconfig does **not** set `skipLibCheck: true` fails to compile:

```text
node_modules/effect/dist/Schema.d.ts(10815,57): error TS2694: Namespace '".../effect/dist/SchemaAST"' has no exported member 'Sentinel'.
```

Observed on `4.0.0-beta.105` and still present on `4.0.0-beta.107`, with both `tsc` (TypeScript 7.0.2) and `tsgo`.

## Expected

The published declaration set should be self-consistent: either `Sentinel` ships in `SchemaAST.d.ts` (drop the `@internal`, or exempt it from stripping since a public type references it), or the `~sentinels` member's declared type avoids the internal name.

## Workaround

Local patch appending the two-field type back onto `dist/SchemaAST.d.ts`:

```ts
export type Sentinel = {
    readonly key: PropertyKey;
    readonly literal: LiteralValue | symbol;
};
```

Reproduction repo to follow.

PR #8162, opened 2026-09-10
fix: remove dangling @internal references from published declarations

### LINKED ISSUE #8161: effect@4.0.0-rc.113: @internal-stripped declarations leave dangling references in published .d.ts (Effectable, Match, Schema) — breaks skipLibCheck: false
## What happens

`effect@4.0.0-rc.113` publishes declaration files that reference names which `@internal` stripping removes from the published `dist`. Three distinct dangling references, all introduced in `rc.113`:

| Referenced from | Symbol | Declared at (src) | Published in `dist`? |
|---|---|---|---|
| `dist/Effectable.d.ts:11` | `EffectTypeId` | `src/internal/core.ts:19` (`/** @internal */`) | ❌ `dist/internal/core.d.ts` is `export {};` |
| `dist/Match.d.ts:21` | `internal.Contextual` | `src/internal/matcher.ts:23` (`/** @internal */`) | ❌ `dist/internal/matcher.d.ts` is `export {};` |
| `dist/Schema.d.ts:11008,11018` | `AnnotationSchemaConstraint` | `src/Schema.ts:15516` (`/** @internal */`) | ❌ stripped, references left in same file |

Any consumer whose tsconfig does **not** set `skipLibCheck: true` fails to compile on a bare `import { Effect } from "effect"`:

```text
node_modules/effect/dist/Effectable.d.ts(11,15): error TS2305: Module '"./internal/core.ts"' has no exported member 'EffectTypeId'.
node_modules/effect/dist/Effectable.d.ts(55,24): error TS2589: Type instantiation is excessively deep and possibly infinite.
node_modules/effect/dist/Match.d.ts(21,41): error TS2694: Namespace '".../effect/dist/internal/matcher"' has no exported member 'Contextual'.
node_modules/effect/dist/Schema.d.ts(11008,72): error TS2304: Cannot find name 'AnnotationSchemaConstraint'.
node_modules/effect/dist/Schema.d.ts(11018,67): error TS2304: Cannot find name 'AnnotationSchemaConstraint'.
```

The `TS2589` at `Effectable.d.ts:55` appears to cascade from the unresolved `EffectTypeId`, which is used as a computed property key at `Effectable.d.ts:58`:

```ts
readonly [EffectTypeId]: AsEffectReturn<this>[typeof EffectTypeId];
```

## Minimal reproduction

`package.json` — only `effect@4.0.0-rc.113` and `typescript@7.0.2` installed.

`index.ts`:
```ts
import { Effect } from "effect";
export const ok = Effect.succeed(1);
```

`tsconfig.json`:
```json
{
  "compilerOptions": {
    "strict": true,
    "noEmit": true,
    "target": "ES2022",
    "module": "ESNext",
    "moduleResolution": "bundler",
    "lib": ["ES2022", "ESNext.Disposable", "DOM", "DOM.Iterable"],
    "skipLibCheck": false
  },
  "files": ["index.ts"]
}
```

`tsc --noEmit` emits exactly the five errors above. No other compiler options are required to trigger it.

## Regression range

Same project, same TypeScript, only the `effect` version swapped:

- `4.0.0-rc.111` — clean
- `4.0.0-rc.112` — clean
- `4.0.0-rc.113` — the five errors above

`rc.111`'s `Effectable.d.ts` did not import from `internal/core.ts` at all, and its `Match.d.ts` had no `Contextual` alias. `AnnotationSchemaConstraint` does not appear anywhere in `rc.111`'s `Schema.d.ts`.

## Scope of the stripping

31 of the 32 files in `dist/internal/` are `export {};` (only `schemaError.d.ts` has content), so any public `.d.ts` that reaches into `internal/` will dangle.

## This is a recurring class

Same root cause as #6431 (`preResponseHandler`, beta.98) and #7187 (`SchemaAST.Sentinel`, beta.107), both fixed. It seems worth a systemic guard rather than another point fix — e.g. a release-time check that type-checks the published `dist` with `skipLibCheck: false` against a trivial consumer, which would catch all three of these before publish.

## Expected

The published declaration set should be self-consistent: either these symbols ship in `dist` (drop `@internal`, or exempt symbols that public declarations reference), or the public declarations avoid referencing internal names.

## Environment

- `effect@4.0.0-rc.113`
- `typescript@7.0.2` (also reproduces with `tsgo`)
- macOS (arm64)

tim-smart:
## Problem

`effect@4.0.0-rc.113` publishes declaration files that reference names removed by `stripInternal`, so any consumer compiling with `skipLibCheck: false` fails on a bare `import { Effect } from "effect"`. Building `packages/effect` with `stripInternal: true` and type-checking a trivial consumer reproduces the five reported errors plus one more in the CLI:

```text
dist/Effectable.d.ts(11,15): error TS2305: Module '"./internal/core.ts"' has no exported member 'EffectTypeId'.
dist/Effectable.d.ts(55,24): error TS2589: Type instantiation is excessively deep and possibly infinite.
dist/Match.d.ts(21,41): error TS2694: Namespace '".../dist/internal/matcher"' has no exported member 'Contextual'.
dist/Schema.d.ts(10470,72): error TS2304: Cannot find name 'AnnotationSchemaConstraint'.
dist/Schema.d.ts(10480,67): error TS2304: Cannot find name 'AnnotationSchemaConstraint'.
dist/unstable/cli/internal/command.d.ts(66,122): error TS2339: Property 'getParamMetadata' does not exist on type 'typeof import(".../dist/unstable/cli/Param")'.
```

## Fix

- `Effectable.ts`: use the public `Effect.TypeId` instead of the internal `EffectTypeId` in the `MixinBase` declaration.
- `Match.ts`: define the `Contextual` conditional locally instead of aliasing the `@internal` type from `internal/matcher.ts`.
- `Schema.ts`: remove `AnnotationSchemaConstraint` and reference the public `Schema.Constraint` through a type-only namespace import in `Annotations.ToArbitrary.DeclarationInput` and `Declaration`. This preserves their generic bounds despite the nested `ToArbitrary.Constraint` name collision.
- `unstable/cli/internal/command.ts`: mark `toFlagDoc` as `@internal`. It is only used by the internal help and command modules, and its signature leaked `Param.getParamMetadata`.

## Guard

This is the third occurrence of this class of bug (#6431, #7187). `scripts/check-dist-types.mjs` type-checks the built `effect` declarations from a generated consumer that imports every published entrypoint with `skipLibCheck: false`. The CI `build` job now runs it via `pnpm check-dist-types` right after the `stripInternal` build. On the original source it fails with the six errors above; after this fix it passes.

## Validation

- `pnpm check`, `pnpm lint`, `pnpm jsdocs --check`
- `pnpm test --run` for `Match`, `Effectable`, `cli/Help`, `cli/Command`
- `pnpm test-types Match Effectable`

Closes EFF-1336
Closes #8161

### DIFF
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -14980,7 +14980,7 @@ export function resolveAnnotations<S extends Constraint>(
 export function resolveAnnotationsKey<S extends Constraint>(schema: S): Annotations.Key<S["Type"]> | undefined {
   return schema.ast.context?.annotations
 }
-/** @internal */
+// Alias for the top-level `Constraint`, which `Annotations.ToArbitrary.Constraint` shadows below.
 type AnnotationSchemaConstraint = Constraint
 
 /**
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -62,8 +62,6 @@ import * as Record_ from "./Record.ts"
 import * as Redacted_ from "./Redacted.ts"
 import * as RegExp_ from "./RegExp.ts"
 import * as Result_ from "./Result.ts"
-// oxlint-disable-next-line import/no-self-import -- Qualifies Constraint where Annotations.ToArbitrary.Constraint shadows it.
-import type * as Schema_ from "./Schema.ts"
 import * as SchemaAST from "./SchemaAST.ts"
 import * as SchemaGetter from "./SchemaGetter.ts"
 import * as SchemaIssue from "./SchemaIssue.ts"
@@ -15300,7 +15298,7 @@ export declare namespace Annotations {
      *
      * @since 4.0.0
      */
-    readonly arbitraryConstraint?: ToArbitrary.Constraint<any> | undefined
+    readonly arbitraryConstraint?: ToArbitrary.FilterConstraint<any> | undefined
     /**
      * Marks the filter as *structural*, meaning it applies to the shape or
      * structure of the container (e.g., array length, object keys) rather than
@@ -15366,7 +15364,7 @@ export declare namespace Annotations {
      * @category models
      * @since 4.0.0
      */
-    interface Constraint<T = unknown> extends GenerationConstraint<T> {
+    interface FilterConstraint<T = unknown> extends GenerationConstraint<T> {
       readonly order?: Order.Order<T> | undefined
     }
     /**
@@ -15375,7 +15373,7 @@ export declare namespace Annotations {
      * @category models
      * @since 4.0.0
      */
-    interface DeclarationInput<T, Parameters extends ReadonlyArray<Schema_.Constraint>> {
+    interface DeclarationInput<T, Parameters extends ReadonlyArray<Constraint>> {
       readonly typeParameters: TypeParameters.Type<Parameters>
       readonly constraint: GenerationConstraint<T> | undefined
     }
@@ -15385,7 +15383,7 @@ export declare namespace Annotations {
      * @category models
      * @since 4.0.0
      */
-    interface Declaration<T, Parameters extends ReadonlyArray<Schema_.Constraint>> {
+    interface Declaration<T, Parameters extends ReadonlyArray<Constraint>> {
       (input: DeclarationInput<T, Parameters>): SchemaAST.Link
     }
   }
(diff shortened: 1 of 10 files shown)


packages/effect/src/Schema.ts
```diff
@@ -14980,7 +14980,7 @@ export function resolveAnnotations<S extends Constraint>(
 export function resolveAnnotationsKey<S extends Constraint>(schema: S): Annotations.Key<S["Type"]> | undefined {
   return schema.ast.context?.annotations
 }
-/** @internal */
+// Alias for the top-level `Constraint`, which `Annotations.ToArbitrary.Constraint` shadows below.
 type AnnotationSchemaConstraint = Constraint
```
gcanti: `AnnotationSchemaConstraint` seems completely unnecessary. I think we can remove it.

PR #8168, opened 2026-09-10
feat(Schema): add UndefinedOrFromNullOr

davetorbeck:
## What

Adds `Schema.UndefinedOrFromNullOr(schema)` and `SchemaTransformation.undefinedOrFromNullOr()`.

The schema is `NullOr(schema)` decoded to `UndefinedOr(toType(schema))`. Decoding maps `null` to `undefined`. Encoding maps `undefined` to `null`. Other values pass through in both directions. It follows the `OptionFromNullOr` / `OptionFromUndefinedOr` / `OptionFromNullishOr` family in shape, naming, JSDoc and placement.

Includes a unit test with `TestSchema.Asserts` (decoding, encoding, arbitrary generation), a `tstyche` type test for the wrapper type and its `Type` / `Encoded` sides, doc examples that run under `pnpm doctest`, and a `patch` changeset for `effect`.

`pnpm check` for `packages/effect`, the Schema unit test file, `pnpm test-types` for the Schema type tests, `pnpm doctest` for both modules, `oxlint` and `dprint check` all pass.

## Why

JSON has no `undefined`. `JSON.stringify` drops a key whose value is `undefined`, so a payload with a stable shape (a health probe, a status response) must carry `null` for absence on the wire. Many programs model absence as `T | undefined` rather than `Option<T>`, following the `Array.find` / `Map.get` idiom. Today that combination has no first-class codec: `NullOr` is `null` on both sides, `UndefinedOr` is `undefined` on both sides, `optional` / `optionalKey` drop the key, and `OptionFromNullOr` requires `Option` in the domain. The user-land answer is a per-field helper such as `jsonNull(value)` at every call site, which is exactly the kind of manual boundary code Schema exists to replace, or a hand-written `encodeTo` with two getters in every codebase. This PR provides the codec once, in the library, with the same name pattern as the `Option` variants so it is discoverable next to them.

https://claude.ai/code/session_01XpmvW3MzCahnWN6PD1UCsc

### DIFF
--- a/.changeset/schema-undefined-or-from-null-or.md
+++ b/.changeset/schema-undefined-or-from-null-or.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add `Schema.UndefinedOrFromNullOr` and `SchemaTransformation.undefinedOrFromNullOr`, the `T | undefined` counterpart of `Schema.OptionFromNullOr`: decoding maps `null` to `undefined`, encoding maps `undefined` back to `null`. Use it when a program models absence as `undefined` but the wire (JSON) must carry `null`, so the codec lives in the payload schema rather than at every call site.
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -12963,6 +12963,55 @@ export function OptionFromNullishOr<S extends Constraint>(
 }
 
 /**
+ * Type-level representation returned by {@link UndefinedOrFromNullOr}.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export interface UndefinedOrFromNullOr<S extends Constraint> extends decodeTo<UndefinedOr<toType<S>>, NullOr<S>> {
+  readonly "Rebuild": UndefinedOrFromNullOr<S>
+}
+/**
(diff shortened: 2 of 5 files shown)


spencerbeggs:
@davetorbeck Very nice! I have been re-implementing this by hand in different projects. It would be a great first-class citizen, imo.

gcanti:
For serialization to and deserialization from JSON, the idiomatic approach in v4 is not to worry about it at all and simply use `toCodecJson`:

```ts

const domainModel = Schema.UndefinedOr(Schema.FiniteFromString)

const codec = Schema.toCodecJson(domainModel)

const decode = Schema.decodeExit(codec)
const encode = Schema.encodeExit(codec)

console.log(String(encode(1))) // Success("1")
console.log(String(encode(undefined))) // Success(null)

console.log(String(decode("1"))) // Success(1)
console.log(String(decode(null))) // Success(undefined)
```

video clip: https://x.com/GiulioCanti/status/2090409195807903924

spencerbeggs:
@gcanti Thank you again for another explainer. That video was helpful. What program did you used to create it? It looked slick.

gcanti:
@spencerbeggs https://www.remotion.dev/

PR #8171, opened 2026-09-10
add `Context.Mixin`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

https://discord.com/channels/795981131316985866/847382157861060618/1546595868386402394

`Context.Mixin` wraps an existing class constructor as a `Context.Service` key without moving that constructor onto a `Context.Service` subclass. Constructor parameters, instance methods, and instanceof checks against the base class are preserved. The mixed class is the service key (`yield*`, `Context.get`, `use / useSync`); instances are not `Effects`.

The only option is `{ fiberCached }`, matching `Service`. `Mixin` does not take `make`; construct instances with new or a static helper on the class.

```ts
import { Context, Effect } from "effect"
class Box {
  constructor(readonly value: number) {}
  double() { return this.value * 2 }
}

class MyText extends Context.Mixin<MyText>("MyText")(Box) {}

const context = Context.make(MyText, new MyText(1))

Context.get(context, MyText).double() // => 2

  const text = yield* MyText
  return text.value
}).pipe(Effect.provideService(MyText, new MyText(2)))
```
## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/context-mixin.md
+++ b/.changeset/context-mixin.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add `Context.Mixin` to wrap an existing class constructor as a `Context.Service` key, preserving the original constructor and instance members.
--- a/packages/effect/src/Context.ts
+++ b/packages/effect/src/Context.ts
@@ -115,6 +115,7 @@ export interface Service<in out Identifier, in out Shape> extends Key<Identifier
  * The class itself is the `Context` key, and its string `key` identifies the
  * service at runtime.
  *
+ * @see {@link Mixin} for wrapping an existing class constructor
  * @see {@link Service} for creating function-style keys or class-style service keys
  *
  * @category services
@@ -193,6 +194,7 @@ export declare namespace ServiceClass {
  * Context.get(config, Config).port // => 8080
  * ```
  *
+ * @see {@link Mixin} for wrapping an existing class constructor
  * @see {@link Reference} for service keys with default values
  *
  * @category services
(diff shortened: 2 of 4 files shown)


tim-smart:
Not sure we need this. Trying to avoid multiple ways of defining services.

KhraksMamtsov:
@tim-smart 
I have a few arguments of varying degrees of liquidity :)

* I’m currently seeing a trend toward using classes for opaque types in the ecosystem, for example with `Schema.Class` and `Data.Class`. Mixins seem like a reasonable approach to composition when working with classes.
* I also think `Context.Mixin` can be viewed as a kind of interop mechanism: given an existing plain class, you can quickly lift it into the ecosystem’s context.
* Plus, @harrysolovay came up with this use case. Maybe he has some other arguments in favor of the feature as well :)

```ts
type Payload_ = typeof Payload_.Type
const Payload_ = S.Struct({
  x402Version: Version,
  accepted: Requirements,
  extensions: ExtensionEnvelopes.pipe(S.optional),
  payload: S.Record(S.String, S.Unknown),
  resource: ResourceInfo.pipe(S.optional),
})

// oxlint-disable-next-line typescript/no-empty-interface
export interface Payload extends Payload_ {}

export const Payload = Object.assign(Context.Service<Payload, Payload>("crosshatch/Payload"), Payload_)

// vs

class Payload extends Context.Mixin<Payload>("crosshatch/Payload")(
    Schema.Class<Payload>("crosshatch/Payload")({
      id: Schema.String
	 // ...
    })
) { }
```

tim-smart:
Hmm doesn't seem convincing enough in my opinion. I think we should stick with a single way of defining services.

KhraksMamtsov:
Yes, I understand.

In any case, it seems that Mixin can be declared in userland code, as all the necessary pieces appear to be exported.

Closing the feature.

PR #8172, opened 2026-09-10
fix(Match): infer valueTags input only from the value (#8167)

### LINKED ISSUE #8167: Match.valueTags loses types and exhaustiveness check

rc 113

```typescript
import { Schema, Match } from "effect"

const Vehicles = Schema.Union([
  Schema.TaggedStruct("Car", {}),
  Schema.TaggedStruct("Bike", {})
])

const myVehicle: typeof Vehicles.Type = {_tag: "Bike"}

const result = Match.valueTags(myVehicle, {
  Plane: () => "plane", // no type error, no exhaustiveness check
})

console.log(result)
```

https://effect.website/play#3f2a2931ea29

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

xia-chao:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`Match.valueTags(value, handlers)` was not type-checking the handler object.

If you passed a `Car | Bike` value and wrote `{ Plane: () => "plane" }`, TypeScript said nothing. At runtime it still blew up with `effect/match/Match/exhaustive: absurd`, because there is no `Plane` handler and the `Bike` case is missing.

The type parameter for the value was being inferred from both arguments. The handlers could drag that type around, so the exhaustiveness check never ran against the real input type.

This change only affects the two-argument form. The input type is now taken from the value (`Types.NoInfer`), then the handlers are checked against that. Unknown tags and missing tags should both error. Runtime behavior is unchanged.

## Verification

- `tstyche packages/effect/typetest/Match.tst.ts --target '>=5.9'` — 22 passed on TypeScript 5.9 and 6.0
- `vitest run packages/effect/test/Match.test.ts` — 5 passed

Changeset included for `effect` (patch).

## Related

- Fixes #8167

### DIFF
--- a/.changeset/fix-match-value-tags-inference.md
+++ b/.changeset/fix-match-value-tags-inference.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Lock `Match.valueTags` input inference so unknown tags and missing handlers are rejected.
--- a/packages/effect/src/Match.ts
+++ b/packages/effect/src/Match.ts
@@ -446,11 +446,11 @@ export const valueTags: {
   <
     const I,
     P extends
-      & ValueTagHandlers<I>
-      & { readonly [Tag in Exclude<keyof P, Types.Tags<"_tag", I>>]: never }
+      & ValueTagHandlers<T.NoInfer<I>>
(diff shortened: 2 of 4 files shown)


tim-smart:
Already fixed on main

PR #8177, opened 2026-09-10
Support TTL callbacks in Effect.cachedWithTTL

tim-smart:
`Effect.cachedWithTTL` now accepts a callback from `Exit<A, E>` to `Duration.Input`, allowing successes and failures to use different cache durations. Both data-first and piped calls infer callback types and preserve the effect's success, error, and service types. Parameterless and explicitly typed callbacks can return duration strings, and union-typed TTL inputs remain supported.

The callback runs once per completed computation, with expiration measured from completion. Pending callers share the computation, cache hits reuse the selected TTL, and zero and infinite durations are supported. Fixed-duration calls retain their existing behavior. Public documentation includes a callback example and explains caching interruption exits. A changeset is included.

Validation (through `nix develop -c`):

- `pnpm test --run packages/effect/test/Effect.test.ts`: 261 passed.
- `pnpm test-types packages/effect/typetest/Effect.tst.ts`: 276 passed across TypeScript 5.9.3 and 6.0.3.
- A temporary compiler probe reproduced the original overload failure and verified callback forms, exact inferred effect types, union inputs, and fixed durations on both versions.
- `pnpm doctest --run packages/effect/src/Effect.ts`: 241 passed.
- `pnpm lint-fix`, `pnpm lint`, `pnpm check`, and `pnpm jsdocs --check`: passed.

Closes EFF-1340


packages/effect/src/internal/effect.ts
```diff
@@ -4393,18 +4397,24 @@ export const cachedInvalidateWithTTL: {
...
         exit = undefined
-        return onExit(self, (exit_) =>
-          sync(() => {
-            running = false
-            expiresAt = clock.currentTimeMillisUnsafe() + ttlMillis
-            exit = exit_
-            latch.openUnsafe()
-          }))
+        return onExit(
+          onExit(self, (exit_) =>
```
tim-smart: Why the double `onExit`?

PR #8190, opened 2026-09-11
fix(http): omit body in toWeb for 204, 205, and 304

### LINKED ISSUE #8189: HttpServerResponse.toWeb throws on Node for 204/205/304 when the response still has a body
## What happens

On **Node** (undici), `HttpServerResponse.toWeb(...)` blows up if the status is **204, 205, or 304** and the Effect response still has a body (`text` / `raw` / `stream` / `uint8Array`).

The error is:

```
TypeError: Response constructor: Invalid response status code 204
```

`empty({ status: 204 })` is fine. `toWeb(..., { withoutBody: true })` is also fine. **Bun does not throw** (it keeps the body). So this is easy to miss if you only try Bun.

Fetch / undici do not allow a body on 204 / 205 / 304. `toWeb` currently picks the body from `_tag` and always passes it into `new Response(...)`. It does not strip the body for those statuses.

## How to reproduce (Node, not Bun)

1. Use current `main`.
2. Run this with **Node**:

```js
import { HttpServerResponse } from "effect/unstable/http"

const emptyOk = HttpServerResponse.toWeb(
  HttpServerResponse.empty({ status: 204 })
)
console.log("empty", emptyOk.status, emptyOk.body)

HttpServerResponse.toWeb(
  HttpServerResponse.text("still a body", { status: 204 })
)
```

3. First call works (`body` is `null`). Second call throws.

Same throw for `raw` / `stream` / `uint8Array` with status 204 / 205 / 304.

## Screenshots (debugger on Node)

**1. `empty({ status: 204 })` — `_tag` is Empty, body stays empty, constructor is happy**

empty 204 ok

**2. `text(..., { status: 204 })` — `_tag` is Uint8Array, still about to pass a body into `new Response`**

text 204 before throw

**3. Debug console — empty works, text throws `Invalid response status code 204`**

console throw

## Expected

`toWeb` should give a Web `Response` that Node accepts: status stays 204 / 205 / 304, **body is empty** (same as `empty` / `withoutBody: true`).

## Where

`packages/effect/src/unstable/http/HttpServerResponse.ts` → `toWeb`

xia-chao:
Closes #8189
Closes EFF-1356

HTTP responses with status 204, 205, or 304 now omit their bodies in `HttpServerResponse.toWeb` and the native Bun/Deno adapters. This prevents rejected Web Response construction and hung Deno requests.

The shared `HttpServerResponse.omitsBody` predicate also governs request-scope ownership in `HttpEffect`. Omitted Effect streams never start, request finalizers run, and omitted raw `ReadableStream` bodies are cancelled. The adapters preserve response headers when constructing the Web response, including HEAD and 304 representation metadata.

Includes patch changeset entries for `effect`, `@effect/platform-bun`, and `@effect/platform-deno`.

Validation, all through `nix develop -c`:

- `pnpm test --run packages/effect/test/unstable/http/HttpServerResponse.test.ts packages/effect/test/unstable/http/HttpEffect.test.ts`: 55 passed.
- `bun run --bun vitest run packages/platform/bun/test/BunHttpServer.test.ts`: 23 passed.
- `deno task test --run packages/platform/deno/test/DenoHttpServer.test.ts`: 39 passed.
- `pnpm lint-fix`, `pnpm check`, and `deno check packages/platform/deno/src/DenoHttpServer.ts packages/platform/deno/test/DenoHttpServer.test.ts`: passed.
- Temporary probes confirmed that omitted text/JSON bodies retain lazy encoding in `toWeb` and the Deno adapter for HEAD/204/205/304, and that `toWeb` succeeds when the `ReadableStream` global is absent. Both issues were reproduced before the guard fix.

Coverage verifies body omission, request finalization, raw-stream cancellation, HEAD headers, and direct-conversion 304 headers. Native-adapter tests avoid asserting a 304 Content-Length value because Bun and Deno rewrite it during HTTP serialization.

### DIFF
--- a/.changeset/fix-http-toweb-null-body-status.md
+++ b/.changeset/fix-http-toweb-null-body-status.md
@@ -4,4 +4,4 @@
 "@effect/platform-deno": patch
 ---
 
-Stop `HttpServerResponse.toWeb` from passing a body into the Web `Response` constructor for status 204, 205, and 304. Node/undici rejects those statuses when a body is present; empty and `{ withoutBody: true }` already worked.
+Omit response bodies for statuses 204, 205, and 304 in `HttpServerResponse.toWeb` and the Bun/Deno HTTP adapters, preventing invalid Web responses and hung requests. Cancel omitted raw `ReadableStream` bodies, and finalize request resources without starting omitted Effect streams.
--- a/.changeset/fix-http-toweb-null-body-status.md
+++ b/.changeset/fix-http-toweb-null-body-status.md
@@ -1,5 +1,7 @@
 ---
 "effect": patch
+"@effect/platform-bun": patch
+"@effect/platform-deno": patch
 ---
 
 Stop `HttpServerResponse.toWeb` from passing a body into the Web `Response` constructor for status 204, 205, and 304. Node/undici rejects those statuses when a body is present; empty and `{ withoutBody: true }` already worked.
--- a/.changeset/fix-http-toweb-null-body-status.md
+++ b/.changeset/fix-http-toweb-null-body-status.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Stop `HttpServerResponse.toWeb` from passing a body into the Web `Response` constructor for status 204, 205, and 304. Node/undici rejects those statuses when a body is present; empty and `{ withoutBody: true }` already worked.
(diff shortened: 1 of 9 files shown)


xia-chao:
@tim-smart 
Please review

PR #8191, opened 2026-09-11
fix(http): keep outer status when toWeb wraps a raw Response

xia-chao:
`HttpServerResponse.toWeb` used to copy headers onto an inner Web `Response` and then return that same object. A Web `Response` cannot change its status after it is created, so a call like `raw(inner201, { status: 404 })` still came out as 201.

`toWeb` now builds a new Web `Response`: outer status and statusText, inner body, outer headers laid on top of the inner headers.

Includes a patch changeset for `effect`.

Validation:

- `vitest run packages/effect/test/unstable/http/HttpServerResponse.test.ts`: 9 passed


xia-chao:
@tim-smart 
Please review

tim-smart:
This behaviour is intentional the the raw response should be sent untouched.

PR #8194, opened 2026-09-11
feat(sql-pg): add startupParameters to connection config

### LINKED ISSUE #8193: feat(sql-pg): add support for startupParameters in connection config

After migrating to the native PostgreSQL driver introduced in `@effect/sql-pg@4.0.0-rc.113`, we lost the ability to configure session defaults during connection startup.

Previously, we supplied `statement_timeout`, `random_page_cost`, and `search_path` through the connection URL's `options` parameter. The node-postgres-backed implementation [forwarded these options](https://github.com/brianc/node-postgres/blob/458903894f3514465b74d623f6b92f713b6b952f/packages/pg/lib/client.js#L567-L569) in PostgreSQL's startup packet.

The native driver's [startup call](https://github.com/Effect-TS/effect/blob/4f73f9e0bd9538260f0f7166e0e6abdbd507a0e8/packages/sql/pg/src/PgConnection.ts#L2017-L2025) only forwards `user`, `database`, and `application_name`, with no public configuration for other startup parameters.

We currently work around this with parameterized `set_config` calls after acquiring a connection. That adds a round trip and doesn't establish the same session defaults for `RESET`.

Expose a structured `startupParameters` field on the native connection/client configuration, for example:

```ts
PgClient.layer({
  url,
  startupParameters: {
    statement_timeout: "10s",
    random_page_cost: "4",
    search_path: "app,public",
  },
})
```

A `Readonly<Record<string, string>>` would accommodate PostgreSQL's built-in settings and extension-defined settings without requiring a driver-specific field for each one.

The resolved parameters would be passed to `PgProtocol.encodeStartupMessage` for every new physical connection, including pooled replacements and standalone connections. Existing connection identity fields and driver-required encoding settings should have clearly defined precedence or be protected from overriding.

The protocol support appears to already exist: [`StartupParameters`](https://github.com/Effect-TS/effect/blob/4f73f9e0bd9538260f0f7166e0e6abdbd507a0e8/packages/sql/pg/src/PgProtocol.ts#L1088-L1100) accepts arbitrary string-valued parameters, and [the encoder writes them into the startup packet](https://github.com/Effect-TS/effect/blob/4f73f9e0bd9538260f0f7166e0e6abdbd507a0e8/packages/sql/pg/src/PgProtocol.ts#L1137-L1153). This would expose that capability through the connection configuration.

PostgreSQL's [StartupMessage specification](https://www.postgresql.org/docs/18/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-STARTUPMESSAGE) explicitly supports individual runtime parameters and recommends them over the command-line-style `options` string. They become session defaults during startup, without an additional SQL round trip.

- **Forwarding `options?: string`:** this would restore compatibility with existing connection URLs, but a structured `startupParameters` map is preferable for new code. It avoids constructing and escaping strings such as `-cstatement_timeout=10s -csearch_path=app,public`. URL `options` support could still be added separately for compatibility.
- **Application-owned `set_config` initialization:** our current workaround. It adds a query and connection lifecycle handling, and SQL-applied settings differ from startup defaults when [`RESET`](https://www.postgresql.org/docs/18/sql-reset.html) is used.
- **Role/database defaults:** useful for fixed deployments, but they move application-specific configuration into database administration and don't provide per-client settings.

eugene1g:
Closes https://github.com/Effect-TS/effect/issues/8193

Expose `startupParameters` on `PgConnection.Config` and `PgClientConfig`, allowing callers to send session defaults in PostgreSQL's startup packet without additional SQL.

```ts
PgClient.layer({
  url,
  startupParameters: {
    statement_timeout: "10s",
    random_page_cost: "4",
    search_path: "app,public"
  }
})
```

The native driver's protocol encoder already supports arbitrary startup parameters, but the connection configuration previously only forwarded identity and application name. This adds the missing configuration path; `PgPool.Config` inherits it automatically. No protocol encoder, query execution, or pool lifecycle changes are required.

## Behavior

- Parameters are forwarded when each physical connection starts, including replacement pooled connections.
- PostgreSQL applies them as session defaults, so `RESET ALL` restores them.
- `user` and `database` remain reserved; callers use the existing identity options.
- Application-name precedence is explicit `applicationName`, then `startupParameters.application_name`, then the URL value, then the driver default. An explicitly empty name is preserved.
- `client_encoding` accepts `UTF8` and `UTF-8`, case-insensitively, and sends canonical `UTF8`. Other encodings fail before connecting because the driver encodes and decodes text as UTF-8.
- Startup parameter names are normalized to lowercase.
- Empty parameter names and NUL bytes in names or values fail before connecting. Other setting validation remains PostgreSQL's responsibility.
- No legacy URL `options` parsing or environment-variable behavior is added.
- Includes an `@effect/sql-pg` patch changeset.

## References

- [PostgreSQL startup parameters](https://www.postgresql.org/docs/18/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-STARTUPMESSAGE)
- [PostgreSQL RESET semantics](https://www.postgresql.org/docs/18/sql-reset.html)


eugene1g:
I think I'll rework this PR to restore full support for PG `options` in the new native driver, for better ecosystem compatibility.

eugene1g:
This was reimplemented in https://github.com/Effect-TS/effect/pull/8234

PR #8202, opened 2026-09-11
fix(RcRef): keep idleTimeToLive when the value is 0

xia-chao:
`RcRef.make({ idleTimeToLive: 0 })` now preserves numeric zero as a duration, matching `Duration.zero` and `"0 millis"` in RcRef.

When the last reference to a resource that has not been invalidated is released, a zero TTL runs the zero-duration sleep and resource release in a forked fiber. The scope releasing that reference does not await the release. If `idleTimeToLive` is omitted or the resource has been invalidated with `RcRef.invalidate`, that scope releases the resource and awaits completion.

The regression verifies both the zero-duration sleep and completed resource release. All 8 tests in `packages/effect/test/RcRef.test.ts` pass, and restoring the original truthiness check makes the regression fail.

Closes EFF-1358

### DIFF
--- a/.changeset/fix-rcref-idle-ttl-zero.md
+++ b/.changeset/fix-rcref-idle-ttl-zero.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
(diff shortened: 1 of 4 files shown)


xia-chao:
@tim-smart 
Please review

PR #8207, opened 2026-09-12
fix(http): fail active multipart file parts on upstream body failure

### LINKED ISSUE #8203: BunMultipart.stream hangs when the request body errors during an active file

- `effect@4.0.0-rc.115`
- `@effect/platform-bun@4.0.0-rc.115`
- Bun `1.4.2`, Linux x64 (also reproduced on Bun `1.3.13`)

Create an empty directory with this `package.json`:

```json
{
  "private": true,
  "type": "module",
  "dependencies": {
    "effect": "4.0.0-rc.115",
    "@effect/platform-bun": "4.0.0-rc.115"
  }
}
```

Save the following as `repro.mjs`. The request supplies a valid multipart file header and five file bytes, then its body stream errors while that file is active. `highWaterMark: 0` keeps the error on the next pull rather than prefetching it before the buffered data is consumed.

```js
import { Effect, Stream } from "effect";
import * as BunMultipart from "@effect/platform-bun/BunMultipart";

const body = new ReadableStream({
  start(controller) {
    controller.enqueue(new TextEncoder().encode(
      '--b\r\nContent-Disposition: form-data; name="file"; filename="a.txt"\r\n\r\nhello',
    ));
  },
  pull(controller) {
    controller.error(new Error("body-read-failed"));
  },
}, { highWaterMark: 0 });

const request = new Request("http://localhost/upload", {
  method: "POST",
  headers: { "content-type": "multipart/form-data; boundary=b" },
  body,
});

let bytesRead = 0;
const result = await Effect.runPromiseExit(
  BunMultipart.stream(request).pipe(
    Stream.runForEach((part) => part._tag === "File"
      ? Stream.runForEach(part.content, (chunk) => Effect.sync(() => { bytesRead += chunk.length; }))
      : Effect.void),
    Effect.timeout("1 second"),
  ),
);

console.log({ bun: Bun.version, bytesRead });
console.log(JSON.stringify(result, null, 2));
```

Run:

```sh
bun install --ignore-scripts
bun --no-install repro.mjs
```

Consuming the active file should promptly fail with the request-body read error, wrapped in `MultipartError`. The outer multipart-consumption effect should fail without reaching the watchdog timeout.

### What do you see instead?

Five file bytes are consumed, then the operation fails only when the outer timeout expires:

```text
{ bun: "1.4.2", bytesRead: 5 }
```

The resulting failure is `TimeoutError`, with the message `Operation timed out after '1s'`, rather than the injected `body-read-failed` error.

As a control, consuming an identically constructed body through `BunStream.fromReadableStream` propagates the read error without timing out.

### Additional information

In [`Multipart.makeChannel`](https://github.com/Effect-TS/effect/blob/8feeb3c4cdbd0f2e57fa25bb0d6fe3b9509117db/packages/effect/src/unstable/http/Multipart.ts#L479-L543), the file-content pull loop checks buffered chunks and `finished`. `pump` records an upstream failure in `exit`, but the nested file loop does not check that value before pumping again. The outer part loop checks it, but the consumer is still awaiting the active file.

Related history: #7155 proposed propagating upstream failures to active files and was closed in favor of #7156. The latter addresses parser limits/truncated input. This reproduction uses an explicit request-body read failure with no configured parser limits and fails before any closing boundary is received. #6284 is a related limit-triggered hang report.

mnkprs:
Fixes #8203

## Summary

- fail an active multipart file part when the request body stream errors, instead of pumping a dead upstream forever
- normalize the upstream cause to a `MultipartError` so `File.content` and `File.contentEffect` keep their declared error type
- add regression tests for a failing body during an active file, for `contentEffect`, and for a file part that is never consumed
- add a patch changeset

## Root cause

In `Multipart.makeChannel`, `pump` records an upstream failure in `exit` and the outer part loop checks it, but the per-file pull loop created in `onFile` only checked its own `finished` flag:

```ts
return finished ? Cause.done() : Effect.flatMap(pump, loop)
```

The parser is never told about an upstream failure (`parser.end()` only runs on `Done`), so it cannot send the active file callback its terminating `null`. `finished` therefore stays `false` and a consumer draining `part.content` calls `pump` on an already-failed upstream forever. The loop never yields, so the failure surfaces only when an outer watchdog fires, and in the tests here even `Effect.timeout` inside the same fiber never gets a chance to run.

This is a different trigger from #7155 / #7156. Those covered parser limit and truncated-input failures, which the parser can observe and which it now terminates by sending `null` to the active file. An upstream stream failure is invisible to the parser, so it has to be handled in the wrapper.

## Change

`makeChannel` keeps the upstream cause in a new `upstreamFailure` slot next to the existing `exit`, mapped so a cause that is already a `MultipartError` passes through and anything else becomes `InternalError`. The file pull loop drains any buffered chunks first, then ends on `finished`, and otherwise fails with that cause rather than pumping again.

Because the file channel can now fail, its error type widens from `never` to `MultipartError`. This is internal only: the public `File.content` and `File.contentEffect` were already typed as failing with `MultipartError`. `FileImpl.contentEffect` therefore no longer needs to wrap the channel error, and `defaultWriteFile` now passes an existing `MultipartError` through instead of re-wrapping it as `InternalError`.

A file that completed before the failure still ends cleanly, so the parser-limit behavior from #7156 is unchanged.

## Validation

- `pnpm test --run packages/effect/test/unstable/http/Multipart.test.ts` (19 tests)
- the two new consuming tests hang and time out on `main`, and pass with the change
- `pnpm test --run packages/effect/test/unstable/http/HttpServerRequest.test.ts packages/effect/test/unstable/http/HttpServer.test.ts packages/effect/test/unstable/http/HttpPlatform.test.ts`
- `pnpm test --run packages/platform/node/test/MultipartParser.test.ts packages/platform/node/test/HttpApi.test.ts` (128 tests)
- `pnpm lint`, `pnpm check`

The issue reproduces through `@effect/platform-bun`, but the hang is in the shared `effect/unstable/http/Multipart` channel that `BunMultipart.stream` pipes through, so the regression tests live with the core module. Bun was not available locally to run the reporter's script directly.

### DIFF
--- a/.changeset/fix-multipart-upstream-failure.md
+++ b/.changeset/fix-multipart-upstream-failure.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fail active multipart file parts when the request body stream errors, instead of hanging until the consumer times out.
--- a/packages/effect/src/unstable/http/Multipart.ts
+++ b/packages/effect/src/unstable/http/Multipart.ts
@@ -281,6 +281,8 @@ export class MultipartError extends Data.TaggedError("MultipartError")<{
   }
 }
 
+const isMultipartError = (u: unknown): u is MultipartError => Predicate.hasProperty(u, MultipartErrorTypeId)
+
 /**
  * Schema type for persisted multipart files.
  *
@@ -469,6 +471,7 @@ export const makeChannel = <IE>(headers: Record<string, string>): Channel.Channe
     Effect.map(makeConfig(headers), (config) => {
       let partsBuffer: Array<Part> = []
       let exit = Option.none<Exit.Exit<never, IE | MultipartError | Cause.Done>>()
+      let upstreamFailure = Option.none<Cause.Cause<MultipartError>>()
       let ended = false
 
       const parser = MP.make({
@@ -480,9 +483,19 @@ export const makeChannel = <IE>(headers: Record<string, string>): Channel.Channe
           let chunks: Array<Uint8Array> = []
           let finished = false
           const pullChunks = Channel.fromPull(
-            Effect.succeed(Effect.suspend(function loop(): Pull.Pull<Arr.NonEmptyReadonlyArray<Uint8Array>> {
+            Effect.succeed(Effect.suspend(function loop(): Pull.Pull<
+              Arr.NonEmptyReadonlyArray<Uint8Array>,
+              MultipartError
+            > {
               if (!Arr.isReadonlyArrayNonEmpty(chunks)) {
-                return finished ? Cause.done() : Effect.flatMap(pump, loop)
+                if (finished) {
+                  return Cause.done()
+                }
+                // The parser never sees an upstream failure, so it cannot end
+                // this file. Fail the part instead of pumping a dead upstream.
+                return Option.isSome(upstreamFailure)
+                  ? Effect.failCause(upstreamFailure.value)
+                  : Effect.flatMap(pump, loop)
               }
               const chunk = chunks
               chunks = []
@@ -523,6 +536,12 @@ export const makeChannel = <IE>(headers: Record<string, string>): Channel.Channe
             }
           } else {
             exit = Option.some(Exit.failCause(cause)) as any
+            upstreamFailure = Option.some(
+              Cause.map(
+                cause,
+                (error) => isMultipartError(error) ? error : MultipartError.fromReason("InternalError", error)
+              )
+            )
           }
           return Effect.void
         })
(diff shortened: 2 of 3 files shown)


mnkprs:
The `Types` failure looks like a runner OOM rather than a type error: `tsc -b tsconfig.json` printed only `Killed` and exited 137, with no diagnostic, and the `deno check` / `test-types` steps were skipped.

The same job failed identically on a push to `main` before this PR existed ([run 34657889492](https://github.com/Effect-TS/effect/actions/runs/34657889492)). `pnpm check` passes locally on this commit, and every other job here is green, including all four test runtimes.

Could someone re-run that job when convenient?

mnkprs:
Superseded by #8206, which landed the same fix. Closing.

PR #8208, opened 2026-09-12
fix(openapi-generator): type binary multipart fields as File | Blob

### LINKED ISSUE #8204: Generated multipart clients type binary file fields as string

- `effect@4.0.0-rc.115`
- `@effect/openapi-generator@4.0.0-rc.115`

With these versions installed, save the following as `openapi.json`:

```json
{
  "openapi": "3.0.3",
  "info": { "title": "Reproduction", "version": "1.0.0" },
  "paths": {
    "/upload": {
      "post": {
        "operationId": "upload",
        "requestBody": {
          "required": true,
          "content": {
            "multipart/form-data": {
              "schema": {
                "type": "object",
                "properties": {
                  "file": { "type": "string", "format": "binary" }
                },
                "required": ["file"],
                "additionalProperties": false
              }
            }
          }
        },
        "responses": {
          "204": { "description": "Uploaded" }
        }
      }
    }
  }
}
```

Generate the client:

```sh
bun run openapigen --spec openapi.json --name Client --format httpclient > client.ts
```

Use the generated payload type with a file:

```ts
import type { UploadRequestFormData } from "./client.ts";

const payload: UploadRequestFormData = {
  file: new File(["hello"], "hello.txt"),
};
```

The generated multipart payload type and schema accept a `File` or `Blob` for the binary file field, allowing a file upload without a type assertion or generated-code patch.

### What do you see instead?

The generator emits a string field:

```ts
export type UploadRequestFormData = {
  readonly file: string;
};

export const UploadRequestFormData = Schema.Struct({
  file: Schema.String.annotate({ format: "binary" }),
});
```

The example therefore fails type checking because `File` is not assignable to `string`. The `binary` format is retained as an annotation on `Schema.String`.

The generated operation serializes this payload using:

```ts
HttpClientRequest.bodyFormDataRecord(options.payload as any);
```

Supplying a string satisfies the generated type but produces a text form field rather than a file part.

### Additional information

`bodyFormDataRecord` already supports native `File` and `Blob` values. As a control, passing `{ file: new File(["hello"], "hello.txt") }` directly to `HttpBody.formDataRecord` preserves the file contents and filename. The generated payload type prevents using that support.

Related: #7769 and #7770 corrected the constructor used for record-based multipart payloads. The remaining problem here is the generated binary-file type.

xia-chao:
## Summary

Generated HTTP clients emitted `string` for `multipart/form-data` fields that use
`format: binary`. `HttpClientRequest.bodyFormDataRecord` accepts `File` / `Blob`
at runtime, so a `File` payload was rejected by the generated type
(`TS2322: Type 'File' is not assignable to type 'string'`) even though it worked
when sent.

Client formats now replace those fields with a generated
`globalThis.File | globalThis.Blob` type by reusing the existing multipart schema
substitution. `--format httpapi` is unchanged.

Closes #8204
Closes EFF-1365

## Before / after (`--format httpclient`)

```ts
// before
export type UploadRequestFormData = { readonly "file": string }

// after
export type UploadRequestFormData = { readonly "file": __ClientMultipartFile }
```

## Test plan

- [x] `vitest --run packages/tools/openapi-generator/test/OpenApiGenerator.test.ts` (36 passed)
- [x] New regression test `types multipart binary fields as File | Blob for generated clients`
      compiles usage assigning `File` and `Blob` to the payload and rejecting strings for both
      `httpclient` and `httpclient-type-only`
- [x] `pnpm check`
- [x] `pnpm lint-fix`

### DIFF
--- a/.changeset/multipart-binary-client-file-type.md
+++ b/.changeset/multipart-binary-client-file-type.md
@@ -1,0 +1,7 @@
+---
+"@effect/openapi-generator": patch
+---
+
+Generated HTTP clients now type `multipart/form-data` fields that use `format: binary` as `globalThis.File | globalThis.Blob`.
+
+A binary field was previously emitted as `string`, so passing a `File` (the value accepted by `HttpClientRequest.bodyFormDataRecord`) failed to type-check even though it worked at runtime. `httpapi` output is unchanged.
--- a/packages/tools/openapi-generator/src/JsonSchemaGenerator.ts
+++ b/packages/tools/openapi-generator/src/JsonSchemaGenerator.ts
@@ -35,8 +35,19 @@ interface MultipartSchemaRefs {
   readonly files: string
 }
 
-interface GenerateHttpApiOptions extends GenerateOptions {
+interface GenerateHttpApiOptions extends GenerateOptions, RenderSchemaTypeAndRuntimeOptions {
   readonly multipartSchemaRefs?: MultipartSchemaRefs | undefined
+}
+
+interface RenderSchemaTypeAndRuntimeOptions {
+  readonly multipartSchemaRefs?: MultipartSchemaRefs | undefined
+  /**
+   * Synthetic `$ref` name used for binary multipart fields in generated
+   * clients. Clients build request bodies from `File` / `Blob` values, so the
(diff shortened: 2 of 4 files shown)


xia-chao:
@tim-smart 
Please review

PR #8209, opened 2026-09-12
feat(http): expose the original Web Response on HttpClientResponse

xia-chao:
## What was wrong

`HttpClientResponse` already holds the original Web `Response`, but the field was `private`, so nobody could get to it. The only public body API was the Effect `Stream`.

That bites on workerd. R2 refuses a stream unless it knows the length up front:

```text
Provided readable stream must have a known length (request/response body or readable half of FixedLengthStream)
```

The original response body knows its length. `Stream.toReadableStreamEffect(response.stream)` builds a brand new stream and throws that knowledge away. So this fails:

```ts
const body = yield* Stream.toReadableStreamEffect(response.stream)
yield* Effect.tryPromise(() => env.BACKUP.put(key, body))
```

## What this PR does

Adds `HttpClientResponse.toWeb` and exposes `source`, so you can hand R2 the original body:

```ts
const source = HttpClientResponse.toWeb(response)
if (source?.body) yield* Effect.tryPromise(() => env.BACKUP.put(key, source.body))
```

`HttpClientRequest`, `HttpServerRequest` and `HttpServerResponse` already had `toWeb`. This just fills the gap.

## Why not fix the stream conversion instead

Because it can't be fixed. The length is a marker on native workerd streams, and any `new ReadableStream({...})` drops it. A `Content-Length` header does not bring it back. On top of that, `HttpClient.execute` wraps the body in `Stream.ensuring(..., controller.abort())` to abort the request on interruption, and that finalizer has to keep running — we can't just pass the original stream through.

Same bytes, put into R2 in workerd:

| what you put in R2 | result |
| --- | --- |
| `new Response(bytes).body` | accepted |
| the original response body | accepted |
| `body.tee()[0]` | accepted |
| `new ReadableStream({ start })` | rejected |
| `new Response(stream, { "content-length": "3" }).body` | rejected |
| `toReadableStreamEffect(response.stream)` | rejected |

## Tests

I ran the whole repo suite from the repo root, the same way CI does (`pnpm test --run --max-concurrency=10`): **11848 passed**.

The only failures are environment problems that already fail on `main`:

- `platform-node-shared > NodeChildProcessSpawner`: the process-group kill tests also fail on `main` on this machine (started from a clean checkout of `main`)
- `effect > cluster/ClusterWorkflowEngine`: one flaky timeout under full-suite parallelism — it passes when run on its own, and 3 out of 3 on repeat

Plus:

- new `HttpClientResponse.test.ts` checks `toWeb` returns the very same `Response`, also through `HttpClient.execute`
- `pnpm check`, `test-types`, `oxlint`, `dprint check` → pass
- verified in workerd: `toWeb(response).body` accepted by R2, `toReadableStreamEffect(response.stream)` rejected
- `pnpm jsdocs --check` fails on `unstable/ai/McpSchema.ts` both before and after this change

### DIFF
--- a/.changeset/httpclient-response-to-web.md
+++ b/.changeset/httpclient-response-to-web.md
@@ -1,0 +1,7 @@
+---
+"effect": patch
+---
+
+Add `HttpClientResponse.toWeb`, which returns the original Web `Response` backing a client response when the client exposes one.
+
+The returned response keeps runtime-specific body capabilities that a stream rebuilt in JavaScript cannot carry. On workerd, `HttpClientResponse.toWeb(response)?.body` reports a known length to `R2Bucket.put`, while `Stream.toReadableStreamEffect(response.stream)` produces a rebuilt stream that does not.
--- a/packages/effect/src/unstable/http/HttpClient.ts
+++ b/packages/effect/src/unstable/http/HttpClient.ts
@@ -1721,6 +1721,10 @@ class InterruptibleResponse implements HttpClientResponse.HttpClientResponse, Pi
     return this.original.request
   }
 
+  get source() {
+    return this.original.source
+  }
+
   get url() {
     return this.original.url
   }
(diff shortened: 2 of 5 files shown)


xia-chao:
@tim-smart 
Please review

PR #8217, opened 2026-09-13
fix(cluster): clean up RunnerServer streams on caller disconnect

### LINKED ISSUE #8214: cluster: RunnerServer leaks the entity handler when a non-persisted stream's caller disappears
## What happens

`RunnerServer.layerHandlers`'s `Stream` handler (`packages/effect/src/unstable/cluster/RunnerServer.ts`, the non-persisted branch) forwards the request with `sharding.send(message)` and returns a reply queue. Nothing runs when that handler ends.

For a persisted stream request that is fine: replies go through `MessageStorage` and a reconnecting caller resumes. For a non-persisted one, the caller's connection is the request's whole lifetime. When the caller's process dies or the runner-to-runner transport drops, the HTTP protocol layer interrupts the `RunnerServer` handler fiber (`RpcServer.makeProtocolWithHttpEffect` writes an `Interrupt` for every request id when the request scope closes), but the entity is never told. A `Rpc.fork`'d streaming handler on the entity then stays parked forever: `entityManager` drops a request only when its stream completes, and the server-side `respond` is a `Queue.offerUnsafe` that cannot fail, so no error reaches the handler's stream either.

Result: one leaked handler fiber (plus whatever it holds, e.g. a PubSub subscription) per dropped connection, for the life of the runner process.

A clean client-side close does not leak: the client's `Stream` interruption sends `Envelope.Interrupt`, and `entityManager` interrupts the forked handler. Only the "caller vanished without saying so" case leaks.

## Reproduction

- An `Entity` with a non-persisted rpc whose success is a stream, handler `Rpc.fork(Stream.fromPubSub(pubsub))` (an infinite subscription feed).
- Runner A hosts the entity (`HttpRunner.layerHttp`); runner B is client-only (`HttpRunner.layerHttpClientOnly`) and calls `Entity.client(id).Subscribe(...)`.
- Kill B's consumer fiber without letting its scope close cleanly (simulating a process death), then publish to the pubsub.
- Observe: the pubsub still has the subscriber; the entity's `activeRequests` still holds the request; nothing ever removes it.

Under `TestRunner.layer` with `RunnerServer.layerHandlers` composed over a `Sharding` hosting the entity, a test that asserts the subscriber count returns to 0 after the caller is taken away fails (`Expected: 0, Received: 1`).

## Proposed fix

In the non-persisted branch of the `Stream` handler, observe the handler fiber and send an `Envelope.Interrupt` for the request when it ends. An interrupt for an already-completed request is a no-op in `entityManager`. Reusing the request id as the envelope id avoids adding a `Snowflake.Generator` requirement to `layerHandlers`; a `Scope` finalizer would be tidier but would add `Scope` to the layer's requirements.

```ts
// RunnerServer.ts, Stream handler, non-persisted branch
Effect.andThen(
  sharding.send(message),
  Effect.withFiber((fiber) => {
    const runFork = Effect.runForkWith(fiber.context)
    fiber.addObserver(() => {
      runFork(Effect.ignore(sharding.send(
        new Message.IncomingEnvelope({
          envelope: new Envelope.Interrupt({
            id: request.requestId,
            address: request.address,
            requestId: request.requestId
          })
        })
      )))
    })
    return Effect.void
  })
)
```

Persisted stream requests should stay untouched, since a broken connection there is a resume, not a death.

Version: effect 4.0.0-rc.112; the same handler is present on `main` at the time of filing.

dexhunter:
Fixes #8214

When a caller disconnects from a non-persisted RunnerServer stream, the server now finalizes the entity handler by routing an internal interrupt through Sharding. The cleanup is attached to the caller-side stream lifecycle, while persisted requests keep their existing storage-backed behavior. The Sharding interrupt path is restricted to interrupt envelopes and does not create a missing entity manager.

The paired RunnerServer replay used the same five-test file against the current base production files and the candidate production files. The base run passed 3 of 5 tests and timed out on the ordinary caller-disconnect test and the admitted-send-return test; the candidate passed all 5. The maintained Sharding suite passed 55 tests, and the full workspace check and lint passed. The change makes no latency or throughput claim.

Weco Observe record: https://dashboard.weco.ai/share/n_Mk80-KCqwzrNKTHLiu2ByDYrOwQ22E

AI-assisted implementation: Codex, GPT-5.6-Luna.


tim-smart:
Already fixed on main :)

PR #7265, opened 2026-08-15
feat: add v2026-07-28 protocol adapter

lloydrichards:
## Type

- [x] Refactor
- [x] Feature
- [x] Bug Fix

## Description

Add support the `McpProtocol.v2026_07_28` which introduced the 'modern' stateless mcp architecture, while still supporting the previous session-based protocol versions.  To support both lifecycles, we extract the protocol lifecycle handling into dedicated runtimes. `McpServer` selects the appropriate runtime for each adapter: older adapters keep their stateful initialization and session behavior, while `v2026_07_28` selects and decodes the protocol from each request without creating session state.

The main user-facing additions are:

- Stateless `server/discover` negotiation, with no initialization handshake or session state.
- `McpRequestContext` gives handlers access to request and client capabilities neutral to the runtime of the protocol.
- Multi-round-trip tool results (MRTR) where handlers can return `McpSchema.InputRequired`, and clients can provide keyed elicitation, sampling, or roots responses on a later request.
- `subscriptions/listen`, delivering filtered server change notifications over stdio or Server-Sent Events.
- Tool output schemas and `structuredContent` for every JSON value in the 2026-07-28 adapter. Earlier adapters continue to project only the shapes they support.
- Request-selected streaming RPC HTTP responses and typed server-to-client notifications.

<details>
<summary>A toy example of the stateless server and JSON tool output:</summary>

```ts
const Greet = Tool.make("greet", {
  description: "Return a greeting for the supplied name",
  parameters: Schema.Struct({ name: Schema.String }),
  success: Schema.Struct({ greeting: Schema.String })
})

// New in 2026-07-28: a tool output schema may describe any JSON value, not
// only an object. The adapter returns this array as `structuredContent`.
const ProjectFiles = Tool.make("project_files", {
  description: "Return JSON array output and the request-scoped client name",
  parameters: Tool.EmptyParams,
  success: Schema.Array(Schema.String),
  dependencies: [McpSchema.McpRequestContext]
})

const InspectorToolkit = Toolkit.make(Greet, ProjectFiles)

// New in 2026-07-28: the server selects this adapter from each request's
// metadata and headers; there is no initialize handshake or session ID.
const McpHttp = McpServer.layerHttp({
  name: "effect-mcp-2026-07-28-scratchpad",
  version: "1.0.0",
  description: "Modern-only MCP server for Inspector interoperability checks",
  path: "/mcp",
  protocols: [McpProtocol.v2026_07_28]
[...]

PR #8228, opened 2026-09-13
fix(ai): McpServer issues with tool strictness, server instructions and prompt titles

### LINKED ISSUE #8218: McpServer: unknown tool-call keys are silently stripped; Tool.StrictJsonSchema is not honoured and there is no way to decode strictly

4.0.0-rc.115 (also current `main`, `packages/effect/src/unstable/ai/McpServer.ts` and `Toolkit.ts`)

Register a tool with a struct parameter schema through `McpServer.toolkit` and call it with an extra, unknown key (for example a misspelled parameter).

```ts
import { Effect, Layer, Schema } from "effect"
import { McpServer, Tool, Toolkit } from "effect/unstable/ai"

const Echo = Tool.make("echo", {
  parameters: { text: Schema.String, limit: Schema.optionalKey(Schema.Number) },
  success: Schema.String
})

const toolkit = Toolkit.make(Echo)

const ToolkitLayer = toolkit.toLayer({
  echo: ({ text }) => Effect.succeed(text)
})

// Serve over stdio with McpServer.layerStdio and call:
// { "method": "tools/call", "params": { "name": "echo",
//   "arguments": { "text": "hi", "limti": 5 } } }
```

Some way for the server to reject the call, since `limti` is not an accepted parameter. Either:

1. `McpServer` honours the existing `Tool.StrictJsonSchema` annotation: when a tool is strict, `tools/list` serves `additionalProperties: false` on every object level of `inputSchema`, and `tools/call` decodes with `onExcessProperty: "error"` and returns the same `InvalidParams` / `isError` path used for a wrong type; or
2. `McpServer.toolkit` (or `Toolkit.toLayer`) accepts `Schema` parse options so a server can opt in.

### What do you see instead?

The call succeeds. `Toolkit.ts` decodes parameters with a bare `Schema.decodeUnknownEffect(tool.parametersSchema)` (line 254), so Schema's default `onExcessProperty: "ignore"` silently drops `limti`, and the handler runs with `limit` unset. The served `inputSchema` has no `additionalProperties`, so clients cannot detect the problem either.

`Tool.StrictJsonSchema` exists but is only consulted by LLM provider clients (it maps to OpenAI's `strict: true`); `McpServer` never reads it.

### Additional information

This matters more for an MCP server than for an ordinary API: the caller is a model that cannot see that a key was dropped, so a typo turns into a silently different tool call with a default value. We worked around it with a custom registration over `McpServer.McpServer.addTool` that walks the raw payload for unknown keys and rewrites the served schema with `additionalProperties: false`; happy to upstream that shape if option 1 is the preferred direction.

Related but distinct: #6355 is about a client rejecting `additionalProperties` in a served schema; this issue is about the server accepting keys it never declared.

### LINKED ISSUE #8219: McpServer: no way to set InitializeResult.instructions

4.0.0-rc.115 (also current `main`)

Build a server with `McpServer.layerStdio({ name, version, description, websiteUrl, icons, protocols, extensions })` and try to set the MCP `instructions` field.

`InitializeResult.instructions` is part of the MCP spec (all supported protocol versions, 2025-03-26 onward): a string the client is expected to feed into the model's system context describing how to use the server. `McpServer.layer`, `layerStdio` and `layerHttp` should accept an `instructions?: string` option and return it from `initialize`.

### What do you see instead?

There is no way to set it. The options type for `layer` / `layerStdio` / `layerHttp` (and `layerHandlers`, `McpServer.ts` line 2195) has `name`, `version`, `description`, `websiteUrl`, `icons`, `extensions`, but no `instructions`, and the string `instructions` does not appear anywhere in `McpServer.ts`. `description` is a different field (it lands in `serverInfo`, a human-facing label), so it is not a substitute.

### Additional information

Clients such as Claude Code and Claude Desktop surface `instructions` to the model at connect time. A server built on Effect currently has to leave that channel empty or embed usage guidance in every tool description instead.

### LINKED ISSUE #8220: McpServer.prompt: no way to set the prompt title served in prompts/list

4.0.0-rc.115 (also current `main`)

Register a prompt with `McpServer.prompt({ name, description, parameters, content })` and inspect the `prompts/list` response.

The MCP spec added an optional `title` to `Prompt` in 2025-06-18 (a human-readable display name, distinct from the programmatic `name`), and `McpSchema.Prompt` already models it (`McpSchema.ts` line 1225, `title: optional(Schema.String)`). `McpServer.prompt` should accept `title?: string` and emit it in `prompts/list`, the same way tools accept `Tool.Title` and emit `title` in `tools/list`.

### What do you see instead?

`McpServer.prompt`'s options (`McpServer.ts` line 2066) are `name`, `description`, `parameters`, `completion`, `content`, `annotations`. There is no `title`, and no annotation the server reads for prompts, so the served `Prompt` never carries one.

### Additional information

Small gap, but it is the only field of the served `Prompt` record a user cannot set, and clients that render a prompt picker fall back to the snake_case `name`.

lloydrichards:
## Type
- [x] Bug Fix

## Description

I'm still in the flow from #7265 so took a look at these recent issues together:

- #8218 -> Strict unknown-key rejection. From the MCP tools specification says servers must validate tool inputs against the tool’s advertised inputSchema and since the Tool.Strict is an Effect-specific API contract this needs to Effect.die when an invalid combination is used.
- #8219 -> Server instructions. String is part of InitializeResult from 2024-11-05 through 2025-11-25 but in 2026-07-28 it was removed and the field moved to DiscoverResult.instructions.
- #8220 -> Prompt title. Introduced in 2025-06-18 through the prompt’s BaseMetadata.

## Related

- Closes #8218
- Closes #8219
- Closes #8220 

Closes EFF-1369


guillempuche:
You're pushing MCP in Effect, thanks Lloyd!

PR #8233, opened 2026-09-14
feat(Crypto): add ULID support

### LINKED ISSUE #8213: feat(Crypto): add ULID support

Add  `readonly randomULID` to `interface Crypto`

ULIDs are better than UUIDs in some cases. (shorter, lex-sortable, no hyphens, etc)
ULIDs are also case-insensitive, contains no confusing characters (like O/0 or I/1), and is easily double-click selectable in terminal windows or URLs (again, no hyphens!).

https://github.com/ulid/javascript

Thank you :)

gold-beyond:
Closes #8213
Closes EFF-1385

Adds `Crypto.randomULID` to generate ULIDs through the platform Crypto service.

```ts
import { Crypto, Effect } from "effect"

  const crypto = yield* Crypto.Crypto
  return yield* crypto.randomULID
})
```

Each identifier contains 26 uppercase Crockford base32 characters: 10 encode the Clock timestamp and 16 encode 80 bits from `randomBytes(10)`. ULIDs sort by timestamp; ordering within the same millisecond is not guaranteed.

Timestamp normalization matches UUIDv7: fractions are truncated, negative values clamp to zero, values above `2 ** 48 - 1` clamp to that maximum, and `NaN` encodes as zero. Negative and positive infinity clamp to the respective endpoints.

`internal/ulid.ts` encodes the timestamp and random bytes directly, requiring exactly 10 random bytes. The timestamp loop appends characters from most to least significant. `Crypto.make` reads the Clock and obtains fresh randomness on each execution.

## Validation

- 26 tests across `Crypto.test.ts`, `internal/Ulid.test.ts`, and `internal/Uuid.test.ts` cover exact encodings, valid boundaries, UUIDv7-compatible normalization, timestamp ordering, and invalid randomness lengths.
- `pnpm lint-fix`, `pnpm lint`, and `pnpm check` pass.
- `pnpm jsdocs --check` reports existing diagnostics in `src/unstable/ai/McpSchema.ts`, outside this PR.


packages/effect/src/internal/uuid.ts
```diff
@@ -53,3 +57,35 @@ export function v7Bytes(timestampMillis: number, bytes: Uint8Array = randomBytes
...
+  return out
+}
+
+// A ULID is a 48-bit big-endian timestamp followed by 80 bits of randomness,
+// rendered as 26 Crockford base32 characters.
+/** @internal */
+export const ulidString = (timestampMillis: number, bytes: Uint8Array): string => {
+  const buffer = new Uint8Array(16)
+  writeTimestamp(timestampMillis, buffer)
+  buffer.set(bytes.subarray(0, 10), 6)
```
SwannHERRERA: Since ULIDs require exactly 80 random bits, could we validate that `bytes.length === 10` and then use buffer.set(bytes, 6) directly? The current subarray silently truncates oversized input and silently zero-pads undersized input. The latter produces a valid-looking ULID with reduced entropy. A length check also avoids allocating the subarray view.

packages/effect/src/internal/uuid.ts
```diff
@@ -53,3 +57,35 @@ export function v7Bytes(timestampMillis: number, bytes: Uint8Array = randomBytes
...
+    accumulator &= (1 << bits) - 1
+  }
+
+  return out
+}
+
+// A ULID is a 48-bit big-endian timestamp followed by 80 bits of randomness,
+// rendered as 26 Crockford base32 characters.
+/** @internal */
+export const ulidString = (timestampMillis: number, bytes: Uint8Array): string => {
```
SwannHERRERA: I suggest to move the ULID implementation to internal/ulid.ts? ULID is a separate identifier standard with its own encoding and layout, while this module otherwise only implements UUID variants.

I see that this will require to move elements or to duplicate `base32` and `writeTimestamp` but they might be useful elsewhere

packages/effect/src/internal/uuid.ts
gold-beyond: You are right that the zero-padding case is the dangerous one: a short buffer still produced a well-formed 26 character ULID, just with fewer bits of entropy behind it. `ulidString` now rejects anything that is not exactly 10 bytes and then uses `buffer.set(bytes, 6)` directly, and there is a test covering 0/9/11/16 byte inputs (a length check loosened to `!== 11` makes it fail).

packages/effect/src/internal/uuid.ts
gold-beyond: Moved it to `internal/ulid.ts`.

Since ULID no longer needs anything from `uuid.ts`, I reverted that module and its test back to `main`, so the UUID code is untouched by this PR again and `v7Bytes` keeps writing its timestamp inline. The Crockford base32 encoder and the timestamp writer now live in `ulid.ts`; the writer is the same six lines as in `v7Bytes`. If you would rather not carry that duplication, I can lift it into a shared internal module instead - say the word and I will do it in a follow-up commit.

PR #8234, opened 2026-09-14
feat(sql-pg): support startup parameters and opaque options

### LINKED ISSUE #8193: feat(sql-pg): add support for startupParameters in connection config

tim-smart:
The native PostgreSQL driver ignored URL `options`, so callers could not establish session defaults at connection startup. Add structured `startupParameters` and opaque `startupOptions` to `PgConnection.Config` and `PgClientConfig`. Explicit `startupOptions` overrides URL `options`; the URL query key and wire field remain `options`. Every physical connection receives the resolved defaults, including pooled replacements.

Normalize parameter names, restrict client encoding to UTF-8, and reject reserved names, empty names, and NUL bytes before connecting. PostgreSQL validates other GUCs. Document application-name precedence and the requirement to avoid setting the same GUC in named parameters and opaque options. Includes API documentation, a README example, and a patch changeset.

Validation:

- All 74 tests pass across `PgConnection.test.ts`, `PgConnection.in-process.test.ts`, and `StartupParameters.integration.test.ts`: 66 unchanged pre-existing tests, six representative validation cases, one forwarding/precedence test using the existing TCP helper, and one PostgreSQL regression for startup defaults, replacement, and `RESET ALL`.
- `pnpm check`, `pnpm lint-fix`, `pnpm lint`, and `git diff --check` pass.
- `pnpm jsdocs --check` reports two existing errors in `packages/effect/src/unstable/ai/McpSchema.ts` (short-description paragraph and section-heading rules). That file is unchanged from main; no SQL-PG diagnostics.

Closes #8193
Closes EFF-1373


AVtheking:
Hey @tim-smart, I saw we marked `replication` param as reserved. Just want to know if we are not going to support replication connection through it ?

 I was trying to build a pg logical replication client in effect and have to use `node-pg` library instead of `@effect/sql-pg `just due to this .

PR #8250, opened 2026-09-15
fix(Schema): Class rejects Struct schemas piped through encodeKeys

estilles:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Fixes a TypeScript overload error when a `Schema.Struct` transformed by `Schema.encodeKeys` is passed directly to `Schema.Class`.

```ts
class Foo extends Schema.Class<Foo>('Foo')(
  Schema.Struct({
    a: Schema.String,
  }).pipe(Schema.encodeKeys({ a: 'c' }))
) {}
```

fails to type-check:

```text
No overload matches this call.
  Type 'encodeKeys<Struct<...>, ...>' is missing the following
  properties from type 'Struct<Fields>': fields, mapFields
```

despite working correctly at runtime — `Schema.decodeUnknownSync(Foo)`/`Schema.is(Foo)` etc. all behave as expected once the type error is silenced. The root cause is that `Schema.Class`'s struct overload requires the literal `Struct<Fields>` interface, but `encodeKeys`'s return type only extends `decodeTo<To, From>`, which doesn't expose `fields`/`mapFields` — even though `encodeKeys`'s `to` side is structurally just the original struct.

> Note: this can be worked around by applying `encodeKeys` *after* constructing the class instead of inside the field definition.

```ts
class Foo extends Schema.Class<Foo>('Foo')({
  a: Schema.String,
  b: Schema.optionalKey(Schema.Finite),
}) {}
const FooWithEncodeKeys = Foo.pipe(Schema.encodeKeys({ a: 'OriginalA' }))
```

That pattern is fine when a single top-level rename is needed and the caller controls the decode/encode call site directly. It doesn't work when the rename needs to live on a *nested* field's own struct definition (e.g. a child class embedded inside a parent, where the child's schema is what's passed to the parent's `Class` constructor) or when multiple nested schemas need `encodeKeys`. These are the use cases this PR aims to fix.

```ts
class Foo extends Schema.Class<Foo>('Foo')({
  a: Schema.String,
  b: Schema.optionalKey(Schema.Finite),
}) {}
const FooWithEncodeKeys = Foo.pipe(Schema.encodeKeys({ a: 'OriginalA' }))

class Bar extends Schema.Class<Bar>('Bar')({
  c: Schema.String,
  foo: Foo,
}) {}
const BarWithEncodedKeys = Bar.pipe(Schema.encodeKeys({ c: 'OriginalC' }))
```

## Changes

- `encodeKeys`'s returned schema now carries a real `.fields` property (mirroring the original struct's fields) and a private type-id brand (`EncodeKeysTypeId`, following the existing `flip`/`FlipTypeId` pattern), so it satisfies the structural shape `Class` needs.
- `Class`'s struct overload, its interface, `makeClass`, and the internal `isStruct` guard are widened to accept `Struct<Fields> | encodeKeys<Struct<Fields>, any>` in the relevant positions.
- `Class.mapFields` now delegates to the underlying struct (`struct.to`) when built from an `encodeKeys` schema, since `encodeKeys` results don't have their own runtime `mapFields`.
- `Class.extend` explicitly throws when called on a class built from an `encodeKeys` schema, rather than silently reconstructing the class without the key mapping. `extend` merges fields via `SchemaAST.struct(...)` from scratch, which has no way to carry the original rename forward — attempting to support it correctly would mean re-deriving a new `encodeKeys` mapping that also covers newly added fields, which is a separate, larger feature. I'd rather fail loudly here than ship something that silently drops the rename.

## Scope

This fixes `Schema.Class` only. `Schema.TaggedClass`, `Schema.Error`, and `Schema.TaggedError` share the same overload shape and would need the identical treatment if this pattern is wanted there too — happy to extend this PR to cover them if that's preferred, or leave it to a follow-up.

## Verification

- `Schema.isSchema` recognizes a class built from an `encodeKeys` schema.
- Full construction / `make` / `Equal.equals` / arbitrary generation / decode success & failure / encode success & failure round-trip for a class built from an `encodeKeys` schema.
- `mapFields` works correctly on such a class.
- `extend` throws with a clear error message on such a class.

```sh
> pnpx vitest run packages/effect/test/schema/Schema.test.ts

 Test Files  1 passed (1)
      Tests  600 passed (600)
```


gcanti:
I'm not convinced there is an actual missing use case here. Can't the nested case already be handled by defining `FooWithEncodeKeys = Foo.pipe(Schema.encodeKeys(...))` and using `FooWithEncodeKeys` as the field in `Bar`?

The PR's example creates `FooWithEncodeKeys` but then uses `foo: Foo`, so naturally the mapping is not applied.

estilles:
> I'm not convinced there is an actual missing use case here. Can't the nested case already be handled by defining `FooWithEncodeKeys = Foo.pipe(Schema.encodeKeys(...))` and using `FooWithEncodeKeys` as the field in `Bar`?
> 
> The PR's example creates `FooWithEncodeKeys` but then uses `foo: Foo`, so naturally the mapping is not applied.

I understand what you're saying. You're right. I screwed up my explanation and my example. `Bar` should be using `FooWithEncodeKeys`. Despite my error, I would argue that if

```ts
class Foo extends Schema.Class<Foo>('Foo')(
  Schema.Struct({
    a: Schema.String,
  })
) {}
```

is valid syntax, it would be logical to assume that

```ts
class Foo extends Schema.Class<Foo>('Foo')(
  Schema.Struct({
    a: Schema.String,
  }).pipe(Schema.encodeKeys({ a: 'c' }))
) {}
```

would likewise be valid syntax, if one needs different encode keys.

estilles:
I ended up encountering a few other Schema use cases, besides `Schema.encodeKeys()` that are not accepted by `Schema.Class()`. I'm abandoning this PR for now, and may revisit sometime in the future.

Thanks for the comments!

PR #8257, opened 2026-09-16
fix(sqlite-do): propagate native transaction completion failures

danieljvdm:
Native Durable Object transactions can reject after the SQL body succeeds, causing `withTransaction` to report success despite rollback. Await native completion before releasing the connection and propagate rejection as a `SqlError` with the original cause.

One regression test delays native rejection after a successful body, checks that a competing query stays blocked until completion, and verifies the typed error, original cause, and rolled-back rows. Pre-existing tests are unchanged.

Validation: all 10 client tests, root typecheck, and lint pass.

Closes EFF-1387


danieljvdm:
The regression changed from `Success("committed")` after rollback to a typed `SqlError` retaining the original native cause. A temporary public-RPC proof against real SQLite Durable Object storage in workerd confirmed the same failure and fix. Native body-failure and interruption paths retain their original outcomes, finish rollback, and exit cleanly. All 10 SQLite Durable Object package tests pass, along with the owning repository’s typecheck and lint/format checks. The regression cites the [introducing commit](https://github.com/Effect-TS/effect/commit/fe1b2d53b7374db7a690ba96cc46392789d8ffe0).

PR #8261, opened 2026-09-16
feat(http): add QUERY method support

nikhilsnayak:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add HTTP `QUERY` support to clients, routers, HttpApi endpoints, and AI request metadata. Configure CORS support through `allowedMethods`; defaults are unchanged.

OpenAPI 3.1 output represents `QUERY` through `x-oai-additionalOperations`, which requires consumer support for that extension. The OpenAPI generator accepts the extension and the native OpenAPI 3.2 `query` field.

## Related

Closes EFF-1389


lloydrichards:
https://www.rfc-editor.org/info/rfc10008/

I'm pretty sure this is still only a proposal 🤔 don't get me wrong, it looks exciting but I wonder what happens when an `HttpClient` sends a `QUERY` to something that doesn't support it yet

PR #8273, opened 2026-09-17
Refactor cluster caller cancellation using public Scope operations

fubhy:
Cluster caller cancellation currently inspects Scope state and manually inserts and removes finalizers. Use `Scope.fork`, `Scope.addFinalizer`, and `Scope.close` to own admission cleanup and handler cancellation registrations instead.

Register cancellation before checking and inserting a request into the mailbox. Close the child scope when admission is rejected or interrupted, and detach each handler binding when its effect finishes. Caller disconnect still removes queued requests and signals interruption without waiting for handler teardown; replayed handler effects are rejected when caller closure has already started. The existing persistence and uninterruptibility policy is preserved.

This removes EntityManager's dependency on Scope state checks and internal finalizer functions. Child scopes introduce an allocation per cancellation registration. No changeset is required because this is an internal refactor preserving the existing public interface and lifecycle behavior.

Validation:

- 122 tests pass across RunnerServer, Entity, interruptors, Sharding, and ClusterWorkflowEngine.
- New regressions cover blocked handler cleanup, admission during pending caller finalization, and replay while a later caller finalizer is blocked.
- `pnpm check`, `pnpm lint-fix`, and `git diff --check` pass.


fubhy:
@tim-smart Before you say it, yes, this is slop. Just trying to get rid of the supossedly "private" Scope state reads.

PR #8281, opened 2026-09-17
Fix BigDecimal arbitrary magnitude coverage and constrained shrinking

adrian-gierakowski:
## Problem

A one-sided BigDecimal bound fixes every generated value at scale 20. With the default size 10, a strictly positive decimal consequently has only 101 possible values, all between 1e-20 and 1.01e-18. Increasing the sample count cannot improve that range. Unconstrained decimals also miss large magnitudes at small sizes.

## Change

Mix ordinary decimal amounts with representative wide scales and explicit coefficient ranges. This exercises both signs, values beyond JavaScript floating-point range, and nonzero values below its minimum, while keeping ordinary amounts common.

Project inclusive and exclusive bounds onto each scale using exact integer rounding, omit empty grids, and include boundary scales for narrow or high-precision intervals. Generate offsets from the permitted boundary nearest zero so even coarse-scale samples shrink to the exact inclusive boundary.

This PR is independent of the separate BigInt arbitrary change.

## Validation

- 183 focused tests: `BigDecimal.test.ts` and `Arbitrary.test.ts`.
- Coverage at sizes 0, 1 and 10; positive and negative bounds; narrow, exclusive, singleton and extreme-scale intervals; shrink validity, exact inclusive-boundary shrinking and replay.
- `pnpm check`, `pnpm lint-fix` and `pnpm lint` passed in the repository's Nix development shell.

The commit was validated against Effect 4.0.0-rc.115 (`4a05d4914`).


gcanti:
Closing in favor of https://github.com/Effect-TS/effect/pull/8296

PR #8282, opened 2026-09-17
Broaden BigInt arbitraries independently of size and shrink large roots

adrian-gierakowski:
## Problem

Unbounded BigInt arbitraries currently use roughly [-size², size²]. At the default size 10, only 201 integers are possible. One-sided constraints inherit a similarly narrow range, so ordinary samples miss unsafe integer conversions and very large integer arithmetic.

Simply widening one uniform interval would mostly generate huge values. Those values also expose a shrinking limitation: repeated halving exhausts the default 100-attempt budget before reaching useful small counterexamples.

## Change

Mix small, ordinary and large ranges up to 2048 bits independently of collection size. Respect one-sided bounds and explicitly exercise their boundaries even when they lie outside the default magnitudes. Generation with both bounds supplied retains its existing behavior.

Before halving a large initial shrink gap, try smaller power-of-two offsets while retaining the previous passing candidate. This lets common failures shrink to useful boundaries within the default budget, including with large nonzero bounds.

This PR contains only the BigInt change and does not depend on the separate BigDecimal fix.

## Validation

- 183 focused tests: `BigInt.test.ts` and `Arbitrary.test.ts`, with the original BigDecimal implementation, plus `pnpm check`.
- 210 focused tests with both changes applied.
- Coverage at sizes 0, 1 and 10; both signs, near-zero values and extreme magnitudes; inclusive/exclusive one-sided bounds beyond the default ranges; shrink validity, default-budget failure reduction and replay.
- `pnpm lint-fix` and `pnpm lint` passed in the repository's Nix development shell.

The independent validation used Effect 4.0.0-rc.115 (`4a05d4914`) as the base.


gcanti:
Closing in favor of https://github.com/Effect-TS/effect/pull/8296

adrian-gierakowski:
> Closing in favor of #8296

Thanks!

PR #8290, opened 2026-09-18
fix(JsonSchema): preserve boolean schema false in fromSchemaDraft2020_12

### LINKED ISSUE #8289: bug(JsonSchema): boolean schema 'false' is destructured into empty object schema {}
### Problem

According to the JSON Schema spec (Draft 2020-12 §4.3.2 / Draft-07 §3.2), a boolean schema `false` should reject every input value (it is unsatisfiable, equivalent to `Schema.Never`).

However, right now if you pass `false` to `JsonSchema.fromSchemaDraft2020_12(false)` (or `fromSchemaDraft07` / `fromSchemaOpenApi3_1`), it silently turns into `{}` (an empty schema that accepts everything).

---

### Step-by-step Reproduction & Root Cause

I stepped through the source code with the debugger, and here is exactly what happens:

#### 1. Entry point: rest destructuring on a boolean turns it into `{}`

In `packages/effect/src/JsonSchema.ts` line 320:

```typescript
export function fromSchemaDraft2020_12(js: JsonSchema): Document<"draft-2020-12"> {
  const { $defs, ...schema } = js
  return {
    dialect: "draft-2020-12",
    schema,
    definitions: Predicate.isObject($defs) ? ($defs as Definitions) : {}
  }
}
```

When `js` is `false`, running `const { $defs, ...schema } = false` evaluates `schema` to `{}` (empty object).

Image

As seen in the debugger above, `js` is `false`, but after line 320, `schema` becomes `{}`. The document returned is `{ dialect: "draft-2020-12", schema: {}, definitions: {} }`.

#### 2. Downstream consumer: skips `if (input === false)` and accepts everything

The downstream parser in `packages/effect/src/internal/schema/fromJsonSchemaDocument.ts` line 868 already intends to support `false`:

```typescript
function translateSchema(input: unknown, path: Path, isRoot = false): ImportedJsonSchemaRepresentation {
  if (input === false) {
    return never
  }
  if (!isObject(input)) return unknown
  ...
```

However, because the document's `schema` was already flattened into `{}` by `fromSchemaDraft2020_12`, `translateSchema` receives `input = {}`:

Image

As shown in the screenshot, `input` is `{}` so `input === false` evaluates to `false`. The `return never` branch is completely missed, and it falls through to generic object handling, turning it into a type that accepts any input.

---

### Minimal Reproducible Example

```typescript
import { JsonSchema, Schema, SchemaRepresentation } from "effect"

// 1. Raw boolean schema: false (should reject everything)
const raw = false

// 2. Convert to document
const doc = JsonSchema.fromSchemaDraft2020_12(raw)
console.log(doc.schema) // Output: {} (Expected: false)

// 3. Convert back to Effect Schema
const effectSchema = SchemaRepresentation.fromJsonSchemaDocument(doc)

// 4. Test validation with any random string
console.log(Schema.is(effectSchema)("hello world"))
// Actual: true  (Accepts anything!)
// Expected: false (Schema.Never should reject everything)
```

---

### Proposed Fix

Handle boolean schema upfront before rest destructuring in `packages/effect/src/JsonSchema.ts`:

```diff
 export function fromSchemaDraft2020_12(js: JsonSchema): Document<"draft-2020-12"> {
+  if (typeof js === "boolean") {
+    return {
+      dialect: "draft-2020-12",
+      schema: js,
+      definitions: {}
+    }
+  }
   const { $defs, ...schema } = js
   return {
     dialect: "draft-2020-12",
     schema,
     definitions: Predicate.isObject($defs) ? ($defs as Definitions) : {}
   }
 }
```

xia-chao:
Closes #8289

## Problem

`fromSchemaDraft2020_12` uses object rest destructuring on its input:

```ts
const { $defs, ...schema } = js
```

When `js` is the boolean `false`, JavaScript silently spreads it into an empty object `{}`. This means a schema that should reject every input (equivalent to `never`) is instead treated as one that accepts everything.

## Steps to Reproduce

```ts
import { JsonSchema } from "effect"

const result = JsonSchema.fromSchemaDraft2020_12(false)
console.log(result.schema)
// Bug:      {}     ← accepts everything
// Expected: false  ← rejects everything
```

## Fix

Add a `typeof js === "boolean"` guard before the destructuring so boolean schemas are returned as-is:

```ts
export function fromSchemaDraft2020_12(js: JsonSchema): Document<"draft-2020-12"> {
  if (typeof js === "boolean") {
    return { dialect: "draft-2020-12", schema: js, definitions: {} }
  }
  const { $defs, ...schema } = js
  // ...
}
```

## Tests

Two regression test suites added:
- `test/JsonSchema.test.ts`: directly tests `fromSchemaDraft2020_12(false)` and `fromSchemaDraft2020_12(true)`
- `test/schema/representation/fromJsonSchemaDocument.test.ts`: end-to-end roundtrip test verifying `Schema.Never` is preserved correctly


xia-chao:
If you find anything in this PR that needs to be changed, feel free to commit directly to this branch instead of opening a new PR. I'd prefer to keep the changes in this PR.
@tim-smart Thank you very much for editing my PR during this period

xia-chao:
JsonSchema is an object type, so false is outside the input contract.
So I closed this PR

PR #8300, opened 2026-09-18
Deprecate span event emission API

yordis:
OpenTelemetry has deprecated the Span Event API in favour of emitting events through the Logs API, correlated to their span by trace and span id. Effect currently documents the opposite: `Logger.tracerLogger` tells readers that recording logs as span events is how you make them visible in OpenTelemetry, Jaeger, or Zipkin, which now points new instrumentation away from the supported path.

- `OtlpLogger` already stamps the trace and span id onto every log record, so the replacement is available today and needs no new API. Installing it alongside `tracerLogger` exports each log twice, once as a log record and once as a span event, which is the duplication the deprecation exists to remove.
- Reading events back off a span is explicitly retained by the spec, so the accessors, exporters, and the `@effect/opentelemetry` bridge are untouched, and neither `Span#event` nor `tracerLogger` is removed.
- `Span#event` is user-implementable, so the tag surfaces on custom `Span` implementations as well as call sites. That is the accurate signal, but happy to drop it to `tracerLogger` alone if the noise is unwelcome.

Deliberately scoped to the deprecation. Whether `tracerLogger` should stay in the default logger set, and whether `@effect/opentelemetry`'s `NodeSdk` should stop installing a log exporter beside it, are behaviour changes worth deciding separately, and the release candidate is the cheap moment to make them.

References:

- [Deprecating Span Events API](https://opentelemetry.io/blog/2026/deprecating-span-events/)
- [OTEP 4430: Span Event API deprecation plan](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4430-span-event-api-deprecation-plan.md)


yordis:
Related to https://github.com/pingdotgg/t3code/pull/12493

PR #8301, opened 2026-09-18
Struct.evolve tweaks

harrysolovay:
Previously a transform for a key that does not exist on the struct was silently accepted and ignored at the type level. The evolver constraint now maps such keys to `never`, so a misspelled key is a compile error.

The check is expressed as a single generic mapped type rather than an intersection with `NoExcessProperties`, because TypeScript can only derive a contextual type for nested calls such as
`evolve(s, { a: evolve({ ... }) })` from a plain mapped type. Using an intersection loses inner-key inference and editor completions.

## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update


devinjameson:
This is closer to [`evo`](https://github.com/foldkit/foldkit/blob/a5413bbc0fc8f6578285632d7bbd7847efcf1a09/packages/foldkit/src/struct/index.ts#L3-L33), but `evo` also requires each transform’s return type to be assignable to the corresponding field’s existing type (the return type is covariant in the field type). Maybe that particular functionality would be better as a new `Struct.modifyFields`? (That’s also what `evo` is being renamed to in Foldkit.)

I think this change depends on whether `Struct.evolve` intentionally accepts and ignores transforms for fields that don’t exist on the source struct:

```ts
const transforms = {
  name: normalizeName,
  email: normalizeEmail,
}

// This becomes a type error after this change.
Struct.evolve({ name: "Alice" }, transforms)
```

harrysolovay:
> ```ts
> const transforms = {
>   name: normalizeName,
>   email: normalizeEmail,
> }
>
> Struct.evolve({ name: "Alice" }, transforms)
> ```

I can't think of use cases in which I'd want to group transforms to evolve differently-shaped structs. Any that come to mind?

My sense is allowing extra fields could be a foot gun (one could easily misspell a key).

Re. constraining the return type, this might be a more sensible default given the contexts in which Struct.evolve is typically used (reducers, setters, etc.––these all typically act on typed structures anyways.

harrysolovay:
@devinjameson I went ahead and updated this PR with a separate `Struct.modifyFields` that enforces the given field transform's return type conform to that of the input. I left `Struct.evolve`'s transform fields as allowing any return type. Also, I kept `Struct.evolve` as not allowing excess property keys. Thoughts?

devinjameson:
> @devinjameson I went ahead and updated this PR with a separate `Struct.modifyFields` that enforces the given field transform's return type conform to that of the input. I left `Struct.evolve`'s transform fields as allowing any return type. Also, I kept `Struct.evolve` as not allowing excess property keys. Thoughts?

Makes sense to me! I think modifyFields is valuable. Would be cool to move that upstream. Re: evolve change, it just depends what the intended contract is.

Issue #7374, opened 2026-08-20
socket protocol utilities for http apis


harrysolovay:
### What is the problem this feature would solve?

Streamline including socket protocols in the client, and parsing socket protocols in http APIs.

### What is the feature you are proposing to solve the problem?

Some minimal socket protocol utils.

```ts
export const SocketProtocolsKey = "Sec-WebSocket-Protocol" as const

const SocketProtocols = S.String.pipe(
  S.decodeTo(S.Array(S.Trim), {
    decode: SchemaGetter.split({ separator: "," }),
    encode: SchemaGetter.transform((arr) => arr.join(",")),
  }),
)
```

Ideally there'd also be a requirement that we could optionally provide to any `SocketConstructor`.

```ts
mySocketClient.pipe(
  Layer.provide(Socket.layerWebSocketConstructorGlobal.pipe(Layer.provide(SocketProtocols.layer(["a", "b", "c"]))))
)
```

And a piece of http middleware for parsing the protocols from an `HttpServerRequest`.

```ts
const handler = Effect.gen(function*() {
  const protocols = yield* SocketProtocols
})
```

ArjunCodess:
hi, i'd like to work on this.

i'd add a socketprotocols schema that splits sec-websocket-protocol, plus a layer for the constructor and http middleware that yields the parsed list.

should the middleware fail closed if the header is missing, or yield an empty array?

harrysolovay:
I actually just submitted this yesterday. Would you be willing to provide a review?: https://github.com/Effect-TS/effect/pull/8302

ArjunCodess:
yeah, left a review on #8302.

schema + the request layer look right. missing header as undefined is actually the nicer fail-open. empty header still becomes [].

the client SocketConstructor layer is still missing, and `a,,b` will leave a hole in the array. details are on the pr.

i'll drop the claim. this is yours.

PR #8302, opened 2026-09-18
feat: socket-protocol-parsing service

harrysolovay:
I kept running into the need to extract socket protocol tokens from the given `HttpServerRequest`. Figured I'd propose this util; I'm not sure if `unstable/socket` is the correct place for this (vs. in the http module).

Adds a SocketProtocols service to `effect/unstable/socket` for reading and parsing the Sec-WebSocket-Protocol request header.


packages/effect/src/unstable/socket/SocketProtocols.ts
```diff
@@ -0,0 +1,86 @@
...
+      )
+    )
+  )
+}) {
+  /**
+   * Layer that derives the service from the current `HttpServerRequest`.
+   *
+   * @since 4.0.0
+   */
+  static readonly layer: Layer.Layer<SocketProtocols, S.SchemaError, HttpServerRequest.HttpServerRequest> = Layer
```
harrysolovay, line 84: Perhaps worth moving this to the module-level so that it's referenced as `SocketProtocol.layer` instead of `SocketProtocol.SocketProtocol.layer`... thoughts?

ArjunCodess:
looked through this. schema + the request layer are the same shape i was going to add.

missing header yields undefined, which is actually the nicer fail-open. empty header still becomes []. that answers the question i asked on #7374.

two gaps vs the original issue:
1. the client SocketConstructor layer is still missing. the ticket wanted `SocketProtocols.layer(["a", "b", "c"])` on the websocket constructor. this pr only covers the server read.
2. `"a,,b"` will decode to a hole in the array. worth filtering empty tokens.

on your note: yeah, export `layer` at module level. `SocketProtocols.layer` reads better than `SocketProtocols.SocketProtocols.layer`.

and the socket package pulling `HttpServerRequest` is a bit sideways. if the client piece lands later, maybe keep the schema in socket and the request layer in http.

tests look good for the happy path. i'd add the empty-token case and a lowercase header, since browsers send `sec-websocket-protocol`.

i'll drop my claim on #7374. this is yours.

harrysolovay:
@ArjunCodess I just addressed your (excellent) feedback – ready for another review

PR #8310, opened 2026-09-19
fix(Pool): release retired usage-TTL resources

williamrobertson13:
## Summary

Fix a memory leak in `Pool.makeWithTTL` with the usage strategy. Retired resources and consumed acquisition errors could remain in the TTL queue when the pool reached its target or minimum size.

The usage strategy no longer keeps its own queue of items. `state.items` is a `Set`, which iterates in insertion order, so the strategy retires the oldest non-invalidated entry of `state.items` directly and `reclaim` moves a reclaimed item to the back by re-adding it. Nothing outside the pool's own state references an item, so it is released the moment it leaves the pool, whether by invalidation, a consumed acquisition failure, or shutdown. The pool core no longer contains any strategy-specific hooks; its only change is skipping `onAcquire` for a failed acquisition that a borrower consumed while it was finalizing. Public types, last-lease finalization, FIFO retirement, and minimum-size behavior are preserved. Each retirement skips over invalidated busy items ahead of it in insertion order, which is the one cost of not keeping a separate list.

Tests cover four GC retention cases, last-lease finalization, and FIFO/minimum-size behavior after head, middle, or tail removal. Pool and McpServer tests share a GC helper; GC cases are skipped on Bun and Deno. Includes an `effect` patch changeset.

## Validation

Validation on implementation commit `6c79fc749f`:

- Pool suite: 46 passed, including all four GC regressions.
- McpServer suite: 69 passed.
- `pnpm lint` and `GOMEMLIMIT=4GiB GOMAXPROCS=4 pnpm check` passed.

Earlier failing-before verification substituted pre-fix `Pool.ts` from `3d59ae6d5f`: all four GC regressions failed and the other 39 Pool tests passed. With no strategy-owned queue there is no enqueue-after-removal window left to cover.

Toolchain commands ran through `nix develop -c`.

Closes EFF-1425


williamrobertson13:
thanks so much for fixing this up (and apologies for the slop PR submission)!

PR #8314, opened 2026-09-19
fix(SqlEventJournal): keep writeFromRemote callback errors

xia-chao:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

SQL `writeFromRemote` wraps your callback error in `EventJournalError`. The in-memory journal does not.

If `effect` or `compact` fails with your own tagged error:

- memory: you can catch that error by tag
- SQL: you get `EventJournalError` instead. Your error is only in `cause`

`write` and `withRemoteUncommited` already keep the callback error (#7844). This method was missed.

This PR stops the wrap. Your error comes back as-is. Real SQL journal failures still become `EventJournalError`.

New tests in `SqlEventJournal.test.ts` check that the original error object comes back from both `effect` and `compact`.

## Related

- Related Issue: none
- Closes: none

---

<details>
<summary><b>📸 Reproduction walkthrough — breakpoint evidence from the real sources (3 screenshots)</b></summary>

<br>

### Probe

One harness drives the public API with a callback that fails with the caller's own tagged error:

```ts
class ProbeError extends Data.TaggedError("ProbeError")<{ message: string }> {}

journal.writeFromRemote({
  remoteId: EventJournal.makeRemoteIdUnsafe(),
  entries: [new EventJournal.RemoteEntry({ remoteSequence: 0, entry })],
  effect: () => Effect.fail(new ProbeError({ message: "probe" }))
})
```

The same probe is run against both public backends.

### 1. SQL — where the callback error is wrapped

The debugger is paused at **`packages/effect/src/unstable/eventlog/SqlEventJournal.ts:245`**, inside the `catchIf` handler:

```ts
writeFromRemote: (options) =>
  writeFromRemote(options).pipe(
    withTracerDisabled,
    Effect.catchIf(
      (e) => e._tag !== "EventJournalError",                                                             // ← 244
      (cause) => Effect.fail(new EventJournal.EventJournalError({ cause, method: "writeFromRemote" }))   // ← 245
    )
  ),
```

Debugger paused at SqlEventJournal.ts:245 - VARIABLES shows cause = ProbeError

What the screenshot shows:

- **VARIABLES → `Local`** — `cause = ProbeError: probe`. The caller's own tagged error has reached this line and is one step away from being wrapped.
- **CALL STACK** — top frame is `SqlEventJournal.ts 245:1`, i.e. a real source file of this repository.
- The predicate one line above (`e._tag !== "EventJournalError"`) means **everything whose tag is not `EventJournalError` gets wrapped** — callback errors included.

### 2. memory — the callback error is passed straight through

Same probe, same callback failure. The debugger is paused at **`packages/effect/src/unstable/eventlog/EventJournal.ts:453`**:

```ts
for (const originEntry of compacted) {
  ...
  yield* options.effect({ entry: originEntry, conflicts })   // ← 453
  break
}
```

Debugger paused at EventJournal.ts:453 - memory yields the callback with no wrap

What the screenshot shows:

- **CALL STACK** — top frame is `EventJournal.ts 453:1` (real source), and the stack runs straight from memory's `writeFromRemote` into the Effect runtime. **There is no `catchIf` / `catchTag` frame anywhere on it.**
- The surrounding code (visible in the same view, from the function head at line 419 down to line 453) call the callback and yield it directly — nothing wraps `options.effect` or `options.compact`.

### 3. The observable difference

Debug Console - SQL observes EventJournalError, MEMORY observes ProbeError

```
callback error thrown: ProbeError
SQL    journal observes: EventJournalError  (cause: ProbeError)
MEMORY journal observes: ProbeError
```

Same callback, same `ProbeError`. On SQL the caller sees `EventJournalError`; on memory the caller sees their own error.

### Why this is a defect

`writeFromRemote` is one public method with one declared error channel. `write` and `withRemoteUncommited` were already aligned by #7844 — they map only the journal's own SQL failures and leave callback errors alone. `writeFromRemote` kept a blanket `catchIf` over the whole operation, so:

- `Effect.catchTag("MyError", ...)` around `writeFromRemote` works on memory and silently does not work on SQL;
- the caller's error is only reachable through `cause`, so tag-based recovery has to be rewritten per backend.

This PR removes the blanket `catchIf` and maps only the journal's own SELECT/INSERT, matching `write` and `withRemoteUncommited`.

</details>


xia-chao:
@tim-smart 
Please review

tim-smart:
These methods don't need to have generic errors :)

PR #8322, opened 2026-09-20
warn on duplicate cluster workflow tag registration

### LINKED ISSUE #8297: [ClusterWorkflowEngine] silent first-wins on duplicate workflow tag registration — restore re-registration warning
## What happened

Two workflow definitions sharing one tag (`OrganizationDeploymentWorkflow`) coexisted in our program: an unannotated definition used by submitters and an `inWorkflowShardGroup`-annotated duplicate with a different payload shape used by the runner. The result was a silent stall — submits journaled (202 accepted), no runner ever activated, no error anywhere. It took a full live-debug session (role-log markers, journal dumps, DB snapshots) to localize, across 53 same-tag pairs.

## Root cause in v4 source

`packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts` (`ensureEntity`):

- Registry is `Map<string, Workflow.Any>` keyed by tag only; a second definition with the same tag is silently ignored (first-wins, no overwrite, no diagnostic).
- Routing uses the *submitter's* annotations (`options.workflow.annotations` → shard group), while payload decode uses the *registered* definition's schema — so a same-tag/different-shape pair can route to an unserved group AND decode against the wrong schema, with neither side aware of the other.

## History

Old `@effect/cluster` defected on re-registration; #5693 deliberately downgraded that to a warning. The v4 `effect/unstable` rewrite dropped even the warning — the branch is now fully silent.

## Ask

Restore at least the warning on duplicate tag registration (tag + both payload-shape fingerprints), ideally also when the two definitions' annotations or payload schemas differ. A string-keyed global registry is invisible to TypeScript by construction (`register(workflow: Workflow.Any, …)` erases payload to `object`), so a runtime diagnostic is the only guard available. Happy to PR it if the shape of the warning is agreed.

ArjunCodess:
## Summary
- restore a warning when a second workflow definition registers an existing ClusterWorkflowEngine tag
- keep first-wins, and mention payload schema / annotations when those differ

Fixes #8297

## Test plan
- [ ] `packages/effect/test/cluster/duplicateWorkflowRegistration.test.ts`
- [ ] registering two `Workflow.make` definitions with the same tag logs the warning and does not start a second runner


tim-smart:
There is a few issues with this PR, so I will solve the issue in another PR.

PR #8325, opened 2026-09-20
Add native socket formatting helpers to NetAddress

fubhy:
Native socket adapters need platform-specific IPv6 zone formatting. Add `NetAddress.formatNativeHost(address, scopeIds, platform?)` and `NetAddress.formatMulticastInterface(networkInterface, scopeIds, platform?)`, accepting the map returned by `scopeIdsFromInterfaces` directly.

The optional third argument accepts a platform string; omitting it uses non-Windows behavior. Windows retains numeric scope IDs; other platforms use the first matching interface name with numeric fallback. IPv4 formatting and zero-scope behavior are preserved. Both helpers are pure and leave platform detection and interface discovery to callers. The supplied helpers were not present on `main`, so there are no existing call sites to migrate.

Validation: all 44 NetAddress tests pass; `pnpm check`, `pnpm lint-fix`, `pnpm lint`, and `pnpm jsdocs --check` pass. Includes a patch changeset.


packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -1356,6 +1356,67 @@ export const inetAddressFromStringUnsafe = (input: string): InetAddress =>
...
+    : networkInterface === 0
+    ? "::"
+    : `::%${platform === "win32" ? networkInterface : resolveScopeName(networkInterface, scopeIds)}`
+
+const resolveScopeName = (index: number, scopeIds: ReadonlyMap<string, number>): string => {
+  for (const [name, scopeId] of scopeIds) {
+    if (scopeId === index) return name
+  }
+  return String(index)
+}
```
fubhy, line 1419: We can just write this imperatively without the ternary and the inline the lookup instead of an extra function call.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -1356,6 +1356,66 @@ export const inetAddressFromStringUnsafe = (input: string): InetAddress =>
...
+ *
+ * Supply the platform and a scope map from {@link scopeIdsFromInterfaces}; this
+ * function performs no operating-system lookups.
+ *
+ * @category encoding
+ * @since 4.0.0
+ */
+export const formatNativeHost = (
+  self: InetAddress,
+  platform: string,
```
fubhy: make `platform` the third argument and make it optional and maybe even just make it a boolean `win32` that defaults to false.

packages/effect/src/unstable/net/NetAddress.ts
```diff
@@ -1356,6 +1356,66 @@ export const inetAddressFromStringUnsafe = (input: string): InetAddress =>
...
+ *
+ * Supply the platform and a scope map from {@link scopeIdsFromInterfaces}; this
+ * function performs no operating-system lookups.
+ *
+ * @category encoding
+ * @since 4.0.0
+ */
+export const formatMulticastInterface = (
+  networkInterface: Ipv4Address | number,
+  platform: string,
```
fubhy: make `platform` the third argument and make it optional and maybe even just make it a boolean `win32` that defaults to false.

fubhy: Scratch that last bit ... string is fine and more ergonomic (just pass in `os.platform` result)

packages/effect/src/unstable/net/NetAddress.ts
fubhy: Scratch that last bit ... string is fine and more ergonomic (just pass in `os.platform` result)

PR #8329, opened 2026-09-20
Warn on conflicting workflow tag registrations

### LINKED ISSUE #8297: [ClusterWorkflowEngine] silent first-wins on duplicate workflow tag registration — restore re-registration warning

tim-smart:
## Summary

- warn when a different workflow definition reuses an existing tag
- include both payload shapes in the warning
- preserve first-wins registration and keep repeated registration of the same definition silent

## Validation

- `pnpm vitest run packages/effect/test/cluster/ClusterWorkflowEngine.test.ts -t 'duplicate workflow tags|same workflow definition'`
- `pnpm dprint check packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts packages/effect/test/cluster/ClusterWorkflowEngine.test.ts`
- `pnpm oxlint packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts packages/effect/test/cluster/ClusterWorkflowEngine.test.ts`

Closes EFF-1444
Closes #8297


packages/effect/src/unstable/cluster/ClusterWorkflowEngine.ts
```diff
@@ -47,6 +47,15 @@ import * as Reply from "./Reply.ts"
 import * as Sharding from "./Sharding.ts"
 import * as Snowflake from "./Snowflake.ts"
 
+const payloadShape = (workflow: Workflow.Any): string => {
```
tim-smart: Move this to the bottom of the file

packages/effect/test/cluster/ClusterWorkflowEngine.test.ts
```diff
@@ -1446,6 +1447,76 @@ describe.concurrent("ClusterWorkflowEngine", () => {
...
+        .map((message) => globalThis.Array.isArray(message) ? message.join(" ") : String(message))
+        .find((message) => message.includes("DuplicateClassPayloadWorkflow"))
+      assert(warning, "duplicate workflow registration must emit a warning containing its tag")
+      const normalized = warning.toLowerCase()
+      for (const fragment of ["organizationid", "string", "deploymentid", "number"]) {
+        assert.include(normalized, fragment, `warning must identify both class payload shapes: ${warning}`)
+      }
+    }).pipe(Effect.provide(TestWorkflowEngine), Effect.withLogger(logger))
+  })
+  it.effect("does not fail when duplicate workflow payload shapes cannot be rendered", () => {
```
tim-smart, line 1492:
```suggestion
  })

  it.effect("does not fail when duplicate workflow payload shapes cannot be rendered", () => {
```

PR #8334, opened 2026-09-21
feat(effect): Added `Config.flatMap`

nikelborm:
**😉 Ready for review 🫶**

## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added support for `Config.flatMap` to significantly simplify the chaining and branching of configs. Without this operator, you can't provide a fallback config when the other is missing. You'd need to step completely out of the `Config` world and use plain effect combinators instead, losing composability with other config values.


packages/effect/src/Config.ts
```diff
@@ -261,6 +261,83 @@ export const map: {
...
+ * @since 2.0.0
+ */
+export const flatMap: {
+  <A, B>(f: (a: A) => Config<B>): (self: Config<A>) => Config<B>
+  <A, B>(self: Config<A>, f: (a: A) => Config<B>): Config<B>
+} = dual(2, <A, B>(self: Config<A>, f: (a: A) => Config<B>): Config<B> => {
+  return make((provider, pathPrefix) =>
+    Effect.flatMap(evaluateAt(self, provider, pathPrefix), (resolution) =>
+      resolution._tag === "Resolved"
+        ? evaluateAt(f(resolution.value), provider, pathPrefix)
```
tim-smart, line 336: This will lose `resolution.hasInput` from the outer config.

packages/effect/src/Config.ts
```diff
@@ -261,6 +261,83 @@ export const map: {
...
+ *       )
+ *       // prod is much stricter
+ *       : Config.NonEmptyString("PROD_HOST").pipe(
+ *         withAbsenceFallback(Config.NonEmptyString("HOST"))
+ *       )
+ *   )
+ * )
+ * ```
+ *
+ * @see {@link mapEffect} – to avoid propagating absence to the whole config
```
tim-smart, line 321: Not sure about putting `mapEffect` here. Using a config value in a mapEffect will lose the path prefix.

packages/effect/src/Config.ts
```diff
@@ -261,6 +261,83 @@ export const map: {
...
+ * ```ts import.meta.vitest
+ * import { Config, Option } from "effect"
+ *
+ * const EnvVar = (name: string) =>
+ *   Config.Literals([
+ *     ...["prod", "production", "PROD", "PRODUCTION"] as const,
+ *     ...["dev", "development", "DEV", "DEVELOPMENT"] as const
+ *   ], name)
+ *
+ * const withAbsenceFallback = <A, B>(fallback: Config.Config<A>) => (self: Config.Config<B>): Config.Config<A | B> =>
```
tim-smart, line 287: We could probably add `orElseIfAbsent`

tim-smart (changes requested):

packages/effect/src/Config.ts
nikelborm: but that's intentional

packages/effect/src/Config.ts
nikelborm: Do you mean it as a replacement for the whole `flatMap`, or as an addition beside it? I can do either option

packages/effect/src/Config.ts
nikelborm: I tried the implementation that has `preserveInputEvidence(evaluateAt(f(resolution.value), provider, pathPrefix), resolution.hasInput)` here. 
And with it, even the basic test fails.

```
it("flatMap lets an inner Absent stay recoverable through Config.option", async () => {
      const hostConfig = Config.Int("port").pipe(
        Config.flatMap((port) =>
          port === 80
            ? Config.String("prodHost")
            : Config.String("devHost")
        )
      )

      await assertSuccess(
        Config.option(hostConfig),
        ConfigProvider.fromUnknown({ port: "80" }),
        Option.none()
      )
```

```
- Expected
+ Received

  {
    "_id": "Result",
-   "_tag": "Success",
-   "value": {
-     "_id": "Option",
-     "_tag": "None",
+   "_tag": "Failure",
+   "failure": ConfigError {
+     "_tag": "ConfigError",
+     "cause": SchemaError {
+       "message": "Expected string
+   at [\"prodHost\"]",
```

nikelborm: I even specifically test for it:
```ts
it("flatMap doesn't preserve hasInput across nested composition", async () => {
      const config = Config.all({
        flag: Config.Int("port").pipe(
          Config.flatMap(() => Config.option(Config.String("unused")))
        ),
        required: Config.String("required")
      }).pipe(Config.withDefault({ flag: Option.none(), required: "default" }))

      // If you want to get `Expected string at ["required"]` error instead, you
      // can just use `Config.mapEffect` in place of `Config.flatMap`
      await assertSuccess(
        config,
        ConfigProvider.fromUnknown({ port: "80" }),
        { flag: Option.none(), required: "default" }
      )
    })
```

nikelborm: Although, as you noted the advice to use `config.mapEffect` is not good, because of the path prefix

I initially thought about adding something like `.flatMap` and `.flatMapEager` and the second one would have `preserveInputEvidence` and would fail early (eagerly) instead of my previous recommendation for mapEffect.

Do you think a second operator with slightly different behavior is a good idea? Should I choose a different name instead of `.flatMapEager`? Because any other name I considered sounded like exposure of implementation details. Because of that, I haven't added the second one at all, since `Config.flatMap` with the other implementation is useless for my use case, and I couldn't have both with sane names.

tim-smart: Currently Config.flatMap + Config.succeed has different behavior to Config.map. They should probably be aligned with whatever we end up with.

I think this has uncovered an issue with how absence is treated inside Config.all:



I think we want the "target" behavior.

packages/effect/src/Config.ts
tim-smart: An addition in another PR.

packages/effect/src/Config.ts
gcanti: https://github.com/Effect-TS/effect/pull/8488

nikelborm: ok, let me rebase on top of main, since it's merged

packages/effect/src/Config.ts
nikelborm: Ok, I'll do that in a follow-up PR, based on `.flatMap` introduced by this one

packages/effect/src/Config.ts
nikelborm: Removed info about it

packages/effect/src/Config.ts
nikelborm: Rebased on top of #8488 and reworked tests to cover every data flow I could think of. Also made tests more granular so that it's much easier to pinpoint failed assertions.

gcanti (approved):

PR #8344, opened 2026-09-21
Fix composed-hash collisions and equality bugs, and reduce hashing work

front-depiction:
`Hash` and `Equal` already require values to be immutable once they have been hashed or compared, and Effect's collections are persistent, so most of what these functions compute is a fact that stays true. `main` relied on that only partially: object hashes were cached, but `Date` and `RegExp` bypassed the cache, `Equal` kept a cache entry per compared pair, prototype chains were walked again on every hash and comparison, and a new version of a `HashMap` or `Chunk` was hashed from scratch although it shares almost all of its structure with the version before it. Separately, hashes were composed with linear operations that discarded most of their information, which is what makes structured `HashMap` keys degrade toward quadratic behavior.

This PR fixes the composition, fixes four `Hash`/`Equal` correctness bugs found along the way, and caches each established fact at the granularity where immutability makes it reusable:

| fact | reused at |
| --- | --- |
| the hash of an object | the object (`WeakMap`, now consulted before any other work; dates are recomputed instead, as that is cheaper than a cache write) |
| what a class's prototype chain contributes | the prototype |
| the hash of a subtree or subsequence | the `HashMap` / `HashSet` / `Chunk` node, shared across versions |
| a proven equality | the equivalence class (union-find) |

Public APIs are unchanged; hash values change.

## Pathological collisions in composed hashes

Leaf hashes are cheap (a small integer is its own hash), but `main` composed them with linear operations: XOR for records, arrays, `Set` and `HashSet` members, `*53 ^` in `Hash.combine`, `hash ^ (hk + hv)` in `HashMap`, `Trie` and `Graph`. Linear folds of low-entropy inputs lose information: a record `{ x, y }` of small integers hashes to `c ^ x ^ y`, so 10,000 such records share 128 hashes; `[1, 2, 3]` hashes like `[]`; `[a, b]` like `[b, a]`. `HashSet` hashed every set to the same constant.

Collisions among 10,000 distinct values (about 0.02 expected for a well-distributed 31-bit hash), `node packages/effect/runtimeperf/suites/hash/quality.mts`:

| input family | main | this PR |
| --- | ---: | ---: |
| records `{ x, y }`, int pairs `[a, b]` | 9,872 | 0 |
| `HashMap { x, y }` | 9,772 | 0 |
| `HashSet { i, j }` (11,175 values) | 11,174 | 0 |
| sets `{ i, j }`, nested `{ a: [i, j], b: { c: k } }` | 10,920, 10,616 | 0, 0 |
| `Trie`, `Chunk [a, b]`, two-node `Graph` | 9,872 each | 0 |
| custom hasher `Hash.combine(hash(a), hash(b))` | 4,752 | 0 |
| permutations of `[0..6]` (5,040 values) | 5,039 | 0 |
| floats `i * 0.1`, `i * 0.001` | 8,352, 7,416 | 0, 0 |
| `null`/`undefined`/booleans vs their spellings | 4 | 0 |

Collisions are never incorrect, but hashed collections pay for them: colliding keys share collision nodes that are scanned with `Equal`, so every operation degrades with size. `HashMap` with `{ x, y }` record keys:

| N | insert | lookup (hit) | lookup (miss) | remove |
| ---: | --- | --- | --- | --- |
| 16 | 4.4× faster | 2.8× faster | 1.3× faster | 1.5× faster |
| 1,024 | 24× faster | 16× faster | 1.2× faster | 1.7× faster |
| 16,384 | 104× faster | 53× faster | 1.2× faster | 1.4× faster |
| cost per key, growth 16 → 16,384 (main → this PR) | 35.3× → 1.5× | 33.0× → 1.7× | 1.3× → 1.4× | 1.8× → 1.9× |

In `main` the cost per key grows linearly with the map (each operation scans a collision group proportional to N); here it grows with trie depth only. Lookups of absent keys were not affected in this corpus, since they hashed apart from present keys. With 10,000 fresh record or tuple keys, inserting then looking up each is 38× and 34× faster.

This affects every `MutableHashMap`-backed structure keyed by records, tuples or `Data` classes: `Cache`, `ScopedCache`, `RcMap`, `FiberMap`, `RequestResolver` grouping.

**Fix.** Leaf hashes stay as they are; avalanche is applied where hashes are composed, with the MurmurHash3 finalizer (`fmix32`) and block step, in `internal/hash.ts`:

- ordered composition (arrays, typed arrays, `Hash.combine`): the MurmurHash3 body step and finalizer, so position matters and repeated elements do not cancel;
- unordered composition (records, `Map`, `Set`, `HashMap`, `HashSet`, `Trie`, `Graph`): XOR of independently mixed per-entry terms, with the value scrambled before meeting the key so entries cannot trade places;
- non-int32 numbers hash their IEEE-754 bits; `null`, `undefined`, booleans, `NaN` and ±Infinity hash to distinct mixed tags instead of the hashes of their spellings.

`Hash.combine` keeps its signature; its docs no longer describe the formula.

This is the failure and the fix CPython went through for tuples of small integers (bpo-34751, replaced by an xxHash-derived combiner in 3.8) and Clojure 1.6 for small vectors and sets (collection hashing moved to Murmur3). Quality was measured with the corpus report above rather than assumed.

## `Equal` correctness

Four bugs in `main`, each with a regression test:

1. **The per-pair equality cache remembered provisional results.** Comparing cyclic values assumes a pair already being compared is equal; `main` cached inner results computed under that assumption even when the outer comparison then failed. Reproduction (the `x`/`y` swap makes the pair pass `main`'s hash prefilter):

   ```ts
   const A: any = { child: undefined, x: 1, y: 2 }
   const B: any = { child: undefined, x: 2, y: 1 }
   const P = { parent: A }
   const Q = { parent: B }
   A.child = P
   B.child = Q
   Equal.equals(A, B) // false
   Equal.equals(P, Q) // main: true (P and Q are not equal)
   ```

2. **Hashes of cyclic values depended on hashing order**, because hashes computed across a cycle (relative to where the traversal started) were cached. Two isomorphic cyclic graphs could hash differently, and `Equal.equals` rejected them on the hash prefilter. Hashes computed across a cycle are no longer cached.

3. **A throwing custom `[Hash.symbol]` left its object marked as visited**, so later hashes of it silently returned the circular-reference constant. The visited sets are now maintained in `finally` (also in `Equal`).

4. **`MutableHashMap` lost its entries after being used as a key.** Lookups read `referentialKeysCache.get(self)` (keyed by the map) while writes stored `set(key, storedKey)` (keyed by the query key), and the cache was shared by all maps:

   ```ts
   const m1 = MutableHashMap.empty<object, string>()
   const m2 = MutableHashMap.empty<object, string>()
   MutableHashMap.set(m1, { id: 1 }, "v")
   MutableHashMap.set(m2, { id: 1 }, "v")
   const outer = MutableHashMap.empty<object, string>()
   MutableHashMap.set(outer, m2, "outer")
   MutableHashMap.get(outer, m1) // Some("outer"), found through Equal
   MutableHashMap.get(m1, { id: 1 }) // main: None
   ```

   A query-key to stored-key shortcut is only valid per map, and is redundant given cached key hashes and the equivalence cache below, so it is removed.

## `Equal` performance

Profile of `Equal.equals` on fresh nested records in `main`: 37.5% computing the hashes of both operands, 17.9% plus most GC time in the per-pair cache, 9.9% in `getAllObjectKeys` sets.

- **Hashes remain fingerprints.** `Equal` still compares the (cached) hashes of both operands first, now at the lower cost of the hashing changes below: each value is traversed once, and every later comparison with an unequal value is rejected in O(1). Using only hashes computed elsewhere was measured as an alternative: it made one-off comparisons faster, but pairwise comparison of many values (deduplication by `Equal`) then traversed every pair, a regression growing with input size (6.6× to 52× slower than `main` for 100 records), so it was not kept.
- **Equivalence classes replace the pair cache.** Equality is an equivalence relation, so proven equalities are kept with union-find (as in Hopcroft–Karp equivalence checking), which also answers pairs related only by transitivity. Only top-level `true` results are recorded: once the outermost comparison completes, every provisional assumption made for cycles has been discharged. Objects map to class tokens through a `WeakMap`; tokens reference only their parent token.
- **Instances of the same class** share every prototype member, so only own fields and accessors are compared, with the prototype layout computed once per class (shared with `Hash`).
- **Records** compare values before checking that the key sets match, so unequal records return at the first differing value.
- **`Map` and `Set`** entries are matched within their hash group, O(n) expected instead of O(n²), exact because equal values have equal hashes. This also applies to small collections: `Equal` has already hashed both operands, so the element hashes are mostly cached.

| workload | this PR vs main | main → this PR |
| --- | --- | --- |
| fresh nested records, compared once | **1.7× faster** | 8.73 µs → 5.02 µs |
| fresh `Map` / `Set` of 8 entries | **2.1× faster** / **2.3× faster** | 31.89 µs → 15.35 µs / 19.78 µs → 8.79 µs |
| fresh `Map` / `Set` of 1,000 entries | **64× faster** / **135× faster** | 108.13 ms → 1.68 ms / 121.41 ms → 897.99 µs |
| consecutive comparison of 1,000 `Data` class values | within noise | 7.03 ms → 1.77 ms |
| all pairs of 100 large records differing late (fresh / reused) or early | **2.6× faster** / within noise / **2.6× faster** | 1.14 ms → 431.07 µs / 136.67 µs → 135.45 µs / 1.18 ms → 446.51 µs |

## Everyday operations

Public operations whose cost includes hashing and equality, on 1,000 fresh values (records of small integers or `Data` classes, half of them duplicates), plus iteration:

| workload | this PR vs main | main → this PR |
| --- | --- | --- |
| `MutableHashMap` set and get, record keys (backs `Cache`, `RcMap`, `RequestResolver`) | **2.6× faster** | 3.77 ms → 1.43 ms |
| `Array.dedupe`, records / `Data` classes | **1.9× faster** / **1.9× faster** | 1.11 ms → 579.44 µs / 2.22 ms → 1.14 ms |
| `Array.difference` / `Array.intersection`, records | **1.9× faster** / **1.9× faster** | 2.66 ms → 1.38 ms / 2.67 ms → 1.39 ms |
| `Array.dedupeAdjacent`, records | **1.9× faster** | 2.17 ms → 1.15 ms |
| `HashSet.union`, records / `HashMap.union`, `Data` class keys | **2.4× faster** / **2.1× faster** | 3.90 ms → 1.64 ms / 3.08 ms → 1.47 ms |
| `Stream.changes` over `Data` class values (`Effect.runSync`) | **1.6× faster** | 3.34 ms → 2.06 ms |
| iterate a `HashMap` of 10,000 entries: entries / keys | **4.5× faster** / **7.1× faster** | 307.32 µs → 68.40 µs / 327.78 µs → 45.90 µs |
| iterate a `HashSet` of 10,000 values | **6.7× faster** | 332.03 µs → 49.28 µs |

## Persistent structures

`HashMap`, `HashSet` and `Chunk` share structure between versions; their hash summaries now do too, so hashing a new version only visits the nodes it does not share with an already hashed one:

- `HashMap`/`HashSet`: each trie node caches the XOR of its entry terms (a Merkle tree over the HAMT, valid because the summary is a commutative monoid). In-place edits during transient mutation clear the cache of every node they touch. Equality of two maps rejects in O(1) when the cached hashes differ.
- `Chunk`: the summary must be associative but not commutative, so it is a polynomial (Rabin–Karp) hash over mixed element terms, `h(L ++ R) = h(L)·B^|R| + h(R)`: every concatenation tree of the same sequence hashes alike, and `append` combines the cached base summary with one element. Slices are hashed from the summaries of the nodes they cover, without copying. The summary lives in a slot initialized for every chunk.
- `Trie`: the key of an entry is spelled along its path, so subtree summaries are not self-contained and caching does not apply. Hashing folds the tree once with an incremental key hash instead of building key strings, and equality walks both tries in lockstep without materializing them.
- `HashMap.forEach`/`reduce` fold the trie directly, and iteration walks it with one explicit stack instead of an iterator per node, with `keys`/`values` (and so `HashSet`) yielding elements without building entry tuples.

This is the monoid-annotated tree pattern (Merkle trees; measures in finger trees, Hinze & Paterson).

Hashing a new version, by size of the structure:

| workload | 1,000 | 10,000 | 100,000 | growth 1k → 100k (main → this PR) |
| --- | --- | --- | --- | --- |
| `set` one key of a `HashMap`, then compare with the previous version (as `AtomRef` and `Stream.changes` do) | 69× faster | 585× faster | 5,274× faster | 161× → 2.1× |
| hash `Chunk.append(base, x)` | 17× faster | 179× faster | 2,099× faster | 129× → 1.1× |
| hash a slice of a concatenated `Chunk` | 19× faster | 81× faster | 905× faster | 45× → 0.9× |

Hashing and comparing a `Trie` derived from one of 1,000 entries is 5.8× faster and 2.4× faster.2× faster and 2.6× faster.

## Hash immutable values once

- The object cache is consulted first; `Date` and `RegExp` no longer bypass it (and a `Date` subclass implementing `Hash` is honored).
- Prototype member contributions are computed once per class, held weakly by the prototype. Methods and object-valued members contribute through their key, so structural hashes no longer include per-process random function hashes and are identical across processes for `Data` and `Schema` classes.
- Containers whose members are all primitives cannot be part of a cycle and skip cycle tracking (about 48 ns per cold hash).
- Dates hash their timestamp (no ISO string) and are not cached (recomputing costs about as much as a cache lookup and far less than a cache write); regular expressions hash their source and flags, and symbols their description, without building strings; `Map` and `Set` are hashed with `forEach` (no entry tuples); fixed tags in Effect's own hashers are computed once.

| workload | this PR vs main | main → this PR |
| --- | --- | --- |
| warm (cached) record / `Date` / `RegExp` | **2.4× faster** / **31× faster** / **24× faster** | 13.4 ns → 5.6 ns / 368.9 ns → 11.8 ns / 134.0 ns → 5.6 ns |
| mixed corpus of 256 values, warm / cold | **5.0× faster** / **1.3× faster** | 10.23 µs → 2.06 µs / 171.12 µs → 135.20 µs |
| cold record / `Data` class / `Option.some(record)` | **1.3× faster** / **1.7× faster** / **1.2× faster** | 415.6 ns → 309.3 ns / 1.16 µs → 692.1 ns / 628.0 ns → 526.2 ns |
| cold `Date` / `Uint8Array(64)` | **11× faster** / **1.4× faster** | 413.4 ns → 36.4 ns / 526.8 ns → 373.6 ns |
| `HashMap` with 64 `Data` class keys, insert and lookup | **2.0× faster** | 298.29 µs → 149.55 µs |

## Costs

| workload | this PR vs main | main → this PR | cause |
| --- | --- | --- | --- |
| first hash of a `RegExp` | 1.35× slower | 235.9 ns → 318.5 ns | it is now cached like other objects (repeated hashes: 24× faster); 1.1× to 1.35× slower across runs |
| integer / float leaf | within noise / 1.13× slower | 5.4 ns → 6.2 ns / 4.8 ns → 5.4 ns | floats hash their IEEE-754 bits, which removes the float collisions above |
| the same two distinct equal values, compared repeatedly (nested / `HashMap` of 1,000) | 2.07× slower / 1.73× slower | 25.2 ns → 52.2 ns / 25.3 ns → 43.7 ns | a hash lookup per side plus the equivalence-class lookup, where `main` did one pair-cache lookup |

**Bundle size.** The `basic.ts` fixture (`Effect.succeed(123).pipe(Effect.runFork)`) grows from 6.88 KB to 8.28 KB gzipped (+1.40 KB). Every program includes `Hash` and `Equal` (core Effect types implement them), and everything reachable from the two dispatchers is bundled. Subsumed and niche paths were removed to bring this down from +1.85 KB (the `Map`/`Set` quadratic helpers are now only bundled when imported). Most of what remains is the per-class shape and same-class comparison paths; the collision mixing, the equivalence cache and hash-grouped `Map`/`Set` matching account for the rest.

Within noise of `main`: cold `Map`, `Set` and `Chunk` hashing, tuples, nested records, strings, bigints, consecutive `Data` class comparison, and repeated pairwise comparison of the same 100 records.

## Contract notes

- Hash values change. Nothing in Effect persists or transmits `Hash` values (cluster sharding uses its own hash); doctests that pinned values are updated, and the two tests that pinned collisions of the old algorithm (`Request`, `Trie`) use pairs found by search that collide under the new one.
- Class instances hashed or compared structurally, that is without their own `Hash` / `Equal` (for example `Data.Class` and `Schema.Class` instances), have their prototype chain below `Object.prototype` read once per prototype. For those instances only, the existing immutability requirement now extends to that chain: members added, removed or redefined on it after instances were hashed or compared are not reflected. Values implementing `Hash` / `Equal`, plain objects, arrays and other built-ins are unaffected. Documented on `Hash.hash` and `Equal.equals`.
- `Equal` docs describe the equivalence cache and hash-matched `Map`/`Set` comparison instead of the pair cache and O(n²).
- Weak references only: every new cache is keyed by the object, prototype or node it describes (the GC was checked to collect classes, prototypes, compared values and their class tokens, persistent structures and symbol keys). Registered symbols are cached in a `Map`, as the registry never releases them.

## Not included

- `Schema.toEquivalence` for maps and sets still uses the quadratic `Equal.makeCompareMap`/`makeCompareSet`: a schema can override element equivalence (for example case-insensitive keys), and hash grouping requires a hash compatible with that equivalence. Deriving a hash alongside the equivalence would be new API.
- String hashing is unchanged: in V8 it is bound by `charCodeAt` (block-wise FNV and Murmur variants measured no faster than djb2 at lengths 8 to 256).
- `AtomRef` hashes and compares its mutable current value, which contradicts the immutability contract.
- `runtimeperf/compare.mts` trimmed `git diff` output before applying it, corrupting patches ending in a blank context line; fixed here because the comparisons above depend on it.

## Checks

- `pnpm test --run --project effect`: 10,404 passed; `ClusterWorkflowEngine > a parent stays suspended past the activity interrupt retry budget` times out only under full-suite load, as on `main`, and passes alone (45/45)
- `pnpm vitest run --config vitest.docs.ts`: 3,406 passed
- `pnpm check`, `pnpm lint`, `pnpm circular`, `pnpm jsdocs`
- `pnpm test-types`: 2,542 passed
- `pnpm runtimeperf-compare hash --rounds 9 --base upstream/main --head HEAD` (Node v24.3.0, V8 13.6; each round runs both sides in fresh processes, with bootstrap intervals; every multiple above is computed from these runs): all timings above
- `node --test packages/effect/runtimeperf/test/*.test.mts`: 18 passed
- `node packages/effect/runtimeperf/suites/hash/quality.mts`: collision table above
- Tests: a reference model of structural hashing fuzzed over 3,000 random prototype chains; the same-class `Equal` path against a `getAllObjectKeys` reference on 3,000 random chains; differential agreement of `Equal` with `main` on 60,000 generated pairs (run during development, before the same-class path); the equivalence cache against cache-free comparisons (fails when the cache is made to record `false`); Merkle hashes under persistent and transient edits (fails when invalidation is removed); container laws (order-insensitivity of commutative containers, order-sensitivity and associativity for sequences)


packages/effect/src/internal/graph.ts
```diff
@@ -106,17 +113,19 @@ const ProtoGraph = {
...
-    return hash
+    // Nodes and edges are independent mixed terms, XOR-folded (order-insensitive
+    // like graph equality), rather than linear sums that cancel and collide.
+    this.nodes.forEach((nodeData, nodeIndex) => {
+      hash ^= entryTerm(Hash.hash(nodeIndex), Hash.hash(nodeData))
+    })
+    this.edges.forEach((edgeData, edgeIndex) => {
+      hash ^= mix(entryTerm(Hash.hash(edgeIndex), edgeHash(this.type, edgeData)))
+    })
+    return optimize(hash)
```
colelawrence: Consider native JS iterators (as it was before). They are almost always faster than function invocation if your goal is optimization

If you have the option for `for ; ; ;` iteration, that will be even faster for v8.

.changeset/persistent-structure-hashing.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
```
fubhy: Holy changelog. We have to adjust our instructions in the repository so that it stops generating a flood of changesets for intra-pr changes.

front-depiction: indeed, should be fixed now!

tim-smart:
Sorry but this change is way too intrusive to be considered at this stage. It also needs to stay under 7kb for the `basic.ts` bundle size benchmark.

PR #8353, opened 2026-09-21
Make awaitAllChildren interruptible

tim-smart:
## Summary

- make the child wait in `awaitAllChildren` interruptible when the caller is interruptible
- preserve both the wrapped effect's failure and the interruption when interrupted during that wait
- preserve caller masking when `awaitAllChildren` runs inside an uninterruptible region
- add regression coverage for the success, failure, cleanup, and caller-masking paths

## Validation

- `nix develop -c pnpm test packages/effect/test/Effect.test.ts --run -t awaitAllChildren`
- `nix develop -c pnpm lint`
- `nix develop -c pnpm check`
- `git diff --check`

Closes EFF-1458


Voyagerroc-Lab (commented):
Reviewed the interruptibility change in awaitAllChildren, including onExitPrimitive usage, preservation of the wrapped failure cause, interruption propagation to child fibers, and masking under an enclosing uninterruptible region. The regression tests cover success, failure, cleanup, and masking semantics, and the public documentation matches the final behavior. I found no blocking correctness issue.

PR #8357, opened 2026-09-22
feat(sql-clickhouse-native): add clickhouse native support

### LINKED ISSUE #8343: Native TCP transport for @effect/sql-clickhouse-native

`@effect/sql-clickhouse` currently uses ClickHouse's HTTP interface. For workloads that need the ClickHouse Native protocol over TCP, this introduces an unavoidable HTTP transport boundary and prevents use of Native protocol features such as binary block encoding.

I would like to use ClickHouse directly through its Native TCP protocol while retaining the existing Effect SQL abstractions:

- `SqlClient`
- `SqlConnection`
- `Statement`
- typed `SqlError` values
- `Stream`
- scoped connection lifecycle, pooling, interruption, and Effect concurrency

This is particularly useful for high-volume inserts, where ClickHouse Native blocks avoid constructing large SQL `VALUES` statements and align with ClickHouse's official native clients such as `clickhouse-go`, `clickhouse-rs`, and `ch-go`.

Add a Native TCP transport option for ClickHouse SQL, either as:

- a new package, such as `@effect/sql-clickhouse-native`; or
- an additional transport implementation in `@effect/sql-clickhouse`.

The implementation would:

- connect to ClickHouse over the Native TCP port (normally `9000`);
- implement the client/server Native protocol handshake, authentication, query packets, progress packets, exceptions, and end-of-stream packets;
- encode inserts as Native data blocks;
- decode result blocks into Effect SQL result rows;
- expose the standard `SqlClient` API and map server failures into Effect SQL errors such as `ConnectionError`, `AuthenticationError`, `AuthorizationError`,
`SqlSyntaxError`, `StatementTimeoutError`, and `UnknownError`;
- use an Effect-managed connection pool so `Effect.all` and `Effect.forEach` can run independent queries concurrently.

## Important protocol constraint

This should not promise PostgreSQL-style multiplexing over a single Native TCP socket.

A ClickHouse Native connection processes a query exchange sequentially. Concurrency should therefore come from a pool of Native connections; each physical connection should handle one active query at a time. This preserves protocol correctness and avoids interleaving server packets from unrelated queries.

## Proposed API shape

```ts
import { ClickhouseNativeClient } from "@effect/sql-clickhouse-native"
import { Effect, Layer } from "effect"

const ClickhouseLive = ClickhouseNativeClient.layer({
  host: "127.0.0.1",
  port: 9000,
  database: "default",
  username: "default",
  password: "",
  pool: {
    min: 1,
    max: 10
  }
})

  const sql = yield* ClickhouseNativeClient.ClickhouseNativeClient

  return yield* sql`
    SELECT instrument_id, observed_at
    FROM market_data.tsetmc_analysis_market_watch
    ORDER BY observed_at DESC
    LIMIT 100
  `
}).pipe(Effect.provide(ClickhouseLive))
```

For high-volume writes, it would be valuable for the implementation to use Native block encoding internally while preserving the normal SQL client interface.

## What alternatives have you considered?

- Current HTTP ClickHouse client: works well for general use, but does not provide a Native TCP transport.
- Application-specific driver: possible, but duplicates protocol, pooling, error mapping, and Effect resource-management work for each application.
- PostgreSQL-style socket multiplexing: not appropriate for ClickHouse Native protocol; a bounded Native connection pool is the intended concurrency model.

## Additional context

I have implemented and integration-tested a Native TCP client in an Effect application, including authentication, reads, Native block inserts, pool-based concurrent queries, and mapping to Effect SQL abstractions. I would be interested in contributing a focused implementation or extracting a minimal reproduction/prototype if this direction fits the project.

smhmayboudi:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds the `@effect/sql-clickhouse-native` package, providing native ClickHouse support for Effect SQL.

Includes client configuration, pooling, SQL client/connection support, migrations, error classification, package registration, documentation, and integration/unit test coverage.

## Related

- Related Issue #8343
- Closes #8343


smhmayboudi:
There some problems with references which I changed the tsconfig.tests.json. you can search `} from "effect/schema"` on test files, fix them and remove the tsconfig.tests.json extra line.

smhmayboudi:
I will add more tests.

smhmayboudi:
It need more test which I am going to add on next Sunday.

PR #8366, opened 2026-09-22
add `Schema.String.withAutocomplete`

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

```ts
const auto = Schema.String.withAutocomplete(["gpt-4o", "o3-mini"])
(typeof schema)["Type"] // <-- "gpt-4o" | "o3-mini" | (string & {})
auto.make("") // <-- suggests "gpt-4o", "o3-mini" options, but accepts any string
```

## Related
discord https://discord.com/channels/795981131316985866/1125094089281511474/1551638382428291132

- Related Issue #
- Closes #


KhraksMamtsov:
@gcanti , what do you think about this little feature?

gcanti:
Not sure about a specific constructor for this, what about exposing `Schema.StringForLiteralAutocomplete` as a fixed schema for `string & {}`, then composing it with `Literals` through `Union`?

The implementation in `Schema.ts` could be:

```ts
export interface StringForLiteralAutocomplete extends
  Bottom<
    string & {},
    string & {},
    never,
    never,
    SchemaAST.String,
    StringForLiteralAutocomplete
  >
{}

export const StringForLiteralAutocomplete: StringForLiteralAutocomplete =
  String
```

The explicit type annotation exposes `string & {}` for both `Type` and `Encoded`, while reusing the existing `String` instance. No cast is needed.

Usage:

```ts
const Method = Schema.Union([
  Schema.StringForLiteralAutocomplete,
  Schema.Literals(["GET", "POST"])
])

type Method = typeof Method.Type
// (string & {}) | "GET" | "POST"

Method.make("PATCH")
```

The literals still provide editor suggestions and remain available in the AST for introspection and derived tooling:

```ts
const knownMethods = Method.members[1]
knownMethods.literals // ["GET", "POST"]
```

KhraksMamtsov:
@gcanti 
I don’t have the same level of expertise as you do. So if you think this is a better approach, I’m 100% on board. Should I take care of it, or are you going to do it yourself?

gcanti:
> So if you think this is a better approach

Honestly I'm not sure yet, I'm still thinking about it. I definitely would not change `Schema.String`, as I don't want to increase the bundle size.

As for a dedicated constructor, it would be more convenient, but using a specific schema like `StringForLiteralAutocomplete` seems more composable, so for now I prefer that approach. For example, you might already have a union of literals defined somewhere, and you can get the autocomplete trick simply by adding `StringForLiteralAutocomplete` to the union.

KhraksMamtsov:
Yes, a separate schema for this union member (the one that does the autocomplete trick) definitely makes more sense, since it could be more reusable.
I’ll make the changes now.

PR #8368, opened 2026-09-22
Fix remaining Hash/Equal correctness gaps and hash persistent structures incrementally

front-depiction:
This PR includes #8360 (Tim's commits, unchanged) and builds on it. If #8360 lands first, this PR reduces to the commits on top. It also brings in the structural pieces from #8344, rebuilt on #8360's primitives.

| | vs #8360 |
|---|---|
| **Bugs fixed** | 4 that #8360 still has: history-dependent equality cache, a throwing `Equal` poisoning later comparisons, wrong answers on cyclic values (261 of 3001 comparisons disagree with bisimilarity on #8360, 0 here), and singletons hashing like their names |
| **Faster** | `HashMap` equality after a one-key change 479× (10k) and 4,159× (100k); `AtomRef.update` on a 10k map 512×; `Chunk` append 2,043× and slice 974× (100k); native `Map`/`Set` equality 14×/25× (1k); `HashMap` iteration 6.3×; `Trie` hash 3.8× and equality 2.1×; warm object hashing 1.5×; nested structural equality 1.16×. No measured slowdowns |
| **API** | No public signature, export or type changes. Hash values change for `HashSet`, `Chunk`, `Trie` and singletons |

## Correctness

Each fix has a regression test that fails on #8360 and passes here.

### 1. An earlier comparison could change a later result

**What.** `Equal.equals` cached `false` results and read the cache at every depth. A wrong `false` from one call then overrode a later comparison that matches.

```ts
// constant hash, v = 1 on every node, so all four nodes are equal
n0 = { v: 1, a: n2, b: n0 }; n2 = { v: 1, a: n0, b: n0 }
m0 = { v: 1, a: m1, b: m1 }; m1 = { v: 1, a: m1, b: m0 }
Equal.equals(n2, m1)                                  // false on #8360, and cached
Equal.equals({ w: n2, k: 1 }, { w: m0, k: 1 })        // true when evaluated fresh; #8360 returns false after the call above
```

**How.** Only the outermost comparison reads or writes the cache, and only final results are recorded (see fix 3).

**Why.** An outermost `true` has discharged every assumption it made, and assumptions only ever claim equality, so an outermost `false` is final too. A nested result reached under an ancestor's assumption is not.

Test: `Equal.equals > cyclic values > does not let an earlier comparison change a later result`.

### 2. A throwing custom `Equal` left its pair marked as visited

**What.** The visited sets were cleared only when the comparison returned. After one throw, the next comparison of the same objects took the circular-reference shortcut.

```ts
const x = { k: new Flaky(1) }, y = { k: new Flaky(2) }  // Flaky's [Equal.symbol] throws once
Equal.equals(x, y)   // throws
Equal.equals(x, y)   // true on #8360; x.k and y.k are different
```

**How.** Cleanup moved into `try/finally`, the same fix #8360 made in `Hash.hash`.

**Why.** Tracking state that outlives an exception makes every later comparison of those objects depend on a call that failed.

Test: `Equal.equals > custom Equal implementations > clears visited state when a custom Equal throws`.

### 3. Cycles were matched per object, not per pair

**What.** Each side had its own visited set. If both objects had already been visited, the comparison counted a closed cycle even when the two had never been compared with each other; if only one had, it counted a mismatch. The first rule accepts unequal graphs and the second rejects equal ones:

```ts
// a self-loop vs a node that points into an identical self-loop: equal
loop = { v: 0, a: loop, b: loop }; e0 = { v: 0, a: e0, b: e0 }; e1 = { v: 0, a: e0, b: e0 }
Equal.equals(loop, e1)   // false on #8360

// p0 only reaches v = 1, q0 reaches q1 with v = 0: unequal
p0 = { v: 1, a: p0, b: p0 }; p1 = { v: 0, a: p0, b: p0 }
q0 = { v: 1, a: q1, b: q0 }; q1 = { v: 0, a: q0, b: q0 }
Equal.equals(p1, q1)     // true on #8360
```

A property test compares both with a reference bisimilarity check (greatest fixpoint by partition refinement) on 500 random pairs of cyclic graphs with a constant hash (3001 comparisons). #8360 disagrees 261 times (13 false positives, 248 false negatives). This branch agrees on all of them.

**How.** A pair is assumed equal while it is being compared, and a cycle that returns to the same pair closes. This is the standard coinductive check; the pairs on a successful path form a bisimulation. Precedents: the Hopcroft–Karp equivalence algorithm and equality of recursive types (Amadio–Cardelli, Brandt–Henglein). The assumptions are the pairs on the current comparison path: two arrays pushed and popped in lockstep, so a comparison allocates nothing for them and holds nothing afterwards, including when it throws. Outermost results go in one `WeakMap<a, WeakMap<b, boolean>>`, weak on both sides.

This replaces the two visited sets and the pair cache.

Tests: `Equal.equals > cyclic values > assumes cycles equal pair by pair` and `… > agrees with bisimilarity on random cyclic graphs`.

### 4. Hash quality: singletons hashed like their names

A collision, not a correctness bug: `Hash.hash(null) === Hash.hash("null")`, and likewise `undefined`, `true`, `false` and symbols. These now share one tail that mixes their name hash once more. Java's `Boolean.hashCode` also gives `true` and `false` fixed constants rather than hashing their spelling. The change is smaller than the three arms it replaces.

Test: `Hash > collision resistance > does not hash null, undefined, booleans and symbols like their names`.

## Performance

Paired runs, #8360 (base) against this branch (head), fresh processes alternating per round, bootstrap 95% CIs; multiples computed from the reports. The structural rows come from a 9-round run over the full suite. The equality, insert and warm-hash rows come from targeted 3-round runs after the final equality and cache changes, and the `Trie` rows from a 5-round run.

| workload | #8360 → this PR | 95% CI |
|---|---:|---|
| `HashMap` equality after a one-key change, 100k | **4,159×** | 4,119–4,274 |
| `Chunk` append + hash, 100k | **2,043×** | 1,756–2,449 |
| `Chunk` slice + hash, 100k | **974×** | 826–1,176 |
| `AtomRef.update`, 10k-entry `HashMap` | **512×** | 471–527 |
| `HashMap` equality after a one-key change, 10k | **479×** | 469–484 |
| `Chunk` append + hash, 10k | **194×** | 190–197 |
| `HashMap` equality after a one-key change, 1k | **63×** | 62–64 |
| native `Set` / `Map` equality, 1k | **24.9× / 14.2×** | 24.4–26.0 / 13.7–14.3 |
| `HashSet` / `HashMap` keys iteration, 10k | **6.3× / 6.3×** | 6.2–6.4 / 5.9–6.4 |
| `Trie` hash / equality, 1k | **3.8× / 2.1×** | |
| pairwise dedupe, fresh / reused objects | **1.37× / 1.08×** | 1.33–1.38 / 1.06–1.14 |
| same pair compared repeatedly | **1.17×** | 1.15–1.17 |
| nested structural equality, cold | **1.16×** | 1.13–1.26 |
| warm object hashing | **1.5×** | 1.48–1.56 |
| `HashMap` hit / remove, 16–16k | **1.3–1.6×** | |

Scaling: `HashMap` equality after one change grows from 63× at 1k to 479× at 10k and 4,159× at 100k, because the cost now depends on the changed path, not the map size. `Chunk` append goes from 18× at 1k to 194× at 10k and 2,043× at 100k.

#8360 already brings the collision fixes' speedups over main (for example `Atom.family` with struct arguments 16–20×, `HashMap` with record keys 15–30×); those are included.

Each piece stays in the module that owns its data and uses #8360's primitives: `Hash.combine`, the XOR-of-mixed-terms unordered fold, and the rule that a hash which met a back-edge is not cached.

### HashMap and HashSet: cached subtree hashes

**What.** Each trie node caches the XOR of `combine(hash(key), hash(value))` over its subtree, the same fold #8360's `HashMap` hash computes, so `HashMap` hash values are exactly #8360's. A map derived from an already hashed map only hashes the nodes it doesn't share with it.

**How.** A field on the node, declared without an initializer so only hashed nodes carry it and inserts are untouched, reset when an edit session mutates the node in place (`canEdit(edit)`, checked on entry to `set` and `remove`), and not written when `Hash.hash` met a back-edge. `HashSet` hashes its backing map's summary; that is lawful but its values differ from #8360's.

**Why.** It is a Merkle tree over the trie; Scala 2.13's CHAMP nodes likewise cache a hash over their subtree's keys. `AtomRef.update` compares each new map with the previous one, and #8360 rehashes all n entries on every write.

Tests: `Hash > hash law > hashes HashMaps and HashSets independently of insertion order`, `… > hashes a HashMap like a fresh build after persistent and in-place edits`, and `Hash > cyclic values > does not cache a HashMap subtree hash computed from a circular sentinel`. Removing either the invalidation or the back-edge guard makes the matching test fail.

### Chunk: a sequence hash that composes over concatenation

**What.** `Chunk` hashed by flattening itself (which rewrites its backing) and folding `Hash.array`. It now hashes as a polynomial over mixed element terms, `h(xs) = Σ tᵢ·Bⁿ⁻¹⁻ⁱ mod 2³²`, so `h(L ++ R) = h(L)·B^|R| + h(R)`: the Rabin–Karp rolling hash used as a monoid, cached per node with the same back-edge rule.

**Why.** Appending to a hashed chunk hashes only the new nodes, and hashing no longer flattens as a side effect. Chunk's `Equal` only accepts chunks, so the requirement is Equal ⇒ same hash across every representation.

Tests: `Hash > hash law > hashes a Chunk independently of how it was built` (append, prepend, random concatenation trees, `take`/`drop` slices) and `Hash > cyclic values > does not cache a Chunk hash computed from a circular sentinel`.

### Trie: key hashes carried along the path

**What.** The `Trie` hash rebuilt every key string and hashed it from scratch, and equality materialized both tries into arrays.

**How.** Each key's hash is extended one character at a time down the path it is spelled on, and entries fold as XOR of mixed terms, so the result doesn't depend on the tree's shape. Equality walks both tries in order in lockstep and stops at the first difference.

**Why.** Keys share prefixes along the path, so a key is never rehashed from its first character. Tries with equal entries iterate in the same order, so no arrays are needed.

Test: `Hash > hash law > hashes and compares Tries independently of insertion order` (including distinct keys hashing apart). The crafted collision in `Trie > equality rejects tries with different numbers of entries after a hash collision` is re-solved for the new fold, so it still exercises the equality path.

### HashMap iteration: one explicit stack

**What.** Every trie node built its own iterator, so each entry passed through one iterator per level, and `keys`/`values` wrapped the entry iterator again.

**How.** One iterator walks the trie depth-first with an explicit stack, pushing children right to left so the order is unchanged, and projects keys, values or entries directly. The per-node iterator methods are deleted.

**Why.** One `next` call per entry instead of one per level plus a wrapper, as Clojure's and Scala's HAMT iterators do.

Test: `HashMap > iterators and getters > iterates every entry once across all node kinds`.

### Equal on native Map and Set: grouping by hash

**What.** Native `Map` and `Set` equality scanned one side for each entry of the other: O(n²).

**How.** Entries are grouped by hash (by key for maps) and matched one-to-one within their group, as in a hash join. `makeCompareMap`/`makeCompareSet` stay for `Schema`'s custom equivalences, which have no hash.

**Why.** Equal values have equal hashes and `compareBoth` already rejects on a hash mismatch, so a group holds exactly the pairs the scan could match: O(n) for well-distributed hashes, same results.

## Validation

- Hash, Equal, HashMap, HashSet, Chunk, MutableHashMap, Trie, Graph and Request tests: 574 passed after rebasing onto #8360's latest head. Every test added here fails on #8360 where it covers a bug, and all of #8360's tests pass.
- Full `effect` project: no new failures versus #8360's head. On the macOS machine used, 76 `test/schema/*` tests fail on both, because the filesystem ignores case (`effect/schema` resolves to `Schema.ts`); Linux CI doesn't have this. The ClusterWorkflowEngine "retry budget" test times out under full-suite load on both and passes alone.
- Doctests for the touched modules pass; `tsc -b`, `oxlint`, `dprint` and `effect-jsdocs --check` are clean on the changed files.

## Costs

- No measured slowdowns against #8360. `HashMap` insert is at parity (1.04×, CI 0.99–1.06).
- Hash values change for `HashSet`, `Chunk`, `Trie` and `null`, `undefined`, booleans and symbols. `HashMap` values are unchanged from #8360.
- Memory: one number per hashed `HashMap` trie node and per hashed `Chunk`, and one weak entry per outermost comparison result.

## Question for maintainers

`Hash` and `Equal` already assume in practice that values are immutable once hashed or compared: both cache results per object. If that contract were written down, `Date` and `RegExp` could be hashed cache-first like other objects (~25–31× faster on warm hashes in our measurements). This PR doesn't change how they are hashed or cached. Would you like that as a follow-up?


front-depiction:
Subsumed by #8360.

PR #8369, opened 2026-09-22
Precompute constant tag hashes as literals (stacked on #8368)

front-depiction:
Stacked on #8368 (which includes #8360). Only the last commit is new here: "Precompute constant tag hashes as literals".

**What.** Several `[Hash.symbol]` implementations rehashed a constant tag name on every call: `Hash.string("Some")`, `Hash.string(this._tag)` for `Fail`/`Die`, `Hash.string("HashMap")`, `Hash.string(options.op)` for `Exit`, and so on. Some built a template string first (`` `${this._tag}:${this.waiting}` ``, `` `Named:${this.id}` ``). Each of these is now a numeric literal computed once, at authoring time.

**How.** Literals replace the calls in `Option`, `Result`, `Exit`, the `Fail`/`Die`/`Interrupt` reasons, `DateTime` zones, `NetAddress`, `AsyncResult` (a table over tag × `waiting`), `HashMap`, `HashSet`, `Trie`, `Graph` (including its `directed`/`undirected` type), and the `Map`/`Set`/`Invalid Date`/`[Circular]` constants in `Hash`. Each literal carries a comment naming the string it stands for.

**Why.** A constant subexpression evaluated on every call is work a partial evaluator would fold; here it's folded by hand, so there is no runtime cost left, not even at module initialization. Hash values are unchanged, except `Interrupt` and `DateTime` zones: those now combine the tag literal with the id or offset instead of hashing a freshly built string, so they no longer allocate per call.

**Guard.** A test (`Hash > precomputed tag hashes > match the hashes of their tag names`) recomputes every literal from its name through the public API, so a wrong literal fails CI. Changing one literal by one makes it fail.

No public API changes.


front-depiction:
Closing: measured no speed benefit, and the literal hashes compress worse than the calls they replace (+58 B gzip on basic.ts). #8368 stays as the single PR.

PR #8370, opened 2026-09-22
Stop shipping module-init work that programs never use

front-depiction:
Lets bundlers drop several module-init-time tables and prototype objects when the module they belong to is imported but the feature they back is never used.

Rollup can only remove a top-level `const` when it can prove the initializer has no side effects. Two expression shapes defeat that proof every time: a top-level loop that fills a table (a loop is a side effect it has to run), and a spread of another value's own properties or a member read off a call result (it can't prove a spread is getter-free, or that a call has no side effects, so it keeps the call and everything it pulls in). Several modules built their internal state this way, so importing anything from them paid for the whole thing regardless of whether it was ever used.

- **`Mime`**: the extension/type lookup tables were built by top-level `for` loops over the MIME database at module evaluation. Any import of the module ran the loops and kept the tables reachable. They're now built lazily, on first lookup, behind a `getTables()` accessor.
- **`SchemaRepresentation`**: the check annotation schema was defined as `...RepresentationAnnotationSchema.fields`. The spread keeps `RepresentationAnnotationSchema` (and its schema fields) reachable, since Rollup can't prove the spread never reads a getter. It now lists the two fields (`id`, `payload`) directly, with no spread.
- **`HttpRouter.disableLogger`**, SQL's `StatementProto`, and the internal prototype objects backing **`ai/Toolkit`**, **`cli/Prompt`**, and **`process/ChildProcess`** all ended their top-level `const` in a trailing member read or spread on a call result — `middleware(...).layer`, `...Effectable.Prototype(...)`. Wrapping each in an immediately invoked function expression turns the initializer into a single call. `babel-plugin-annotate-pure-calls`, which the build already runs (`babel dist --plugins annotate-pure-calls`), marks that call `@__PURE__` for exactly this shape (a call in assignment position, executed during initialization), so Rollup can drop the whole binding when nothing references it. Each wrapped body builds and returns a fresh object and has no other observable effect, so the annotation is honest, not a workaround. One exception: `disableLogger`'s body constructs a `MiddlewareImpl`, which increments the module-private `middlewareId` counter. That counter only makes context keys unique, and middleware order comes from insertion order, not from the id. Dropping an unused `disableLogger` therefore changes only internal key strings.

Bundle size changes for the CI fixtures are in the bundle-size comment below.

## Test plan

- [x] `vitest` for every touched module's tests (`Mime`, `SchemaRepresentation`, `HttpRouter`, SQL `Statement`, `ai/Toolkit` (via `Tool`, `Chat`, `LanguageModel*`, `Response`), `cli/Prompt`, `process/ChildProcess`) plus the two new regression tests (`HttpRouter.disableLogger`, `Statement`/`StatementProto`): 275 tests in 14 files, all passing
- [x] `tsc -b` clean
- [x] `oxlint` and `dprint check` clean on every touched file
- [x] All 33 bundle fixtures measured base vs. head: no fixture larger (see the CI bundle-size comment)
- Perf: skipped — none of the touched code paths are hot; `Mime`, SQL `Statement`, and `HttpRouter`'s `disableLogger` are one-time module-init construction, and the three additional candidates are the same shape (a prototype object built once, not per-call)


packages/effect/src/http/Mime.ts
```diff
@@ -7,27 +7,44 @@
...
-    if (index === 0) {
-      typeToExtension.set(type, extension)
-    }
-    if (!starred) {
-      extensionToType.set(extension, type)
+// Built on first lookup rather than at module evaluation, so bundles that import this
+// module without calling a lookup can drop the tables and the MIME data they index.
+let tables: Tables | undefined
+
+const getTables = (): Tables => {
```
tim-smart, line 20: Will just keep the mime changes please

PR #8374, opened 2026-09-22
Fix Stream losing elements and hanging in share/broadcast, concurrent mapEffect and callback

front-depiction:
Five bug fixes, three performance changes to `Stream`/`Sink`, one commit each. Order, failure and interruption are unchanged except where a fix corrects the result: `Stream.share`'s first consumer sees every element, late `share`/`broadcast`/`broadcastN` subscribers end, a synchronous throw in concurrent `mapEffect` fails the stream, a failing `callback` registration fails the stream, and the crash in 4 returns the full result. One timing effect: a never-suspending loop takes fewer run-loop steps/element, so an interrupt lands later.

## Summary

| # | change | kind | complexity | result |
|---|---|---|---|---|
| 1 | `Stream.share` loses elements and hangs; late subscribers of `Stream.broadcast` / `broadcastN` / `share` hang | bug fix | per pull: one extra run-loop step | lost elements and hangs → every element, every subscriber ends |
| 2 | concurrent `Stream.mapEffect` (and `tap`, `tapBoth`, `bindEffect`, `partitionEffect`, `Channel.tap`) hangs when `f` throws | bug fix | per element: one extra run-loop step | hang → fails like `Effect.die` |
| 3 | `Stream.callback` (and `Channel.callbackArray`) hangs when the registration effect fails or dies | bug fix | per stream: one `catchCause` | hang → fails with the registration's cause |
| 4 | `Sink.collect`, `Sink.take`, `Stream.mapAccum` (and `groupedWithin`) throw `RangeError` on large chunks | bug fix | stack use O(chunk size) → O(1) | `RangeError` at 1M elements → correct result |
| 5 | `Stream.forever` / `Stream.repeat` slow down and leak with every repetition | bug fix | time O(n²) → O(n); memory O(n) → O(1) | 308× / 224× faster at 16k repetitions; ~1.4 KB → ~4 B retained per repetition |
| 6 | `Stream.concat` chains run as one flat sequence | algorithmic | O(n²) → O(n) | 68.8× at 1k, 624× at 10k concatenations |
| 7 | sequential `Stream.mapEffect` (and `tap`, `flattenEffect`) in one stage | constant factor, engine-sensitive | O(n), unchanged | 1.51× flat across 1k / 10k / 100k |
| 8 | effectful folds: `Sink.reduceWhileEffect` (and `reduceEffect`, `findEffect`), `Sink.takeWhileEffect` (and `takeUntilEffect`), `Sink.takeWhileFilterEffect`, `Channel.runFoldEffect` | constant factor | O(n), unchanged | 1.19–3.18× faster; 137–550 B less memory per element at chunk size 1 |

Costs: `share` 2–4% slower/pull (one-element chunks; five paired: 1.8%, 2.1%, 3.0%, 3.5%, 4.1%; same-side control 1.003× [0.987, 1.014]), 0 B more/element; concurrent `mapEffect` 136 B more/element (~3 KB/fiber), time at noise floor — paired 0.96×–1.05×, same-side 0.95×–1.10×; `collect`/`take` unchanged; `callback` +848 B/stream, 0 B/element, 0.97× [0.95, 1.00] at 100k and at the µs limit below; fusion in 7: 0.98× [0.97, 0.99] and 0.98× [0.96, 1.01] on concurrent `mapEffect` before fix 2.

Growth factors: larger-size time ÷ smaller, per side, confirming complexity class. Multiples: base ÷ head — below 1× is slower.

## 1. `Stream.share` loses elements, late subscribers hang (bug fix)

Two lost wake-ups: `share` started upstream before its first consumer subscribed — at budgets 3–64, `share(Stream.make(1, 2, 3))`'s first consumer missed elements or never ended at 53 of 62 (`broadcast`: 16 of 62); its end was only a message to subscribers, so late `broadcast`/`share(idleTimeToLive)` subscribers, or `dropping`-dropped ones (also `broadcastN`), never ended. Fix: one hub owns subscriptions/upstream, subscribing then starting in one uninterruptible step, storing the result as its state before the end publishes. Class: lost wake-up / parent-guarantee re-check.

- 9 new tests (budgets 3–64, 3–12×0–16, 3–6×0–90) fail base, pass head; deliberate breaks (ordering/timing/missing `uninterruptible`) each fail a test, the last stalling every later consumer (2/364 points).
- Base-vs-head probe, 14 scenarios (budgets 2048/5; 2 consumers, capacity 1): identical except fixed cases (`share`+`take(2)`@5: `[3]`→`[1,2]`; late `broadcast`: hang→`[]`/failure).

| case (one-element chunks, capacity 16) | base growth, 1k → 10k → 100k | head growth | multiples at 1k, 10k, 100k (95% CI) |
|---|---|---|---|
| `share` + `runCount` | ×9.9, ×10.1 | ×9.9, ×10.1 | 0.97× [0.96, 0.97], 0.98× [0.97, 0.99], 0.97× [0.95, 0.99] |
| `broadcastN` (n = 1) + `runCount` | ×9.9, ×10.2 | ×9.9, ×10.2 | 1.00× [0.99, 1.01], 1.00× [0.97, 1.01], 0.98× [0.97, 1.00] |

- Cost: hub-state read (one run-loop step/pull) costs 2–4% on one-element chunks (3% of 66 ms/100k pulls ≈ 20 ns/pull); dropping it recovers 3.3% on 64-element chunks (0.967× [0.957, 0.989]) — same-side 1.01× [1.00, 1.01], 1.00× [0.97, 1.01], 1.00× [0.99, 1.01], above noise. Allocation: 3,377 → 3,381 B at 1k, 2,079.6 → 2,080.0 B at 10k (~3.5 KB fixed, 0 B/element).

## 2. Concurrent `Stream.mapEffect` hangs when `f` throws (bug fix)

Above `concurrency: 1`, `mapEffect` called `f` in its pull loop and forked the result; a throw failed that loop, waiting behind in-flight effects (joins ordered, permits unordered), hanging if one never completed — `Effect.die` failed instantly. Fix: `f` runs inside the forked fiber (`runFork(Effect.suspend(() => f(value, index)))`), taking `die`'s path; `tap`, `tapBoth`, `bindEffect`, `partitionEffect`, `Channel.tap` share it. Class: per-call invariant.

- 2 new tests (ordered/unordered, `concurrency: 3`, `f` throws at element 2 while element 1 never completes) fail base (`TestClock` timeout), pass head: defect, interrupt at element 1, upstream release, `die`-matching log; deliberate breaks (ordered/unordered reverted, unordered suspending `handle(f(…))` not `f`) each fail the matching test.
- Base-vs-head probe, 16 forms (6 settings × combinators above, ordered/unordered + sequential): 14 hang on base, match `die` here (interrupted, released); the 2 already matching unchanged.

| case (`concurrency: 4`) | base growth, 1k → 10k → 100k | head growth | multiples at 1k, 10k, 100k (95% CI) |
|---|---|---|---|
| ordered | ×19.0, ×8.1 | ×18.2, ×8.2 | 0.982× [0.978, 0.992], 1.04× [0.99, 1.13], 0.988× [0.954, 0.998] |
| `unordered: true` | ×9.9, ×10.2 | ×10.1, ×10.2 | 0.994× [0.966, 0.995], 0.959× [0.946, 0.981], 0.977× [0.970, 0.982] |

- Cost: one `suspend`+closure/element, 136 B more (interpreted 1k: 4,234 → 4,370 B ordered, 3,583 → 3,719 B unordered; optimized 10k: +131 B ordered, +135 B unordered). At noise floor: paired 0.96×–1.05×, same-side 0.95×–1.10× (ordered 100k: 0.99× [0.97, 1.10] head-vs-head, 1.00× [0.95, 1.11] base-vs-base). ×19 step 1k→10k ordered, both sides; not investigated.

## 3. `Stream.callback` hangs when registration fails (bug fix)

`callback` (and `Channel.callbackArray`) forks the registration effect but never reads its result: a failed/died registration before ending the queue left the stream waiting forever (a throw already failed it). Fix: the queue owns the outcome — registration failure fails it (`catchCause(cause => Queue.failCause(queue, cause))`); offered elements emit, success leaves it open. Class: parent-guarantee re-check.

| deleted | what it compensated for | what carries it now |
|---|---|---|
| nothing; one arrow added: the registration's failure → the queue | the stream had no reader for the registration's exit | the queue, which the pull already reads |

- 3 new tests: failure after offered elements fails the stream after them; die/throw/`callbackArray` die → same `Exit.die`; success keeps it open (first two fail base via `yieldNow`+poll). Deliberate breaks (no relay; ending not failing it) fail 2 tests each; success-ending fails the third.
- Base-vs-head probe, 10 cases (various fail/die/throw/queue-state combinations): identical except the 3 fixed cases (hang→fail); finalizer logs match.
- Cost: one `catchCause`/stream — 848 B/stream (interpreted, 99.5 → 100.4 B/element at 1k), nothing/element; 0.97× [0.95, 1.00] at 100k, 2–11 µs at 10/1,000 elements (timer resolution). `onError` cost 1,488 B with no reader for its re-failure; `catchCause` was used.

## 4. Stack overflow on large chunks (bug fix)

`xs.push(...ys)` passes every element as an argument, overflowing the stack for large `ys` (Node 24: 120,000 succeeded, 150,000 threw); `fromIterable(array)` emits the array as one chunk, so `collect`, `take(n)` (n ≥ chunk length), `groupedWithin`, and a large `mapAccum` step threw at 1M elements. Fix: `collect`/`take` spread up to 8,192 elements (default 4,096), slicing larger ones in 8,192-element pieces; `mapAccum` loops instead.

- 3 new tests on a 1M-element chunk fail base (`RangeError`), pass head; base-vs-head probe (chunk boundaries, `transduce`, `groupedWithin`, edge inputs, `mapAccum` variants) identical except 1M cases (now full/ordered).

| case | base growth | head growth | multiples (95% CI) |
|---|---|---|---|
| `Sink.collect`, 1k → 10k → 100k → 1M | ×7.1, ×37.0, ×9.8 | ×7.2, ×35.0, ×8.6 | 1.02× [0.99, 1.05], 0.99× [0.92, 1.15], 0.99× [0.93, 1.06], 1.18× [0.82, 1.32] |
| `Sink.take`, 1k → 10k → 100k | ×7.2, ×10.7 | ×7.2, ×10.1 | 0.98× [0.46, 1.12], 1.02× [0.80, 1.06], 1.03× [1.00, 1.06] |
| `mapAccum`, 1k → 10k → 100k → 1M | ×10.0, ×11.0, ×9.2 | ×8.9, ×16.7, ×7.0 | 2.18× [2.13, 2.21], 2.46× [2.43, 2.47], 1.60× [1.59, 1.61], 2.09× [2.08, 2.14] |
| `mapAccum`, running sum kept small | ×10.1, ×9.8, ×9.5 | ×9.7, ×9.6, ×9.4 | 1.84× [1.79, 1.85], 1.92× [1.89, 1.99], 1.95× [1.93, 1.98], 1.95× [1.75, 2.22] |

- `collect`/`take`: no speed claim: every CI to 100k includes 1×; the ×35–37 step in `collect` (10k→100k) appears on both sides, not investigated.
- `mapAccum` is faster everywhere via a spread-to-loop swap, so engine-sensitive: the head's ×16.7 step (10k→100k) coincides with the running sum leaving V8's small-int range at 65,536; kept small, both grow linearly, multiple flat.

## 5. `Stream.forever` / `Stream.repeat` slow down and leak (bug fix)

`forever`/`repeat` implemented `X = s ++ X` as `concatWith(self, () => forever(self))`, nesting an O(n) layer per repetition. Fix: one loop over a single pull — close scope, step schedule, fork new scope, repeat; `concatWith` unchanged. Class: hylo.

| case | predicted | base growth | head growth | multiples (95% CI) |
|---|---|---|---|---|
| `forever`, `take` 1k → 4k → 16k | base ×16 per 4×, head ×4 | ×17.1, ×17.5 | ×3.7, ×4.0 | 15.2× [14.8, 15.7], 69.8× [68.8, 70.2], 308× [281, 323] |
| `repeat(Schedule.forever)`, same sizes | base ×16 per 4×, head ×4 | ×14.7, ×21.3 | ×4.0, ×4.1 | 11.5× [11.3, 11.9], 43.4× [42.5, 43.8], 224× [216, 227] |

| deterministic measure | predicted | base | head |
|---|---|---|---|
| run-loop operations, `forever`, 1k → 2k repetitions | base ×4, head ×2 | ×3.93 | ×2.00 |
| run-loop operations, `repeat`, 1k → 2k repetitions | base ×4, head ×2 | ×3.90 | ×2.00 |
| retained heap per repetition, `forever` (1k → 16k, after full GC) | base grows, head ≈ control | 1,440 B | 3.9 B |
| retained heap per repetition, `repeat` | base grows, head ≈ control | 1,380 B | 3.9 B |
| retained heap per element, control (plain infinite stream) | ≈ constant | 5.2 B | 4.2 B |

- Operation-count tests fail base, pass head; release-order/schedule-metadata pass both, fail under a break; base-vs-head probes (23 cases incl. `recurs(0/3)` + 13 interruption cases) identical.
- Base `repeat` growth ×21.3 at 4k→16k, above quadratic ×16, not investigated.

## 6. `Stream.concat` as one flat sequence (algorithmic)

`concat(a, b)` was `flatten(fromArray([a, b]))`: each concat added a flatten layer, so `n` concatenations cost O(n²). Associative, so concat now collects its tree's leaves once (explicit stack) into a single `flatten`; two non-concat streams run the old pipeline. Class: quadratic assoc.

| case | predicted | base growth | head growth | multiple (95% CI) |
|---|---|---|---|---|
| left-nested chain, 1k → 10k | base ×100, head ×10 | ×87.9 | ×9.8 | 68.8× [68.4, 70.2], 624× [588, 634] |

| case | multiple (95% CI) |
|---|---|
| 100 concatenations, left-nested | 10.54× [10.26, 10.76] |
| 100 concatenations, right-nested | 9.56× [9.40, 9.69] |
| 1,000 concatenations, left-nested | 65.85× [65.25, 66.71] |
| build and run a 10-part chain each time | 2.47× [2.40, 3.70] |
| one concat of two 5k ranges | 1.02× [0.98, 1.06], unchanged |
| one concat of two small streams | 0.98× [0.97, 1.02], unchanged |

- New tests on 1,000-part concatenations (left/right/balanced) check order, release timing, mid-chain failure, reuse, scoping — pass both; base-vs-head probe (1,000-deep nesting, failure/defect/interruption/laziness) identical elements, acquire/emit/release logs.

## 7. Sequential `Stream.mapEffect` in one stage (constant factor)

Sequential `mapEffect` (also `tap`, `flattenEffect`) was three pull stages — flatten to elements, run the effect, rewrap as one-element arrays. Now one stage (fusion), still one element/chunk; concurrent path unchanged apart from a shared check. Same O(n): a constant-factor gain.

| case | base growth | head growth | multiple (95% CI) |
|---|---|---|---|
| `mapEffect`, 1k → 10k → 100k | ×9.7, ×10.0 | ×9.8, ×9.9 | 1.51× [1.36, 1.53], 1.51× [1.48, 1.54], 1.52× [1.50, 1.53] |

| case | multiple (95% CI) |
|---|---|
| `mapEffect`, 10k one-element chunks | 1.42× [1.39, 1.43] |
| `tap`, 10k | 1.34× [1.33, 1.37] |
| `flattenEffect`, 10k | 1.46× [1.43, 1.47] |
| `mapEffect`, `concurrency: 4` | 0.98× [0.97, 0.99]; an earlier run 0.98× [0.96, 1.01] |
| `filterEffect` (does not use `mapEffect`) | 0.97× [0.90, 1.01], unchanged |

- New tests (effect order, per-run index restart, one element/chunk, failure/interruption/`tap` cases) pass both; base-vs-head probe (all concurrency settings, failure/nesting/async/interruption variants) identical output, effect/pull/release logs.

## 8. Effectful folds: one step per element (constant factor)

`whileLoop` runs one effect/element, handing the result to `step`; `Sink` folds instead wrapped each element in `suspend`+`flatMap`+a continuation duplicating `step`'s job, and `takeWhileEffect` also built a map+tuple/element. Now the body calls the user function, `step` updates state; a stopped fold sets a flag the loop checks next call. Class: hylo.

| case | chunk size | base growth, 1k → 10k → 100k | head growth | multiples at 1k, 10k, 100k (95% CI) |
|---|---|---|---|---|
| `reduceWhileEffect` | 4096 | ×10.1, ×9.4 | ×10.0, ×9.6 | 2.02× [2.00, 2.05], 2.04× [1.96, 2.07], 2.01× [1.96, 2.05] |
| `reduceWhileEffect` | 1 | ×11.3, ×11.7 | ×10.1, ×11.6 | 1.40× [1.33, 1.45], 1.35× [1.11, 1.82], 1.34× [1.08, 1.74] |
| `takeWhileEffect` | 4096 | ×9.9, ×10.4 | ×9.7, ×10.6 | 3.14× [3.08, 3.18], 3.18× [3.13, 3.31], 3.14× [3.00, 3.17] |
| `takeWhileEffect` | 1 | ×10.5, ×9.6 | ×10.6, ×9.8 | 1.68× [1.59, 1.82], 1.65× [1.38, 1.77], 1.64× [1.60, 1.73] |
| `takeWhileFilterEffect` | 4096 | ×10.0, ×9.8 | ×9.8, ×10.4 | 1.92× [1.81, 2.00], 1.99× [1.92, 2.04], 1.86× [1.68, 1.93] |
| `takeWhileFilterEffect` | 1 | ×10.0, ×10.6 | ×9.9, ×10.4 | 1.30× [1.18, 1.36], 1.28× [1.28, 1.39], 1.34× [1.28, 1.41] |

| bytes per element (1k → 100k slope) | chunk size 1, base → head | chunk size 4096, base → head |
|---|---|---|
| `reduceWhileEffect` | 650 → 449 B | 49.8 → 49.8 B |
| `takeWhileEffect` | 1,017 → 471 B | 413 → 70 B |
| `takeWhileFilterEffect` | 705 → 503 B | 102 → 101 B |

- Linear growth confirmed for time/memory (savings exceed prediction — each removed step also carried a closure or arguments object); new tests (mid-chunk stop, failures/throws, interruption) pass both, each deliberate break fails a test; base-vs-head probes (1,200 sink + 1,080 scoped-resource cases, incl. mid-chunk stop while upstream holds a resource) identical.
- Timing: a never-suspending fold, interrupted from another fiber, folded 509 elements on base and 2,035 on head before the interrupt landed.

`Channel.runFoldEffect` gets the same change (fold state updated in `whileLoop`'s `step` instead of `map`ping each result):

| `Channel.runFoldEffect` | chunk size 1 | chunk size 4096 |
|---|---|---|
| time at 1k / 10k / 100k | 1.23× / 1.22× / 1.19× | within noise |
| bytes per element, base → head | 827 → 690 B | 49.8 → 49.8 B |
| run-loop steps per 100k elements | fewer by one continuation per element | 100,339 → 100,288 (−0.05%) |

- At chunk 4096, paired runs read 0.90–0.99×: an engine artifact — deterministic-compilation flags narrow the gap to <0.1% CPU time; alone, this fold's background compilation sometimes settles ~0.7% heavier, disappearing once other folds share the call site.

## Questions for review

- **`Stream.broadcast` start.** Starts the upstream at creation, dropping pre-subscription elements unless `replay` keeps them (documented example does); starting on first subscription instead, as `share` does, would change upstream timing — left unchanged.
- **`Channel.embedInput` defects.** A throw in the input function fails the channel; a forked-fiber die drops like any `forkIn` fiber, hanging a waiting `self`. Fix needs a race per pull, a `merge` (queue hop/element — per-pull checks can't wake it), or running it inside the fork (dropping throws too, free); both change public behavior — which one?
- **`Stream.toPubSubTake` end.** Its end is a published `PubSub` message — a late subscriber, or one `dropping` drops, waits forever (section 1's hub no longer uses it); it has no state for the end — keeping it in the replay buffer or shutting it down after changes what subscribers see — left unchanged.

## Method

Node v24.3.0 (V8 13.6), Apple M4 Pro. Sections 1–3: paired CPU-time runs, base/head, fresh processes, ABAB, 7 rounds, one run/sample, median of ratios w/ bootstrap 95% CI, `--no-concurrent-recompilation` (load avg 18–45); `Stream.ts`/`Channel.ts` byte-identical to `upstream/main`. `runtimeperf`: base/head alternate, 5 rounds, 300ms/100ms warmup, bootstrap 95% CI (incl. 1×=unchanged); growth factors: one fixed batch size/size. Load avg 16–26/14 cores; pairing cancels most. Section 5: built `dist` output (load avg 19–36); memory=V8 young-gen bytes (1 GiB semi-space, no in-window GC), as slope over repeated runs.


tim-smart (changes requested):
One focused change per PR please

PR #8384, opened 2026-09-22
Compile Schema once per schema instead of once per value

front-depiction:
Five commits removing work Schema repeated on every decode, comparison or format although the result can only change when the schema changes. No confirmed bug was found here, so nothing leads with a fix. Decoded values, issue trees, error messages, equivalence results, formatter output and JSON Schema documents are unchanged; the behavior differences are below.

## Summary

| # | commit | change | kind | cost before → after |
|---|---|---|---|---|
| 1 | `cbf79cf201` | the cached candidate-index lookup allocated the index builder's closure context | constant factor | one context per union decode → none |
| 2 | `0ec7c07977` | a union with no literal members mapped and filtered its members on every decode | algorithmic in member count `m` | O(m) + 2 arrays + 1 closure per decode → O(1), no allocation |
| 3 | `18afd4845b` | `toEquivalence` recompiled a `Suspend` body at every level of the compared values, each level capturing a copy of an unused `path` array | algorithmic in value depth `D` | compile O(D·n_body), retained O(D²) → O(n_body), O(1) |
| 4 | `7fcbaed367` | `toFormatter` recompiled a `Suspend` body at every level of the formatted value | algorithmic in `D` | compile O(D·n_body), retained O(D) → O(n_body), O(1) |
| 5 | `17d3fbc9b6` | the tuple rest split `const [head, ...tail] = rest` ran per comparison and per format | constant factor | one array per `Arrays` node per call → none |

Costs: across the 33 CI bundle fixtures no file moves more than 0.05 KB either way — 15 report 0.00 KB, 31 of 33 sit within ±0.03 KB, and the extremes are `schema-toEquivalence.ts` −0.04 KB and `schema-toFormatter.ts` +0.05 KB (ground-truth rebuild of both dists with `annotate-pure-calls` and `stripInternal: true`, built identically).

Two reviewer notes:

- **Commit 4 is the one addition rather than a deletion.** The other four delete work; `toFormatter` has no memoized entry point to reuse — it takes per-call `options` — so it gets one share table per call, scoped to the returned formatter.
- **`toFormatter`'s `onBefore` hook** (public, on `Schema.toFormatter`) now fires once per suspended target instead of once per level of the formatted value. Output is unchanged; a side-effecting hook observes fewer calls. See question 3.

A third difference: the `@internal` `SchemaAST.getCandidates` now returns the same frozen array on every call for a union with no literal members, as the literal-only and single-tag indexes already did, instead of a fresh mutable one. Every consumer in the repo only reads `.length` and indexes — the union parser and its `iterateEager` driver, the generated JIT/AOT code, `toEquivalence`, `toFormatter` — so a caller that pushed onto the result would previously have extended its private copy and now silently does nothing.

Growth factors below are the larger size's measurement divided by the smaller's, per side.

## 1–2. Union candidates

`getIndex` both looked up the cached index and built it on a miss, so the builder's closures shared one function context with the lookup, which V8 allocated on entry just to return a cached index. The lookup is now `candidateIndexCache.get(types) ?? makeIndex(types)`.

For a union not discriminated by literal tags, every decode then ran `otherwise[runtimeType].map(i => types[i]).filter(filterLiterals(input))`: two arrays, a closure and one `toCandidate` lookup per member — a per-call construction of an invariant, since it depends only on the input's runtime type. And with no literal member, `filterLiterals` returns `true` for everything, so the filter is a no-op on a degenerate case. Candidates are now resolved once per runtime type, frozen and returned directly; unions with any literal member keep the per-input filter, and the discriminated index reuses the same shared arrays for its fallbacks.

Predicted: base time linear in `m`, head flat, so the multiple grows with `m`; head allocates nothing per decode.

| decode, m = 2 / 8 / 32 | base | head | multiple | growth base | growth head |
|---|---|---|---|---|---|
| time per decode | 62.7 / 106.4 / 305.1 ns | 40.0 / 40.9 / 41.2 ns | 1.57× / 2.60× / 7.41× | ×1.70, ×2.87 | ×1.02, ×1.01 |
| bytes per decode | 856 / 904 / 1,456 B | 392 / 392 / 392 B | 2.18× / 2.31× / 3.71× | ×1.06, ×1.61 | ×1.00, ×1.00 |

AGREE on both axes: base grows with `m`, head is flat.

| other shapes | base | head | result |
|---|---|---|---|
| `NullOr(Struct)` decode, time | 49.1 ns | 29.4 ns | 1.67× |
| `NullOr(Struct)` decode, bytes | 616 B | 160 B | −456 B |
| `String \| Number \| Boolean`, string input, bytes (time not measured) | 544 B | 88 B | −456 B |
| `Literal \| Literal \| String`, non-matching string, bytes (time not measured) | 560 B | 496 B | −64 B |

The last row is the filtered path, deliberately unchanged: only commit 1's closure context (64 B) is removed. An earlier draft that also shared the filtered arrays measured 2.9× *slower* there, because V8's `Array.prototype.filter` is much slower on frozen arrays; filtered unions therefore keep main's fresh map + filter.

DISAGREE, disclosed: the `runtimeperf` suite on the previous base reported 1.46× / 2.37× / 6.94× and 1.25× for `NullOr(Struct)`. My re-measurement gives larger multiples because my fixtures use minimal single-field members, shrinking the parsing work the saving is divided by. Same direction and shape; the suite's are the conservative numbers.

## 3–4. Recursive equivalence and formatter compiled once per value level

`Suspend` held `equivalence ??= recur(ast.thunk(), path)`, but each level of the value reached a *different* `Suspend` closure, so the body was compiled again at every level. In `toEquivalence` a `path` array was threaded through every case and copied at every level (`[...path, i]`), never read by anything — these compilers return booleans and strings, not paths. `Suspend` now resolves its target through the already-memoized `toEquivalence` entry point and `path` is deleted; `toFormatter` gets the per-call share table above.

Predicted: retained memory O(D²) for equivalence (D closures each holding an O(D) path copy) and O(D) for the formatter, both → O(1); compile O(D·n_body) → O(n_body); steady-state time unchanged.

| recursive tree, D = 8 / 64 / 512 | base | head | multiple | growth base | growth head |
|---|---|---|---|---|---|
| `toEquivalence` retained per compiled equivalence | 13,568 / 185,152 / 6,976,832 B | 944 / 944 / 944 B | 14.4× / 196× / 7,391× | ×13.6, ×37.7 | ×1.00, ×1.00 |
| `toFormatter` retained per compiled formatter | 5,656 / 39,704 / 312,376 B | 1,104 / 1,104 / 1,104 B | 5.12× / 36.0× / 283× | ×7.02, ×7.87 | ×1.00, ×1.00 |
| cold: fresh schema → compile → one comparison | 12.29 / 43.92 / 873.42 µs | 3.92 / 8.54 / 53.92 µs | 3.14× / 5.14× / 16.20× | ×3.57, ×19.9 | ×2.18, ×6.31 |

AGREE: base retained memory grows toward quadratic in depth (×13.6 then ×37.7 for 8× the depth), head is byte-identical at every depth. Not an engine artefact — all six retained numbers are byte-identical under `--no-opt --no-maglev`. It is the dead `path`: level k captured a copy of a path of length ≈2k. ALGORITHMIC, checked: at D = 512 under those flags the cold comparison is still 1,175.38 µs → 161.00 µs, a 7.30× gain.

| steady state, one compiled interpreter reused | base | head | delta | label |
|---|---|---|---|---|
| `toEquivalence` D = 64, default tiers | 5.79 µs | 3.99 µs | −31% | ENGINE-SENSITIVE |
| `toEquivalence` D = 512, default tiers | 58.66 µs | 34.13 µs | −42% | ENGINE-SENSITIVE |
| `toEquivalence` D = 64, `--no-opt --no-maglev` | 23.32 µs | 19.41 µs | −17% | same case, tiers off |
| `toEquivalence` D = 512, `--no-opt --no-maglev` | 184.07 µs | 157.15 µs | −15% | same case, tiers off |
| `toFormatter` D = 64, default tiers | 19.77 µs | 19.14 µs | −3% | neutral, not claimed |

The warm gain is ENGINE-SENSITIVE: it shrinks from −42% to −15% once the optimizing tiers are removed, because base creates one distinct closure per value level so the call sites stay polymorphic and V8 cannot inline them, while head reuses one closure per AST node. The formatter's −3% is inside noise at that size.

DISAGREE, disclosed, two:

- The `runtimeperf` cold numbers (a true first-ever compile, batch 1) were base 10.15 µs / 78.10 µs / 3.01 ms → head 6.44 / 11.85 / 51.10 µs: −36% / −86% / −98%. My probe warms the compiler itself for 20 iterations before timing a fresh schema and gets −68% / −81% / −94%. The head sides agree closely (51.10 vs 53.92 µs at D = 512); the base side differs 3.4× there because a true first compile also pays to compile the compiler, and base does D times more of that work. The table quotes the conservative warm-compiler figures; on a genuine first compile the gain is larger.
- Retained memory was earlier reported as 896 B and 1,056 B, measured before commit 5 landed. On the full branch they are 944 B and 1,104 B — +48 B each, the hoisted `tail` array now retained per `Arrays` node. Still flat in depth, still three orders of magnitude below base at D = 512.

## 5. Tuple rest split

`const [head, ...tail] = rest` ran inside the comparison and format callbacks, allocating one array per `Arrays` node per call although `rest` is fixed per schema. Hoisted to compile time.

| warm allocation per call | base | head | delta |
|---|---|---|---|
| recursive tree equivalence, D = 8 / 64 / 512 | 2,520 / 18,200 / 53,320 B | 936 / 6,760 / 53,320 B | −1,584 / −11,440 / 0 B |
| recursive tree formatter, D = 8 / 64 / 512 | 16,032 / 174,400 / 4,977,152 B | 14,448 / 160,840 / 4,977,152 B | −1,584 / −13,560 / 0 B |
| `TupleWithRest` (3 rest elements) equivalence, n = 4 / 32 / 256 | 1,136 / 4,048 / 27,344 B | 728 / 3,640 / 26,936 B | −408 B, constant in n |
| `TupleWithRest` (3 rest elements) formatter, n = 4 / 32 / 256 | 4,872 / 21,128 / 154,168 B | 4,464 / 20,736 / 153,880 B | −408 / −392 / −288 B |
| `Array(Number)` (1 rest element, empty tail), n = 4 / 32 | 592 / 3,504 B | 416 / 3,328 B | −176 B, constant in n |

ENGINE-SENSITIVE because of the D = 512 row: base and head allocate identically there, since once that closure is optimized V8's escape analysis already elides the rest array, while at D = 8 and 64 it does not. The `TupleWithRest` rows show the saving is constant per call, not growing with tuple length — a small constant win, not claimed as more.

Small delta diagnosed both ways: `Array(Number)` at n = 256 reads base 26,144 B (min) / 26,800 B (median) against head 26,624 B (min = median) — below base's median, above its min, base non-deterministic in that window and head not. Confirmed as neither an improvement nor a regression.

## Behavior evidence

- `packages/effect/test/schema`: 2,260 passed, 76 failed. The same 76 fail identically on `upstream/main` (7 `SchemaCompiler*` / `SchemaJITCompiler` / `VariantSchema` files, failing on import or setup, not on assertions); the two failing sets were diffed and are identical.
- Four new tests fail on `upstream/main` and pass here: two pin the shared candidate arrays and their immunity to mutation, two pin that a recursive equivalence and a recursive formatter compile their body once (`9 !== 1` on main) rather than once per value level.
- A fifth new test guards commit 4's share table rather than main's behavior: two mutually recursive schemas give two *distinct* suspended targets with the same AST tag, and the test pins the formatted output. Keying the table on `ast._tag` instead of the AST node yields `{ "tag": "odd", "n": undefined, "odd": undefined }` — an audit found that mis-keying passed the whole suite before this test existed.
- Five deliberate breaks, each caught: dropping the candidate memo fails 2 tests; dropping the literal filter, 3 tests and 114 probe mismatches; taking the rest split from `elements`, 2 tests and 178 mismatches; reverting the equivalence `Suspend` to a per-level compile, 1 test; the tag-keyed share table, the new test and 84 mismatches.
- Differential probe, base and head module graphs in one process, 28 schemas over a shared pool of shaped and random values: **72,389 observations compared, 0 mismatches**, in both the interpreter and the JIT (`SchemaJITCompiler/enable`) configurations. Per pair it compares `decodeUnknownResult` and `encodeUnknownResult` under `errors: "first"` and `"all"` (whole issue tree, serialized structurally), messages from `SchemaIssue.makeFormatterDefault` and `makeFormatterStandardSchemaV1`, `is`, `asserts` and the thrown issue's message, `makeOption`, `toFormatter` with and without an `onBefore` hook, `getCandidates` in both constructor modes, `toJsonSchemaDocument` in draft-2020-12 / 07 / 04 under both `onExcessProperty` modes, and `toEquivalence` over every ordered pair of schema-valid values on one shared compiled equivalence. The corpus covers unions with and without literals in both orders, tagged, mixed and refined unions, tagged tuples, tuples with rest elements before and after the rest, recursive trees and union lists, mutually recursive schemas with two distinct suspended targets, and a recursive union of structs with no discriminator.
- Staleness probe, 7 scenarios, 47 observations, 46 matching: annotating after the first compile, one suspended target reached from two roots, a `suspend` thunk returning a fresh AST per call, `toFormatter` with and without options interleaved, the candidate index driven through interleaved runtime types, repeated compiles of one recursive union, and a caller mutating the returned candidate array. The single difference is the intended one above. Nothing memoized can go stale: the `fallbacks` memo lives inside the pre-existing `WeakMap`-cached index under the same immutable `types`; `toFormatter`'s share table lives and dies with one returned formatter; `toEquivalence`'s memo already existed, is keyed on AST identity, and `annotate` builds a new AST.

## Method

- Node v24.3.0 (V8 13.6), macOS, 14 cores, shared with other jobs; load average 25–36, recorded around every run. base = `upstream/main` 061b9611d4, head = the committed branch.
- Time: one process per side per round, alternating, 5 rounds (3 for the tier-flag runs), median of per-round medians; each round is 40 cold samples, or 25 batched steady-state samples after 20 warm-ups.
- Memory: the validated young-generation meter (1 GiB semi-space so no GC falls inside a window, windows that saw a GC discarded, slope over r vs 2r repetitions). Every allocation row has min = median; retained bytes come from two full GCs per side, with the schema's own cost subtracted.
- Gates: `oxlint -f unix` and `dprint check` clean; `tsc -b tsconfig.json` reports 527 errors, byte-for-byte the same 527 as `upstream/main` and none in `packages/effect/src`; the 33 bundle fixtures rebuilt from both dists as CI does.

## Questions for review

1. **Inlined representations.** `toRepresentation` rebuilds an inlined (unreferenced) node at every occurrence, so `toJsonSchemaDocument` pays the unfolded size rather than the AST size. A throwaway prototype measured −90% on a shared-subschema DAG (21.01 ms → 2.09 ms at depth 12) and −65% on a 512-field struct, with identical JSON — exploration-branch figures, not part of this PR's evidence, and none of that code is here. Sharing one representation object across occurrences would make `$defs`-free output arrays (`required`, `enum`, …) shared too, because `rewriteRefs` copies schema objects but not their array values; base already shares annotation values with the user's annotation object. Is that acceptable?
2. **Union member selection in `toEquivalence` / `toFormatter`.** Each union level runs a full `is(a) && is(b)` on the remaining value, so comparing two 512-element recursive union lists costs 25.9 ms against 63 µs to decode one — quadratic in length. (Exploration-branch figures, on the code as it stands after this PR, which does not change the selection.) Narrowing to the candidate index alone, or fusing membership into the comparison, would change which member is selected for a value that satisfies the type but fails a refinement. Is that acceptable?
3. **`toFormatter(options.onBefore)`.** Commit 4 makes the hook fire once per suspended target rather than once per level of the formatted value. Output is unchanged, but a hook with side effects sees fewer calls. Is the call count intended to be observable?
4. **`standardSchema`'s `jsonSchema.input` / `output`.** These recompile the whole `toJsonSchemaDocument` pipeline per call. Caching per (AST, target) would make repeats O(1), but callers mutate the returned object (`schema.$defs = …` inside `toBaseStandardJSONSchemaV1`), so the cache would have to hand out copies. Worth doing?


front-depiction:
Replaced by six focused PRs, one change each: #8428, #8429, #8430, #8434, #8435, #8436.

PR #8391, opened 2026-09-22
Fix a permit leak in PartitionedSemaphore and make transactional waiters interruptible

front-depiction:
Two ways of taking a permit go wrong when the taker is interrupted.
`PartitionedSemaphore` takes the permits one step before it installs the
release, so an interrupt in between keeps them. `TxSemaphore` and
`TxReentrantLock` take the opposite position and make the whole acquire
uninterruptible, including the wait, so a parked waiter cannot be cancelled at
all. Both are the same rule read twice: taking the resource and installing its
release are one step, and the wait before it is not part of that step.

Each bug has a test that fails on `main`. `Scheduler.MaxOpsBeforeYield` set to a
small value makes a fiber yield at every point between two steps, which turns a
window into a deterministic case; a budget below 3 never makes progress, so the
sweeps run 3..64.

## Bugs

| # | primitive | what went wrong | symptom | fails on main at | test |
|---|---|---|---|---|---|
| 1 | `PartitionedSemaphore.withPermits` | acquires inside `restore`, installs the release after the wait returns | permits kept with no holder: `available` drops to 0 and every later `withPermits` waits forever | budget 7 | `an interrupted withPermits returns its permit at every yield` |
| 2 | `PartitionedSemaphore.withPermitsIfAvailable` | takes the permits in `suspend`, installs `ensuring` one step later | same leak, same deadlock | budget 4 | `an interrupted withPermitsIfAvailable returns its permit at every yield` |
| 3 | `TxSemaphore.withPermit`, `withPermits`, `withPermitScoped` | the acquiring transaction parks on `Effect.txRetry` inside the uninterruptible acquire of `acquireUseRelease` or `acquireRelease` | interrupt, `timeout` and `race` cannot cancel a waiter until a permit frees | deterministic; each test hangs until the test timeout on main | `interrupting a withPermit waiter completes, and leaks no permit`, `a timeout cancels a withPermits waiter`, `interrupting a withPermitScoped waiter completes` |
| 4 | `TxReentrantLock.withReadLock`, `withWriteLock`, `readLock`, `writeLock` | same shape | same: a waiting reader or writer cannot be cancelled | deterministic; hangs until the test timeout on main | `interrupting a withWriteLock waiter completes, and the lock stays usable`, `interrupting a readLock waiter completes` |

Bug 2 is not in the report this branch started from; it is the same shape one
function further down the file.

## Fixes

| # | fix | who owns the fact now |
|---|---|---|
| 1, 2 | one masked step takes what is available and registers a waiter for the rest, and installs the release in that same step; only the wait for the rest is interruptible | the call's waiter record: it holds `requested - owed` permits at every moment, granted or not, so one release settles either outcome |
| 3, 4 | the acquiring transaction never retries: it takes the resource or reports that it could not, and the release is installed in the same uninterruptible step. Waiting is a separate transaction that only reads and retries | the taking transaction, which is atomic by construction, and the scope or exit frame that holds the release |

The Tx fix is `txAcquireUseRelease` / `txAcquireScoped`, shared by both modules.
A waiter parked in the read-only transaction is an ordinary interruptible
`txRetry` park, which is what `TxQueue.take` already relies on.

**Behaviour this changes on purpose.** Splitting the transactional acquire into
"wait, then take" means a woken waiter can find the resource gone and park
again, so the order in which several `TxSemaphore` waiters woken by one release
are granted is not the same as on `main`. `TxSemaphore` documents no ordering
guarantee and `main`'s own order already varies with the op budget (a probe with
three waiters and one `releaseN` gives `a,b,c` on `main` at budgets `-` and 7
and `a,c,b` at 4, 9, 13 and 30; head gives `c,a,b`, `a,c,b` and `a,b,c` at the
same budgets), and permit accounting is exact on both sides in every case.
`PartitionedSemaphore`'s documented round-robin grant order across partitions
is unchanged.

## Deletion matrix

| deleted | what it was compensating for | what carries the guarantee now |
|---|---|---|
| `PartitionedSemaphore.withPermits`'s `flatMap(restore(take), … ensuring(release))` | the take and the release living in two steps | one masked step that takes and installs the release together |
| `take`'s inline registration body (now `enterUnsafe`) and its private cleanup closure | the same registration written once per caller | `enterUnsafe` and `leaveUnsafe`, shared by `take` and `withPermits` |
| `withPermitsIfAvailable`'s `suspend` + `ensuring` | the same split | the same masked step |
| `acquireUseRelease` / `acquireRelease` in seven `TxSemaphore` and `TxReentrantLock` combinators | an acquire that had to be uninterruptible because it could park | a transaction that never parks, and a separate interruptible wait |
| `acquireWrite`'s re-lookup of the reader count it was already iterating, and the grant written inline in each acquire | one spelling of the rule per entry point | `canRead` / `canWrite` predicates and `grantRead` / `grantWrite`, each written once and used by the acquire, the try and the wait |
| two of the three copies of the "Number of permits must be positive" defect | the same literal repeated per entry point | one `invalidPermits` |

Added: `internal/txAcquire.ts`, 47 lines, the two combinators above, with two
callers and no state of its own.

## Evidence

| check | result |
|---|---|
| tests that fail on main | 13 new tests; 7 fail on main (bugs 1–4), reproduced on a worktree carrying only the new test files. The other 6 pin behaviour that was already correct and pass on both sides: a woken waiter still gets its permits and a permit handed to it, a non-interrupted waiter still gets the permit and the lock, `withPermits` still rejects a non-positive count without running the effect, and the acquire-to-release step stays atomic under a budget sweep |
| deliberate breaks | taking the permits outside the `withPermits` mask, dropping `restore` around its body, forgetting the permits a waiter is still owed, unmasking `withPermitsIfAvailable`, making the transactional wait uninterruptible, installing the transactional release a step after the acquire, and dropping the non-positive-count guard each fail at least one named test |
| differential probe | 125 seeded scenarios over the three primitives, mixing holders, interrupts of granted holders, `take`, `tryAcquire`, scoped acquisition and releases, at budgets 7 and 30 and unbudgeted, recording both the counters and the grant order: traces identical on base and head. A second probe of 84 scenarios (grant order under simultaneous contention, permit accounting across failing and dying bodies, over-capacity requests, zero and negative counts, reentrancy, nesting inside an outer `Effect.tx` including a rolled-back transaction, and scoped acquisition) differs only in the `TxSemaphore` grant order described above |
| nested transactions | `withPermit` inside an outer `Effect.tx` commits, parks, rolls back and completes identically on base and head |
| full `packages/effect` suite | the same 84 failing tests, by name, as base (network and environment); 13 more tests pass |

A contract slip found and fixed during the work: the first version of the Tx
fix ran the body before dying for `withPermits(sem, 0, …)`, where `main` dies
without running it. `withPermitsImpl` now rejects a non-positive count before
anything else; the pinning test above fails if that guard is removed.

## Cost

Measured with the runtimeperf harness, base `upstream/main` against the branch
head, batch size pinned to 1 across sizes, 5 alternating rounds in fresh
processes. The box was loaded (load average 10–35); the control on the noisiest
fixture was -0.8%, CI [-14.4%, +5.2%]. Per-call op counts and allocated bytes
are deterministic and measured separately with the validated meter.

Predicted before measuring: the uncontended `withPermits` no longer builds a
waiter record or goes through the `callback` protocol, so it should lose most of
its steps and about half its bytes, as a per-call constant; `TxSemaphore` should
lose the generator acquire and its `acquireUseRelease` wrapper, a smaller
constant; `TxReentrantLock` keeps the same number of steps, so it should be
neutral. Measured: agrees on all three.

| fixture | base 1k | head 1k | base 10k | head 10k | base 100k | head 100k | delta | ops/call base → head | bytes/call base → head |
|---|---|---|---|---|---|---|---|---|---|
| `PartitionedSemaphore.withPermits`, free | 616.8µs | 225.0µs | 4.74ms | 2.17ms | 45.45ms | 24.56ms | -48.6% to -51.7% | 11 → 5 | 1633 → 769 |
| `PartitionedSemaphore.withPermitsIfAvailable`, free | 280.4µs | 259.4µs | 2.89ms | 2.77ms | 29.77ms | 27.67ms | -6.3% to -7.9% | 8 → 7 | 865 → 825 |
| `TxSemaphore.withPermit`, free | 3.00ms | 1.79ms | 26.64ms | 17.76ms | 271.65ms | 208.89ms | -32.8% to -41.4% | 49 → 39 | 10788 → 8876 |
| `TxReentrantLock.withWriteLock`, free | 6.97ms | 7.23ms | 78.21ms | 71.49ms | 412.57ms | 458.97ms | -14.0% to +3.2%, every CI spans 0: neutral | 60 → 60 | 13806 → 13864 |

Growth is linear on both sides at every fixture (×10 per size step in a quieter
run: base ×9.9, ×10.4 and head ×10.1, ×10.4 for `withPermits`), so the deltas
are per-call constants.

An independent 5-round re-run at 10k on a quieter box, same base and head:
`PartitionedSemaphore.withPermits` 4.03ms → 1.99ms (-50.2%),
`withPermitsIfAvailable` 2.77ms → 2.61ms (-6.3%), `TxSemaphore.withPermit`
26.39ms → 17.29ms (-35.1%), `TxReentrantLock.withWriteLock` 37.65ms → 37.66ms
(+1.8%, CI [-2.5%, +2.6%]). Both runs agree.

The uncontended `withPermits` no longer goes through the waiter callback: taking
a free permit is now a synchronous check inside the mask, which is six fewer
steps and half the bytes. `TxSemaphore.withPermit` takes its permit with a
single `TxRef.modify` transaction instead of a generator-based acquire wrapped
in `acquireUseRelease`. `TxReentrantLock.withWriteLock` is unchanged in steps
and 58 bytes per call larger (+0.4%); that row varies by about ±40 B between
meter runs, and its timing deltas sit inside the control's band.

Bundle: the one added file is internal and has no public export, so CI's fixture
sizes are the size story.

## Not fixed here

`TxQueue.offerAll` on a bounded queue with more values than capacity never
completes, even with a draining consumer, because the batch is one transaction
that can never fit. `TxPubSub.publishAll` has the same shape. That is a
liveness question about batch semantics, not an interrupt window, so it is left
for its own change.


tim-smart (changes requested):
Could we only have one focused change per PR please

